Issue #596, opened 2014-05-27
Maximum times a user can vote on a poll


Polarbear541:
So as per the original roadmap thread, I have been working on adding an option which allows you to specify the maximum number of times a user can vote on a poll. But after seeing #583 I'm wondering whether I've misunderstood the idea.

My interpretation is shown below:
![image](https://cloud.githubusercontent.com/assets/2014518/3095400/67c5a522-e5c7-11e3-9234-ae599ece2151.png)
Where the user can vote multiple times on the same poll.

My question is will this be useful and should I continue to work on it? :P

JN-Jones:
@mybb/developers @mybb/management @mybb/sqa As we're already at 1.8.8 without ever getting this done I'm not sure whether we really need this. Especially since we decided to concentrate on 2.0 and stop adding features to 1.8. While @Polarbear541 had done some work on this (#1095) it doesn't look like that PR was near done. So I'd propose closing this and instead keep it in mind for 2.0.

Stefan-ST:
It's certainly not a feature that has been requested many times.

JN-Jones:
@euantorano Fine with closing this too?

euantorano:
Yep, no objections here.

PR #1095, opened 2014-08-02
WIP Feature #596

Polarbear541:
This is what I have so far for feature #596. It's still a WIP, it doesn't work with multiple option polls yet (though I'm not sure why you'd want to vote multiple times on a multiple options poll anyway) and undoing votes doesn't work either.

Also I'm still yet to sort out guest voting with the IP address restrictions.


Destroy666x:
Closing this since I don't think @Polarbear541 will fix the conflict anytime soon.

PR #1370, opened 2014-09-03
fix: inline error for username taken (guest post/thread), refs #841

Nik101010:
#841

### DIFF
--- a/newreply.php
+++ b/newreply.php
@@ -310,27 +310,18 @@ if($mybb->input['action'] == "do_newreply" && $mybb->request_method == "post")
 	// If this isn't a logged in user, then we need to do some special validation.
 	if($mybb->user['uid'] == 0)
 	{
-		// Check if username exists.
-		if(username_exists($mybb->get_input('username')))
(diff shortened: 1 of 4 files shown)


Sama34:
@Nik101010

Nik101010:
@Sama34 sanatizing usernames in the build_postbit function would make sense, but this could lead to stuff beeing double escaped... 
I reviewed the whole files, I found one path were the username was not sanatized... added it in this case

PaulBender:
Everything looks good to me, if no one else has any issues, this can be merged.

newreply.php
```diff
@@ -1074,7 +1055,7 @@
 			$post = $db->fetch_array($query);
 			if(!$mybb->user['uid'] || !$post['username'])
 			{
-				$post['username'] = $mybb->get_input('username');
+				$post['username'] = htmlspecialchars_uni($mybb->get_input('username'));
```
Sama34: Please revert this. the `build_postbit` function will handle this on another PR.

Nik101010:
@Sama34 I reverted the last commit ...

Issue #1296, opened 2014-08-23
Hello World! 2.0


Sama34:
I think we need to update the example plugin file provided within the pack. It is very outdated.

PR #1389

PR #1389, opened 2014-09-05
Implements #1296 Hello World 2.0

Sama34:
Implements #1296 Hello World 2.0

### DIFF
--- a/inc/plugins/hello.php
+++ b/inc/plugins/hello.php
@@ -276,12 +276,12 @@ function hello_activate()
 
 		// Filter valid entries.
 		$setting = array_intersect_key($setting,
-									   array(
-										   'title' => 0,
-										   'description' => 0,
-										   'optionscode' => 0,
-										   'value' => 0,
-										   ));
+			array(
+				'title' => 0,
+				'description' => 0,
+				'optionscode' => 0,
+				'value' => 0,
+		));
 
 		// Escape input values.
 		$setting = array_map(array($db, 'escape_string'), $setting);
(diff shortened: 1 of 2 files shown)


inc/plugins/hello.php
```diff
@@ -8,17 +8,58 @@
...
+	
+// cache templates - this is important when it comes to performance
+// THIS_SCRIPT is defined by some of the MyBB scripts, including index.php
+if(defined('THIS_SCRIPT'))
+{
+    global $templatelist;
+
+    if(isset($templatelist))
+    {
+        $templatelist .= ',';
```
JordanMussi: Wouldn't this break something by adding a comma when it might not be used?
Better to have a comma on the two lines below instead. Also there shouldn't be spaces for consistency.
```php
	if(THIS_SCRIPT== 'index.php')
	{
		$templatelist .= ',hello_index,hello_message';
	}
	elseif(THIS_SCRIPT== 'showthread.php')
	{
		$templatelist .= ',hello_post,hello_message';
	}```

Comment (Destroy666x) [inline inc/plugins/hello.php]: I don't think it breaks anything, it only tries to select a template with empty title, which is already done here anyways: https://github.com/mybb/mybb/blob/feature/inc/class_templates.php#L49

But I agree that your code is more preferred.

**Comment by Sama34** [comment]
You are right. Though there shouldn't be any breaking. Please open a new issue since this is already closed.

File: inc/plugins/hello.php [hunk 2]
```diff
@@ -8,17 +8,58 @@
...
+    }
 
-$plugins->add_hook("pre_output_page", "hello_world");
-$plugins->add_hook("postbit", "hello_world_postbit");
+	if(THIS_SCRIPT== 'index.php')
```
WildcardSearch: Why not

```php
if(THIS_SCRIPT == 'index.php' || THIS_SCRIPT == 'showthread.php')
{
	$templatelist .= 'hello_index, hello_message';
}
```

...or even `if in_array(THIS_SCRIPT, array('index.php', 'showthread.php'))`

JordanMussi: Just [without the space](https://github.com/mybb/mybb/pull/1389#r25568895) in the templatelist :wink:

WildcardSearch: lol My comment was for line 28 👅

WildcardSearch: Also, there needs to be a space before the `==` in both cases.

PR #1455, opened 2014-09-18
Fix #1453

### LINKED ISSUE #1453: Smilie breaks JS when ending with backslash
Hi all

After upgrading my forum to 1.8 I noticed that no js based stuff worked at all and I tracked it back to one of my smilies, which had the code ":-\" at the time. That code is inserted in the smilie function in js and escapes the closing quote =&gt; js syntax error.
I've changed to ":-/" for now so it works but the smilie codes should be checked before inserting them in js.

Bye.

Original thread: [Smilie breaks JS when ending with backslash](http://community.mybb.com/thread-159684.html)

Destroy666x:
Used addslashes() to prevent smilie JS errors (not sure if it's the best
solution, but it was used in misc.php only for some reason).

Also added spaces in misc.php inserter for consistency and listed the
replacements in new lines instead of separating them with space/comma
which looks kinda weird.

https://github.com/mybb/mybb/issues/1453

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -887,8 +887,8 @@ EOF;
 					$finds_count = count($finds);
 					
 					// Only show the first text to replace in the box
-					$find = htmlspecialchars_uni($finds[0]);
-					$image = htmlspecialchars_uni($image);
+					$find = addslashes(htmlspecialchars_uni($finds[0]));
+					$image = addslashes(htmlspecialchars_uni($image));
 					if(substr($image, 0, 4) != "http")
 					{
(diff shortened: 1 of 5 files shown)


PaulBender:
Looks good to me, just need to fix the merge conflicts.

Sama34:
I tested this and also worked fine.

ATofighi:
@Destroy666x

PaulBender:
Merge conflicts still need fixing

JN-Jones:
Merged manually, changed template version to `1804`.

PR #1521, opened 2014-10-14
Redirect_profileupdate - small change

Cu8eR:
for more info: http://community.mybb.com/thread-161147.html


Destroy666x:
The capitalization should be corrected, for example:

> You will be now returned to the Username settings.

There is no reason to capitalize only the `Username` part.

Cu8eR:
@so, no capital letters? Or chenge it to "Username Settings"?

Destroy666x:
Either no capital letters or the exact name of the page, for example `Change Username page` (as in the menu).

Cu8eR:
Updated

Destroy666x:
It's ok.

Anyways, some more things:
- few phrases have `Thank you,` at the beginning and the rest not, I'd remove it from all phrases for consistency
- Edit Options redirect string is wrong, it says you're redirected to the subscription list
- https://github.com/Cu8eR/mybb/blob/patch-2/inc/languages/english/usercp.lang.php#L367 could be changed too, it redirects to the subscription list

Cu8eR:
About the second point - you mean L327?

Destroy666x:
Yep.

Cu8eR:
@Destroy666x - should be OK now

Issue #1526, opened 2014-10-17
Make usage of $mybb->get_input() instead of $mybb->input[]


JanMalte:
The input should be accessed by ```$mybb->get_input()``` instead of ```$mybb->input[]```.

JanMalte:
@Sama34 I'm working on it, but as already stated, it is quite a lot of work to be done and tested.

JanMalte:
I think all changes are made and can be merged. Only calls with casting were replaced by the method call alternative.

Destroy666x:
It's still not optimised (UserCP actions), `MyBB::INPUT_STRING` was used inconsistently only in several function calls (the rest is empty and in my opinon all should have stayed empty) and you introduced a syntax error in admin/modules/config/mod_tools.php

JanMalte:
@Destroy666x Could you show me the syntax error?

I've added some more commits to fix the points Destroy666x mentioned. If something still doesn't fit, please tell me and give an example how it should be.

Destroy666x:
I think we can close this too, unless someone has time to work on it properly.

euantorano:
Yep, I'm all for closing it unless somebody is willing to put themselves forwards to work upon it.

Destroy666x:
Closing then, feel free to reopen if you're going to make a PR.

PR #1527, opened 2014-10-17
changed ->input[] to ->get_input()

JanMalte:
PR for issue https://github.com/mybb/mybb/issues/1526


ATofighi:
We can’t automatically merge this pull request.

JanMalte:
The PR should now be able to merge.

Destroy666x:
This PR is still not good. You're:
- not using `get_input()` where it would be nice to see
- using `get_input()` where it's not needed, for example what's the point of calling the function 30 times in usercp.php when the correct value is already assigned to another variable? https://github.com/JanMalte/mybb/blob/feature-1526/usercp.php#L53
- adding the `MyBB::INPUT_STRING` to some places while we decided to skip this parameter since it's default, I think.

JanMalte:
Adding ```MyBB::INPUT_STRING``` was only by merging the changes of @ATofighi 

Assigning the casted value to the input array and accessing it directly isn't the way it was supposed to be used. The code wasn't optimized, as this would make the PR bigger then necessary and would mix different aspects together.
After merging this PR, the code could be optimized by assigning the value to a local variable and remove duplicated method calls. But I would strongly suggest to keep this in a separate PR, as it makes merging much easier.

Using ```MyBB::get_input()``` for multi level array access is not very well tested right now, therefor this was skipped for now, but can come later in a new PR, I think.

ATofighi:
but I don't use `MyBB::INPUT_STRING`...
this is my changes: https://github.com/mybb/mybb/commit/cf871e276d93218e1107f51ec8531e5da174506a
do you see any `MyBB::INPUT_STRING` in it?

Destroy666x:
> Assigning the casted value to the input array and accessing it directly isn't the way it was supposed to be used.

You're missing the "because..." part here. I don't see any reason why not, as long as the proper check is done it doesn't matter if the result is assigned to `$mybb->input['action']` or `$randomvariable`. Not saying that the 2nd isn't better, but the 1st is proper too.

> The code wasn't optimized, as this would make the PR bigger then necessary and would mix different aspects together.

But we won't accept a PR that makes the code much less optimised for no reason, that's the point.

Sama34:
I agree there is no need to replace all `$mybb->input[''action]` with `$mybb->get)input('action')` since, as it was already mentioned it was done that way for performance reasons. And while it may not be a big difference for low frequency visited pages it may be for the rest, `usercp.php` in this case for example.

admin/modules/config/mod_tools.php
```diff
@@ -865,7 +865,8 @@ function checkAction(id)
 
 			if($mybb->get_input('threadprefix', MyBB::INPUT_INT) >= 0)
 			{
-				$thread_options['threadprefix'] = $mybb->get_input('threadprefix', MyBB::INPUT_INT);
+				$thread_options['threadprefix'] = $mybb
+                    ->get_input('threadprefix', MyBB::INPUT_INT);
```
Destroy666x: I meant this line when speaking about the syntax error, but I may have noticed wrong and it possibly doesn't throw anything. But It should be reverted regardless - no need to separate it into a new line.

dequeues:

Destroy666x:
IMO we should just close this, there's been absolutely no progress since we pointed out inconsistencies and suboptimal code 1.5 year ago.

euantorano:
I'm closing this for now due to the above.

PR #1547, opened 2014-10-24
Fix #1545

### LINKED ISSUE #1545: is_member() should always return true for "-1"
The is_member(&#36;groups) function (inc/functions.php) should always return true if &#36;groups is -1.
The new setting type groupselect saves the -1 value for the given field if All groups option was selected, so the function should check the groups accordingly.

Original thread: [is_member() should always return true for "-1"](http://community.mybb.com/thread-159313.html)

Destroy666x:
Add -1 handling to `is_member()` function.
https://github.com/mybb/mybb/issues/1545

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -390,11 +390,7 @@ class PostDataHandler extends DataHandler
 		}
 
 		// Check to see if this person is in a usergroup that is excluded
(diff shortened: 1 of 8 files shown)


Sama34:
So I assume you can now finish this up? :P

Destroy666x:
This is finished now.

PaulBender:
Everything works for me

PR #1574, opened 2014-11-01
Fixes #1573 "Can post replies?" doesn't override "Can only reply to own ...

### LINKED ISSUE #1573: "Can post replies?" doesn't override "Can only reply to own threads?"
"Can post replies?" doesn't override "Can only reply to own threads?"

Stefan-ST:
...threads?"
Fixes #1573


Destroy666x:
Merging this and will open a new issue,

PR #1602, opened 2014-11-09
Base Database Class

euantorano:
http://community.mybb.com/thread-162474.html

This pull request adds a base database class based on the MySQL classes but that works with MySQL, PgSQL and SQLite.

Also fixes potential error in db_pdo where it was using an undeclared variable $user when connecting.

### DIFF
--- a/inc/db_base.php
+++ b/inc/db_base.php
@@ -8,7 +8,7 @@
  *
  */
 
-abstract class DB_Base
+interface DB_Base
 {
     /**
      * Connect to the database server.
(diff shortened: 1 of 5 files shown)


PirataNervo:
@euantorano you forgot to load the DB_Base interface in the core.

euantorano:
Damn. Forgot we don't have such a thing as an autoloader :(

PirataNervo:
I'm creating a new PR now

PirataNervo:
https://github.com/mybb/mybb/issues/1652

Sama34:
Was concerning such thing :P

Issue #1555, opened 2014-10-26
Template Edit Bug


Destroy666x:
Its not that big of a bug but when I went to edit the templates, it looked like this.
[attachment=32260]

Original thread: [Template Edit Bug](http://community.mybb.com/thread-158012.html)

PaulBender:
Fixed in #1611

PR #1611, opened 2014-11-13
Fix #1592 and #1555

### LINKED ISSUE #1592: Search error when code mirror is on
Hi,

Browsers Search option is a necessary tool for designers to find their needed codes between themes and templates codes; But you can't use browsers search option (Ctrl+F) when Code Mirror is Turned on.It6 doesn't work!

Original thread: [Search error when code mirror is on](http://community.mybb.com/thread-158635.html)

Destroy666x:
Allows users to search CodeMirror content with browser by introducing the `viewportMargin: Infinity` option:
https://github.com/mybb/mybb/issues/1592

Ensures that the CodeMirror is always max. 100% width:
https://github.com/mybb/mybb/issues/1555

Also fixed tabbing issue (2 spaces were inserted instead of a tab due to outdated CodeMirror option usage) and a small typo in comment.

### DIFF
--- a/admin/inc/class_form.php
+++ b/admin/inc/class_form.php
@@ -894,7 +894,7 @@ class DefaultFormContainer
 	/**
 	 * Initialise the new form container.
 	 *
-	 * @param string The title of the forum container
+	 * @param string The title of the form container
 	 * @param string An additional class to apply if we have one.
 	 */
 	function __construct($title='', $extra_class='')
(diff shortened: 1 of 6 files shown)


Sama34:
This is the default style sheet with your edits:
clipboard01

After typing:
clipboard0

Destroy666x:
@Sama34, this should fix it after cache refresh.

PaulBender:
All looks good to me. Tested in all major browsers (Firefox, Google Chrome, Opera and Internet Explorer) and can't find anything wrong.

PR #1655, opened 2014-11-22
Fix #1654

### LINKED ISSUE #1654: Hidden forum stats.php appears
Forum permissions closed. But appear in the statistics.

"Forum" view closed.

Original thread: [Hidden forum stats.php appears](http://community.mybb.com/thread-163221.html)

Destroy666x:
Avoid unviewable most popular forum caching by checking them inside the
function instead of relying on a global variable available only in
stats.php (while the function is used in several places).

https://github.com/mybb/mybb/issues/1654

### DIFF
--- a/inc/class_datacache.php
+++ b/inc/class_datacache.php
@@ -659,8 +659,22 @@ class datacache
 	function update_statistics()
 	{
 		global $db;
+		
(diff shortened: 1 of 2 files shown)


Destroy666x:
Going to fix it in https://github.com/mybb/mybb/pull/1671

PR #1675, opened 2014-12-02
Fixes #1674 Custom MyCodes not accepted by editor.

### LINKED ISSUE #1674: Custom MyCodes not accepted by editor.
> Create a new custom MyCode:
> Regular Expression:
> 
> ``` regex
> \[code=gist\](.*?)\[/code\]
> ```
> 
> Now try your new MyCode in a new thread/post. The code gets replaced by a common `[code]` MyCode.
> 
> Full or source editor doesn't makes the difference it seems.

Original thread: [Custom MyCodes not accepted by editor.](http://community.mybb.com/thread-163255.html)

martec:
Fixes #1674

### DIFF
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -263,8 +263,25 @@ $(document).ready(function($) {
 		allowsEmpty: true,
 		isInline: false,
 		allowedChildren: ['#', '#newline'],
-		format: '[php]{0}[/php]',
-		html: '<code class="phpcodeblock">{0}</code>'
+		format: function (element, content) {
+			var	desc = '',
+				$elm  = $(element);
+
+			if($elm.data('desc')) {
+				desc = '=' + $elm.data('desc');
+			}
+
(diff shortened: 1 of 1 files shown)


Cu8eR:
@martec

martec:
@Cu8eR 

plz read https://github.com/mybb/mybb/issues/1674

Sama34:
I think this is not to be fixed.

ATofighi:
@Sama34 
IIRC it fixes the issue for `code` and `php` tags.

Sama34:
But as you mentioned new issues come with the _fix_. I think the overall issue can't be really fixed and users should instead not create custom MyCodes like those.

Destroy666x:
We could do it for every tag as the SCEditor developer said, but I don't think it's a good solution either.. So if noone has a better idea, I'm for not fixing this issue too.

Destroy666x:
I still don't think that doing it for each tag is worth a hassle and noone had objections after my last reply, so closing. Still hoping SCE developers will come up with a better solution.

Issue #1679, opened 2014-12-05
CDN upload function


euantorano:
Currently, all CDN uploads are done inline within other functions, making it difficult to extend the CDN functionality via plugins. There should be a method to copy files to a CDN with plugin hooks.

Cu8eR:
should be considered https://github.com/mybb/mybb/pull/1824

euantorano:
Yes, #1824 requires this patch.

PR #1680, opened 2014-12-05
Creating function to perform copies to CDN.

euantorano:
#1679


inc/class_core.php
```diff
@@ -519,6 +521,8 @@ public function get_asset_url($path = '', $use_cdn = true)
 			$url = $path;
 		}
 
+		$plugins->run_hooks('get_asset_url_end', $url);
```
Sama34: What is the purpose of this hook?

inc/functions.php
```diff
@@ -7850,3 +7850,66 @@ function log_spam_block($username = '', $email = '', $ip_address = '', $data = a
...
+
+					if($success)
+					{
+						$uploaded_path = $cdn_upload_path;
+					}
+				}
+			}
+		}
+
+		$plugins->run_hooks('copy_file_to_cdn_end', $success);
```
Sama34: I'm not sure if I follow you, so if the copy was successful and I edit it to `false` within a plugin even though I can't do anything else, how is that useful?

What I mean is, why isn't this passing some more arguments like `$file_path` and `$uploaded_path`?

inc/functions_upload.php
```diff
@@ -675,12 +675,13 @@ function delete_uploaded_file($path = '')
 	$path = ltrim($path, '/');
 	$cdn_path = realpath($cdn_base_path . '/' . $path);
 
-
 	if($mybb->settings['usecdn'] && !empty($cdn_base_path))
 	{
 		$deleted = $deleted && @unlink($cdn_path);
```
Sama34: This is not allowed in the standars BTW.

inc/functions_upload.php
```diff
@@ -675,12 +675,13 @@ function delete_uploaded_file($path = '')
...
 	}
 
+	$plugins->run_hooks('delete_uploaded_file');
```
Sama34: Should be passing `$path` and `$deleted`.

inc/functions_upload.php
```diff
@@ -708,6 +709,8 @@ function delete_upload_directory($path = '')
 		$deleted = $deleted && @rmdir($cdn_path);
```
Sama34: Same here.

inc/functions_upload.php
```diff
@@ -708,6 +709,8 @@ function delete_upload_directory($path = '')
...
 	}
 
+	$plugins->run_hooks('delete_upload_directory');
```
Sama34: > Should be passing `$path` and `$deleted`.

inc/functions_upload.php
```diff
@@ -740,15 +743,9 @@ function upload_file($file, $path, $filename="")
 	$filename = preg_replace("#/$#", "", $filename); // Make the filename safe
 	$moved = @move_uploaded_file($file['tmp_name'], $path."/".$filename);
 
-	$moved_cdn = false;
-	$cdn_base_path = rtrim($mybb->settings['cdnpath'], '/');
-	$cdn_path = rtrim(realpath($cdn_base_path . '/' . $path), '/');
+	$cdn_path = '';
```
Sama34: If my acknowledgement of PHP is correct, this line is unnecessary because `$cdn_path` is passed by argument.

inc/class_core.php
euantorano: @Sama34 Some CDNs might use different URLs. The main ones (AWS S3, etc) don't, but some might. This hook is there just in case.

inc/functions.php
euantorano: @Sama34 You may want to set it to false if uploading to an external CDN to say that the upload was unsuccessful (eg: due to curl not being available or the external CDN being offline).

It's not passing those arguments because passing multiple arguments with the current plugin system is a complete joke. You can still access the variables as globals.

inc/functions_upload.php
euantorano: @Sama34 And where does it say that? http://docs.mybb.com/1.8/development/standards/

THe alternative is much more long-winded. This is just simple boolean evaluation...

inc/functions_upload.php
euantorano: Agreed on this one.

inc/functions_upload.php
euantorano: IMO it's much more readable this way. It works either way though.

inc/functions.php
Sama34: How can you access variables as globals if you are running the hook inside a function? Or are you assuming no plugins will use this function inside another function with completely different variable names?

euantorano: What? I'm not sure I understand the question. `$real_file_path`, `$file_dir_path` and `$file_name` are all globals. You can access them via plugins by globalising them as with other global variables. I know, because I've already created a plugin for a client that does this and uses this exact hook to upload files to AWS S3...

inc/class_core.php
Sama34: But this is quite an addition IMO and it should be checked.

@mybb/developers @Destroy666x @JN-Jones 

This method runs from 6/7 to 9 nine times a page in a fresh installation.

euantorano: And how would you suggest implementing this then? `run_hooks` is only slow if there are hooks associated, and if there are hooks associated then obviously the hook is useful...

Sama34: > And how would you suggest implementing this then?

I wouldn't, TBH.

That is why I'm seeking for more feedback, it is not up to us only really..

euantorano: We can always remove it as it's not crucial, but it could be nice to have for some developers.

inc/functions.php
Sama34: Oh I see the code now. Well, that was unexpected hook system implementation..

You are ignoring the reason passing by reference was implemented..

But anyways, if no one else opposes then I wouldn't either.

euantorano: I'm not ignoring it, I'm just saying the current implementation of hook parameters is deeply flawed :wink: Plus, these parameters shouldn't really be modified as changing them could easily break other plugins that might be using the same hook.

Sama34: I'm curious about why you call it `flawed`, some interesting reading you could have :P?

JN-Jones:
@euantorano are you still working on this one (there're some comments where you agreed to change something) or should we defer it to 1.8.5?

euantorano:
I really want to get this into 1.8.4. I'll make the changes required after work tonight.

Sama34:
I would like to have some feedback about the hook in `get_aseet_url()` mentioned above.

If no one does I have no problem then.

Sama34:
@euantorano

euantorano:
Going to look at this this afternoon ready to merge ASAP.

inc/class_core.php
euantorano: @mybb/developers Removing this hook for now, but any other thoughts?

inc/functions.php
euantorano: Fixed.

Regarding flawed, the approach allows only one parameter, which is a serious limit and leads to stupid workarounds (such as allocating an array of variable references so multiple variables can be accessed... ew...). It also uses multiple methods to call functions(`call_user_fun` for object functions, dynamic execution of variables as functions), where only one would work (`call_user_fun` handles all use cases, but is deemed as "too slow").

euantorano:
@Sama34 All updated according to your suggestions. Happy now?

PaulBender:
Looks good but I have no real way of testing it as I don't have any CDN set up.

Destroy666x:
Have you tested it yourself with a CDN @euantorano?

euantorano:
Yes. Everything worked as intended.

Sama34:
Ok, I will merge this now.

You should open an new issue with the removed hook for discussion if you feel it could be really helpful IMO.

PR #1705, opened 2014-12-21
Fixes #1704 Banned IPs not being blocked

### LINKED ISSUE #1704: Banned IPs not being blocked
I just got a new user registration from an IP address that is in my blocked list. Also last known IP is also in the blocked list. I am using wild cards to block both IPs such as 103.230.105.*. Yet it still allowed registration to go through. Is 1.6 block list not working with 1.8?

Original thread: [Banned IPs not being blocked](http://community.mybb.com/thread-161124.html)

Stefan-ST:
Fixes #1704

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -6287,7 +6287,7 @@ function is_banned_ip($ip_address, $update_lastuse=false)
 		$ip_range = fetch_ip_range($banned_ip['filter']);
 		if(is_array($ip_range))
 		{
-			if(strcmp($ip_range[0], $ip_address) >= 0 && strcmp($ip_range[1], $ip_address) <= 0)
+			if(strcmp($ip_range[0], $ip_address) <= 0 && strcmp($ip_range[1], $ip_address) >= 0)
 			{
 				$banned = true;
 			}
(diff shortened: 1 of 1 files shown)


PirataNervo:
I can feel @Stefan-ST's pain. What about allowing IPv4 ranges but disallow IPv6 and note that in the description?

Sama34:
Din't somebody commented in the community, that `1.*.*.*` will do what `1.*` does? That is enough IMO.

Destroy666x:
I don't think it's enough if people who are upgrading added a lot of `1.*` ranges which will stop working in 1.8. I still hope it will be fixed for IPv4 at least. Shouldn't be too hard I guess - check if the IP is IPv6, the if it is use the new code and if it's not, use the old `preg_match()` code from 1.6: https://github.com/mybb/mybb/blob/master/inc/functions.php#L5540

Sama34:
Oh so it breaks backwards compatibility, sorry didn't noted that in the discussion. Can't those be updated in the upgrade process?

Stefan-ST:
1.* should work now

Destroy666x:
Looks good now.

PR #1706, opened 2014-12-21
Fixes #1657 Upgrade script loads plugins.

### LINKED ISSUE #1657: Upgrade script loads plugins.
Apparently, and intentionally, plugins are being loaded during the last step of the upgrade script.

__Reference:__ http://mybbhacks.zingaburga.com/showthread.php?tid=1006&pid=17964#pid17964

The issue seems to be the following code:
```PHP
	// Attempt to run an update check
	require_once MYBB_ROOT.'inc/functions_task.php';
	run_task(12);
```

The script doesn't really finish and the `.lock` file is not created. This may lead administrators to the false assumption that the installation script is locked when it isn't.

This is not the first plugin I find to be loaded during upgrades BTW.

Stefan-ST:
Fixes #1657

### DIFF
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -626,7 +626,16 @@ function upgradedone()
 
 	// Attempt to run an update check
 	require_once MYBB_ROOT.'inc/functions_task.php';
-	run_task(12);
+	$query = $db->simple_select('tasks', 'tid', "file='versioncheck'");
+	$update_check = $db->fetch_array($query);
+	if($update_check)
+	{
+		// Load plugin system for update check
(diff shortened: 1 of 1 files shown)


PirataNervo:
I'm a bit confused: what do you mean by "within the parser itself" ? And wasn't the point to stop loading plugins within the upgrade script? The PR actually loads them.

Destroy666x:
Don't the plugins need to be loaded first to work?
https://github.com/mybb/mybb/blob/feature/inc/init.php#L230

I think this PR is fine.

Sama34:
Semem you are right @Destroy666x.

Issue #1354, opened 2014-08-31
"Moderator Options" time difference


PaulBender:
ModCP

AdminCP

Also this is my suggestion to 1.8

Original thread: ["Moderator Options" time difference](http://community.mybb.com/thread-158312.html)

PR #1719, opened 2014-12-25
Improve #1354 "Moderator Options" time difference

Sama34:
Improve #1354 "Moderator Options" time difference

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1375,7 +1375,7 @@ EOF;
 			$expired = nice_time($remaining, array('seconds' => false));
 
 			$color = 'inherit';
-			if($remaining <= 0 || $remaining < 3600)
+			if($remaining < 3600)
 			{
 				$color = 'red';
 			}
@@ -1546,7 +1546,7 @@ EOF;
 			$expired = nice_time($remaining, array('seconds' => false));
 
 			$color = 'inherit';
-			if($remaining <= 0 || $remaining < 3600)
+			if($remaining < 3600)
 			{
 				$color = 'red';
 			}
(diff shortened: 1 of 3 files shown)


admin/modules/user/users.php
```diff
@@ -1526,8 +1542,24 @@ function toggleAction()
 		$mybb->input['moderateposting'] = 1;
 		if($user['moderationtime'] != 0)
 		{
-			$expired = my_date('relative', $user['moderationtime'], '', 2);
-			$existing_info = $lang->sprintf($lang->moderate_length, $expired);
+			$remaining = $user['moderationtime']-TIME_NOW;
+			$expired = nice_time($remaining, array('seconds' => false));
+
+			$color = 'inherit';
+			if($remaining <= 0 || $remaining < 3600)
```
Destroy666x: What's the point of checking `$remaining <= 0`? If it's lower than 3600 it's also lower than 0.

admin/modules/user/users.php
```diff
@@ -1371,8 +1371,24 @@
...
+			{
+				$color = 'red';
+			}
+			elseif($remaining < 86400)
+			{
+				$color = 'maroon';
+			}
+			elseif($remaining < 604800)
+			{
+				$color = 'green';
```
Destroy666x: I think yellow would be more suitable here, green indicates something good and I suppose this should be a warning.

Sama34:
The code was based off (not to say it was copy/pasting) from existing core in (IIRC) ''modcp.php'.

Destroy666x:
@Sama34 , even if there are conditionals like that there is no reason to duplicate the mistakes.

As for the colors, I see, they're consistent now. We could change the green to dark yellow or orange in the 4 files which set it based on time. But I guess this can be done in another issue.

Sama34:
Done.

PR #1720, opened 2014-12-25
Fix #1341 Closed board message should load language string, fallback to custom reason if set.

### LINKED ISSUE #1341: Closed board message should load language string, fallback to custom reason if set.
**Reference:**
http://community.mybb.com/thread-158217-post-1096704.html#pid1096704

Sama34:
Fix #1341 Closed board message should load language string, fallback to custom reason if set.


PirataNervo:
Merging :)

PR #1721, opened 2014-12-26
Fix #1321 Multipage doesn't work if portal.php is renamed

### LINKED ISSUE #1321: Multipage doesn't work if portal.php is renamed
We currently _support_ renaming `portal.php` to `index.php` but if administrators do so the pagination no longer works.

Sama34:
Fix #1321 Multipage doesn't work if portal.php is renamed

### DIFF
--- a/portal.php
+++ b/portal.php
@@ -51,13 +51,7 @@ if($mybb->settings['portal'] == 0)
 
 // Fetch the current URL
 $portal_url = get_current_location();
-
-$file_name = THIS_SCRIPT;
-$file_name_real = my_strtolower(basename($portal_url));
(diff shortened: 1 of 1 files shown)


Sama34:
That should fix it.

Destroy666x:
Looks ok, but what about `THIS_SCRIPT`?

Sama34:
`THIS_SCRIPT` is a constant and shouldn't change for backwards compatibility with plugins that do use it. Renaming the file will make no differences whatsoever over it really.

Destroy666x:
Ah, you're right I guess.

PR #1731, opened 2014-12-30
Fix #1730 right id column used

### LINKED ISSUE #1730: When listing all reports, there is a problem with reputations.
When reputation is reported, there was nick of author missing in modcp.php?action=allreports:

The problem is in selecting id's from database, because there are three id columns - id, id2 and id3.
CONTRIBUTING.md Wrote:Pull Requests should only be sent for confirmed issues.
I'm not sure what to do now. After confirmation, should I create pull request on Github straight away or create Issue first?

Original thread: [When listing all reports, there is a problem with reputations.](http://community.mybb.com/thread-164858.html)

CZ-madA12345:
As title says.


Destroy666x:
Works for me.

Issue #257, opened 2014-03-03
Change to SMTP mail composition to comply with standards


nmalcolm:
> **Problem:**  
> ## When a user writes an Email (PM) to another user, the mail header is constructed in this manner now (only the important lines):  
> 
> Return-Path: <fromemail>  
> To: <toemail>  
> From: <fromemail>  
> ## Reply-To: <fromemail>  
> 
> This (often) results in the mail server sending an error : <ins>_"The mail server does not understand the RCPT TO command. Reason: 553 5.7.1. Sender address rejected: not owned by user XXXXXX."_</ins>  
> As it seems, the attempt to send an email from [user@xxx.com](mailto:user@xxx.com) from the mailserver yyy.com is seen as a potential abuse and is rejected.
> ## **The mail should therefor be constructed (correctly) in this manner:**  
> 
> Return-Path: <Mybb_name> <Mybb_returnmail/adminmail>  
> To: <toname> <toemail>  
> From: <Mybb_name <Mybb_adminmail>  
> ## Reply-To: <fromname> <fromemail>  
> 
> **This scheme should be mandatory for ALL emails sent by the system.**  
> The names in front of the email adresses are required to avoid confusion for mail programs (e.g. mail programs like to use the "from name" when replying to a message when the "reply-to name" is empty (despite of the email).

Original report: http://dev.mybb.com/issues/1720
Reporter: Yuri Taillandier
Status: New

Sama34:
I suppose we can check if the server is respecting the SPF policy before sending a e-mail. Or add a new setting if that is not possible.

NotoriousPyro:
![image](https://cloud.githubusercontent.com/assets/706132/8344077/3814d7b0-1ad4-11e5-92ba-8dd76786689b.png)

Example test email sent by myself to myself using the board. The from field is now filled correctly and the reply-to is also filled with the user's email.

Not sure if this breaks anything though, so be warned.

Evenprime:
I don't think it should be necessary to check if the domain that is used in the "from"-field uses SPF/is protected by DMARC. It is the responsibility of the administrator to use a forum reply-to adress that he/she has control over and therefore can set the correct SPF entries for that domain.

The only situation that I can think of where the from adress is not the forum adress, is the sending of mails from users to users (the initial reason for this issue report). This case should be fixed by passing the correct parameters to the mailing method. The from field should end up being the same adress that is used for all types of forum notifications, the "reply-to" adress should be that of the user that is sending the mail.

[...]

PR #1732, opened 2014-12-30
Update class_mailhandler.php

NotoriousPyro:
When a user changes their password, an email is received from the board administrator. This "functionality" exposes the administrator's email to anyone who registers.

Thus, this change allows a person to set ReturnEmail to donotreply@domain.com and AdminEmail to adminemail@domain.com, the emails from Contact Us, etc will still work fine and be delivered to the administrator, yet any other emails not specifically defined as From anyone will be sent via donotreply@domain.com.

For example, the original file does this:
* A user uses the Contact Us link, the email is delivered to AdminEmail received from the user's own email address.
* A user registers, the email is delivered to the user's own email address, received from **AdminEmail**.

These changes do:
* A user uses the Contact Us link, the email is delivered to AdminEmail received from the user's own email address.
* A user registers, the email is delivered to the user's own Email Address, received from **ReturnEmail**.

### DIFF
--- a/inc/class_mailhandler.php
+++ b/inc/class_mailhandler.php
@@ -122,12 +122,12 @@ class MailHandler
 
 			if($mybb->settings['mail_handler'] == 'smtp')
 			{
-				$this->from = $mybb->settings['adminemail'];
+				$this->from = $mybb->settings['returnemail'];
 			}
 			else
 			{
 				$this->from = '"'.$this->utf8_encode($mybb->settings['bbname']).'"';
-				$this->from .= " <{$mybb->settings['adminemail']}>";
+				$this->from .= " <{$mybb->settings['returnemail']}>";
 			}
 		}
 
--- a/inc/class_mailhandler.php
+++ b/inc/class_mailhandler.php
@@ -92,6 +92,23 @@ class MailHandler
 	public $parse_format = 'text';
 
 	/**
+	 * Selects between AdminEmail and ReturnEmail, dependant on if ReturnEmail is filled.
+	 */
+	function from_select()
+	{
+		global $mybb;
+		
+		if($mybb->settings['returnemail'])
+		{
+			return "returnemail";
+		}
+		else
+		{
+			return "adminemail";
+		}
+	}
+
+	/**
 	 * Builds the whole mail.
 	 * To be used by the different email classes later.
 	 *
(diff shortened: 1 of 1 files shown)


NotoriousPyro:
Forgot to leave a commit message for my last edit, I tidied up the code and reduced code-repeation.

Sama34:
I do feel it is now inconsistent but if it does works.

@mybb/developers @Destroy666x @JN-Jones

Destroy666x:
Not sure if it solves the problem yet, but llinking to a similar issue: https://github.com/mybb/mybb/issues/257

NotoriousPyro:
This was never designed to fix that issue, it was designed to fix the issue of mail being delivered inconsistently between senders.

inc/class_mailhandler.php
```diff
@@ -92,6 +92,23 @@ class MailHandler
 	public $parse_format = 'text';
 
 	/**
+	 * Selects between AdminEmail and ReturnEmail, dependant on if ReturnEmail is filled.
```
JN-Jones: This should get a php doc paramater:
`@return string`

inc/class_mailhandler.php
```diff
@@ -92,6 +92,23 @@ class MailHandler
...
+	 */
+	function from_select()
+	{
+		global $mybb;
+		
+		if($mybb->settings['returnemail'])
```
JN-Jones: Should be checked via `empty` or at least `trim`, just in case the user entered a space.

JN-Jones:
Except those two comments I also think that we shouldn't have a function which needs to be called like this: `$mybb->settings[$this->from_select]`. Instead that function should already return the correct email imho (and should be renamed to something like `get_from_email`).

euantorano:
Definitely agree with @JN-Jones. I'd much prefer a `MailHandler::get_from_email() -> string` method that returns the actual from address, rather than an array key name.

JN-Jones:
@NotoriousPyro

NotoriousPyro:
Requested changes added, not extensively tested so proceed with caution.

JN-Jones:
Can you also fix the Doc Block I've mentioned? Rest seems fine to me.

NotoriousPyro:
Done.

JN-Jones:
Thanks. @Destroy666x @euantorano if both of you are fine with this PR I'll merge it with the others.

Destroy666x:
The current form of the PR renders the `$from` parameter used in many places in MyBB and plugins completely useless...

What should be done:
- don't touch `$this->from` if it's passed to the function
- the initial issue which could be fixed here is caused by invalid `Reply-To` header: https://github.com/NotoriousPyro/mybb/blob/patch-1/inc/class_mailhandler.php#L299 which is based on `$this->return_email` not `$this->from`

NotoriousPyro:
Reverted.

NotoriousPyro:
Probably the best way to check if SPF is in use is to actually ask the administrator. It could be a toggled setting in the email settings, constant DNS lookups I costly, both to PHP and DNS servers.

JN-Jones:
IMHO #257 should be fixed seperatly anyways. Would like to merge this without any fix for it @Destroy666x @NotoriousPyro

Destroy666x:
Yes, as I said above a new PR would be better. And an issue should be correct for this one because there is none.

JN-Jones:
Ok, will create an issue later and merge this.

JN-Jones:
#2054

JN-Jones:
Thanks again for this PR!

Issue #1733, opened 2014-12-31
Update my_serialize not to support objects or resources


PirataNervo:
PHP is known for having issues with resources or objects serializing, especially when it comes to security, allowing for remote code execution.

This has come to our attention by http://community.mybb.com/user-90701.html and the following suggestion was made:
https://github.com/piwik/piwik/blob/4ab8054b8bdd3f8a1f831b0d1a25ec8c6c754563/libs/upgradephp/upgrade.php#L379

We should use Piwik's solution for my_serialize and my_unserialize.

PirataNervo:
Added functions. Only replacement that needs to be done now is serialize by my_unserialize.

PirataNervo:
This should be done :)

PirataNervo:
Merged.

Destroy666x:
Reopened due to @Stefan-ST 's comment in the PR. I also suspect that the current limits may not always work, 4096 character limit isn't that high.

Sama34:
The minimum value for `MAX_SERIALIZED_INPUT_LENGTH` in a fresh installation is `18998`.

Destroy666x:
I still think we should just get rid of the limits.

Stefan-ST:
Just removing the limit may not be the best solution. The custom function probably uses (a lot) more resources than the native PHP function. Hitting the memory limit just because of this is not acceptable.

Destroy666x:
So, should we only keep `my_unserialize()` for users inputs as you said in the PR? As long as someone is going to do all changes - I don't see why not.

PirataNervo:
Don't forget to check the possibilities of:
1. Inserting input A into the database for some field that is not serialized.
2. That field gets cached by some task later or when doing something else not directly related to the page where input A is saved - indirect input.
3. The attack would probably still be possible.

PR #1734, opened 2015-01-02
WIP #1733 Update my_serialize not to support objects or resources

PirataNervo:
#1733

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -2104,8 +2104,8 @@ function _safe_unserialize($str)
  */
 function my_unserialize($str)
 {
-	// ensure we use the byte count for strings even when strlen() is overloaded by mb_strlen()
-	if (function_exists('mb_internal_encoding') && (((int) ini_get('mbstring.func_overload')) & 2))
+	// Ensure we use the byte count for strings even when strlen() is overloaded by mb_strlen()
+	if(function_exists('mb_internal_encoding') && (((int)ini_get('mbstring.func_overload')) & 2))
 	{
 		$mbIntEnc = mb_internal_encoding();
 		mb_internal_encoding('ASCII');
@@ -2113,10 +2113,11 @@ function my_unserialize($str)
 
 	$out = _safe_unserialize($str);
 
-	if (isset($mbIntEnc))
+	if(isset($mbIntEnc))
 	{
 		mb_internal_encoding($mbIntEnc);
 	}
+	
 	return $out;
 }
 
@@ -2182,7 +2183,7 @@ function _safe_serialize( $value )
 function my_serialize( $value )
 {
 	// ensure we use the byte count for strings even when strlen() is overloaded by mb_strlen()
-	if(function_exists('mb_internal_encoding') && (((int) ini_get('mbstring.func_overload')) & 2))
+	if(function_exists('mb_internal_encoding') && (((int)ini_get('mbstring.func_overload')) & 2))
 	{
 		$mbIntEnc = mb_internal_encoding();
 		mb_internal_encoding('ASCII');
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1906,28 +1906,295 @@ function my_set_array_cookie($name, $id, $value, $expires="")
 	$mybb->cookies['mybb'][$name] = $newcookie;
 }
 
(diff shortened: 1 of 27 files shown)


inc/functions.php
```diff
@@ -1906,28 +1906,295 @@ function my_set_array_cookie($name, $id, $value, $expires="")
 	$mybb->cookies['mybb'][$name] = $newcookie;
 }
 
+/*
+ * Arbitrary limits for _safe_unserialize()
+ */
+define('MAX_SERIALIZED_INPUT_LENGTH', 4096);
+define('MAX_SERIALIZED_ARRAY_LENGTH', 256);
+define('MAX_SERIALIZED_ARRAY_DEPTH', 5);
```
Sama34: Shouldn't those be in `class_core.php`?

Destroy666x: @Sama34 , `my_unserialize()` is also used on scripts which don't include `inc/init.php`, for example installation. I think they're safer to use here or in the `_safe_unserialize()` function body.

But are these limits required for fixing the issue anyways?

inc/functions.php
```diff
@@ -1906,28 +1906,295 @@ function my_set_array_cookie($name, $id, $value, $expires="")
...
+ * Credits go to https://github.com/piwik
+ * Wrapper for _safe_unserialize() that handles exceptions and multibyte encoding issue
+ *
+ * @param string $str
+ * @return mixed
+ */
+function my_unserialize($str)
+{
+	// ensure we use the byte count for strings even when strlen() is overloaded by mb_strlen()
+	if (function_exists('mb_internal_encoding') && (((int) ini_get('mbstring.func_overload')) & 2))
```
Destroy666x: Should be changed to use our standards - no space between `if (` and after `(int)`  - also in other places.

Destroy666x:
Also, shouldn't all `serialize()` calls be changed to the new `my_serialize()`?

inc/functions.php
PirataNervo: @Destroy666x I left them there because on boards with small stack this may be a problem so reducing the array length and depth allowed will stop the stack from filling up to a max.

inc/functions.php
PirataNervo: Ah I thought I had changed all, sorry.

PirataNervo:
@Destroy666x yes see here: https://github.com/mybb/mybb/issues/1733#issuecomment-68552597

PirataNervo:
I haven't done the replacements yet because I want to make sure no one actually decided to serialize objects or resources in the core.

PirataNervo:
Alright fixed the issues related to the dev standards. Meanwhile I found that 3rdparty Diff has some serialize calls which seem to be arrays but I'll ignore those as they're 3rdparty and those are the only calls I found by 3rdparty libs.

PirataNervo:
I couldn't find any errors on my test board. All replacements are now in place.

inc/functions.php
Destroy666x: Oh ok.

PirataNervo:
Corrected :) there are conflicts now, hold on

PirataNervo:
I'm merging this now before it breaks again :D

PirataNervo:
If anyone finds any issues, a new PR should be opened :)

ATofighi:
:D I also think `@my_serialize` should replace to `my_serialize` (I'm not sure)

PirataNervo:
@ is only used where it may be possible to obtain warnings (e.g. when the input array is empty). The new function should not generate errors anymore anyway so they are not even needed where I left them :)

Stefan-ST:
@PirataNervo Sorry to say that after merging this but there is probably a major issue. The objects we store in the cache can be considerably larger than 4096 chars (current MAX_SERIALIZED_INPUT_LENGTH).
Although we could increase/remove the limit but the safe functions are probably still much slower.
I would have used the functions just for user inputs (GET, POST, COOKIE). Everything else should be safe in any case?

PirataNervo:
Which cases are those? Would be better to consider those special cases than differentiating between user input and software input (into the DB) IMO.

JN-Jones:
Likely the cache I guess?

Issue #1736, opened 2015-01-03
Use constants like MyBB::INPUT_INT instead of the raw value


JanMalte:

JanMalte:
Is this issue still valid? Or are they all replaced now and the issue can be closed?

ATofighi:
@JanMalte There are afew places that should replace, I'll make other changes.

PR #1737, opened 2015-01-03
changed raw values into constants #1736

JanMalte:


Destroy666x:
https://github.com/mybb/mybb/issues/1736

JanMalte:
Changed manually to make sure only the right values are changed.

PirataNervo:
That's quite a lot of work :P I went through many and didn't find any problems. Still, after reading so many my eyes are likely to miss any problems that may exist. We need more eyes here.

@mybb/developers @JN-Jones @Destroy666x @ATofighi

PirataNervo:
@demonate

Destroy666x:
@ATofighi, your script looks good, but just in case I'd also make the space between `,` and the number optional in case it was forgotten, for example `$mybb->get_input('something',1)` Same for possible space after `(int)`: `(int) $mybb->input['xxx']`

ATofighi:
I'll merge it and make other changes.

PR #1739, opened 2015-01-04
Fix #1728 last occurence of /install/

### LINKED ISSUE #1728: // TODO: Change this to find the last position of /install/
Found this while checking https://github.com/mybb/mybb/issues/1656:
https://github.com/mybb/mybb/blob/feature/install/index.php#L1872 - someone added the comment but never finished it. The fix is simple - `strrpos()` can be used instead of `my_strpos()`

Destroy666x:
Used `strrpos()` to find the last occurence of the folder and cleaned
some redundant stuff to simplify the code.
https://github.com/mybb/mybb/issues/1728

### DIFF
--- a/install/index.php
+++ b/install/index.php
@@ -188,7 +188,7 @@ function intro()
 	global $output, $mybb, $lang;
 
 	$output->print_header($lang->welcome, 'welcome');
-	if(strpos(strtolower($_SERVER['PHP_SELF']), "upload/") !== false)
+	if(strpos(strtolower(get_current_location()), '/upload/') !== false)
 	{
 		echo $lang->sprintf($lang->mybb_incorrect_folder);
 	}
@@ -1823,9 +1823,7 @@ EOF;
 	{
 		$bbname = 'Forums';
 		$cookiedomain = '';
-		$cookiepath = '/';
 		$websitename = 'Your Website';
-		$contactemail = '';
 
 		$protocol = "http://";
 		if((!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != "off"))
@@ -1847,7 +1845,7 @@ EOF;
 
 		if(substr($cookiedomain, 0, 4) == "www.")
 		{
-			$cookiedomain = my_substr($cookiedomain, 4);
+			$cookiedomain = substr($cookiedomain, 4);
 		}
 
 		// IP addresses and hostnames are not valid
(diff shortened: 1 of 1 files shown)


install/index.php
```diff
@@ -1847,7 +1845,7 @@ function revertSetting(defval, inpid)
 
 		if(substr($cookiedomain, 0, 4) == "www.")
 		{
-			$cookiedomain = my_substr($cookiedomain, 4);
+			$cookiedomain = substr($cookiedomain, 4);
```
PirataNervo: Why substr over my_substr?

install/index.php
```diff
@@ -1864,34 +1862,14 @@ function revertSetting(defval, inpid)
...
-		if($currentscript)
-		{
-			$bburl = my_substr($currentscript, 0, my_strpos($currentscript, '/install/'));
-		}
-
-		if($_SERVER['SERVER_ADMIN'])
-		{
-			$contactemail = $_SERVER['SERVER_ADMIN'];
-		}
+		$noinstall = substr($currentlocation, 0, strrpos($currentlocation, '/install/'));
```
PirataNervo: Again why substr over my_substr?

install/index.php
Destroy666x: The only reason to use `my_substr` over `substr` are multi-byte UTF-8 characters, but there is none of them in the `www.` substring, only basic ASCII. So using the slower MyBB alternative here is pointless.

install/index.php
Destroy666x: This part of URL can't contain multi-byte chars either. Anyways, I'll fix the possible `/install/` in query parameters here too.

PirataNervo: Actually it may, depending on the folder the user has put the install folder. I could put MyBB in folder téstèfõldër for example - unless I'm missing something.

install/index.php
PirataNervo: You're right but $cookiedomain may contain so it's probably better to use my_substr in the if condition.

Destroy666x: But the first 4 removed chars are always `www.` because of the conditional above. Doesn't matter if the rest of the domain contains special characters - they won't be affected by the `substr()` at all.

PirataNervo: I don't think that's how it works. When you specify that you want the first 4 characters, substr doesn't take into account that the length of multibyte characters is not one byte therefore the result may not be right. I certainly don't know how the inner workings of substr so I can't be sure about that. I don't think it hurts to use my_substr here though - unless of course you know for sure this is fine.

Destroy666x: You don't need to know the inners of `substr()`, just run this code after the $footer is eval'd in global.php:
```php
$footer .= substr('www.żółć♜.śń.pl', 4);
```
The result is what I'm expecting to see.

Anyways, if `my_substr()` should be used anywhere then in the conditional to make sure the string actually starts with `www.` because `$cookiepath` might start with `www` + a multi-byte character and then it might result in a false positive. So someone made it the other way round. I'll add it to the conditional.

PirataNervo: Sounds good then :)

install/index.php
PirataNervo: @Destroy666x don't miss this one

Destroy666x: Oh, forgot to answer here. Anyways: http://stackoverflow.com/questions/708017/can-a-php-file-name-or-a-dir-in-its-full-path-have-utf-8-characters?answertab=votes#tab-top - don't know how up-to-date it is, but multi-byte characters are not recommended for file/fiolder names and may break more things in MyBB.

And even if someone provides the folder you said, once again, you can test it:
```php
$footer .= substr('/téstèfõldër/install/index.php', 0, strrpos('/téstèfõldër/install/index.php', '/install/')).'/';
$footer .= '<br />';
$footer .= my_substr('/téstèfõldër/install/index.php', 0, strrpos('/téstèfõldër/install/index.php', '/install/')).'/';
```
Results in:
```php
/téstèfõldër/
/téstèfõldër/ins/
```
So my version is correct and using `my_substr() breaks it. Why? Because the last occurence's position is calculated without considering multi-byte characters and same goes for `substr()`, thus the output is good. `my_substr()` on the other hands counts the characters differently and won't work properly unless we define `my_strrpos()` too and use it instead of `strrpos()`.

PirataNervo: Alright, sounds good then ;)

PR #1741, opened 2015-01-04
Fix #1656 finish forcing cookie to ACP dir

### LINKED ISSUE #1656: Restrict ACP session cookie to /admin path (or other)
We should restrict the ACP session cookie to the /$adminfolder folder.

http://community.mybb.com/thread-162999.html

This way, if someone exploits an XSS vulnerability on the front end, they cannot retrieve the ACP session cookie.

Destroy666x:
Makes the forced cookies work with `/` cookiepath for a forum in a
subfolder (and any other combination).
https://github.com/mybb/mybb/issues/1656

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -36,7 +36,7 @@ require_once MYBB_ADMIN_DIR."inc/functions.php";
 require_once MYBB_ROOT."inc/functions_user.php";
 
 // Set cookie path to our admin dir temporarily, i.e. so that it affects the ACP only
-$loc = get_current_location();
+$loc = get_current_location('', '', true);
 $mybb->settings['cookiepath'] = substr($loc, 0, strrpos($loc, "/{$config['admin_dir']}/"))."/{$config['admin_dir']}/";
 
 if(!isset($cp_language))
--- a/admin/index.php
+++ b/admin/index.php
@@ -22,7 +22,7 @@ send_page_headers();
 
 if(!isset($config['admin_dir']) || !file_exists(MYBB_ROOT.$config['admin_dir']."/inc/class_page.php"))
 {
-	$config['admin_dir'] = "admin";
+	$config['admin_dir'] = basename(dirname(__FILE__));
 }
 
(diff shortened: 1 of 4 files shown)


ATofighi:
I think we also should remove `adminsid` and `acploginattempts` cookies from front-end dir.

PirataNervo:
@ATofighi by doing it in admin/index.php it should affect all created cookies unless configuration is re-loaded.

Destroy666x:
I think @ATofighi means unsetting the current cookies. I guess we could do that on the current upgrade. I'll add it if noone has anything against.

PirataNervo:
Ah! Sorry I may have misread it yesterday. We could do it in the upgrade but it's useless because they cannot be used (sessions only last 2h).

I don't like the current fix though. @Destroy666x if you visit this page for example:
http://localhost/mybb18/admin2/index.php?folder=/admin2/
You'd get `/mybb18/admin2/index.php?folder=/admin2/` as final path which is not right. I believe we should only get what's before .php (including the extension) and forget about the parameters.

ATofighi:
> You'd get /mybb18/admin2/index.php?folder=/admin2/ as final path which is not right.

I think we can use
```php
if(!empty($_SERVER['PATH_INFO']))
{
	$loc = htmlspecialchars_uni($_SERVER['PATH_INFO']);
}
elseif(!empty($_ENV['PATH_INFO']))
{
	$loc = htmlspecialchars_uni($_ENV['PATH_INFO']);
}
elseif(!empty($_ENV['PHP_SELF']))
{
	$loc = htmlspecialchars_uni($_ENV['PHP_SELF']);
}
else
{
	$loc = htmlspecialchars_uni($_SERVER['PHP_SELF']);
}
```
instead of 
```php
$loc = get_current_location();
```

PirataNervo:
Rule #1: never trust the intelligence of people!

ATofighi:
@Destroy666x

Destroy666x:
I'll fix all my remaining issues this weekend.

Destroy666x:
This should be finished.

ATofighi:
Looks good to me

Issue #1652, opened 2014-11-21
Base database class


Destroy666x:
Issue for https://github.com/mybb/mybb/pull/1602

PirataNervo:
Re-opening this as DB_Base is never loaded.

PR #1742, opened 2015-01-04
#1652 Load interface

PirataNervo:
#1652

### DIFF
--- a/inc/init.php
+++ b/inc/init.php
@@ -103,6 +103,9 @@ if(is_dir(MYBB_ROOT."install") && !file_exists(MYBB_ROOT."install/lock"))
 	$mybb->trigger_generic_error("install_directory");
 }
 
+// Load DB interface
+require_once MYBB_ROOT."inc/db_base.php";
+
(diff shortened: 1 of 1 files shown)


Destroy666x:
As I said in the other PR, there are few places which don't load inc/init.php. 
- installation
- upgrade: https://github.com/mybb/mybb/blob/feature/install/upgrade.php#L63
- and reloading the file may be needed here (not sure): https://github.com/mybb/mybb/blob/feature/inc/functions.php#L176

Sama34:
Yes, it should be loaded everywhere the DB class is initialized.

PirataNervo:
Done!

ATofighi:
Also https://github.com/PirataNervo/mybb/blob/featurefix-1652/install/index.php#L1255

PirataNervo:
Done :)

Destroy666x:
I think it can be included globally in the installation, otherwise here's a 3rd place that needs it: https://github.com/PirataNervo/mybb/blob/featurefix-1652/install/index.php#L1403

PirataNervo:
Yeah I'm starting to think so too...

PR #1743, opened 2015-01-04
Update captcha.php

dustwolf:
The debugging "test" condition is heavily exploited by spambots. Our website saw a 50% decrease in the number of spambots sucesfully registering after removing this code.

### DIFF
--- a/captcha.php
+++ b/captcha.php
@@ -26,11 +26,7 @@ $min_angle = -30;
 $max_angle = 30;
 
 $mybb->input['imagehash'] = $mybb->get_input('imagehash');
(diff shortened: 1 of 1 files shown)


Destroy666x:
Is it surely a debug message @mybb/developers @JN-Jones ? If yes, should be removed in 1.8.4.

JN-Jones:
I think it was added to make support a bit easier (check whether the captcha is properly generated). But I guess it can be removed without problems

Sama34:
I agree it should be removed.

Destroy666x:
https://github.com/mybb/mybb/issues/1744

PR #1745, opened 2015-01-05
Fixes #1736 -  Use constants like MyBB::INPUT_INT instead of the raw val...

### LINKED ISSUE #1736: Use constants like MyBB::INPUT_INT instead of the raw value

ATofighi:
...ue #1736 
https://gist.github.com/ATofighi/1e77e8001682e32891ae

### DIFF
--- a/admin/modules/config/profile_fields.php
+++ b/admin/modules/config/profile_fields.php
@@ -212,7 +212,7 @@ if($mybb->input['action'] == "add")
 	$selected_values = '';
 	if($mybb->input['viewableby'] != '' && $mybb->input['viewableby'] != -1)
 	{
-		$selected_values = explode(',', $mybb->get_input('viewableby', MyBB::INPUT_STRING));
+		$selected_values = explode(',', $mybb->get_input('viewableby'));
 
 		foreach($selected_values as &$value)
 		{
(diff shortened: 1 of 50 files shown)


Destroy666x:
I don't think it's still complete. At least your script didn't complete the `(int)$mybb->input` conversion, check for example admin/modules/user/groups.php

ATofighi:
@Destroy666x Fixed.
now I only don't change `(int)$mybb->input[....][....]`s.

modcp.php
```diff
@@ -1362,9 +1362,9 @@
 			'uid' => $mybb->user['uid'],
 			'subject' => $mybb->input['title'],
 			'message' => $mybb->input['message'],
-			'allowhtml' => (int)$mybb->input['allowhtml'],
-			'allowmycode' => (int)$mybb->input['allowmycode'],
-			'allowsmilies' => (int)$mybb->input['allowsmilies'],
```
ATofighi: I think it should be bool . :-\
@Destroy666x @mybb/developers

Destroy666x: I think it's ok as it is now.

admin/modules/config/profile_fields.php
```diff
@@ -212,7 +212,7 @@
 	$selected_values = '';
 	if($mybb->input['viewableby'] != '' && $mybb->input['viewableby'] != -1)
 	{
-		$selected_values = explode(',', (string)$mybb->input['viewableby']);
+		$selected_values = explode(',', $mybb->get_input('viewableby', MyBB::INPUT_STRING));
```
Sama34: There is no need for `MyBB::INPUT_STRING`.

ATofighi: you're right.
fixed.

PR #1711, opened 2014-12-21
Fix #1151 Add more hooks v3

Sama34:
Fix #1151 Add more hooks v3

(#66 [Feature request] Subscribe to thread alert)

Issue #1151, opened 2014-08-09
Add more hooks v3


Sama34:
Starting from https://github.com/mybb/mybb/issues/882#issuecomment-50003653

Destroy666x:
@Sama34 @mybb/developers What about the duplicate `admin_tools_mailerrors_prune_commit` hook? I'm quite sure it's a bug and should be changed in 1.8.4 - better now than later at least. Otherwise a developer may not notice it and end up with another bug in their plugin. Additionally, I remembered I found another duplicate some time ago - `admin_forum_management_deletemod`:
https://github.com/mybb/mybb/blob/feature/admin/modules/forum/management.php#L437 - deletemod hook for clearing all permissions
https://github.com/mybb/mybb/blob/feature/admin/modules/forum/management.php#L1877 - and deleting a mod

I also remember there were much bigger change for hooks in 1.6.5: http://community.mybb.com/thread-106729.html and some people complained about them. This is nothing compared to that - 2 rarely used hooks will only be affected.

Also, adding hooks to `viewthreadnotes`, `getpmip` and `getip` actions would be nice:
https://github.com/mybb/mybb/blob/feature/moderation.php#L1190
https://github.com/mybb/mybb/blob/feature/moderation.php#L1151
since all the other tools in moderation.php have hooks, no idea why these actions were the only ones omitted.

PirataNervo:
@Destroy666x yeah the one in the maillogs.php should admin_tools_maillogs_prune_commit

Sama34:
I'm kind of surprised about your decision to -_potentially_- break plugins but I agree so I added them to the PR. I also added some I do consider some not mentioned anywhere in this discussion.

PR #1747, opened 2015-01-08
Fix #307 User Pruning bug

### LINKED ISSUE #307: User Pruning bug
> When user pruning people with less than x posts, the user may in fact have the required amount of posts but they may not show up in the mybb_users postnum field due to the posts not counting in a certain forum, e.g. the "Testing" forum on MyBB.
> 
> An example would be pruning a user who has posted in the /me thread 50 times but no where else, maintaining a post count of 0. If the pruner would be run, they would be deleted.
> 
> I'm guessing this isn't intended, and probably should be looked into.

Original report: http://dev.mybb.com/issues/2139
Reporter: Paul H.
Status: Confirmed

Sama34:
Fix #307 User Pruning bug


inc/tasks/userpruning.php
```diff
@@ -41,11 +41,25 @@ function task_userpruning($task)
 			unset($in_usergroups[$key]);
 		}
 
-		$regdate = TIME_NOW-((int)$mybb->settings['dayspruneregistered']*24*60*60);
-		$query = $db->simple_select("users", "uid", "regdate <= ".(int)$regdate." AND postnum <= ".(int)$mybb->settings['prunepostcount']." AND usergroup IN(".$db->escape_string(implode(',', $in_usergroups)).")");
-		while($user = $db->fetch_array($query))
+		$prunepostcount = (int)$mybb->settings['prunepostcount'];
+
+		$regdate = (int)(TIME_NOW-((int)$mybb->settings['dayspruneregistered']*24*60*60));
+		$query = $db->simple_select('users', 'uid', "regdate<={$regdate} AND postnum<={$prunepostcount} AND usergroup IN({$db->escape_string(implode(',', $in_usergroups))})");
```
Stefan-ST: You can't call a function like that.

inc/tasks/userpruning.php
```diff
@@ -41,11 +41,25 @@ function task_userpruning($task)
 			unset($in_usergroups[$key]);
 		}
 
-		$regdate = TIME_NOW-((int)$mybb->settings['dayspruneregistered']*24*60*60);
-		$query = $db->simple_select("users", "uid", "regdate <= ".(int)$regdate." AND postnum <= ".(int)$mybb->settings['prunepostcount']." AND usergroup IN(".$db->escape_string(implode(',', $in_usergroups)).")");
-		while($user = $db->fetch_array($query))
+		$prunepostcount = (int)$mybb->settings['prunepostcount'];
+
+		$regdate = (int)(TIME_NOW-((int)$mybb->settings['dayspruneregistered']*24*60*60));
```
Stefan-ST: Why do you intval the result of a calculation? It will always be numeric.

inc/tasks/userpruning.php
Sama34: Well, I just did and it worked. Or are you saying I'm not allowed to?

Stefan-ST: Okay, it works (actually I never tried) but you should probably stick with common way which is IMHO more readable.

inc/tasks/userpruning.php
```diff
@@ -41,11 +41,25 @@ function task_userpruning($task)
 			unset($in_usergroups[$key]);
 		}
 
+		$prunepostcount = (int)$mybb->settings['prunepostcount'];
+
 		$regdate = TIME_NOW-((int)$mybb->settings['dayspruneregistered']*24*60*60);
-		$query = $db->simple_select("users", "uid", "regdate <= ".(int)$regdate." AND postnum <= ".(int)$mybb->settings['prunepostcount']." AND usergroup IN(".$db->escape_string(implode(',', $in_usergroups)).")");
-		while($user = $db->fetch_array($query))
+		$query = $db->simple_select('users', 'uid', "regdate<={$regdate} AND postnum<={$prunepostcount} AND usergroup IN({$db->escape_string(implode(',', $in_usergroups))})");
+		while($uid = (int)$db->fetch_field($query, 'uid'))
```
Destroy666x: Casting not needed, uid field in the databse has to be integer.

Sama34: Well, the results are used in the following query so I'm not going to remove this one.

Destroy666x: Not an issue (except increasing the file size and execution time a very tiny bit) but still doesn't make much sense to use it here. It's not an user input so the extra security layer is never needed in this case, unless someone hacks the database, changes the `uid` column to another type, injects malicious strings and then runs this query, which would be the weirdest attack ever.

inc/tasks/userpruning.php
```diff
@@ -41,11 +41,28 @@ function task_userpruning($task)
...
+		$prunepostcount = (int)$mybb->settings['prunepostcount'];
+
 		$regdate = TIME_NOW-((int)$mybb->settings['dayspruneregistered']*24*60*60);
-		$query = $db->simple_select("users", "uid", "regdate <= ".(int)$regdate." AND postnum <= ".(int)$mybb->settings['prunepostcount']." AND usergroup IN(".$db->escape_string(implode(',', $in_usergroups)).")");
-		while($user = $db->fetch_array($query))
+
+		$usergroups = $db->escape_string(implode(',', $in_usergroups));
+
+		$query = $db->simple_select('users', 'uid', "regdate<={$regdate} AND postnum<={$prunepostcount} AND usergroup IN({$usergroups})");
+		while($uid = (int)$db->fetch_field($query, 'uid'))
```
Destroy666x: > Well, the results are used in the following query so I'm not going to remove this one.
It is no real issue at all.

Not an issue (except increasing the file size and execution time a very tiny bit), but still doesn't make much sense to use it here. It's not an user input so the extra security layer is never needed in this case, unless someone hacks the database, changes the uid column to another type, injects malicious string to a record and then runs this query, which would be the weirdest attack ever.

inc/tasks/userpruning.php
Sama34: I'm going to avoid replying every one of your points and just remove it.

Sama34:
Now if somebody can confirm this does works and/or merge I would really appreciate. I would like to fix #1711 which is the last PR of mine and then fix my repository to jump to new issues.

Yeah I know this is tagged for 1.8.5 but I missed that when sending the PR.

Destroy666x:
What about the soft-deleted/unapproved posts and drafts?

Cu8eR:
This needs to be merged before #1151 right?

Sama34:
I have updated this to ignore any deleted/unapproved post but it certainly won't consider the status of the thread it self, so the user can still appear to have  zero posts.

I don't think adding thread status verification to be so necessary, but if you disagree please comment otherwise I would appreciate this being merged because as @Cu8eR mentioned I need this to fix my last PR.

Sama34:
I will get this merged if no one rejects.

Destroy666x:
A peeker should be added: https://github.com/mybb/mybb/blob/feature/admin/modules/config/settings.php#L1617

Except that it looks ok, but I haven't tested it.

Sama34:
That is the first conflict. The file differs from my own.

Sama34:
I have pushed directly.

e6853da2da3f70a12a4e7ea4235a31d64737b887

PR #1754, opened 2015-01-14
Fix #1753 ID is set even when empty.

### LINKED ISSUE #1753: ID is set even when empty.
Because of the following code:
```PHP
    function construct_cell($data, $extra=array())
    {
        foreach(array('class', 'style', 'id') as $field)
        {
            // Common-used fields
            if(!isset($extra[$field]))
            {
                $extra[$field] = '';
            }
        }

        $this->_cells[] = array("data" => $data, "extra" => $extra);
    } 
```

You get directly to the first element with no ID (id="") every time you enter an page that generated an cell with no ID.

Example:
`http://root/admin/index.php?module=style-themes&amp;action=edit&amp;tid=1`

There you go directly to the first sheet that is inherited from MyBB Master Style.

Original thread: [ID is set even when empty.](http://community.mybb.com/thread-161584.html)

Sama34:
Fix #1753 ID is set even when empty.

### DIFF
--- a/admin/inc/class_table.php
+++ b/admin/inc/class_table.php
@@ -36,15 +36,6 @@ class DefaultTable
 	 */
 	function construct_cell($data, $extra=array())
 	{
-		foreach(array('class', 'style') as $field)
-		{
-			// Common-used fields
-			if(!isset($extra[$field]))
-			{
-				$extra[$field] = '';
-			}
-		}
-
 		$this->_cells[] = array("data" => $data, "extra" => $extra);
 	}
 
@@ -58,7 +49,7 @@ class DefaultTable
 		$i = 1;
 		$cells = '';
 
-		foreach(array('class', 'style', 'id', 'rowspan', 'width') as $field)
+		foreach(array('class', 'style', 'rowspan', 'width') as $field)
 		{
 			// Common-used fields
 			if(!isset($extra[$field]))
@@ -258,7 +249,7 @@ class DefaultTable
 		foreach($this->_rows as $key => $table_row)
 		{
 			$table .= "\t\t<tr";
-			if($table_row['extra']['id'])
+			if(isset($table_row['extra']['id']))
 			{
 				$table .= " id=\"{$table_row['extra']['id']}\"";
 			}
--- a/admin/inc/class_table.php
+++ b/admin/inc/class_table.php
@@ -49,19 +49,16 @@ class DefaultTable
 		$i = 1;
 		$cells = '';
 
-		foreach(array('class', 'style', 'rowspan', 'width') as $field)
-		{
-			// Common-used fields
-			if(!isset($extra[$field]))
-			{
-				$extra[$field] = '';
-			}
-		}
-
 		// We construct individual cells here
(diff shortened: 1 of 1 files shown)


Destroy666x:
Is there any reason to set class/style to empty? I'd remove them too if not.

And what about: https://github.com/Sama34/mybb/blob/2617bc0ac0f7017596466aa8c58fa6c85ed4f735/admin/inc/class_table.php#L61

Sama34:
I removed the `id` key from that line but TBH I can't see where the rest is used. `class` is necessary (though may be applies differently) but the others...

Destroy666x:
> `class` is necessary 

Where is the empty class needed? Could you show the code?

Sama34:
https://github.com/Sama34/mybb/blob/2617bc0ac0f7017596466aa8c58fa6c85ed4f735/admin/inc/class_table.php#L71

Though really can be moved to be inside the foreach loop. I'm not sure why it was implemented like this either, I couldn't find mentions of 'rowspan', 'width', etc. anywhere so I didn't removed them.

Destroy666x:
There are `isset()` checks so there should be no problem with removal of the foreach loop, I think. And the loop you linked is for the cells made with `construct_cell()`: https://github.com/mybb/mybb/pull/1754/files#diff-e7ac6f5110e6cc267b6cb360669409c0R39 where you removed it.

Rows seem to be handled in other loop and `!empty()` could be used here:
https://github.com/Sama34/mybb/blob/2617bc0ac0f7017596466aa8c58fa6c85ed4f735/admin/inc/class_table.php#L261
https://github.com/Sama34/mybb/blob/2617bc0ac0f7017596466aa8c58fa6c85ed4f735/admin/inc/class_table.php#L279
to prevent warnings.

Sama34:
What about that?

Destroy666x:
But you're still trying to set the class to empty `class=""` if there is none specified, why? It doesn't matter that much, still no reason to output redundant attributes in my opinion.

Sama34:
Because of the following code:
```PHP
 			if($key == 0)
 			{
 				$cell['extra']['class'] .= " first";
 			}
 			elseif(!isset($this->_cells[$key+1]))
 			{
 				$cell['extra']['class'] .= " last";
 			}
 			if($i == 2)
 			{
 				$cell['extra']['class'] .= " alt_col";
 				$i = 0;
 			}
 			$i++;
			if($cell['extra']['class'])
 			{
 				$cells .= " class=\"".trim($cell['extra']['class'])."\"";
 			}
```

Note the `.=`, every if statement tries to use the variable even if it is not defined. If my understanding of PHP is correct, all variables should be defined before attempting to use them this way. MyBB is probably not going to complaint (it suppresses most errors) but that is the way it should be, IIRC.

So if that is the only issue you see I would really appreciate you merging this so I can jump to make new PRs (:

Destroy666x:
Never mind, I missed this condition, my bad: https://github.com/Sama34/mybb/blob/fix-1753/admin/inc/class_table.php#L273 I thought it's setting the `class=""` regardless of the empty value.

PR #1771, opened 2015-01-17
Fixes #1733 Update my_serialize not to support objects or resources

### LINKED ISSUE #1733: Update my_serialize not to support objects or resources

Stefan-ST:
Fixes #1733

- Slightly increased limit
- Cache uses native functions

### DIFF
--- a/admin/modules/tools/cache.php
+++ b/admin/modules/tools/cache.php
@@ -54,7 +54,7 @@ if($mybb->input['action'] == 'view')
 		admin_redirect("index.php?module=tools-cache");
 	}
 
-	$cachecontents = my_unserialize($cacheitem['cache']);
+	$cachecontents = unserialize($cacheitem['cache']);
 	if(empty($cachecontents))
 	{
 		$cachecontents = $lang->error_empty_cache;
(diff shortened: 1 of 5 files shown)


Destroy666x:
Will this surely fix every problem? There are still lots of `my_unserialize()` calls for non-user inputs such as admin permissions which have quite many options.

Stefan-ST:
@Destroy666x I can't guarantee that. However I think we are pretty safe as there is a lot of space left for plugins and other customizations.

PirataNervo:
What about input is cached?

Stefan-ST:
@PirataNervo There is not much we can do about that without slowing MyBB down. I don't think that's a serious issue tough since there are no unserializable objects in MyBB.

Destroy666x:
Tested it and no problems noticed so far, however, I'd at it at least to the admin permissions too, since they can be extended by plugins.

PaulBender:
Looks good to me as well

PR #1772, opened 2015-01-17
Fixes #1770 Remove sessions from upgrade script

### LINKED ISSUE #1770: Remove sessions from upgrade script
This has caused some issues when upgrading to 1.8.x:
http://community.mybb.com/thread-162865.html
http://community.mybb.com/thread-161647.html
http://community.mybb.com/thread-163795-post-1124167.html
...

Stefan-ST:
Fixes #1770

### DIFF
--- a/inc/class_session.php
+++ b/inc/class_session.php
@@ -32,7 +32,7 @@ class session
 		$this->useragent = $_SERVER['HTTP_USER_AGENT'];
 
 		// Attempt to find a session id in the cookies.
-		if(isset($mybb->cookies['sid']))
+		if(isset($mybb->cookies['sid']) && !defined('IN_UPGRADE'))
 		{
 			$sid = $db->escape_string($mybb->cookies['sid']);
 			// Load the session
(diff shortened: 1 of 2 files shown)


ATofighi:
I test it, it's don't work for me because `$mybb->user` isn't defined.

Stefan-ST:
@ATofighi Oh, that was the reason for using sessions...
Unfortunately that means we need to copy some code since the session system always runs queries.

ATofighi:
It still isn't OK :(
I see `You do not have permissions to run this process. You need administrator permissions to be able to run the upgrade procedure.` error when I login to my account in upgrade...

Stefan-ST:
@ATofighi Wasn't a good solution though it worked for me...

Changing the session system is probably easier.

Destroy666x:
Yeah, I think we should change the IP selection based on 1.4/1.6/1.8 version.

ATofighi:
Merged.

PR #1778, opened 2015-01-25
Apply DST based on the timestamp instead of the current date.

Evenprime:
Related thread on the forums: http://community.mybb.com/newreply.php?tid=162091

When "my_date" applies DST in "Automatically detect DST"-Mode (value 2), it doesn't check/respect the timestamps value at all and simply applies DST always/never, based on the current time. This is certainly wrong and potentially confusing, as it retroactively changes the displayed post timestamps (and other dates) whenever the current date switches to/from DST.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -344,7 +344,7 @@ function my_date($format, $stamp="", $offset="", $ty=1, $adodb=false)
 		if(isset($mybb->user['uid']) && $mybb->user['uid'] != 0 && array_key_exists("timezone", $mybb->user))
 		{
 			$offset = $mybb->user['timezone'];
-			$dstcorrection = $mybb->user['dst'];
+			$dstcorrection = $mybb->user['dstcorrection'];
 		}
 		elseif(defined("IN_ADMINCP"))
 		{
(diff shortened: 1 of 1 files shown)


Destroy666x:
https://github.com/mybb/mybb/issues/1781

inc/functions.php
```diff
@@ -357,8 +357,8 @@ function my_date($format, $stamp="", $offset="", $ty=1, $adodb=false)
 			$dstcorrection = $mybb->settings['dstcorrection'];
 		}
 
-		// If DST correction is enabled, add an additional hour to the timezone.
-		if($dstcorrection == 1)
+		// If DST correction is enabled and the date falls into DST, add an additional hour to the timezone.
+		if($dstcorrection == 1 || ($dstcorrection == 2 && date('I', $stamp) == 1))
```
Stefan-ST: Since there are several different rules regarding daylight saving time (and countries not having it at all), it don't thing date('I', $stamp) is reliable.

Evenprime: I guess you are right. Sadly there is no real way to reliable method to query DST for specific timezones/locations in PHP (that I am aware of). And if there were, it would require users to also tell the country that they live in. :/

So you can probably just close the pull request then and I'll just keep using this method for my forum, where I can rely on almost all guests being from the northern hemisphere.

Destroy666x: Wouldn't something like this work? http://stackoverflow.com/a/14423394/1708567 We set it to GMT everywhere.

Stefan-ST: Yes but we can't map our timezones (just +1, -5,...) to those timezones...

Destroy666x: Right.. So no solution for this or?

Stefan-ST: Probably something we should defer to 2.0.

Evenprime: Probably for the best. 2.0 needs a user setting for the country that would be considered for DST calculations at this point.

PR #1782, opened 2015-01-25
Fixes #1775 Soft delete bug

### LINKED ISSUE #1775: Soft delete bug
I disabled "Soft delete feature" in ACP -&gt; General configuration

But I can still soft delete threads and posts... I can see "Soft dete thread" in forum and when I use this, the thread is soft deleted

Is it a bug?

Original thread: [Soft delete bug](http://community.mybb.com/thread-165872.html)

PaulBender:
Fixes #1775 

Also cleared up some permission errors and moved hardcoded HTML soft delete/restore moderation options to templates.


install/resources/upgrade32.php
```diff
@@ -50,6 +50,7 @@ function upgrade32_dbchanges()
 	$db->delete_query("forumpermissions", "fid NOT IN(SELECT fid FROM {$db->table_prefix}forums)");
 
 	$db->update_query("settings", array('optionscode' => 'select\r\n0=No CAPTCHA\r\n1=MyBB Default CAPTCHA\r\n2=reCAPTCHA\r\n3=Are You a Human\r\n4=NoCAPTCHA reCAPTCHA'), "name='captchaimage'");
+	$db->update_query("settings", array('title' => 'Enable Soft Delete for Users', 'description' => 'If enabled, posts and threads deleted by users will be hidden and can be restored by moderators. Otherwise, these posts and threads will be deleted permanently.'), "name='soft_delete'");
```
Sama34: Doesn't the upgrade script updates this automatically?

JN-Jones: From what I've seen it should even update the optionscode in the line above. Though the code in the `sync_settings` function is a bit weird it should do it.

PaulBender: I've removed it.

Cu8eR:
@PaulBender - there is a conflict

PaulBender:
Fixed

Cu8eR:
Looks good

JN-Jones:
When a moderator has the permission to soft delete threads but not posts (may also happen with hard delete from a look at the code) the Quick Delete button shows but will throw an error as the codeblock from line 115 on checks only post permissions. However it's possible to (soft) delete the thread via the mod dropdown in the thread overview.

JN-Jones:
Rest seems to work, at least I haven't found any other things.

PaulBender:
Fixed

Destroy666x:
What about these?
https://github.com/PaulBender/mybb/blob/fix-1775/editpost.php#L292
https://github.com/PaulBender/mybb/blob/fix-1775/editpost.php#L347
Shouldn't they be changed too?

PaulBender:
They've been changed

Destroy666x:
I still think there are some inconsistencies for mod permissions like `canviewdeleted` (in search.php for example), but this PR looks good and can be merged, the rest can be fixed in 1.8.5.

PaulBender:
Fixed merge conflict.

PR #1786, opened 2015-01-26
Visual fix in postbit_gotopost - CSS

Cu8eR:
Delete underline which is useless. CSS si used.

@Sama34

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -1369,6 +1369,7 @@ a.button.report_user_button span {
 	padding-left: 13px; /* amount of padding needed for image to fully show */
 	margin-top: -3px;
 	border: none;
+	text-decoration: none !important;
 }
 
 .pollbar {


Destroy666x:
The HTML from your previous PR wasn't invalid (empty `<a>` tags are allowed, only `href` is required), but this seems to be the best solution: http://stackoverflow.com/questions/10504544/is-it-ok-to-have-an-empty-anchor-tag?answertab=votes#tab-top

Cu8eR:
I think that css solution is better than just delete space. Anyway both
solutions work fine. I prefer css

Destroy666x:
> I think that css solution is better than just delete space

You're missing the "because" part here. I don't see why adding an additional line to a file that needs to be downloaded by the visitor if it's not cached would be better than removing a redundant HTML code.

JN-Jones:
As both is valid I'd also prefer the HTML solution.

Cu8eR:
Ok, so I will make a new PR with html solution. :)

Cu8eR:
@Destroy666x @JN-Jones @Sama34  - I have made a new PR with HTML solution...

PR #1795, opened 2015-01-28
Fix localized date time input

Evenprime:
Related Thread on Forum: http://community.mybb.com/thread-165257.html

The target of this pull request was two-fold:

1. Respect the "$mybb->settings['timeformat']" value (set in Admin-CP) at all locations that display time or allow the input of time as text (and don't/can't use my_date for date/time formatting). This affects:
Admin-CP Forum Announcement creation/editing
Admin-CP User banning preset bantimes selectbox entries
Admin-CP Mass Mail creation/editing
Mod-CP Forum Announcement creation/editing
Delayed Moderation

2. Respect current user timezone and dst when entering date/time. This affects:
Admin-CP Mass Mail creation/editing
Mod-CP Forum Announcement creation/editing
Delayed Moderation (original reason for this pull request as mentioned in the corresponding thread)
(Admin-CP Forum Announcement creation was intentionally left "GMT" only, because there is a label that states that inputs are always interpreted as GMT.)
(The Calendar already offers to set the used timezone explicitely, therefore I didn't touch the logic there. I only shortened the time patterns at two  locations, because the hour/minute part that is generated isn't used at these locations at all.)

3. In addition minor related bugs are fixed:
Admin-CP Mass Mail added the current users timezone when saving data. If at all, the timezone data should have been subtracted at that point. This bug lead to strange behaviour, where repeatedly editing and saving the same mass-mail would push its delivery time further into the future (when timezone offset is positive) or the past (the other case).
Admin-CP Mass Mail would interpret "11:00 PM" as "11:00", because the matching for "pm" only checked for lowercase "pm". Everywhere else the same logic was case insensitive, now it is here too.
The function  get_event_date($event) in functions.php would have applied timezones twice, if the server has a timezone/location set that's not "+0", because mktime is used where gmmktime should've been used (The correct timezone information is applied one line later during a call to "my_date"). This method seems to not be used at all currently.

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -945,7 +945,7 @@ if($mybb->input['action'] == "editevent")
 		{
 			$privatecheck = '';
 		}
-		$start_date = explode("-", gmdate("j-n-Y-g:i A", $event['starttime']+$event['timezone']*3600));
+		$start_date = explode("-", gmdate("j-n-Y", $event['starttime']+$event['timezone']*3600));
 		$single_day = $start_date[0];
 		$single_month[$start_date[1]] = " selected=\"selected\"";
 		$single_year = $start_date[2];
@@ -962,7 +962,7 @@ if($mybb->input['action'] == "editevent")
 		}
 		if($event['endtime'])
 		{
-			$end_date = explode("-", gmdate("j-n-Y-g:i A", $event['endtime']+$event['timezone']*3600));
+			$end_date = explode("-", gmdate("j-n-Y", $event['endtime']+$event['timezone']*3600));
 			$end_day = $end_date[0];
 			$end_month[$end_date[1]] = " selected=\"selected\"";
 			$end_year = $end_date[2];
--- a/admin/modules/forum/announcements.php
+++ b/admin/modules/forum/announcements.php
@@ -230,14 +230,12 @@ if($mybb->input['action'] == "add")
 
 	if(!$mybb->input['starttime_time'])
 	{
-		$start_time = explode("-", gmdate("g-i-a", TIME_NOW));
-		$mybb->input['starttime_time'] = $start_time[0].":".$start_time[1]." ".$start_time[2];
+		$mybb->input['starttime_time'] = gmdate($mybb->settings['timeformat'], TIME_NOW);
 	}
 
 	if(!$mybb->input['endtime_time'])
 	{
-		$end_time = explode("-", gmdate("g-i-a", TIME_NOW));
-		$mybb->input['endtime_time'] = $end_time[0].":".$end_time[1]." ".$end_time[2];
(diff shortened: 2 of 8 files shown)


Evenprime:
Fix for #1796

JN-Jones:
Any reason you added the feedback label here @JoshHarmon? (Also note that those labels should be added to the issue not the PR ;))

JoshHarmon:
Was probably me interpreting the thing the label was for as something other than it really is. My interpretation was that feedback is for the verification/review/suggestions of/to a PR before merging.


On Jun 21, 2015, at 11:46 PM, Jones <notifications@github.com> wrote:

Any reason you added the feedback label here @JoshHarmon? (Also note that those labels should be added to the issue not the PR ;))

JN-Jones:
Here's a list of labels and how they should be used: http://community.mybb.com/thread-153813.html :P

calendar.php
```diff
@@ -899,7 +899,7 @@
 		{
 			$privatecheck = '';
 		}
-		$start_date = explode("-", gmdate("j-n-Y-g:i A", $event['starttime']+$event['timezone']*3600));
+		$start_date = explode("-", gmdate("j-n-Y", $event['starttime']+$event['timezone']*3600));
```
Destroy666x: Any reason time was completely removed instead of using the `timeformat` setting? Same applies to the line below.

JN-Jones: @Evenprime

Evenprime: The date string that is generated by the gmdate() at these locations is immediatly split at the '-' characters.

Of the resulting array only the first (day of month), second (month) and third (year) elements are used in the code afterwards. The last element of the array (time) isn't used anywhere, therefore I removed it in all of these locations.

JN-Jones: Yep, only `0`, `1` and `2` are accessed. @Destroy666x was there anything else or is this merge ready?

Destroy666x: I don't know, I haven't tested it and checked whether all time inputs are covered yet

JN-Jones:
Ok, whoever checks this then (don't have time to look at it atm) should also check whether the places where `gmdate` is used should probably use `my_date`. It's probably not needed for date select boxes but especially where time is used it should be rechecked.

Cu8eR:
@Evenprime - can you please solve conflicts? It seems to be fine, at least for me

dequeues:
This PR has merge conflicts that need to be resolved.This PR has merge conflicts that need to be resolved.

Evenprime:
Squashed and rebased changes, fixed merge conflicts

dequeues:
Just looked through and all looks fine to me. @Destroy666x see anything else before I merge?

Destroy666x:
If you tested (not only looked at) it and it works - I don't see why not.

PR #1800, opened 2015-01-31
Finish fixing #1453 smilies breaking

Destroy666x:
Changed `addslashes()` to `str_replace()` with only

Added missing `get_asset_url()` to some places (but still missing in
some other places in front-end, will create another issue for this) and missing `{theme}` replacement to ACP SCEditor smilies.

https://github.com/mybb/mybb/issues/1453


PaulBender:
Looks good to me

PR #1806, opened 2015-02-02
Fix #1325 add missing setting peekers

### LINKED ISSUE #1325: Lots of setting peekers missing
As the title says, nearly no setting in 1.8 got their peekers when they would be nice to have. I have a list of them, will push the fix when I come back.

Destroy666x:
Adds a possibilty to nest the peekers (so marking as an enhancement).

Fixes a bug where inputs without `id` attribute (like
groupselects/forumselects) couldn't have peekers for some unknown reason (`id` required but not used by/for anything).

Adds lots of missing settings peekers and groups them for having less
code to execute.

Moves some settings to other more proper groups (will be done
automatically by upgrade, right?).

https://github.com/mybb/mybb/issues/1325

### DIFF
--- a/admin/jscripts/peeker.js
+++ b/admin/jscripts/peeker.js
@@ -1,10 +1,11 @@
 /**
  * Peeker controls the visibility of an element based on the value of an input
  *
- * @example
- *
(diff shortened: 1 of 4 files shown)


PaulBender:
Looks good to me

PR #1808, opened 2015-02-05
Fix #349 Perfomance problems on large sites with user logins

### LINKED ISSUE #349: Perfomance problems on large sites with user logins
> Currently the username login is using <acronym title="">LOWER</acronym> in the SQL query. This means the index is not used on some Database backends.
> 
> Forums such as ours (150k+ users) slow down when spammers try to login with spambots all the time.
> 
> We suggest you remove the LOWER from the SQL so the index is used and it becomes fast again.
> 
> Its used in a number of places in the code so we would rather this was fixed upstream.

Original report: http://dev.mybb.com/issues/2290
Reporter: zag zag
Status: New

Sama34:
Fix #349 Perfomance problems on large sites with user logins


inc/functions_search.php
```diff
@@ -1035,10 +1076,10 @@ function perform_search_mysql($search)
 				$lang->error_minsearchlength = $lang->sprintf($lang->error_minsearchlength, $mybb->settings['minsearchword']);
 				error($lang->error_minsearchlength);
 			}
-			$subject_lookin = " AND LOWER(t.subject) LIKE '%{$keywords}%'";
+			$subject_lookin = " AND {$tfield} LIKE '%{$keywords}%'";
 			if($search['postthread'] == 1)
 			{
-				$message_lookin = " AND LOWER(p.message) LIKE '%{$keywords}%'";
+				$message_lookin = " AND{$pfield} LIKE '%{$keywords}%'";
```
Destroy666x: Missed a space

Sama34:
Surprisingly enough I was able to fix the conflict :O

PR #1809, opened 2015-02-05
Fix #1437 Copy/Paste Undo editor bug

### LINKED ISSUE #1437: Copy/Paste Undo editor bug
Copy and paste something, paste it in to your editor. Continue typing. Then control+Z back so the editor is empty. What you pasted will stay regardless of it's place. 

Steps to reproduce:

Type sentence.

Copy this post and paste it into your post.

Attempt to undo.

I've not tried any other browsers or OS.

Windows 7, Chrome Version 37.0.2062.103.

Thank you.

Original thread: [Copy/Paste Undo editor bug](http://community.mybb.com/thread-158872.html)

Sama34:
Fix #1437 Copy/Paste Undo editor bug


Destroy666x:
What about ACP? Should be loaded there too.

install/resources/mybb_theme.xml
```diff
@@ -13763,10 +13763,11 @@ if(use_xmlhttprequest == "1")
 		<template name="codebuttons" version="1804"><![CDATA[<link rel="stylesheet" href="{$mybb->asset_url}/jscripts/sceditor/editor_themes/{$theme['editortheme']}" type="text/css" media="all" />
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/sceditor/jquery.sceditor.bbcode.min.js"></script>
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/bbcodes_sceditor.js?ver=1804"></script>
+<script type="text/javascript" src="{$mybb->asset_url}/jscripts/sceditor/editor_plugins/undo.js?ver=1804"></script>
```
Destroy666x: I might be wrong, but I don't think it needs to be loaded this way. At least the bbcode plugin doesn't seem to.

jscripts/sceditor/editor_plugins/undo.js
```diff
@@ -0,0 +1,184 @@
+(function ($) {
```
Destroy666x: The file should probably be minified, at least other core SCEditor plugins are: https://github.com/mybb/mybb/blob/feature/jscripts/sceditor/editor_plugins/bbcode.js And, not connected with the PR, but format/xhtml plugin files can actually be removed btw since they're not used anywhere.

install/resources/mybb_theme.xml
Sama34: That is because a custom plugin file is loaded just one line above that we don't load the default file.

jscripts/sceditor/editor_plugins/undo.js
Sama34: How was that file compressed?

install/resources/mybb_theme.xml
Sama34: BTW, it didn't worked for me till I included the file in the code.

jscripts/sceditor/editor_plugins/undo.js
Destroy666x: http://jscompress.com/

install/resources/mybb_theme.xml
Destroy666x: Ok then.

Sama34:
Compressed. Not tested after it but should work nonetheless unless there is an issue with the compression tool itself.

Destroy666x:
Still has to be added to ACP though: https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L979

Sama34:
That should do it.

PaulBender:
I don't see anything wrong.

PR #1812, opened 2015-02-06
Fix #1811 upgrade issues

### LINKED ISSUE #1811: Upgrade Issues 1.8.3 -> 1.8.4
Updating Database.. PHP warning

Updating Database.. PHP warning

The solution seems to be to suppress the error in `inc/functions.php` line#1849

Clicking next shows me this:
Duplicated header?

Upgrade not so completed?
Upgrade not so completed?

Destroy666x:
Moved the cookie unset code to avoid headers already sent warnings.

Added forgotten `$plugins` globalization to the `upgradedone()` function
to prevent `add_hook()` error.

https://github.com/mybb/mybb/issues/1811

### DIFF
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -596,7 +596,7 @@ function buildcaches()
 
 function upgradedone()
 {
-	global $db, $output, $mybb, $lang, $config;
+	global $db, $output, $mybb, $lang, $config, $plugins;
 
 	ob_start();
 	$output->print_header("Upgrade Complete");
--- a/install/resources/upgrade32.php
+++ b/install/resources/upgrade32.php
@@ -23,6 +23,11 @@ $upgrade_detail = array(
 function upgrade32_dbchanges()
 {
 	global $db, $output;
+	
+	// Unset old ACP cookies from front-end since they're not needed anymore
+	my_unsetcookie('adminsid');
+	my_unsetcookie('acploginattempts');
+	my_unsetcookie('acp_view');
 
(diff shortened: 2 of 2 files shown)


install/upgrade.php
```diff
@@ -596,7 +596,7 @@ function buildcaches()
 
 function upgradedone()
 {
-	global $db, $output, $mybb, $lang, $config;
+	global $db, $output, $mybb, $lang, $config, $plugins;
```
Sama34: So that was it, lol :D

Sama34:
There is just one missing, the duplicated header which I couldn't find how was it caused.

Destroy666x:
Which duplicate header warning do you mean? The ones caused by `my_unsetcookies()` should be fixed (at least I didn't get them when I tested now), is there any other?

Sama34:
There are three issues mentioned in the issue, one of them is a duplicate header in the "Reverted Templates" page.

Destroy666x:
Oh that, missed it. I'll look into it now. You enabled the `revert_all_templates` option manually btw? Or was it done automatically?

Sama34:
I simply ran the upgrade script as normally, I think that is the expected message?

PaulBender:
Tested the fix, and everything worked fine. The third issue (the doubling of the Templates Reverted page) didn't occur for me. If no one else can reproduce it, this can be merged.

Destroy666x:
Forgot to reply, couldn't reproduce the 3rd issue either.

PR #1819, opened 2015-02-07
Fix #1818 users not removed from groupleaders table

### LINKED ISSUE #1818: Userhandler doesn't delete users from groupleaders table
If you delete an account of an user who's a group leader, the uid stays in the `groupleaders` table and the whole managegroup.php page becomes buggy (noone can't manage users, leaders not displayed correctly and maybe more). Check for more info: http://community.mybb.com/thread-166365-post-1137889.html#pid1137889

Marking as 1.8.4 since it would be nice to fix it ASAP and I'll submit a fix in few minutes.

Destroy666x:
Fixes a possible buggy behavior after user deletion caused by the lack
of record deletion from `groupleaders` table when the whole account is
pruned.
https://github.com/mybb/mybb/issues/1818

Also removed some other forgotten stuff (3 things are still missing but
I'll open a new issue for that).

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -1509,7 +1509,7 @@ class UserDataHandler extends DataHandler
 	}
 
 	/**
-	 * Provides a method to delete an users content
+	 * Provides a method to delete users' content
 	 *
 	 * @param array Array of user ids, false if they're already set (eg when using the delete_user function)
 	 */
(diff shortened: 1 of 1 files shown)


PaulBender:
Looks good to me.

inc/datahandlers/user.php
```diff
@@ -1536,17 +1536,24 @@ function delete_content($delete_uids=false)
 		$plugins->run_hooks('datahandler_user_delete_content', $this);
 
 		$db->delete_query('userfields', "ufid IN({$this->delete_uids})");
-		$db->delete_query('privatemessages', "uid IN({$this->delete_uids})");
+		$db->delete_query('privatemessages', "uid IN({$this->delete_uids}) OR fromid IN({$this->delete_uids})");
```
Stefan-ST: @Destroy666x While purge spammer should delete PMs other users received this method is also used by the user pruning task where PMs shouldn't be deleted.

Destroy666x: @Stefan-ST, why not? The function says to delete user's content and outgoing PM's belong to that too.

Stefan-ST: @Destroy666x If you delete inactive users you may want to keep the PMs they sent when they were active. It doesn't delete posts and threads.

Destroy666x: Maybe. But where should it be moved then? `delete_posts()` would be the more proper alternative perhaps since it takes care of posts/threads but the name of the function might be confusing.

Stefan-ST: Perhaps we can add another function? Updating the counters requires more code anyway...

Destroy666x: Ok. Added it here then: https://github.com/mybb/mybb/issues/1820 and will revert the OR.

PR #1824, opened 2015-02-08
Fix #1773

### LINKED ISSUE #1773: Avatars pushed to CDN are not resized
http://community.mybb.com/thread-165704.html

This is likely an issue with order of processing. Will have to look into the cause.

euantorano:
Copy scaled avatar image to CDN.

### DIFF
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -267,6 +267,8 @@ function upload_avatar($avatar=array(), $uid=0)
 				}
 				else
 				{
+					// Copy scaled image to CDN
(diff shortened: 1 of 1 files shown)


Sama34:
Can somebody with an CDN up confirm this does works?

Cu8eR:
I think this should be in 1.8.4 - this "bug" has a side effect on site´s performance

Sama34:
I suppose we should trust @euantorano on this one since he seems to be the only one with the proper setup for this kind of tests.

Cu8eR:
I agree

PaulBender:
Looks good.

PR #1825, opened 2015-02-09
Fixes #1767

### LINKED ISSUE #1767: ACP attachments have wrong input values after an error
If you enter conditions which will never happen and try to find attachments, after the page refresh the date will be changed to UNIX time and the KB size to bytes. Old values should be kept.

Devilshakerz:
Removed overwriting `$mybb->input` values

### DIFF
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -669,9 +669,9 @@ if(!$mybb->input['action'])
 
 		// LESS THAN or GREATER THAN
 		$direction_fields = array(
-			"dateuploaded" => $mybb->get_input('dateuploaded', 1),
-			"filesize"     => $mybb->get_input('filesize', 1),
-			"downloads"    => $mybb->get_input('downloads', 1)
+			"dateuploaded" => $mybb->get_input('dateuploaded', MyBB::INPUT_INT),
+			"filesize"     => $mybb->get_input('filesize', MyBB::INPUT_INT),
+			"downloads"    => $mybb->get_input('downloads', MyBB::INPUT_INT)
 		);
 
 		if($mybb->input['dateuploaded'] && $mybb->request_method == "post")
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -680,7 +680,7 @@ if(!$mybb->input['action'])
 		}
 		if($mybb->input['filesize'] && $mybb->request_method == "post")
 		{
-			$direction_fields['filesize'] = $direction_fields['filesize']*102;
+			$direction_fields['filesize'] *= 1024;
 		}
 
 		foreach($direction_fields as $field_name => $field_content)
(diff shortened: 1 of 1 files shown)


Destroy666x:
Looks ok, but our 1.x coding standards don't allow ternary for some weird reason, so please change it to conditonals.

Sama34:
I think you need to use `get_input` here and cast to integer instead of using `escape_string`. Curious about why did you choose `false` instead of `0` too.

Sama34:
Is there a existing issue for this BTW?

Devilshakerz:
Yes, it takes care of #1767. I was rebuilding the logic only so the original input data is kept; it doesn't really matter what the value would be if the input wasn't provided.

Since ternary operators are not allowed, the `$direction_fields` array would be defined with these 3 values with `$mybb->get_input('name', 1)` first, and the two `if`s would be restored but would overwrite the assigned values instead of modifying `$mybb->input` directly. Type casting or escaping would not be necessary since the values would be already filtered by `MyBB::get_input()`.

admin/modules/forum/attachments.php
```diff
@@ -668,20 +668,23 @@ function() {
...
-		if($mybb->input['dateuploaded'] && $mybb->request_method == "post")
-		{
-			$mybb->input['dateuploaded'] = TIME_NOW-$mybb->input['dateuploaded']*60*60*24;
+		$direction_fields = array(
+			"dateuploaded" => $mybb->get_input('dateuploaded', 1),
+			"filesize"     => $mybb->get_input('filesize', 1),
+			"downloads"    => $mybb->get_input('downloads', 1),
+		);
+
+		if ($mybb->input['dateuploaded'] && $mybb->request_method == "post") {
```
Destroy666x: This and a similar line below don't comply with the standards yet - `{` should be in a new line and no space between `if` and `(`.

admin/modules/forum/attachments.php
```diff
@@ -668,20 +668,25 @@ function() {
...
+
 		if($mybb->input['dateuploaded'] && $mybb->request_method == "post")
 		{
-			$mybb->input['dateuploaded'] = TIME_NOW-$mybb->input['dateuploaded']*60*60*24;
+			$direction_fields['dateuploaded'] = TIME_NOW-$direction_fields['dateuploaded']*60*60*24;
 		}
 		if($mybb->input['filesize'] && $mybb->request_method == "post")
 		{
-			$mybb->input['filesize'] *= 1024;
+			$direction_fields['filesize'] = $direction_fields['filesize']*102;
```
Destroy666x: Missed `4` at the end and the old shorter `*=` could be used.

admin/modules/forum/attachments.php
```diff
@@ -668,20 +668,25 @@ function() {
 		}
 
 		// LESS THAN or GREATER THAN
+		$direction_fields = array(
+			"dateuploaded" => $mybb->get_input('dateuploaded', 1),
```
Destroy666x: Oh, forgot to mention one last little thing.. - we decided to go with constants for the 2nd parameter: https://github.com/mybb/mybb/issues/1736 and they're used everywhere, so should be here too. Then it can be merged since there are no other issues.

PR #1833, opened 2015-02-14
Fix #1830 installation/upgrade error

### LINKED ISSUE #1830: Another upgrade error
After merging the copy to CDN function PR there is another error on upgrade: http://community.mybb.com/thread-164693-post-1139166.html#pid1139166

`$plugins` should be initlialized and globalized here:
https://github.com/mybb/mybb/blob/feature/install/upgrade.php#L408
And also in the following files: index.php (installation), upgrade12.php, upgrade17.php, upgrade30.php

I'll submit the fix later today.

Destroy666x:
Fix installation/error caused by `$plugins->run_hook()` usage without
access to `$plugins` by checking if the object exists. I think it's
better than defining `$plugins` in several files.

https://github.com/mybb/mybb/issues/1830

### DIFF
--- a/admin/inc/functions_themes.php
+++ b/admin/inc/functions_themes.php
@@ -965,8 +965,11 @@ function update_theme_stylesheet_list($tid, $theme = false, $update_disporders =
 				}
 			}
 		}
-
-		$plugins->run_hooks('update_theme_stylesheet_list_set_css_url', $css_url);
+		
+		if(is_object($plugins))
+		{
+			$plugins->run_hooks('update_theme_stylesheet_list_set_css_url', $css_url);
+		}
 
 		$attachedto = $stylesheet['attachedto'];
 		if(!$attachedto)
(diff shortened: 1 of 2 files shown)


Sama34:
Seems fine to me. But future issues will not be handled by this. Shouldn't we just initialize the plugin system in the upgrade process?

@mybb/developers @JN-Jones @Stefan-ST

Destroy666x:
Well, we can initialize it globally on install/upgrade (not sure if we should though, I'll wait for other opinions), but this should stay since I noticed that the functions are also used in the `recachestylesheets` task.

Sama34:
Yes but is_object($plugins) is just an workaround IMO.

Cu8eR:
Maybe merge this and open a new issue which will be fixed for MyBB 1.8.5

Stefan-ST:
Merging this to get 1.8.4 ready.

Cu8eR:
@Stefan-ST - please do not forget to open a new issue for ideas above. 

Have a nice day, I am looking for 1.8.4

PR #1850, opened 2015-02-18
Fix #1760 Deprecate update_password

### LINKED ISSUE #1760: Deprecate update_password
> MyBB has update_password in ./inc/functions_user.php, which is only used to update a user's password when it is reset.

> This should be deprecated and the user datahandler should be used instead, as is standard elsewhere in MyBB where user fields are updated.

http://community.mybb.com/thread-165506.html

Sama34:
Fix #1760 Deprecate update_password


Sama34:
@tommm I'm unsure if this is how you envisioned it. I'm unsure how to handle the deprecated funcion (just added a comment) since the parameters passed aren't the same the data-handler needs.

euantorano:
@Sama34 probably better to use the [@Depracated](http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.deprecated.pkg.html) PHPDoc tag...

Sama34:
There you have it.

member.php
```diff
@@ -1599,8 +1600,29 @@
...
+		$userhandler->set_data(array(
+			'uid'		=> $user['uid'],
+			'username'	=> $user['username'],
+			'email'		=> $user['email'],
+			'password'	=> random_str($password_length)
+		));
+
+		if(!$userhandler->validate_user())
+		{
+			error($lang->error_unknownerror);
```
Destroy666x: Why "Unknown error."? This won't help the user with anything. `get_friendly_errors()` could be used. Imploded with `<br />` if you don't want to switch to `inline_error()`.

member.php
```diff
@@ -1599,8 +1600,29 @@
...
 
-		$password = random_str($password_length);
-		$logindetails = update_password($user['uid'], md5($password), $user['salt']);
+		// Set up user handler.
+		require_once 'inc/datahandlers/user.php';
+		$userhandler = new UserDataHandler('update');
+
+		$userhandler->set_data(array(
+			'uid'		=> $user['uid'],
+			'username'	=> $user['username'],
```
Destroy666x: Is username verification required on a password change?

member.php
```diff
@@ -1599,8 +1600,29 @@
...
+			'email'		=> $user['email'],
+			'password'	=> random_str($password_length)
+		));
+
+		if(!$userhandler->validate_user())
```
Destroy666x: I don't think the whole `validate_user()` needs to run, only `verify_password()` is required because nothing apart from password is changed.

member.php
Sama34: > This won't help the user

There shouldn't be such error anyways.

Destroy666x: With the previous version of code they were likely (because username and emails were verified too). Now they're very unlikely because only this can happen after the complex password issue will be fixed: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L182

Anyways, now that you display the error `error($lang->error_unknownerror);` and the lang string can be removed because it will never happen (false is only returned with an error).

And the indendation is icorrent (spaces instead of tabs).

Sama34: > false is always returned with an error

false != text message.

> This won't help the user with anything.

Destroy666x: Yes, but I mean that the friendly errors and `false` are always returned in pairs: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L193-194

So if `verify_password()` returns false, `$errors[0]` will be always set and `error($errors[0]);` executed, thus `error($lang->error_unknownerror);` will be never used.

member.php
```diff
@@ -1599,8 +1600,35 @@
...
+		if(!$userhandler->verify_password())
+		{
+			$errors = $userhandler->get_friendly_errors();
+			if(!empty($errors[0]))
+			{
+				error($errors[0]);
+			}
+
+			error($lang->error_unknownerror);
+		}
```
ATofighi: I think we don't need verify a random password!

JoshH99: Technically, we'd shave off like a millisecond by not verifying it, but from a logical standpoint, I personally like the idea of keeping the check to avoid an odd edge case.

Sama34: If the fails fails will the user still be able to activate their account @ATofighi ?

inc/functions.php
```diff
@@ -5156,7 +5156,7 @@ function convert_through_utf8($str, $to=true)
 }
 
 /**
- * DEPRECATED! Please use other alternatives.
+ * @deprecated deprecated since version 1.8.4 Please use other alternatives.
```
JN-Jones: Should be `1.8.6` now. Also note that this is an invalid doc block as per #2026 and will be changed in that PR.

inc/functions_user.php
```diff
@@ -123,12 +123,7 @@ function validate_password_from_uid($uid, $password, $user = array())
 }
 
 /**
- * Updates a user's password.
- *
- * @param int The user's id.
- * @param string The md5()'ed password.
- * @param string (Optional) The salt of the user.
- * @return array The new password.
+ * @deprecated deprecated since version 1.8.5 Please use other alternatives.
```
JN-Jones: Though it's deprecated the rest shouldn't be removed imho. Also `1.8.6` would be correct here too.

Sama34:
What would the correct deprecate notice be like?

JN-Jones:
The deprecate notice itself is correct. But IMHO the "real" doc block which describes the function (params, return) should still be in place.

inc/functions.php
Sama34: > this is an invalid doc block

I meant about this comment specially. What would the correct doc block be then?

JN-Jones: https://github.com/mybb/mybb/pull/2027/files#diff-7fe6e55397c77ab9a0f5d57bc4cbe5b9R5198

The function still has a parameter and returns something

JN-Jones:

Sama34:
Should be ok now.

JN-Jones:
Seems fine to me @Destroy666x @euantorano

member.php
```diff
@@ -1601,8 +1602,35 @@
...
-		$logindetails = update_password($user['uid'], md5($password), $user['salt']);
+		// Set up user handler.
+		require_once 'inc/datahandlers/user.php';
+		$userhandler = new UserDataHandler('update');
+
+		$userhandler->set_data(array(
+			'uid'		=> $user['uid'],
+			'username'	=> $user['username'],
+			'email'		=> $user['email'],
+			'password'	=> random_str($password_length)
```
Destroy666x: Should have the complex password setting as the 2nd parameter.

JN-Jones: Should also fix the other issues from #1868 (eg the password length above). Good catch.

member.php
Destroy666x: I guess a better solution than throwing an error would be simply looping the password verification until it generates a proper password. Something like:
```php
while(!$userhandler->verify_password())
{
		$userhandler->set_data(array(
			'uid'		=> $user['uid'],
			'username'	=> $user['username'],
			'email'		=> $user['email'],
			'password'	=> random_str($password_length)
		));
}
```
Then surely no problems will be caused and there won't be very rare edge cases either.

member.php
Sama34: I don't think those are the same lines, that is why. Otherwise GH handles it correctly or causes a conflict.

So those are the only lines I should "replicate"?
https://github.com/mybb/mybb/pull/1868/files#diff-c6480c1a3fe8c1a339ac23fcdc05d97dR132
https://github.com/mybb/mybb/pull/1868/files#diff-c6480c1a3fe8c1a339ac23fcdc05d97dR135

Destroy666x: Yes

JN-Jones:
@Sama34 Can you fix those lines? Would like to merge this ;)

Sama34:
Done.

JN-Jones:
Looks fine too. Potential merge conflicts etc should be rechecked though after this has been merged. @euantorano @Destroy666x

member.php
Sama34: I missed this. @JN-Jones @euantorano

JN-Jones: It could get in an endless loop but it's very very unlikely. I wouldn't consider that case tbh.

Destroy666x: Why would it generate a password which would match the username or Email all the time? That's the only edge case left: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L190 after the latest commit. I don't think that chances for even more than 1 loop are more than 0.000000000001%.

JN-Jones: @Sama34 Can you change this loop?

Sama34:
@JN-Jones What about now?

JN-Jones:
Can you remove the unknown error language string? It's not used anymore. Apart from that it looks fine.

Sama34:
Done.

JN-Jones:
Looks fine now @Destroy666x @euantorano

member.php
```diff
@@ -1598,11 +1599,30 @@
...
 		if($password_length < 8)
 		{
-			$password_length = 8;
+			$password_length = min(8, (int)$mybb->settings['maxpasswordlength']);
 		}
 
-		$password = random_str($password_length);
-		$logindetails = update_password($user['uid'], md5($password), $user['salt']);
+		// Set up user handler.
+		require_once 'inc/datahandlers/user.php';
```
Destroy666x: Shouldn't there be `MYBB_ROOT` concantenation before the rest of the path to follow the standard?

JN-Jones: Shouldn't matter here, but as we use it everywhere else too, it should be changed. @Sama34

Sama34:
There it is.

JN-Jones:
Would be nice to get this in 1.8.6 @Destroy666x @euantorano

Destroy666x:
I guess noone tested it yet (I just did) because there's one more serious issue - `update_user()` requires `get_validated()` to pass, which requires `set_validated(true)`. Adding `$userhandler->set_validated(true);` right before updating the user should be fine.

JN-Jones:
@Sama34 Can you take a look at those issues?

Cu8eR:
@Sama34

Sama34:
I think I god everything @Destroy666x mentioned.

euantorano:
Yep, looks good to me now, can you confirm @Destroy666x? There's nothing obvious jumping out at me.

Sama34:
I will merge this later if nobody contributes new issues.

PR #1851, opened 2015-02-18
Fix #1752 Bypassing Theme Permissions

### LINKED ISSUE #1752: Bypassing Theme Permissions
This isn't particularly critical but I figured it was still worth bringing up.

We have a Christmas theme that we enabled for the month of December as our default and then disabled it (by clearing the Allowed User Groups section and forcing anyone currently set to it manually back to the default) on January 1. A couple of days ago, I noticed one user inexplicably back on that theme and while he claims that it's likely a cache issue on his end still allowing him to select it, I started doing a little experimenting and found that, if you know the theme's ID, you can arbitrarily switch to any theme by editing the ID into an available theme in the User CP or quick select box (e.g. via inspect element or something like Firebug).

Obviously, this requires knowing the correct theme ID but it's trivial to bypass theme permissions once you do. A simple check when setting the theme as well as an automatic reset to default when a user group is removed from the allowed list would likely fix this but I'll leave that end up to you guys.

Thanks for reading and let me know if you need any additional details!

Original thread: [Bypassing Theme Permissions](http://community.mybb.com/thread-165419.html)

Sama34:
Fix #1752 Bypassing Theme Permissions

### DIFF
--- a/global.php
+++ b/global.php
@@ -241,18 +241,37 @@ if(empty($loadstyle))
 }
 
 // Fetch the theme to load from the cache
+if($loadstyle != "def='1'")
+{
+	$query = $db->simple_select('themes', 'name, tid, properties, stylesheets, allowedgroups', $loadstyle, array('limit' => 1));
+	$theme = $db->fetch_array($query);
+
+	if(isset($theme['tid']) && !$load_from_forum && !is_member($theme['allowedgroups']) && $theme['allowedgroups'] != 'all')
+	{
+		if($load_from_user == 1)
+		{
+			$db->update_query('users', array('style' => 0), "style = '{$mybb->user['style']}'");
+		}
+
+		if(isset($mybb->cookies['mybbtheme']))
+		{
+			my_unsetcookie('mybbtheme');
+		}
+
+		$loadstyle = "def='1'";
+	}
+}
+
 if($loadstyle == "def='1'")
 {
 	if(!$cache->read('default_theme'))
 	{
 		$cache->update_default_theme();
 	}
-	$theme = $cache->read('default_theme');
-}
-else
-{
-	$query = $db->simple_select('themes', 'name, tid, properties, stylesheets', $loadstyle, array('limit' => 1));
-	$theme = $db->fetch_array($query);
+
+	$theme = $cache->read('default_theme', true);
+
+	$load_from_forum = $load_from_user = 0;
 }
 
 // No theme was found - we attempt to load the master or any other theme
@@ -378,39 +397,39 @@ if(my_substr($theme['imgdir'], 0, 7) == 'http://' || my_substr($theme['imgdir'],
 }
 else
 {
-    $img_directory = $theme['imgdir'];
+	$img_directory = $theme['imgdir'];
 
-    if($mybb->settings['usecdn'] && !empty($mybb->settings['cdnpath']))
-    {
-        $img_directory = rtrim($mybb->settings['cdnpath'], '/') . '/' . ltrim($theme['imgdir'], '/');
-    }
+	if($mybb->settings['usecdn'] && !empty($mybb->settings['cdnpath']))
+	{
+		$img_directory = rtrim($mybb->settings['cdnpath'], '/') . '/' . ltrim($theme['imgdir'], '/');
+	}
 
-    if(!@is_dir($img_directory))
-    {
-        $theme['imgdir'] = 'images';
-    }
+	if(!@is_dir($img_directory))
+	{
+		$theme['imgdir'] = 'images';
+	}
 
-    // If a language directory for the current language exists within the theme - we use it
-    if(!empty($mybb->user['language']) && is_dir($img_directory.'/'.$mybb->user['language']))
-    {
-        $theme['imglangdir'] = $theme['imgdir'].'/'.$mybb->user['language'];
-    }
-    else
-    {
-        // Check if a custom language directory exists for this theme
-        if(is_dir($img_directory.'/'.$mybb->settings['bblanguage']))
-        {
-            $theme['imglangdir'] = $theme['imgdir'].'/'.$mybb->settings['bblanguage'];
-        }
-        // Otherwise, the image language directory is the same as the language directory for the theme
-        else
-        {
-            $theme['imglangdir'] = $theme['imgdir'];
-        }
-    }
+	// If a language directory for the current language exists within the theme - we use it
+	if(!empty($mybb->user['language']) && is_dir($img_directory.'/'.$mybb->user['language']))
+	{
+		$theme['imglangdir'] = $theme['imgdir'].'/'.$mybb->user['language'];
+	}
+	else
+	{
+		// Check if a custom language directory exists for this theme
+		if(is_dir($img_directory.'/'.$mybb->settings['bblanguage']))
+		{
+			$theme['imglangdir'] = $theme['imgdir'].'/'.$mybb->settings['bblanguage'];
+		}
+		// Otherwise, the image language directory is the same as the language directory for the theme
+		else
+		{
+			$theme['imglangdir'] = $theme['imgdir'];
+		}
+	}
 
-    $theme['imgdir'] = $mybb->get_asset_url($theme['imgdir']);
-    $theme['imglangdir'] = $mybb->get_asset_url($theme['imglangdir']);
+	$theme['imgdir'] = $mybb->get_asset_url($theme['imgdir']);
+	$theme['imglangdir'] = $mybb->get_asset_url($theme['imglangdir']);
 }
 
 // Theme logo - is it a relative URL to the forum root? Append bburl
@@ -606,7 +625,7 @@ if($mybb->usergroup['cancp'] == 1 || ($mybb->user['ismoderator'] && $mybb->userg
 	{
 		$can_access_moderationqueue = false;
 	}
-	
+
 	if($can_access_moderationqueue || ($mybb->user['ismoderator'] && $mybb->usergroup['canmodcp'] == 1 && $mybb->usergroup['canmanagereportedcontent'] == 1))
 	{
 		// Read the reported content cache
@@ -764,7 +783,7 @@ if($mybb->settings['awactialert'] == 1 && $mybb->usergroup['cancp'] == 1)
 	{
 		$awaitingusers = my_number_format($awaitingusers);
 	}
-	
+
 	if($awaitingusers > 0)
 	{
 		if($awaitingusers == 1)
@@ -892,13 +911,13 @@ $archive_url = build_archive_link();
 if(is_banned_ip($session->ipaddress, true))
 {
 	if($mybb->user['uid'])
-    {
+	{
 		$db->delete_query('sessions', "ip = ".$db->escape_binary($session->packedip)." OR uid='{$mybb->user['uid']}'");
-    }
-    else
-    {
+	}
+	else
+	{
 		$db->delete_query('sessions', "ip = ".$db->escape_binary($session->packedip));
-    }
+	}
 	error($lang->error_banned);
 }
 
@@ -922,7 +941,7 @@ if($mybb->settings['boardclosed'] == 1 && $mybb->usergroup['canviewboardclosed']
 
 	$lang->error_boardclosed .= "<blockquote>{$mybb->settings['boardclosed_reason']}</blockquote>";
 
-	if(!$mybb->get_input('modal')) 
+	if(!$mybb->get_input('modal'))
 	{
 		error($lang->error_boardclosed);
 	}
--- a/global.php
+++ b/global.php
@@ -124,9 +124,7 @@ if(isset($mybb->input['theme']) && verify_post_check($mybb->get_input('my_post_k
 
 	if($userhandler->verify_style())
 	{
-		$userhandler->update_user();
-
-		$mybb->user['style'] = $user['theme'];
+		$mybb->user['style'] = $user['style'];
 
 		// If user is logged in, update their theme selection with the new one
 		if($mybb->user['uid'])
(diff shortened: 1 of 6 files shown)


global.php
```diff
@@ -111,21 +111,36 @@
...
+	$userhandler = new UserDataHandler('update');
+
+	$user = array(
+		'uid'	=> $mybb->user['uid'],
+		'style'	=> $mybb->get_input('theme', MyBB::INPUT_INT)
+	);
+
+	$userhandler->set_data($user);
+
+	if($userhandler->validate_user())
```
Destroy666x: Whole validation not needed here either, only `verify_style()`

inc/functions.php
```diff
@@ -5033,6 +5017,48 @@ function build_theme_select($name, $selected="", $tid=0, $depth="", $usergroup_o
...
+		foreach($themes as $theme)
+		{
+			if($tid == $theme['tid'])
+			{
+				$s_theme = $theme;
+				break;
+			}
+		}
+
+		if($s_theme)
```
Destroy666x: `break 2;` could be used above to break the main foreach.

inc/functions.php
```diff
@@ -5033,6 +5017,48 @@ function build_theme_select($name, $selected="", $tid=0, $depth="", $usergroup_o
...
+	$s_theme = false;
+
+	foreach($tcache as $themes)
+	{
+		foreach($themes as $theme)
+		{
+			if($tid == $theme['tid'])
+			{
+				$s_theme = $theme;
+				break;
```
Destroy666x: `break 2;` could be used here to break both foreach structures instead of the additional `if` below.

Sama34:
Done.

global.php
```diff
@@ -111,21 +111,36 @@
...
+		'uid'	=> $mybb->user['uid'],
+		'style'	=> $mybb->get_input('theme', MyBB::INPUT_INT)
+	);
+
+	$userhandler->set_data($user);
+
+	if($userhandler->verify_style())
 	{
-		if(isset($mybb->cookies['mybbtheme']))
+		$userhandler->update_user();
```
Stefan-ST: Why do you call update_user() also for guests?

Sama34:
Done.

inc/functions.php
```diff
@@ -5033,6 +5017,43 @@ function build_theme_select($name, $selected="", $tid=0, $depth="", $usergroup_o
...
+ * @param int The theme id of the theme.
+ * @return array The theme data
+ */
+function get_theme($tid)
+{
+	global $tcache, $db;
+
+	if(!is_array($tcache))
+	{
+		$query = $db->simple_select('themes', 'tid, name, pid, allowedgroups', "pid!='0'", array('order_by' => 'pid, name'));
```
Destroy666x: Just noticed, any reason for `array('order_by' => 'pid, name')` here? Why would there be a need to sort the results by any of these fields? The `while` loop will go through all of them anyways and do the same thing. So it only slows down the query.

inc/datahandlers/user.php
```diff
@@ -868,6 +868,31 @@ function verify_language()
...
+	{
+		global $lang;
+
+		$tid = &$this->data['style'];
+
+		if($tid != 0 && !defined('IN_ADMINCP'))
+		{
+			$theme = get_theme($tid);
+
+			if(!is_member($theme['allowedgroups']) && $theme['allowedgroups'] != 'all')
```
Destroy666x: For better compatibility with plugins/3rd party registrations/etc. I think that `usergroup` and `additionalgroups` should be passed to `is_member()`, just like it's done in the profile fields code: https://github.com/Sama34/mybb/blob/fix-1752/inc/datahandlers/user.php#L51 Otherwise it will be limited to the current user only.

inc/functions.php
Sama34: It was copied from `build_theme_select()`.

Destroy666x: Well, doesn't matter if you copied it or not - it doesn't make sense <b>here</b>. In theme selector sorting them is not surprising (they're displayed), but here it doesn't do anything.

global.php
```diff
@@ -111,21 +111,36 @@
...
+			$userhandler->update_user();
+		}
+		// Guest = cookie
+		else
+		{
+			my_setcookie('mybbtheme', $mybb->get_input('theme'));
+		}
 	}
 }
 // Cookied theme!
```
Stefan-ST: @Sama34 Since you can manipulate cookies easily shouldn't the be checked, too?

global.php
```diff
@@ -111,21 +111,36 @@
...
 {
-	$mybb->user['style'] = $mybb->get_input('theme');
-	// If user is logged in, update their theme selection with the new one
-	if($mybb->user['uid'])
+	// Set up user handler.
+	require_once MYBB_ROOT.'inc/datahandlers/user.php';
+	$userhandler = new UserDataHandler('update');
+
+	$user = array(
+		'uid'	=> $mybb->user['uid'],
```
Destroy666x: Need to pass `usergroup` and `displaygroup` here. In usercp.php's `do_options` action too.

global.php
Sama34: Exactly how do you want to verify the cookie? On every page load seems the correct but not best approach.

Stefan-ST: Well, the fix makes not much sense when you still can bypass the permissions.
Perhaps we can cache the themes available for guests?

JN-Jones:
We also load the theme in `xmlhttp.php`. Also this PR assumes that neither the usergroup nor the allowedgroup setting change. If one of them changes the permissions would be ignored again as long as the user doesn't submit a form which updates the theme.

IMHO we shouldn't check permissions on saving, they should be changed when loading a theme. Somewhere around https://github.com/mybb/mybb/blob/feature/global.php#L240 a simple `is_member` check (which should probably be skipped if a theme is loaded for a forum with override user style enabled) which updates the user setting to the default theme if he doesn't have permission to use that theme.

Destroy666x:
@Sama34, `usergroup` and `additionalgroups` are still not passed in UCP -> Edit Options: https://github.com/mybb/mybb/blob/feature/usercp.php#L789-821 after the last commit.

> If one of them changes the permissions would be ignored again as long as the user doesn't submit a form which updates the theme.

Yep, definitely agreed it should be prevented too.

> IMHO we shouldn't check permissions on saving, they should be changed when loading a theme. 

Why not both? I think an error should show up when trying to choose incorrect theme, in User CP at least. So the current code should stay in my opinion.

> which updates the user setting to the default theme

The problem I see with that is that admins may delete the default theme deliberately so that no groups should use it. What will we do in that case?

JN-Jones:
There's always one forum wide default theme ;) Same happens when user1 selects theme2 and the admin deletes theme2.

JN-Jones:
Another thing that needs to be considered: Atm it's possible to set permissions on the default theme. Which may cause some serious issues eg if only the default theme is available. Permissions should be ignored imho then.

Destroy666x:
@JN-Jones, yep, that's what I meant above. Currently you can delete the default theme and add 2 custom themes with split permissions, for example Theme1 for 50% of groups and Theme2 for the rest. The solution then would be to check if you can use the current theme and if not enforce the 2nd. But that won't work if there are 3 or more themes with split permissions with possible duplicates - then if group X can only use Theme5 and Theme6 and none of them is set as default, which one should be enforced if the user currently has Theme1? Or as you said, the worst thing that can be made - 1 theme without full permissions - this should definitely be disallowed, not sure about the rest I mentioned, split permissions may be useful on some forums.

global.php
Sama34: Done.

Sama34:
I have pushed some changes to the code.

I would like to know how the `xmlhttp.php` file checks for forum themes though.

@Destroy666x sorry but can you elaborate? Is that possible now and if so it is still possible with my edits?

global.php
```diff
@@ -224,18 +241,37 @@
...
+if($loadstyle != "def='1'")
+{
+	$query = $db->simple_select('themes', 'name, tid, properties, stylesheets, allowedgroups', $loadstyle, array('limit' => 1));
+	$theme = $db->fetch_array($query);
+
+	if(isset($theme['tid']) && !$load_from_forum && !is_member($theme['allowedgroups']) && $theme['allowedgroups'] != 'all')
+	{
+		if($load_from_user == 1)
+		{
+			$db->update_query('users', array('style' => 0), "style = '{$mybb->user['style']}'");
```
JN-Jones: This query will update **all** users which use that specific style - that's sure not what it should do?

global.php
```diff
@@ -224,18 +241,37 @@
...
 		$cache->update_default_theme();
 	}
-	$theme = $cache->read('default_theme');
-}
-else
-{
-	$query = $db->simple_select('themes', 'name, tid, properties, stylesheets', $loadstyle, array('limit' => 1));
-	$theme = $db->fetch_array($query);
+
+	$theme = $cache->read('default_theme', true);
```
JN-Jones: Any reason why you added the second parameter? That'd mean that the database cache gets useless.

xmlhttp.php
```diff
@@ -84,18 +84,30 @@
...
 		$cache->update_default_theme();
 	}
-	$theme = $cache->read('default_theme');
-}
-else
-{
-	$query = $db->simple_select("themes", "name, tid, properties", $loadstyle);
-	$theme = $db->fetch_array($query);
+
+	$theme = $cache->read('default_theme', true);
```
JN-Jones: Same here

inc/functions.php
```diff
@@ -5033,6 +5017,43 @@ function build_theme_select($name, $selected="", $tid=0, $depth="", $usergroup_o
 }
 
 /**
+ * Get the theme data of a theme id.
+ *
+ * @param int The theme id of the theme.
```
JN-Jones: Should be `@param int $tid ...`

inc/functions.php
```diff
@@ -5033,6 +5017,43 @@ function build_theme_select($name, $selected="", $tid=0, $depth="", $usergroup_o
...
+ * @return array The theme data
```
JN-Jones: The function can also return `false` (`bool`) if no theme is found

inc/datahandlers/user.php
```diff
@@ -868,6 +868,31 @@ function verify_language()
...
+	 *
+	 * @return boolean True when valid, false when invalid.
+	 */
+	function verify_style()
+	{
+		global $lang;
+
+		$user = &$this->data;
+
+		if($user['style'] && !defined('IN_ADMINCP'))
```
JN-Jones: Any reason why the ACP doesn't check permissions? As an admin I would like to get an error if I select an invalid theme for an user.

inc/datahandlers/user.php
```diff
@@ -868,6 +868,31 @@ function verify_language()
...
+		{
+			$theme = get_theme($user['style']);
+
+			if(empty($theme) || !is_member($theme['allowedgroups'], array('usergroup' => $user['usergroup'], 'additionalgroups' => $user['additionalgroups'])) && $theme['allowedgroups'] != 'all')
```
JN-Jones: Why not simply pass `$user` to `is_member`?

inc/datahandlers/user.php
Sama34: I don't think that is how the ACP works.

inc/datahandlers/user.php
Sama34: Why would I pass everything but not only what it needs?

inc/datahandlers/user.php
JN-Jones: Should be checked there though. Also what's the reason for this check then?

inc/datahandlers/user.php
JN-Jones: Why would I create a new array and overcomplicate everything?

JN-Jones: Otherwise feel free to change everytime we call functions similar to this.

inc/datahandlers/user.php
Sama34: If I change the style of X user from the ACP shouldn't I be allowed to do so? I feel that is how the users edit module works. Go and try it.

I may be incorrect or have the incorrect impression.

inc/datahandlers/user.php
Sama34: @euantorano @Destroy666x

inc/datahandlers/user.php
Sama34: @euantorano @Destroy666x

inc/datahandlers/user.php
Destroy666x: As I said here: https://github.com/mybb/mybb/pull/2052 I don't mind either way, but I'd also prefer this.

inc/datahandlers/user.php
Destroy666x: The PR now changes the theme immediately on the fly if the user can't use it during a page visit, right? In that case it doesn't matter that the admin will be able to switch to a disallowed theme because it will get changed to a default theme anyways. So I'd show an error.

Sama34: Makes sense.

Sama34:
Fixed the issues.

inc/datahandlers/user.php
Sama34: I have to agree in that it looks easier to read just passing `$user`. I will do as I please within my plugins.

inc/functions.php
```diff
@@ -5033,6 +5017,43 @@ function build_theme_select($name, $selected="", $tid=0, $depth="", $usergroup_o
 }
 
 /**
+ * Get the theme data of a theme id.
+ *
+ * @param int $tid The theme id of the theme.
+ * @return boolean False if no valid theme, array With the theme data otherwise
```
JN-Jones: Should be `@return boolean|array False if no valid theme, Array with the theme data otherwise`

Sama34: Should have suggested that since the beginning..

I checked your PR trying to replicate the best syntax and this was the result >__>

Will update later.

JN-Jones: @Sama34 ? ;)

Sama34:
Done.

JN-Jones:
Looks good too @Destroy666x @euantorano

JN-Jones:
Any other comment? Otherwise I'll merge this tomorrow.

PR #1852, opened 2015-02-19
Fixes #1821 Account settings missing in Admin CP

### LINKED ISSUE #1821: Account settings missing in Admin CP
The account settings **Receive PM notifications for new buddy requests.** and **Automatically accept buddy requests (if the above checkbox is ticked, a PM is sent informing of the new buddy connection)** are not available when editing a user in the Admin CP.

PaulBender:
Fixes #1821


Destroy666x:
They also should be handled in the user datahandler: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L1037
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L637

PaulBender:
Done. Also added the options to the install file (they were missing there as well).

PR #1853, opened 2015-02-19
Fixes #1822 Username autocomplete for Group Leaders

### LINKED ISSUE #1822: Username autocomplete for Group Leaders
Add a username autocomplete to the username field when adding a group leader.

Also add it to Search PM.

PaulBender:
Fixes #1822

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -487,7 +487,7 @@ if($mybb->input['action'] == "leaders")
 	<script type="text/javascript">
 	<!--
 	$("#username").select2({
-		placeholder: "Search for a user",
+		placeholder: "'.$lang->search_for_a_user.'",
 		minimumInputLength: 3,
 		maximumSelectionSize: 3,
 		multiple: false,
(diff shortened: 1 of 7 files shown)


admin/modules/user/groups.php
```diff
@@ -478,9 +478,48 @@
...
-
 	$form_container->end();
+
+	// Autocompletion for usernames
+	echo '
+	<link rel="stylesheet" href="../jscripts/select2/select2.css">
+	<script type="text/javascript" src="../jscripts/select2/select2.min.js?ver=1804"></script>
+	<script type="text/javascript">
+	<!--
+	$("#username").select2({
```
ATofighi: It should be translatable.

admin/modules/user/groups.php
```diff
@@ -478,9 +478,48 @@
...
+		placeholder: "Search for a user",
```
ATofighi: It should be translatable.

PaulBender: Done.

Also made it translatable in several other places.

PR #1854, opened 2015-02-19
Fix #1810 jGrowl alert types style

### LINKED ISSUE #1810: jGrowl alert types style
jGrowl allows alerts to have custom styles, this would be nice to differentiate an error alert from a success one.

**Reference:** http://community.mybb.com/thread-164732.html

Sama34:
Fix #1810 jGrowl alert types style


jscripts/rating.js
```diff
@@ -88,7 +88,7 @@ var Rating = {
 		{
 			$.each(json.errors, function(i, error)
 			{
-				$.jGrowl(lang.ratings_update_error + ' ' + error);
+				$.jGrowl(lang.ratings_update_error + ' ' + error, {theme:'jgrowl_error'});
```
Destroy666x: Missed this: https://github.com/Sama34/mybb/blob/fix-1810/jscripts/rating.js#L100

Sama34: Done.

Destroy666x:
The classes seem to be added everywhere where they're needed now, but what about making an actual use of it in the default ACP/front-end themes?

Something like this should work:
```css
.jgrowl_error
{
	color: red !important;
}

.jgrowl_process
{
	color: yellow !important;
}

.jgrowl_success
{
	color: green !important;
}
```
Also, the message backgrounds could be made consistent - in front-end they're black and in ACP they're yellow for some reason.

Sama34:
I would rather change the background color to fit the message type instead of just the text. What do you think?

Also, updating the main `global.css` in the `mybb_theme.xml` file will automatically update the master style sheet but not the custom ones, am I right?

@Stefan-ST

Destroy666x:
Well, I tried backgrounds first and they didn't look good in my opinion, but if you want to try - go ahead.

JoshH99:
Yeah I think some gradient attempts would be good. Potentially referencing Audentio's Dark Ice notification bar styles as inspiration for how to color the jGrowl boxes.

Stefan-ST:
@Sama34 
> Also, updating the main `global.css` in the `mybb_theme.xml` file will automatically update the master style sheet but not the custom ones, am I right?

Correct but it's probably not an issue here.

Sama34:
That should do it for demonstration.

Destroy666x:
Looks too bright in my opinion, especially the green one:

and I still think that only the font (+ possibly the border) should be changed as in the example above.

Also, whichever style will be chosen, `main.css` ver needs to be bumped in one file.

Sama34:
Feel free to suggest a green code. The red and yellow one are copied from the default style, to be consistent. I totally forgot about the close all one, will probably apply the yellow one over it too.

JN-Jones:
Can you bump the version codes to `1806` again? ;)

Sama34:
Done.

Destroy666x:
IMO the colors are still not matching our front-end and back-end theme. Not sure why you used the same yellow color for the progress notification and close bar either. But if you feel they're better than the old black background + colored fonts and borders, feel free to merge, the rest is ok.

Sama34:
I'm not particularly good with design, honestly. If you could share the colors you feel appropriate I would simply update the code.

I didn't updated as you suggested in https://github.com/mybb/mybb/pull/1854#issuecomment-84166474 because of the same reason.

Destroy666x:
I shared a very simple color solution with a screenshot above:
```css
.jgrowl_error
{
    color: red !important;
}

.jgrowl_process
{
    color: yellow !important;
}

.jgrowl_success
{
    color: lightgreen !important;
}
```
Possibly with borders:
```css
.jgrowl_error
{
    color: red !important;
    border: red !improtant;
}

.jgrowl_process
{
    color: yellow !important;
    border: yellow !improtant;
}

.jgrowl_success
{
    color: lightgreen !important;
    border: lightgreen !improtant;
}
```
With other changes except `border-radius` reverted. Makes them distinguishable and easy to differentiate and keeps the old dark background style we used.

dequeues:

euantorano:
We also need to bump the versions. If you don't get a chance to do so @Sama34, I'll take this over.

Sama34:
I will update my PRs tomorrow.

Sama34:
This is done.

PR #1856, opened 2015-02-19
Fix #1855 Incorrect integer value: '' for column '__' at row 1

### LINKED ISSUE #1855: Incorrect integer value: '' for column '__' at row 1
There are some annoying warning if using MySQL on strict mode.

Sama34:
Fix #1855 Incorrect integer value: '' for column '__' at row 1


inc/datahandlers/user.php
```diff
@@ -1084,11 +1084,11 @@ function insert_user()
...
-			"showcodebuttons" => $user['options']['showcodebuttons'],
-			"sourceeditor" => $user['options']['sourceeditor'],
-			"away" => $user['away']['away'],
-			"awaydate" => $user['away']['date'],
-			"returndate" => $user['away']['returndate'],
+			"showcodebuttons" => (int)$user['options']['showcodebuttons'],
+			"sourceeditor" => (int)$user['options']['sourceeditor'],
+			"away" => (int)$user['away']['away'],
+			"awaydate" => (int)$user['away']['date'],
+			"returndate" => (int)$user['away']['returndate'],
```
Destroy666x: Returndate is not an integer:
https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_tables.php#L1093
http://community.mybb.com/thread-158326-post-1136805.html#pid1136805

Sama34: It only saves digits to it anyways so there is no difference with the casting.

Destroy666x: Not only digits, I don't think `(int)` will keep the `-`:

Sama34: That makes no sense, why is it saved like that? Anyways, I reverted it.

Destroy666x: > why is it saved like that?

Check the 2nd link in my comment above.

PR #1857, opened 2015-02-19
Fix #1239 UX improvement for showing a user's reputations

### LINKED ISSUE #1239: UX improvement for showing a user's reputations
Even if a post a user got a reputation for has been deleted the links to the posts are still shown, so "username's post" leads to the post and "in threadname" leads to the corresponding thread (thread link is hidden for hard deleted posts). 

Using "Rating given for deleted post in threadname" or using "Rating given for deleted post in deleted thread" would increase the consistency of the display style (always show post and thread info) and would improve the user experience, because there are no links to content that is not available any more (or to this user in case of a soft deleted post). 

This issue is especially annoying for an user without the right to see soft deleted content (see http://community.mybb.com/thread-156131.html)

Original thread: [UX improvement for showing a user's reputations](http://community.mybb.com/thread-156133.html)

Sama34:
Fix #1239 UX improvement for showing a user's reputations

### DIFF
--- a/reputation.php
+++ b/reputation.php
@@ -864,17 +864,51 @@ if(!$mybb->input['action'])
 
 	if(!empty($post_cache))
 	{
-		$sql = implode(',', $post_cache);
+		$pids = implode(',', $post_cache);
+
+		$sql = array("p.pid IN ({$pids})");
+
+		// get forums user cannot view
+		$unviewable = get_unviewable_forums(true);
+		if($unviewable)
+		{
+			$sql[] = "p.fid NOT IN ({$unviewable})";
+		}
+
+		// get inactive forums
+		$inactive = get_inactive_forums();
+		if($inactive)
+		{
+			$sql[] = "p.fid NOT IN ({$inactive})";
+		}
+
+		if(!$mybb->user['ismoderator'])
+		{
+			$sql[] = "p.visible='1'";
+			$sql[] = "t.visible='1'";
+		}
+
+		$sql = implode(' AND ', $sql);
 
 		$query = $db->query("
-			SELECT p.pid, p.uid, p.message, t.tid, t.subject
+			SELECT p.pid, p.uid, p.fid, p.visible, p.message, t.tid, t.subject, t.visible AS thread_visible
 			FROM ".TABLE_PREFIX."posts p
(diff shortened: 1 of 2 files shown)


reputation.php
```diff
@@ -936,21 +970,16 @@
...
 
 		// Is this rating specific to a post?
-		$postrep_given = '';
+		$postrep_given = $lang->sprintf($lang->postrep_given_nolink, $user['username']);
 		if($reputation_vote['pid'])
 		{
-			$link = get_post_link($reputation_vote['pid'])."#pid{$reputation_vote['pid']}";
-
-			$thread_link = '';
-			if(isset($post_reputation[$reputation_vote['pid']]))
```
Destroy666x: Why was `isset()` check removed?

reputation.php
```diff
@@ -864,17 +864,51 @@
...
+		}
+
+		// get inactive forums
+		$inactive = get_inactive_forums();
+		if($inactive)
+		{
+			$sql[] = "p.fid NOT IN ({$inactive})";
+		}
+
+		if(!$mybb->user['ismoderator'])
```
Destroy666x: I don't think this part is needed with the 2 conditionals below.

Also, the `canonlyviewownthreads` permission check is still missing,

reputation.php
```diff
@@ -936,21 +970,16 @@
 		$last_updated = $lang->sprintf($lang->last_updated, $last_updated_date);
 
 		// Is this rating specific to a post?
-		$postrep_given = '';
+		$postrep_given = $lang->sprintf($lang->postrep_given_nolink, $user['username']);
```
Destroy666x: This + lack of `isset()` introduce even more noticeable bugs than the fixed issue - every non-post reputation has now incorrect `Rating given for xxx's post` note and the post reputations which should be hidden lead to blank posts.

reputation.php
Sama34: Doesn't `get_unviewable_forums()` handles `canonlyviewownthreads`?

Destroy666x: It never did, only `canviewthreads`

Sama34: I pushed a change to should handle those.

Sama34: > I don't think it makes the code faster either.

For posts may be not, but threads don't have an index on `visible` so I just added it.

PR #1858, opened 2015-02-19
Fix #298 Login per E-Mail

### LINKED ISSUE #298: Login per E-Mail
> *I would like to create a ticket for 1.6.7 but it seems that this is not possible. The dropdown doesn't show up the 1.6.7 version.....*
> 
> I have just tried to login via my e-mail address but didn't succeed. The reason is that there is a second user existing, using the same e-mail address (it's my test account). Once I changed this address I had been able to login.
> 
> It seems that the e-mail login only works for e-mail addresses that exist only one time. Might be necessary to give an according note in the ACP.

Original report: http://dev.mybb.com/issues/1980
Reporter: Joachim K.
Status: Confirmed

Sama34:
Fix #298 Login per E-Mail


Sama34:
To be honest I don't think we can do much about this.

admin/modules/config/settings.php
```diff
@@ -974,6 +974,12 @@
 			}
 		}
 
+		// Administrator is changing the login method.
+		if($mybb->settings['username_method'] != $mybb->input['upsetting']['username_method'] && $mybb->input['upsetting']['username_method'] == 1 || $mybb->input['upsetting']['username_method'] == 2)
+		{
+			$db->update_query('settings', array('value' => 0), "name='allowmultipleemails'");
```
Destroy666x: Not a big deal, but it could be moved here: https://github.com/Sama34/mybb/blob/fix-298/admin/modules/config/settings.php#L925 and the query changed to `$mybb->input['upsetting']['allowmultipleemails']` to possibly save a query.

Also, what if the user tries to change the multiple Emails setting after saving the username method? That wouldn't be blocked.

Sama34:
That should do it.

Stefan-ST:
@Sama34 What happens when using emails multiple times was enabled before?

Destroy666x:
Agreed with the above, should add something like this to upgrade too:
```php
if($mybb->settings['allowmultipleemails'] == 0 && $mybb->settings['username_method'] == 1 || $mybb->settings['username_method'] == 2)
{
// query for updating the setting to 0 here
}
```

Stefan-ST:
@Destroy666x It still won't work when there are already accounts with the same mail.

Destroy666x:
But what can we do with that? In my opinon it's a past feature design flaw that can't be corrected anymore.

We can't force the owners to change their Email on every forum.
We can't really change the login method - how would we be able to detect who's trying to log in with Email only option turned on?

Any other options?

Stefan-ST:
We can at least hide/block the setting in case there are accounts with the same email. Breaking the login for existing accounts is not acceptable.

Destroy666x:
> We can at least hide/block the setting in case there are accounts with the same email. Breaking the login for existing accounts is not acceptable.

Well, on one forum there can be 2 accounts with the same Email and 5000 that use the Email login function regularly. I don't think changing the login method setting to Username only out of sudden is acceptable either.

Especially that people who had the same Email probably aren't using their account anymore anyways (or asked the admin to change it for them), so they won't even notice that they're able to login normally now.

Stefan-ST:
What about not allowing to change it to anything beside Username only in that case? But keeping it for those who have already changed it.

Destroy666x:
@Stefan-ST, not sure what you mean at the moment. The PR does something like that already - it sets multiple Emails setting to 0 if the admin decides to enable Email/both logging methods, which prevents the bug from happening for new accounts. It only has to be added to the upgrade too.

We can of course do it other way round as you said - set the logging method setting to 0 (Username only) if multiple Emails are enabled, but will that make any difference? Except that logging method setting is changed more often, I think. At least it seems to be more useful than allowing multiple users with the same Email, in my opinion.

Stefan-ST:
@Destroy666x 
> but will that make any difference?

It makes a big difference as it shouldn't be able to activate login with email when there are multiple accounts with the same email. The important point is that it should check the existing accounts not the other setting.

Destroy666x:
Oh, I understand what you meant now. Well, yea, checking that would be good too, but only on the setting change, not on upgrade as said earlier.

Sama34:
This is done.

install/resources/upgrade34.php
```diff
@@ -0,0 +1,46 @@
...
+	"revert_all_themes" => 0,
+	"revert_all_settings" => 0
+);
+
+@set_time_limit(0);
+
+function upgrade34_dbchanges()
+{
+	global $db, $output;
+
```
JN-Jones: You need to call `print_headers` too. Otherwise the update is broken.

```php
	$output->print_header("Updating Database");
	echo "<p>Performing necessary upgrade queries...</p>";
	flush();
```

install/resources/upgrade34.php
```diff
@@ -0,0 +1,46 @@
...
+	$query = $db->simple_select('settings', 'value', "name='username_method'");
```
Destroy666x: Is this query needed? Why not simply globalize `$mybb` and check `$mybb->settings['username_method']`?

install/resources/settings.xml
```diff
@@ -959,7 +959,8 @@ disabled=Disable this feature]]></optionscode>
 		</setting>
 		<setting name="username_method">
 			<title>Allowed Login Methods</title>
-			<description><![CDATA[The login methods you wish to allow for the username field. Username only, Email only, or allow both.]]></description>
+			<description><![CDATA[The login methods you wish to allow for the username field. Username only, Email only, or allow both.<br />
+<strong>Note:</strong> This will only work for the account that registered the e-mail first if the "Allow emails to be registered multiple times?" setting was set to "Yes" before changing this.]]></description>
```
Destroy666x: The note is not up-to-date anymore, is it? The setting is not allowed to be changed when there are duplicates.

admin/modules/config/settings.php
```diff
@@ -927,6 +927,21 @@
...
+		{
+			$query = $db->simple_select('users', 'email, COUNT(email) AS duplicates', "email!=''", array('group_by' => 'email HAVING duplicates>1'));
+			if($db->num_rows($query))
+			{
+				$mybb->input['upsetting']['username_method'] = 0;
+				$lang->success_settings_updated .= $lang->success_settings_updated_username_method;
+			}
+			else
+			{
+				$mybb->input['upsetting']['allowmultipleemails'] = 0;
```
Destroy666x: Would be nice to add a `$lang->success_settings_updated` concantenated note for this too.

install/resources/upgrade34.php
JN-Jones: The query also checks duplicated settings while `$mybb->settings` only checks one. Though I'm not sure why we would need to consider that.

Destroy666x: If that's the reason, there's no point in verifying possible buggy duplicate records because only `$mybb->settings['username_method']` is checked when you login in anyways.

Sama34: Why would I check a cached value instead of fetching it? It is an upgrade script not a production script.

I'm sure we do not check `$mybb->settings` when updating settings in any other version script.

Destroy666x: > Why would I check a cached value instead of fetching it?

Because there is no reason to fetch it from the database, correct me if I'm wrong.

> It is an upgrade script not a production script.

That doesn't mean it shouldn't be optimal.

> I'm sure we do not check $mybb->settings when updating settings in any other version script.

https://github.com/mybb/mybb/blob/feature/install/resources/upgrade12.php#L800-814
We don't select them from db anywhere though.

Sama34: The `$settings` array is still build from a cache file which content may differ from that of the DB.

Just because somebody else opted to ignore the slight possibilities of that being the case doesn't mean I will do the same.

Please provide realistic reasons as to why I should take the time to modify something that I considered to be the most effective approach other that just one single meaningless query resources. Which I already thought about before pushing my changes.

Now if somebody else supports you I will not resist.

BTW those references only make use of the setting value not actually update values as this case does.

Destroy666x: > The $settings array is still build from a cache file which content may differ from that of the DB.

https://github.com/mybb/mybb/blob/feature/install/upgrade.php#L94-128 - what exactly will change the setting value between running this code and the 1.8.6 upgrade part (ACP and front-end are locked, just to remind)? Could you show an example?

> BTW those references only make use of the setting value not actually update values as this case does.

?
The 1st example also runs a query that updates stuff in database based on the setting's value. I'm of course only talking about the redundacy of `simple_select()`, not the update queries.

Sama34: The part of code you reference only takes place if the cache file does not exists.

> The 1st example also runs a query that updates stuff in database

Not settings.

Destroy666x: > The part of code you reference only takes place if the cache file does not exists.

Well, if we really want to ensure the settings match database in every case during the upgrade, not just randomly in one update script, the file cache check should be removed and one global setting query should be ran instead of wasting one query per every setting. Otherwise it means the cache is trustable enough to not do that.

Sama34: Nah. You got me with that last line. I will update the code later.

> Otherwise it means __the cache is trustable enough__ to not do that.

If no case has been reported yet then it means it indeed is trust-able enough.

Destroy666x: Right.. So we're stuck yet again and it's time to wait for other opinions then, I guess. @JN-Jones @Stefan-ST @euantorano @ATofighi @Devilshakerz

Devilshakerz: I don't see a reason why the cache shouldn't be used; if the settings are modified during the upgrade process, the module modifying them should take care of that cache as well.

ATofighi: I also think we can use `$mybb->settings`.

Sama34: > So we're stuck yet again

Incorrect. I said in my last comment I was going to update the code later... >__>??

Destroy666x: I thought you were sarcastic there. But if not - my bad, never mind then, Still nice to hear other opinions though.

euantorano: Yes, I'd just use `$mybb->settings`. There's always the chance of the setting changing between you reading it and using it, as you cannot properly lock the database access. We should try to at least follow the standards already in use throughout 1.x.

Sama34:
Should be correct now.

install/resources/upgrade34.php
```diff
@@ -0,0 +1,46 @@
...
+
+function upgrade34_dbchanges()
+{
+	global $db, $output, $mybb;
+
+	$output->print_header("Updating Database");
+	echo "<p>Performing necessary upgrade queries...</p>";
+	flush();
+
+	if($mybb->settings['value'] == 1 || $mybb->settings['value'] == 2)
```
Destroy666x: `$mybb->settings['value']`?

install/resources/settings.xml
```diff
@@ -959,7 +959,8 @@ disabled=Disable this feature]]></optionscode>
 		</setting>
 		<setting name="username_method">
 			<title>Allowed Login Methods</title>
-			<description><![CDATA[The login methods you wish to allow for the username field. Username only, Email only, or allow both.]]></description>
+			<description><![CDATA[The login methods you wish to allow for the username field. Username only, Email only, or allow both.<br />
+<strong>Note:</strong> This will only work if there are no duplicate accounts using the same e-mail address.]]></description>
```
Destroy666x: `This` should rather be `Email options` to be more clear, because changing to `Username only` will always work fine.

admin/modules/config/settings.php
```diff
@@ -927,6 +927,21 @@
...
+			$query = $db->simple_select('users', 'email, COUNT(email) AS duplicates', "email!=''", array('group_by' => 'email HAVING duplicates>1'));
+			if($db->num_rows($query))
+			{
+				$mybb->input['upsetting']['username_method'] = 0;
+			}
+			else
+			{
+				$mybb->input['upsetting']['allowmultipleemails'] = 0;
+			}
+			$lang->success_settings_updated .= $lang->success_settings_updated_username_method;
```
Destroy666x: Can't use the same error for both cases.

`success_settings_updated_username_method` says that `Allowed Login Methods` wasn't updated due to duplicate Emails, there should be another error for the other setting.

JN-Jones:
@Sama34

Sama34:
Should be correct now.

JN-Jones:
@euantorano @Destroy666x Looks good from a quick look. Haven't fully tested yet but would be nice to include this in 1.8.6.

JN-Jones:
@Sama34 Can you fix the conflicts?

dequeues:

Sama34:
Fixed.

Destroy666x:
upgrade36.php is the current upgrade file

Sama34:
Updated.

PR #1861, opened 2015-02-19
Fix #1837 Setting safe_html to 1 for the parser though it isn't supported

### LINKED ISSUE #1837: Setting safe_html to 1 for the parser though it isn't supported
http://crossreference.mybboard.de/nav.html?inc/datahandlers/pm.php.source.html#l660
http://crossreference.mybboard.de/nav.html?inc/datahandlers/post.php.source.html#l1105

Both pass `'safe_html' => 1` as option to the parser, however the parser doesn't have that option at all.

Sama34:
Fix #1837 Setting safe_html to 1 for the parser though it isn't supported

### DIFF
--- a/inc/datahandlers/pm.php
+++ b/inc/datahandlers/pm.php
@@ -657,7 +657,13 @@ class PMDataHandler extends DataHandler
 
 				require_once MYBB_ROOT.'inc/class_parser.php';
 				$parser = new Postparser;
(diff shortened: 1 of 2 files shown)


JN-Jones:
Any reason you added a new variable? The old code was okay apart from the unused variable.

Sama34:
Do you mean `$parser_options`? If so, it seems more readable like that. If you disagree I can revert it.

PR #1862, opened 2015-02-19
Fix #1836 split used in class_parser.php

### LINKED ISSUE #1836: split used in class_parser.php
PHP Code:
case&nbsp;"dailymotion":&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;list(&#36;id,&nbsp;)&nbsp;=&nbsp;split("_",&nbsp;&#36;path[2],&nbsp;1);&nbsp;//&nbsp;http://www.dailymotion.com/video/fds123_title-goes-here&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;break;&nbsp;

split is deprecated since php 5.3 so it should be updated to make use of explode.

Original thread: [split used in class_parser.php](http://community.mybb.com/thread-167112.html)

Sama34:
Fix #1836 split used in class_parser.php

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1342,8 +1342,7 @@ class postParser
 		switch($video)
 		{
 			case "dailymotion":
-				$id = explode('_', $path[2]);
-				$id = $id[0]; // http://www.dailymotion.com/video/fds123_title-goes-here
+				list($id, ) = explode('_', $path[2], 1); // http://www.dailymotion.com/video/fds123_title-goes-here
 				break;
 			case "metacafe":
 				$id = $path[2]; // http://www.metacafe.com/watch/fds123/title_goes_here/
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1342,7 +1342,8 @@ class postParser
 		switch($video)
 		{
 			case "dailymotion":
-				list($id, ) = split("_", $path[2], 1); // http://www.dailymotion.com/video/fds123_title-goes-here
+				$id = explode('_', $path[2]);
+				$id = $id[0]; // http://www.dailymotion.com/video/fds123_title-goes-here
 				break;
 			case "metacafe":
 				$id = $path[2]; // http://www.metacafe.com/watch/fds123/title_goes_here/


Sama34:
Using `http://www.dailymotion.com/video/fds123_title-goes-here` as example results in`$id` being `fds123_title-goes-here` before and`fds123` now. Seems to work.

JN-Jones:
same here: the old code was fine, simply replacing split with explode would have been enough (explode also has a third parameter)

Sama34:
I somehow correlated the usage of `list()` with `split()`.

ATofighi:
@JN-Jones Old code doesn't work fine...
```php
list($id, ) = split('_', 'http://www.dailymotion.com/video/fds123_title-goes-here', 1);
```
or
```php
list($id, ) = explode('_', 'http://www.dailymotion.com/video/fds123_title-goes-here', 1);
```
is same as
```
$id = 'http://www.dailymotion.com/video/fds123_title-goes-here';
```
I think it should be
```php
list($id, ) = explode('_', 'http://www.dailymotion.com/video/fds123_title-goes-here', 2);
```

JN-Jones:
Only `fds123_title-goes-here` is passed to explode - so with my code the same will be returned, with yours `fds123`; both seems to work so probably it's better to ether use
```php
$id = htmlspecialchars_uni($path[2]);
```

Destroy666x:
Both `$id = $path[2];` and `list($id, ) = explode('_', $path[2], 2);` seem to work, but I think it's safer to use the 2nd one with video ID only - looks to be a way preferred by Dailymotion that won't change anytime soon.

Sama34:
I don't understand the issue, I tested and nothing really changed from what we got before and what we got now. The video displayed correctly for me too..

Destroy666x:
Well, `list($id, ) = explode('_', $path[2], 1);` does exactly the same thing as `$id = $path[2];` so no reason to not use the latter (2 useless function calls less). 

`list($id) = explode('_', $path[2], 2);` on the other hand gets only the actual ID as it was meant to in the first place and which is preferred by Dailymotion it seems (every embed code I found at least seems to use only ID), so I'd go with that.

PR #1863, opened 2015-02-19
Fix #1827 Update CONTRIBUTING.md

### LINKED ISSUE #1827: Update CONTRIBUTING.md
https://github.com/mybb/mybb/blob/feature/CONTRIBUTING.md - 1.6 bugs shouldn't be reported anymore and the linked forum is closed, so the info should be updated.

Sama34:
Fix #1827 Update CONTRIBUTING.md


Stefan-ST:
@Sama34 As I already said on the issue there is no need to create a PR for this. The file is not part of releases hence we can skip SQA checks,...

PR #1865, opened 2015-02-19
Fix #1210 soft deleted posts with showthread.php?pid=x#pidx

### LINKED ISSUE #1210: soft deleted posts with showthread.php?pid=x#pidx
Default behaviour: Goto the corresponding thread and jump to post with pid x

Behaviour for hard deleted posts: Display error message (The specified thread does not exist.). 

Behaviour for soft deleted posts and user with right to see soft deleted posts: see default behaviour

Behaviour for soft deleted posts and user without right to see soft deleted posts: the default behaviour is used, so the user is shown the thread the post was in, but jumping to the right post using the # url fails (post is not displayed). Behaviour for hard deleted posts would be right in my point of view, but showing the thread and displaying an inline error that the specified post was deleted would also be ok.

Original thread: [soft deleted posts with showthread.php?pid=x#pidx](http://community.mybb.com/thread-156131.html)

Sama34:
Fix #1210 soft deleted posts with showthread.php?pid=x#pidx

### DIFF
--- a/showthread.php
+++ b/showthread.php
@@ -891,7 +891,11 @@ if($mybb->input['action'] == "thread")
 		if(!empty($mybb->input['pid']))
 		{
 			$post = get_post($mybb->input['pid']);
-			if($post)
+			if(empty($post) || ($post['visible'] == 0 && !is_moderator($fid, 'canviewunapprove')) || ($post['visible'] == -1 && !is_moderator($fid, 'canviewdeleted')))
+			{
+				$footer .= '<script type="text/javascript">$(document).ready(function() { $.jGrowl(\''.$lang->error_invalidpost.'\', {theme: \'jgrowl_error\'}); });</script>';
+			}
+			else
 			{
 				$query = $db->query("
 					SELECT COUNT(p.dateline) AS count FROM ".TABLE_PREFIX."posts p
--- a/showthread.php
+++ b/showthread.php
@@ -44,10 +44,10 @@ if(!empty($mybb->input['pid']) && !isset($mybb->input['tid']))
 		$options = array(
 			"limit" => 1
 		);
-		$query = $db->simple_select("posts", "tid", "pid=".$mybb->get_input('pid', MyBB::INPUT_INT), $options);
+		$query = $db->simple_select("posts", "tid, visible", "pid=".$mybb->get_input('pid', MyBB::INPUT_INT), $options);
 		$post = $db->fetch_array($query);
-		
-		if(empty($post))
+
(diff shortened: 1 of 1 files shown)


Sama34:
I'm unsure what the `if(isset($style) && $style['pid'] == $mybb->get_input('pid', MyBB::INPUT_INT) && $style['tid'])` is supposed to check. `$style['pid']` seems to be the parent id of the theme not the post id and `tid` the theme id not the thread id.

That or I'm confused.

Destroy666x:
Looks ok, but a similar check would be nice here: https://github.com/Sama34/mybb/blob/fix-1210/showthread.php#L894 to avoid redirecting to pages with soft deleted/unapproved posts. And an `else` with `error($lang->error_invalidpost);` would make it more consistent too.

Sama34:
I see. But the thread is still visible, isn't it? @JN-Jones @mybb/developers 

`http://localhost/showthread.php?tid=3&pid=5#pid5` will be disabled even if the thread exists.

With SEF URLs the link could look like `http://localhost/thread-3-post-5.html#pid5` or `http://localhost/thread-subject&pid=612#pid612`.

I'm not sure if that should be blocked. Is pagination somehow affected by this @Destroy666x ?

Destroy666x:
The thread is visible but the user is brought to a page with the soft seleted/unapproved post. It may cause misunderstanding because the user thinks he's linked to the post on top of it.

An inline error saying the linked post is removed/soft deleted/unapproved would be the best solution as an `else` there.

Destroy666x:
Quick example:
- soft deleted/unapproved post: http://nooo.hol.es/mybb16/showthread.php?tid=16&pid=21#pid21
- removed/non-existent post: http://nooo.hol.es/mybb16/showthread.php?tid=16&pid=24#pid24

Sama34:
I like that instead.

Sama34:
I still would like to know what is the meaning of the usage of `$style` here. From `global.php` it is an array containing a forum data not a thread nor post data at all.

Destroy666x:
I still think that errors should be separated to show deleted (if non-existent) / no permission (if exists but can be viewed) errors as in the example above. Why? Because we do that everywhere. We don't throw "Thread is invalid" error if a group can't view it because it would be uninformative and not user-friendly.

As for `$style` check, no idea. I guess that code gets never executed because I can't see where `$style['tid']` is set. It's probably some kind of outdated code (similar thing in editpost.php), a new issue shuld probably be created for it.

Sama34:
Just noticed that the lastpost action doesn't check for this neither.

> I still think that errors should be separated to show deleted

I can't find the code where this is currently applied. So I would like to have it separated from this issue. Sounds like more files will need editing. `error_no_permission()` is used only for forum permissions in `showthread.php`.

Destroy666x:
Well, then merging and I'll open a new issue for the non-existent/no perms separation because it's indeed not user-friendly in more places than I thought.

PR #1866, opened 2015-02-19
Fix #1676 $fid in toggle_thread_softdelete

### LINKED ISSUE #1676: $fid in toggle_thread_softdelete
https://github.com/mybb/mybb/blob/feature/inc/class_moderation.php#L2929-L2959
The function has a `$fid`parameter which is passed later to `soft_delete_threads` or `restore_threads`, however both functions doesn't have the parameter. Haven't searched the places that function is called but IMHO that parameter should be removed.

Sama34:
Fix #1676 $fid in toggle_thread_softdelete

### DIFF
--- a/inc/class_moderation.php
+++ b/inc/class_moderation.php
@@ -2926,13 +2926,12 @@ class Moderation
 	 * @param int Forum ID
 	 * @return boolean true
 	 */
-	function toggle_thread_softdelete($tids, $fid)
+	function toggle_thread_softdelete($tids)
 	{
 		global $db;
 
 		// Make sure we only have valid values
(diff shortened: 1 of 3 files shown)


inc/class_moderation.php
```diff
@@ -2926,13 +2926,12 @@ function toggle_thread_visibility($tids, $fid)
 	 * @param int Forum ID
```
Stefan-ST: Not removing the comment?

Sama34: Done.

PaulBender:
Looks good.

There is a bug with the custom mod tools and soft delete which I'll create a new issue for.

PR #1868, opened 2015-02-19
Fixes #1793 - Random generated password can throw error

### LINKED ISSUE #1793: Random generated password can throw error
When you register and sending a random password is selected it's possible that the password is not accepted.
The random password has always 8 chars and is never complex however there will be an error if you require a longer or a complex password. Since there is no way to chose another password it's impossible to register.
```php
$mybb->input['password'] = random_str();
```

ATofighi:
#1793

### DIFF
--- a/member.php
+++ b/member.php
@@ -130,10 +130,10 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 		$password_length = (int)$mybb->settings['minpasswordlength'];
 		if($password_length < 8)
 		{
-			$password_length = min(8, $mybb->settings['maxpasswordlength']);
+			$password_length = min(8, (int)$mybb->settings['maxpasswordlength']);
 		}
 
-		$mybb->input['password'] = random_str($password_length);
+		$mybb->input['password'] = random_str($password_length, true);
 		$mybb->input['password2'] = $mybb->input['password'];
 	}
 
--- a/member.php
+++ b/member.php
@@ -125,7 +125,6 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 
 	if($mybb->settings['regtype'] == "randompass")
 	{
-		$mybb->settings['requirecomplexpasswords'] = 0;
 
 		$password_length = (int)$mybb->settings['minpasswordlength'];
 		if($password_length < 8)
@@ -133,7 +132,7 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 			$password_length = min(8, (int)$mybb->settings['maxpasswordlength']);
 		}
 
-		$mybb->input['password'] = random_str($password_length, true);
+		$mybb->input['password'] = random_str($password_length, $mybb->settings['requirecomplexpasswords']);
 		$mybb->input['password2'] = $mybb->input['password'];
 	}
 
--- a/member.php
+++ b/member.php
@@ -125,7 +125,14 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 
 	if($mybb->settings['regtype'] == "randompass")
 	{
-		$mybb->input['password'] = random_str();
+		$mybb->settings['requirecomplexpasswords'] = 0;
+
+		$password_length = (int)$mybb->settings['minpasswordlength'];
+		if($password_length < 8)
+		{
+			$password_length = 8;
+		}
+		$mybb->input['password'] = random_str($password_length);
 		$mybb->input['password2'] = $mybb->input['password'];
 	}
 
--- a/member.php
+++ b/member.php
@@ -130,8 +130,9 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 		$password_length = (int)$mybb->settings['minpasswordlength'];
 		if($password_length < 8)
 		{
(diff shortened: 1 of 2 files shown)


member.php
```diff
@@ -125,7 +125,14 @@
...
 	{
-		$mybb->input['password'] = random_str();
+		$mybb->settings['requirecomplexpasswords'] = 0;
+
+		$password_length = (int)$mybb->settings['minpasswordlength'];
+		if($password_length < 8)
+		{
+			$password_length = 8;
+		}
+		$mybb->input['password'] = random_str($password_length);
```
Stefan-ST: This does not cover the rare but possible case having maxpasswordlength lower than 8 (the generated password being to long).

ATofighi: fixed.

JN-Jones:
Looks good to me. @Destroy666x @euantorano ? Otherwise this can be merged from my point.

Destroy666x:
As I said, I'd rather like to see the function to return a proper string in case someone requires that. Here a quick prototype which may be optimised a bit (not that it matters much) and also changed so that it's compatible with PHP 5.2 (if I'm not mistaken and functions in array indexes aren't allowed):
```php
function random_str($length="8", $complex=false)
{
	$set = array_merge(range(0, 9), range('A', 'Z'), range('a', 'z'));
	$howmany = count($set)-1;
	$str = '';
	
	// Always at least 3 characters, even if $length < 3
	if($complex == true)
	{
		$howmanybig = count(range('A', 'Z'));

		// At least one number
		$char_arr[] = $set[my_rand(0, 9)];

		// At least one big letter
		$char_arr[] = $set[my_rand(10, 10 + $howmanybig)];

		// At least one small letter
		$char_arr[] = $set[my_rand(11 + $howmanybig, $howmany)];

		for($i = 1; $i <= ($length - 3); ++$i)
		{
			$char_arr[] = $set[my_rand(0, $howmany)];
		}

		// Make sure they're in random order and convert them to a string	
		shuffle($char_arr);
		$str = implode($char_arr);
	}
	else
	{
		for($i = 1; $i <= $length; ++$i)
		{
			$str .= $set[my_rand(0, $howmany)];
		}
	}

	return $str;
}
```

JN-Jones:
First of all I would remove `$howmanybig` (it's always 26) and I'd change the function so that we only have on `for` (both are nearly the same now, we can simply generate an array in both cases and shuffle it).

However I'm with @Stefan-ST here, you should never trust a generated password which is send via email and always change it anyways. But if the rest wants to change that I'm fine with your function.

Destroy666x:
> However I'm with @Stefan-ST here, you should never trust a generated password which is send via email and always change it anyways.

I agree, but that doesn't change the fact that someone may have integration with other software in which the lack of complex passwords chosen in ACP may break something before it gets changed by the user. And we don't lose anything by generating a proper complex password when needed.

Anyways, here's the slightly optimised and shorter funciton:
```php
function random_str($length="8", $complex=false)
{
    $set = array_merge(range(0, 9), range('A', 'Z'), range('a', 'z'));
    $str = '';

    // Always at least 3 characters, even if $length < 3
    if($complex == true)
    {
        // At least one number
        $char_arr[] = $set[my_rand(0, 9)];

        // At least one big letter
        $char_arr[] = $set[my_rand(10, 36)];

        // At least one small letter
        $char_arr[] = $set[my_rand(37, 61)];

       $length -= 3;
    }

   for($i = 1; $i <= $length; ++$i)
   {
      $char_arr[] = $set[my_rand(0, 61)];
   }

   // Make sure they're in random order and convert them to a string   
   shuffle($char_arr);
   $str = implode($char_arr);

   return $str;
}
```

JN-Jones:
You'd need to consider invalid passwords anyways in those cases. But yeah, I got your point, that's why I said I'm fine with it as long as others think so too ;)

Stefan-ST:
I'm fine with that, too, if it's reliable.

member.php
```diff
@@ -125,7 +125,15 @@
 
 	if($mybb->settings['regtype'] == "randompass")
 	{
-		$mybb->input['password'] = random_str();
+		$mybb->settings['requirecomplexpasswords'] = 0;
```
Destroy666x: Shouldn't be needed anymore.

member.php
```diff
@@ -125,7 +125,15 @@
...
-		$mybb->input['password'] = random_str();
+		$mybb->settings['requirecomplexpasswords'] = 0;
+
+		$password_length = (int)$mybb->settings['minpasswordlength'];
+		if($password_length < 8)
+		{
+			$password_length = min(8, (int)$mybb->settings['maxpasswordlength']);
+		}
+
+		$mybb->input['password'] = random_str($password_length, true);
```
Destroy666x: The 2nd parameter could be `false`/`true` based on `$mybb->settings['requirecomplexpasswords']` value (ofc after removing the `0` assignment I commented above).

ATofighi:
fixed

JN-Jones:
No issues for me, feel free to merge.

JN-Jones:

JN-Jones:
Fine with merging this. @Destroy666x @euantorano

ATofighi:
Thanks

PR #1870, opened 2015-02-19
Fixes #1844 Inline edit subject check

### LINKED ISSUE #1844: Inline edit subject check
Add a check to the inline edit subject function to see if the subject has actually been changed to prevent false logging if you accidentally trigger the edit subject function (as I have a habit of doing).

PaulBender:
Fixes #1844

### DIFF
--- a/xmlhttp.php
+++ b/xmlhttp.php
@@ -333,37 +333,41 @@ else if($mybb->input['action'] == "edit_subject" && $mybb->request_method == "po
 		}
 	}
 
-	// Set up posthandler.
-	require_once MYBB_ROOT."inc/datahandlers/post.php";
-	$posthandler = new PostDataHandler("update");
-	$posthandler->action = "post";
-
-	// Set the post data that came from the input to the $post array.
-	$updatepost = array(
-		"pid" => $post['pid'],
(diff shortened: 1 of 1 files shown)


xmlhttp.php
```diff
@@ -333,37 +333,41 @@
...
-		"tid" => $thread['tid'],
-		"subject" => $subject,
-		"edit_uid" => $mybb->user['uid']
-	);
-	$posthandler->set_data($updatepost);
-
-	// Now let the post handler do all the hard work.
-	if(!$posthandler->validate_post())
+	// Only edit subject if subject has actually been changed
+	if($thread['subject'] != $subject)
```
Destroy666x: else with `xmlhttp_error()` saying that the thread title was the same and wasn't updated would be nice.

Sama34: @Destroy666x what for? If I accidentally trigger the in-line edit subject feature and just click anywhere else to finish it and an error pops up it would just make it more annoying.

Destroy666x: I guess you're right, it doesn't really matter here since nothing happens, but I wouldn't say it's annoying either.

PR #1878, opened 2015-02-20
Fixes #1874 - Nocaptcha missing on newreply.php and newthread.php

### LINKED ISSUE #1874: Nocaptcha missing on newreply.php and newthread.php
Nocaptcha is not visible on newreply.php and newthread.php therefor guests can't post. Apparently $post_captcha->type == 4 is not handled in both files.

ATofighi:
#1874

### DIFF
--- a/member.php
+++ b/member.php
@@ -1782,7 +1782,7 @@ if($mybb->input['action'] == "login")
 				$captcha = $login_captcha->build_hidden_captcha();
 			}
 		}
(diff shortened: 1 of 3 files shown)


Destroy666x:
Same should be done for logging in: https://github.com/mybb/mybb/blob/feature/member.php#L1785

ATofighi:
Done.

PR #1883, opened 2015-02-21
Fixes #1882 Soft Delete option in custom mod tools doesn't work

### LINKED ISSUE #1882: Soft Delete option in custom mod tools doesn't work
The Soft Delete option in custom mod tools doesn't work when you run that tool. Affects both Thread and Post tools.

PaulBender:
Fixes #1882


Destroy666x:
Works for me

PR #1896, opened 2015-03-03
Update task.php

crazycatdevs:
Proposition of better cli-check

cf http://community.mybb.com/thread-167793-post-1144776.html#pid1144776 and following, sometimes checking just 'cli' isn't enough

### DIFF
--- a/task.php
+++ b/task.php
@@ -32,7 +32,7 @@ require_once MYBB_ROOT."inc/functions_task.php";
 
 // Are tasks set to run via cron instead & are we accessing this file via the CLI?
 // php task.php [tid]
-if(PHP_SAPI == "cli")
+if (is_cli()) {
 {
 	// Passing a specific task ID
 	if($_SERVER['argc'] == 2)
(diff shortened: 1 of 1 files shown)


task.php
```diff
@@ -32,7 +32,7 @@
 
 // Are tasks set to run via cron instead & are we accessing this file via the CLI?
 // php task.php [tid]
-if(PHP_SAPI == "cli")
+if (is_cli()) {
```
Destroy666x: There should be no space between `if` and `(`.

Also, the function isn't needed at all (only slows code down a tiny bit), you can just change the conditional.

Stefan-ST: It not just ignores our coding standards but introduces a syntax error.

RageSmirk:
Maybe just: if(defined('STDIN'))

Sama34:
@crazycatdevs your code doesn't work at all. Note @Stefan-ST comment.

Destroy666x:
@crazycatdevs

PR #1900, opened 2015-03-05
Fix #1899 double <br />

### LINKED ISSUE #1899: Extra <br> when all Contact Details set to none
I know it's not crucial, but I found it when I was disabling all of the Contact Details, one of the &lt;br&gt;'s doesn't get removed, so there is a large space above Away information.

And the extra br.

Original thread: [Extra <br> when all Contact Details set to none](http://community.mybb.com/thread-167368.html)

Destroy666x:
Move misplaced `<br />` to beginning of another template to avoid
duplicate.

https://github.com/mybb/mybb/issues/1899


Cu8eR:
Looks fine

PR #1904, opened 2015-03-08
Fix #1903 invalid find_replace_templatesets() $sid code

### LINKED ISSUE #1903: Incorrect find_replace_templatesets() functionality
The following code is incorrect and mixed: https://github.com/mybb/mybb/blob/feature/inc/adminfunctions_templates.php#L33-55 Only the 2 first lines are correct inside the conditional.

And, in my opinion, global templates should always be modified by the function (unless a specific `$sid` is chosen). So this should consider `-1` too: https://github.com/mybb/mybb/blob/feature/inc/adminfunctions_templates.php#L29

More info:
http://community.mybb.com/thread-168112.html
http://community.mybb.com/thread-168051.html

Destroy666x:
As the title says. Makes it possible to modify global templates again.

https://github.com/mybb/mybb/issues/1903

### DIFF
--- a/inc/adminfunctions_templates.php
+++ b/inc/adminfunctions_templates.php
@@ -24,34 +24,17 @@ function find_replace_templatesets($title, $find, $replace, $autocreate=1, $sid=
 {
 	global $db, $mybb;
 
-	$return = false;
+	$return = false;	
+	$template_sets = array(-2, -1);	
+	
+	// Select all templates with that title (including global) if not working on a specific template set
+	$sqlwhere = '>0 OR sid=-1';
 
-	$sqlwhere = '>0';
-
-	$template_sets = array(-2, -1);
-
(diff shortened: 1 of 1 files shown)


inc/adminfunctions_templates.php
```diff
@@ -88,6 +71,8 @@ function find_replace_templatesets($title, $find, $replace, $autocreate=1, $sid=
 		if($master_template['new_template'] != $master_template['template'])
 		{
 			// Update the rest of our template sets that are currently inheriting this template from our master set
+			$sqlwhere = str_replace(' OR sid=-1', '', $sqlwhere);
```
Sama34: I think this two lines could be better, for instance `array_unique` is unnecessary even if there are repeated SIDs it wouldn't matter.

Destroy666x:
No idea why I used `array_unique()` there, corrected.

PR #1910, opened 2015-03-15
Fix #1210 soft deleted posts with showthread.php?pid=x#pidx 2

### LINKED ISSUE #1210: soft deleted posts with showthread.php?pid=x#pidx

Sama34:
Fix #1210 soft deleted posts with showthread.php?pid=x#pidx

I'm unsure why `$visibleonly` and `$visibleonly2` weren't set in the core for normal users?

https://github.com/Sama34/mybb/blob/fix-1210/showthread.php#L744

### DIFF
--- a/showthread.php
+++ b/showthread.php
@@ -120,7 +120,9 @@ if(is_moderator($fid))
 }
 else
 {
-	$ismod = false;
+	$ismod = false;	$ismod = false;
+	$visibleonly = " AND visible=1";
+	$visibleonly2 = "AND p.visible=1 AND t.visible=1";
 }
 
 // Make sure we are looking at a real thread here.
--- a/showthread.php
+++ b/showthread.php
@@ -120,7 +120,7 @@ if(is_moderator($fid))
 }
 else
 {
-	$ismod = false;	$ismod = false;
+	$ismod = false;
 	$visibleonly = " AND visible=1";
 	$visibleonly2 = "AND p.visible=1 AND t.visible=1";
 }


showthread.php
```diff
@@ -120,7 +120,9 @@
 }
 else
 {
-	$ismod = false;
+	$ismod = false;	$ismod = false;
```
Destroy666x: Duplicate.

Also, this has nothing to do with that issue either, but whatever..

Sama34:
It has everything to do with the issue @Destroy666x ...

With the new code implemented in the PR:
https://github.com/mybb/mybb/pull/1865/files#diff-7e30b0b016de305a3f054a785bc62fb6R50

The lastpost action can now result on a existing post which the user can't view. You get the jGrowl error message when landing then. Which makes zero sense.

If you check the code for lastpost action:
https://github.com/Sama34/mybb/blob/fix-1210/showthread.php#L265

You will see it checks for `$visibleonly` and `$visibleonly2` but those aren't set unless you are a moderator.

I come across this while debugging an issue with one of my plugins. A coincidence.

I think this makes the point of we needing more SQA members reviewing modifications.

Destroy666x:
If you set `pid` parameter you don't set action to lastpost... So yes, not connected. But doesn't matter, good catch anyways.

Sama34:
> So yes, not connected.

It is not directly  connected to the original issue but the fix to such issue causes this new.

That is the connection...

Destroy666x:
> but the fix to such issue causes this new.

It didn't cause that issue - the code was wrong regardless of that fix. It only showed you that another bug existed.

Issue #1766, opened 2015-01-17
Greater Than 0 doesn't work for mass mail conditions and find attachment downloads


Destroy666x:
As the title says, the conditions check if the value is other than 0 for some reason, so you can't look for attachments with at least 1 download for example.

https://github.com/mybb/mybb/blob/21c2697a4594178ad5de9f3fabf00fcce87f66cf/inc/functions_massmail.php#L43
https://github.com/mybb/mybb/blob/21c2697a4594178ad5de9f3fabf00fcce87f66cf/admin/modules/forum/attachments.php#L684

It may also affect other places in ACP, haven't checked yet.

effone:
Whats the issue here? Why can't we search for 1 provided 1 is greater than 0 ?

Eldenroot:
I am a little bit confused - in my opinion we should have an ability to search for attachments with total downloads >= 1

Eldenroot:
@Sama34 @euantorano - is it still valid report? Trying to reproduce but without success...

Sama34:
I think he meant the following, take a look at the image, only the `File size is` input should be ignored if `0` value is provided (unless you can upload `0` KB files):
![image](https://user-images.githubusercontent.com/1786584/111883463-cd659700-8978-11eb-99f8-15c70310d747.png)

So the code should perhaps use `isset()` for both `dateuploaded` and `downloads` but `!empty()` for `filesize`.

Something similar should be for mass mail:
![image](https://user-images.githubusercontent.com/1786584/111883643-cf7c2580-8979-11eb-9ce8-7c3a050461fa.png)

For mass mail conditions are stored in serialized format in a `TEXT` column, so using `isset()` should be possible. To ignore these fields they should be left empty by the admin (not `0`).

PR #1548, opened 2014-10-24
Feature #1544 generate_numeric_field improvements

PaulBender:
Feature #1544

PR #1915, opened 2015-03-17
Fixes #1832 Numeric fields - default values lower than minimum

### LINKED ISSUE #1832: Numeric fields - default values lower than minimum
There is still something annoying about the numeric fields - some of them have their default value lower than the minimum. For example Results per page in Find Users is 0 by default while the minimum is 1, so the browser throws an "error" if the admin doesn't change it.

Stefan-ST:
Fixes #1832

### DIFF
--- a/admin/inc/class_form.php
+++ b/admin/inc/class_form.php
@@ -130,7 +130,16 @@ class DefaultForm
 	 */
 	function generate_numeric_field($name, $value="", $options=array())
 	{
-		$input = "<input type=\"number\" name=\"".$name."\" value=\"".(int)$value."\"";
+		if(is_numeric($value))
+		{
+			$value = (float)$value;
+		}
+		else
+		{
+			$value = '';
+		}
+
+		$input = "<input type=\"number\" name=\"{$name}\" value=\"{$value}\"";
 		if(isset($options['min']))
 		{
 			$input .= " min=\"".$options['min']."\"";


Destroy666x:
ACP -> Configuration -> Warning System -> Add New Warning Level (and Edit Warning Level)
The default value and minimum of percentage should be 1 to match the description. Same applies to banning/suspension/moderation length - 0 doesn't make any sense as minimum and default value for them.

JN-Jones:
@Stefan-ST

JN-Jones:
@Destroy666x Do you know any other fields? PR looks good to me so far.

JN-Jones:
Add new profilefield:
- maximum length (required field but "0" doesn't make sense. If "0" means ignored it's not protocoled)
- minimum post count
- display order (1 is imho better than 0 here)
- field length

Edit user:
- ICQ number: defaults to 0 if empty
- Date of birth: year defaults to 0 if empty
- Return date: same

Mass mailing:
- Step 3 (define recipients): The three numeric inputs should be empty instead of 0

Add new group promotion:
- Inputs default to 0 instead of empty

Destroy666x:
I already mentioned this in an old issue:

> default to 0 instead of empty

But can we do anything about that besides making the fields text again? That's the magic of numeric fields in browsers - a number is always displayed and there is no way to get rid of it, I think. And since 0s don't do anything in the fields you listed (ICQ is not shown, birthday and return date are hidden too and mass mail conditions with 0 don't work because of weird conditions: https://github.com/mybb/mybb/issues/1766), I'd say there is no need to change them.

JN-Jones:
At least Opera allows empty fields. But I'm fine with leaving them for now, however we'll have the same problem again when someone is going to fix #1766.

Destroy666x:
If it's cross browser (which i recall not to be, but it may have changed), then it's just a matter of editing this line: https://github.com/mybb/mybb/blob/feature/admin/inc/class_form.php#L133 to something like:
```php
if(is_numeric($value))
{
	$value = (float)$value;
}
else
{
	$value = '';
}
$input = "<input type=\"number\" name=\"{$name}\" value=\"{$value}";
```
which will also allow the usage of decimals (no idea why were they skipped).

JN-Jones:
I've tested it with IE, Opera and Chrome and haven't seen any issues so far. Need to retested with your code when I'm home again.

Sama34:
> (no idea why they were rounded on display, it's not an integer field)

Default `min` value is set to `1` in most (if not all) browsers and the value defines other behaviors apart from that of the _minimum_ input value.

https://blog.isotoma.com/2012/03/html5-input-typenumber-and-decimalsfloats-in-chrome/

Destroy666x:
> The `step` attribute value defines those behaviors.

?

I was of course talking about the `(int)` casting.

Sama34:
Duh! Of course.

JN-Jones:
@Stefan-ST 
> Add new profilefield:
- maximum length (required field but "0" doesn't make sense. If "0" means ignored it's not protocoled)
- minimum post count
- display order (1 is imho better than 0 here)
- field length

Can you fix these? The rest should be retested as mentioned above.

Stefan-ST:
@JN-Jones Wouldn't it make more sense to retest the other fields first?

JN-Jones:
> I've tested it with IE, Opera and Chrome and haven't seen any issues so far. Need to retested with your code when I'm home again.

Also tested empty numeric fields with FF and no problems there too. Also checked this change (https://github.com/mybb/mybb/pull/1915#issuecomment-97746000) which seems to work too.

Stefan-ST:
@JN-Jones So we should apply that instead of fixing all the affected fields individually?

Destroy666x:
@Stefan-ST, you can't fix them individually without correcting the function which inserts `0` due to `(int)` casting of empty values.

JN-Jones:
That's up to you - either apply the fix and check all fields or create a separate issue and only fix the non-empty fields here. But remember that 1.8.5 should be released ASAP.

Stefan-ST:
I've updated the PR.

admin/inc/class_form.php
```diff
@@ -130,7 +130,16 @@ function generate_text_box($name, $value="", $options=array())
...
+		if(is_numeric($value))
+		{
+			$value = (float)$value;
+		}
+		else
+		{
+			$value = '';
+		}
+
+		$input = "<input type=\"number\" name=\"{$name}\" value=\"{$value}";
```
Destroy666x: Just noticed that a closing " is missing for the `value` attribute

JN-Jones: There was a commit which seems to be removed, yes.

JN-Jones:
@Stefan-ST Any comment here? As this already defers 1.8.5 I'd suggest to postpoint this if you don't have time to work on it.

JN-Jones:
@Stefan-ST There were some discussions about wrong default values above, eg the warning points. What's with them?

Stefan-ST:
@JN-Jones I haven't fixed these issues yet.

JN-Jones:
That's why I asked...

JN-Jones:
There's also a "Results per page" field in "Forums -> Attachments" which default to `0`

Stefan-ST:
@JN-Jones "Results per page" is empty for me. From all the fields mentioned here only group promotion fields display a 0 instead of being empty.

Destroy666x:
This should be still fixed:

> ACP -> Configuration -> Warning System -> Add New Warning Level (and Edit Warning Level)
The default value and minimum of percentage should be 1 to match the description. Same applies to banning/suspension/moderation length - 0 doesn't make any sense as minimum and default value for them.

Stefan-ST:
@Destroy666x These fields are not 0 either. They are empty for me which seams to be fine and consistent.

Destroy666x:
But they are required so `1` would be smarter in them by default IMO.

Stefan-ST:
Unless a certain value is recommended we shouldn't set values by default. 1 is actually pretty unlikely for warning points.

Regarding min (and max) values I propose creating a new ticket since it's a much bigger task and low priority.

Stefan-ST:
It would be nice to have this fixed in 1.8.7. ;)

Destroy666x:
> Regarding min (and max) values I propose creating a new ticket since it's a much bigger task and low priority.

Most fields that need to have it already have it: https://github.com/mybb/mybb/pull/1548/files Adding it to these few missing fields isn't big work.

Stefan-ST:
This fixes really annoying issues while min/max isn't important at all. Therefore I prefer to keep this as simple as possible so it doesn't get delayed even further.

Cu8eR:
I agree with @Stefan-ST

PR #1917, opened 2015-03-17
Fixes #1916 Forum Permissions/Group Moderators not deleted if group is d...

### LINKED ISSUE #1916: Forum Permissions/Group Moderators not deleted if group is deleted
If you delete a usergroup, any forum permissions set for that usergroup are not deleted.

The same for group moderators.

PaulBender:
...eleted

Fixes #1916

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -1259,8 +1259,17 @@ if($mybb->input['action'] == "delete")
 
 	if($mybb->request_method == "post")
 	{
-		// Move any users back to the registered group
-		$updated_users = array("usergroup" => 2);
+		if($usergroup['isbannedgroup'] == 1)
+		{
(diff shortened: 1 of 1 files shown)


Destroy666x:
There are 2 more that should be cleaned:
- calendarpermissions: https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_tables.php#L177
- joinrequests: https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_tables.php#L362

Also banned group should be updated to the default banned group or the bans removed:
https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_tables.php#L138
I guess the best thing here would be to add radio buttons to let admin choose 1 out of these 2 actions.

PaulBender:
Fixed.

For banned, I just moved the users to the default banned group.

PR #1920, opened 2015-03-18
Fix #1609 Incomplete forum deletion

### LINKED ISSUE #1609: Incomplete forum deletion
Continuation of https://github.com/mybb/mybb/pull/1605

If you delete a forum, the following things are incorrect:
- [x] all threads and posts are deleted, but records associated with them such as forumsread/threadratings/threadsubscriptions/reportedcontent/polls/polloptions/attachments/etc. are left as orphaned litter in the database. Not good if you delete a forum with many threads. Deleting users works much much better: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L1426
- [x] global threads/posts stats are recounted, but stats for the parent forum(s) and for users that posted in the deleted forum (if the forum settings don't disable user post/thread counting) aren't recounted
- [x] `get_forum()` cache read is used to get the name of the deleted forum, but that doesn't make any sense since it can be selected in the forums table query 2 lines below too

Sama34:
Fix #1609 Incomplete forum deletion

### DIFF
--- a/admin/modules/forum/management.php
+++ b/admin/modules/forum/management.php
@@ -1989,8 +1989,21 @@ if($mybb->input['action'] == "delete")
 				$db->delete_query("forums", "CONCAT(',',parentlist,',') LIKE '%,$fid,%'");
 		}
 
-		$db->delete_query("threads", "fid='{$fid}' {$delquery}");
-		$db->delete_query("posts", "fid='{$fid}' {$delquery}");
+		require_once MYBB_ROOT.'inc/class_moderation.php';
+		$moderation = new Moderation();
+
+		$query = $db->simple_select("threads", "tid", "fid='{$fid}' {$delquery}");
+		while($tid = $db->fetch_field($query, 'tid'))
+		{
+			$moderation->delete_thread($tid);
+		}
+
+		$query = $db->simple_select('posts', 'pid', "fid='{$fid}' {$delquery}");
+		while($pid = $db->fetch_field($query, 'pid'))
+		{
+			$moderation->delete_post($pid);
+		}
+
 		$db->delete_query("moderators", "fid='{$fid}' {$delquery}");
 		$db->delete_query("forumsubscriptions", "fid='{$fid}' {$delquery}");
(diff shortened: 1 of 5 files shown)


admin/modules/forum/management.php
```diff
@@ -1989,11 +1989,25 @@
...
+		while($tid = $db->fetch_field($query, 'tid'))
+		{
+			$moderation->delete_thread($tid);
+		}
+
+		$query = $db->simple_select('posts', 'pid', "fid='{$fid}' {$delquery}");
+		while($pid = $db->fetch_field($query, 'pid'))
+		{
+			$moderation->delete_post($pid);
+		}
```
JN-Jones: The `delete_thread` function already deletes the posts for those threads. So normally that forum shouldn't have any more posts after the threads are deleted.

Destroy666x: Yep, not needed. Except that, you forgot about `forumsread`: https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_tables.php#L308 And the old code that updates global stats is not needed anymore due to: https://github.com/mybb/mybb/blob/50e5a961708a004457dad05589db2d2770b0419c/inc/functions.php#L2568

Sama34: Yeah, probably a pagination would be nice. There could be 23+ queries per thread deletion and 5+ hooks.

That or replicate pretty much the moderation code for forum deletions.

@JN-Jones

Destroy666x: > That or replicate pretty much the moderation code for forum deletions.

We could replicate the code with some `IN` queries indeed, but I think pagination would be better so that plugins that hook into `delete_thread()` would do their stuff too.

JN-Jones: Yep, simply modify the `check_proceed` function would be fine.

JN-Jones:
@Sama34

JN-Jones:
I'd like to get this included in 1.8.6 @Sama34

JN-Jones:
@Sama34 ?

JN-Jones:
Closing this as Omar is too busy atm to work on it and I'll take it over to get it in 1.8.9

PR #1921, opened 2015-03-18
Fix #1589 More proper URL validation

### LINKED ISSUE #1589: More proper URL validation
Some places don't validate URLs properly (they don't take `https://` and other cases into consideration). For example:
https://github.com/mybb/mybb/blob/feature/inc/functions.php#L3721
https://github.com/mybb/mybb/blob/feature/admin/modules/config/smilies.php#L637 (and one more in that file)
https://github.com/mybb/mybb/blob/feature/admin/modules/config/post_icons.php#L468
https://github.com/PaulBender/mybb/blob/ce5804c7e30eaaa583c8bd7b4d61e3c9eda1b9a6/admin/modules/config/attachment_types.php#L322
https://github.com/mybb/mybb/blob/feature/admin/modules/user/users.php#L1455
https://github.com/mybb/mybb/blob/feature/usercp.php#L2250

There may be more of them.

Sama34:
Fix #1589 More proper URL validation


Sama34:
Seems like conflicts arise.. unsure how to fix from phpStorm... xD

admin/modules/config/smilies.php
```diff
@@ -643,7 +643,7 @@
 	while($smilie = $db->fetch_array($query))
 	{
 		$smilie['image'] = str_replace("{theme}", "images", $smilie['image']);
-		if(my_strpos($smilie['image'], "p://") || substr($smilie['image'], 0, 1) == "/")
+		if(filter_var($smilie['image'], FILTER_VALIDATE_URL))
```
Destroy666x: `FILTER_VALIDATE_URL` doesn't allow relative paths (2nd part of the conditional), does it?

Destroy666x:
Also, AFAIK `filter_var()` is not very reliable - for example checking IDNs may result in `false`: http://www.d-mueller.de/blog/why-url-validation-with-filter_var-might-not-be-a-good-idea/

inc/datahandlers/user.php
```diff
@@ -294,23 +294,10 @@ function verify_website()
...
-
-		// Does the website start with http(s)://?
-		if(my_strtolower(substr($website, 0, 4)) != "http")
-		{
-			// Website does not start with http://, let's see if the user forgot.
-			$website = "http://".$website;
-		}
-
-		if(!filter_var($website, FILTER_VALIDATE_URL))
+		if($website && !filter_var($website, FILTER_VALIDATE_URL))
```
Destroy666x: And this is quite a big change - earlier URLs without protocol like `www.something.com` could be entered by less advanced users + not sure why you changed `false` to `true`.

Sama34: > Could be entered by less advanced users

Invalid URLs as well.

Sama34:
Oh I see.

inc/datahandlers/user.php
Destroy666x: Well, `filter_var()` doesn't allow more valid URLs. In the link I posted you can find some examples like `not://valid`, another reason to use a regex

Sama34: I will probably add a custom function.

Destroy666x:
Here's an interesting regex comparison I found recently, BTW: https://mathiasbynens.be/demo/url-regex

Sama34:
I would assume that regex checking (if ever added) would occur only if `filter_var` fails, am I correct?

I fear the load this may cause. If any reasonable.

Destroy666x:
> I would assume that regex checking (if ever added) would occur only if filter_var fails, am I correct?

`filter_var` doesn't need to be used at all if there's a good regex, as you can see in the comparison.

> I fear the load this may cause. If any reasonably noticeable.

?
I don't think it will be a matter of more than several miliseconds for one call per page, which is the case in all changed lines in this PR.

Sama34:
Plugins might do use of the added function.

euantorano:
I hadn't realised the issues with `filter_var` and it was me who suggested it a while ago - oops! Having seen the link @Destroy666x provided, it's probably best to use a Regular Expression in this case. Additionally, as this is a commonly used validation, it might be good to add a function `my_validate_url` or something to `inc/functions.php`.

Destroy666x:
> Plugins might do use of the added function.

And if they think the fully proper regex check is too slow to be called 1000 times on a page, they can still use their own suitable check instead. They're not limited to our solution.

JN-Jones:
So what's the conclusion? Adding a function `my_check_url` (or something like that) which uses a Regex and is used instead of `filter_var`? @Sama34 @Destroy666x @euantorano

euantorano:
That would be my preferred approach, yes.

Destroy666x:
Yep.

JN-Jones:
@Sama34

inc/functions.php
```diff
@@ -962,7 +962,7 @@ function redirect($url, $message="", $title="", $force_redirect=false)
 
 		run_shutdown();
 
-		if(my_substr($url, 0, 7) !== 'http://' && my_substr($url, 0, 8) !== 'https://' && my_substr($url, 0, 1) !== '/')
+		if(!!filter_var($url, FILTER_VALIDATE_URL))
```
Sama34: Oops nobody noticed :P

Sama34:
Which regrex do you want me to use? This?
https://gist.github.com/dperini/729294#file-regex-weburl-js

I'm not that good with regex.

Destroy666x:
The diegoperini one got the highest score, so probably that. You used it as I see but you didn't include flags - why? The case insensitivity and Unicode flags are required for proper functionality, not sure about `S` but it seems to be some kind of optimization.

Sama34:
> why?

Cause I have no idea when it comes to REGEX / preg_* functions :P

> relative paths should also work in some cases

Why? From the code it says "if URL then use URL -> relative otherwise"

> no reason to remove the datahandler code responsible for adding `http://`

No reason to keep it either. If the user entered an protocol less URL then the user will just need to try again.

Any user "owning" a "website" would know the basics.

Destroy666x:
> Any user "owning" a "website" would know the basics.

I see many people posting links without the scheme/protocol and just with `www.` at the beginning, even forum owners (!) in our support forums. I don't think many of them know how a proper URL should look.

So since there are no detailed errors like "Please begin your URL with...", this difference will matter and I disagree that
> No reason to keep it either.

JN-Jones:
@Destroy666x @Sama34 What exactly is missing here? Only the special `http://` handling?

JN-Jones:
Bump

Destroy666x:
> What exactly is missing here?

The regex has no required flags, there's no reason to change the code in the datahandler to make it less user-friendly, some relative paths were incorrectly removed and there may be more places where this function should be used too.

euantorano:
So does this work as is, we just need to make sure it's used everywhere? if so, can we try and do so before the next release (so ASAP)? I've tested the patched version and it does seem to work.

Destroy666x:
Not quite, I mentioned several problems one comment above.

euantorano:
Okay, I'll take a look and see if I can fix those issues up.

Sama34:
I updated the code to allow relative paths and to allow users to insert websites without protocol in the UCP.

inc/functions.php
```diff
@@ -8239,3 +8239,22 @@ function copy_file_to_cdn($file_path = '', &$uploaded_path = null)
...
+ * Validate an url
+ *
+ * @param string $url The url to validate.
+ * @param bool $relative_path Whether or not the url could be a relative path.
+ *
+ * @return bool Whether this is a valid url.
+ */
+function my_validate_url($url, $relative_path=false)
+{
+	if($relative_path && my_substr($url, 0, 1) == '/' || preg_match('^(?:(?:https?|ftp)://)(?:\S+(?::\S*)?@|\d{1,3}(?:\.\d{1,3}){3}|(?:(?:[a-z\d\x{00a1}-\x{ffff}]+-?)*[a-z\d\x{00a1}-\x{ffff}]+)(?:\.(?:[a-z\d\x{00a1}-\x{ffff}]+-?)*[a-z\d\x{00a1}-\x{ffff}]+)*(?:\.[a-z\x{00a1}-\x{ffff}]{2,6}))(?::\d+)?(?:[^\s]*)?$', $url))
```
Destroy666x: The regex is still missing flags though. There's no reason to skip them and as I mentioned right now it doesn't allow e.g. HTTP://GOOGLE.COM since only lowercase letters are allowed.

admin/modules/config/smilies.php
```diff
@@ -726,10 +726,9 @@
 	while($smilie = $db->fetch_array($query))
 	{
 		$smilie['image'] = str_replace("{theme}", "images", $smilie['image']);
-		if(my_strpos($smilie['image'], "p://") || substr($smilie['image'], 0, 1) == "/")
+		if(my_validate_url($smilie['image'], true))
 		{
 			$image = $smilie['image'];
-			$smilie['image'] = str_replace("{theme}", "images", $smilie['image']);
```
Destroy666x: Why was this removed?

global.php
```diff
@@ -847,7 +847,7 @@
 $contact_us = '';
 if(($mybb->settings['contactlink'] == "contact.php" && $mybb->settings['contact'] == 1 && ($mybb->settings['contact_guests'] != 1 && $mybb->user['uid'] == 0 || $mybb->user['uid'] > 0)) || $mybb->settings['contactlink'] != "contact.php")
 {
-	if(my_substr($mybb->settings['contactlink'], 0, 1) != '/' && my_substr($mybb->settings['contactlink'], 0, 7) != 'http://' && my_substr($mybb->settings['contactlink'], 0, 8) != 'https://' && my_substr($mybb->settings['contactlink'], 0, 7) != 'mailto:')
+	if(!my_validate_url($mybb->settings['contactlink'], true))
```
Destroy666x: This doesn't consider `mailto:` from the previous conditional

inc/functions.php
Sama34: Yep, you are correct. I updated the code, tested and seems to work now except for two false positives.

According to the above page the following should be invalid URLs:
```
http://a.b--c.de/
http://www.foo.bar./
```

But even the code supplied by them marks them as valid. Tough the first one seems valid to myself so..

Destroy666x:
Seems to be all ok now, but I'll let anyone else recheck.

PR #1923, opened 2015-03-18
Fix #1440 Soft delete doesn't show up in mod tools when searching thread...

### LINKED ISSUE #1440: Soft delete doesn't show up in mod tools when searching threads
Here's what it looks like in forum display:
http://i.imgur.com/wpXUNVJ.png

Here's what it looks like when searching:
http://i.imgur.com/4xpiUIV.png

Original thread: [Soft delete doesn't show up in mod tools when searching threads](http://community.mybb.com/thread-159443.html)

Sama34:
Fix #1440 Soft delete doesn't show up in mod tools when searching threads


Stefan-ST:
The version codes of the templates you've modified needs to be updated.

Sama34:
Done.

Please note that even though support for post tools was there to begin to, they need to be tested because I noted some strange behaviors, specially in split post.

If somebody can confirm that would be great.

install/resources/mybb_theme.xml
```diff
@@ -10953,7 +10953,7 @@ if(use_xmlhttprequest == "1")
 </body>
 </html>]]></template>
 		<template name="search_results_posts_inlinecheck" version="120"><![CDATA[<td class="{$bgcolor}" align="center" style="white-space: nowrap"><input type="checkbox" class="checkbox" name="inlinemod_{$post['pid']}" id="inlinemod_{$post['pid']}" value="1" style="vertical-align: middle;" {$inlinecheck}  /></td>]]></template>
-		<template name="search_results_posts_inlinemoderation" version="1804"><![CDATA[<script type="text/javascript" src="{$mybb->asset_url}/jscripts/inline_moderation.js?ver=1804"></script>
+		<template name="search_results_posts_inlinemoderation" version="1805"><![CDATA[<script type="text/javascript" src="{$mybb->asset_url}/jscripts/inline_moderation.js?ver=1804"></script>
```
JN-Jones: Should be `1806` now. Same below.

Destroy666x:
> because I noted some strange behaviors, specially in split post

Aren't these mulitpost tools be supposed to be for a specific thread only?

Sama34:
Done.

Destroy666x:
> Please note that even though support for post tools was there to begin to, they need to be tested because I noted some strange behaviors, specially in split post.
If somebody can confirm that would be great.

I didn't notice any problems with any tool, tested on both posts in different/the same thread(s). Feel free to opan a new issue if you can still reproduce it though.

PR #1924, opened 2015-03-18
Fix #1906 Wrong information in profile/reputation report PMs

### LINKED ISSUE #1906: Wrong information in profile/reputation report PMs
As the title says. Some conditionals are needed for this line (and one above):
https://github.com/mybb/mybb/blob/feature/inc/functions_modcp.php#L202
for `$post['subject']`, `$thread['subject']` and `get_post_link()`

EDIT or actually both lang strings should be under a conditional because they say about post only:
https://github.com/mybb/mybb/blob/feature/inc/languages/english/messages.lang.php#L92
https://github.com/mybb/mybb/blob/feature/inc/languages/english/messages.lang.php#L151

http://community.mybb.com/thread-164380-post-1147102.html#pid1147102

Sama34:
Fix #1906 Wrong information in profile/reputation report PMs


inc/functions_modcp.php
```diff
@@ -181,11 +181,36 @@ function send_report($report)
 		}
 	}
 
-	while($mod = $db->fetch_array($query))
+	$lang_string_subject = "emailsubject_report{$report_type}";
+	$lang_string_message = "email_report{$report_type}";
+
+	if(!$lang_string_subject || !$lang_string_message)
```
Destroy666x: Any reason for this check? It will never happen, these strings are always set to `emailsubject_report` and `email_report` values above.

Sama34: Plugin reporttypes.

Destroy666x: What about them? Even if there are 100 plugins installed that change report types `$lang_string_subject` and `$lang_string_message` will be always filled in the 2 lines above.. So that check doesn't do anything useful.

It's something like:
```php
$yes = true;

if($yes === false)
{
    //...
}
```

Stefan-ST: @Sama34 I think you meant something like $lang->$lang_string_subject.

Sama34: Oops.

> add a hook to that function

Forgot to add variables to global scope. You can hook at `report_do_report_start`.

inc/functions_modcp.php
```diff
@@ -140,9 +140,9 @@ function fetch_forum_announcements($pid=0, $depth=1)
  * @param array Array of reported content
  * @return bool True if PM sent
  */
-function send_report($report)
+function send_report($report, $report_type='post')
```
JN-Jones: The new parameter needs to be added to the doc block above

inc/functions_modcp.php
```diff
@@ -228,7 +252,7 @@ function send_report($report)
 		return $pminfo;
 	}
 
-	return false;
+	return true;
```
JN-Jones: Any reason why you changed this? First it may break things (plugins) and secondly the doc block is wrong then too.

JN-Jones:

Sama34:
Should be correct now.

JN-Jones:
You still need to add the new parameter to the doc block ;)

JN-Jones:
@Sama34

Sama34:
Done.

JN-Jones:
Seems to look good @euantorano @Destroy666x

JN-Jones:
I'll merge this tomorrow too if noone has any objections.

PR #1927, opened 2015-03-20
Fixes #1926 Upgrades from <=1.2.14 fails

### LINKED ISSUE #1926: Upgrades from <=1.2.14 fail
In upgrade12.php rebuild_stats() is called which queries deletedthreads and deletedposts causing a MySQL error (unknown column).

Stefan-ST:
Fixes #1926

### DIFF
--- a/install/resources/upgrade12.php
+++ b/install/resources/upgrade12.php
@@ -1063,8 +1063,13 @@ function upgrade12_dbchanges4()
 	$db->write_query("ALTER TABLE ".TABLE_PREFIX."adminoptions ADD defaultviews TEXT NOT NULL");
 	$db->update_query("adminoptions", array('defaultviews' => my_serialize(array('user' => 1))));
 
-	require_once MYBB_ROOT."inc/functions_rebuild.php";
-	rebuild_stats();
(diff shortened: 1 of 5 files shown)


Sama34:
That was an interesting one. I didn't test everything after the upgrade but there is an issue with the global.css style sheet right after upgrading:
mybb1214

The administrator needs to revert the style sheet for the default 1.8 theme to be correctly displayed (assuming the default 1.2.14 theme is being used as in my test environment).

I don't consider the need for that to be fixed considering administrators will probably use a new theme or something so I will merge this.

Stefan-ST:
@Sama34 Upgrading from 1.6.x to 1.8.x doesn't reset the stylesheets either, does it?

JN-Jones:
It only updates the master stylesheet, if you haven't modified it it should be updated otherwise not.

Sama34:
I didn't edit `global.css` at all before upgrading so I would expect it should still be inhered from its parent after the upgrade. I had to revert it for the theme to work.

But there as always been issues like this in all upgrades IIRC anyways.

ATofighi:
@Sama34 do you clean the cache of your browser?

Sama34:
I'm pretty sure cache was not the issue, yes.

PR #1935, opened 2015-03-25
Fix #1934 wrong awaitngactivation time check

### LINKED ISSUE #1934: Wrong awaitingactivation check
https://github.com/mybb/mybb/blob/feature/global.php#L727
It doesn't make any sense because `$awaitingusers['time']` will always be smaller than tommorow's time, thus the 2 cache rebuild queries are executed on every page load.

Destroy666x:
The 24 hours were added to the wrong site of comparison and resulted in
2 queries on every page load.
http://github.com/mybb/mybb/issues/1934

### DIFF
--- a/global.php
+++ b/global.php
@@ -724,7 +724,7 @@ if($mybb->settings['awactialert'] == 1 && $mybb->usergroup['cancp'] == 1)
 {
 	$awaitingusers = $cache->read('awaitingactivation');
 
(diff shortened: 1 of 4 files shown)


Sama34:
Makes sense.

Destroy666x:
Also fixed the stats refresh time, so this is finished.

PR #1937, opened 2015-03-28
Fix #1930 typo in validation selectors

### LINKED ISSUE #1930: Bug in checkbox validation
Hi,

If a textbox profile field required to be filled in during registration, it will break checkbox validation for password, image verification and security question boxes.(with or without regular expression)

In this test, Location box break the checkbox validation.
[attachment=34030]

Original thread: [Bug in checkbox validation](http://community.mybb.com/thread-168753.html)

Destroy666x:
Added missing closing `]` to required fields selectors so that the rest
of the validator doesn't break.

https://github.com/mybb/mybb/issues/1930

### DIFF
--- a/member.php
+++ b/member.php
@@ -1035,19 +1035,19 @@ if($mybb->input['action'] == "register")
 					{
 						if($type == "textarea")
 						{
-							$inp_selector = "$('textarea[name=\"profile_fields[{$field}]\"')";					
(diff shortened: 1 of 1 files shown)


Sama34:
Seems like all browsers but Firefox in my machine give a banana about this.

PR #1938, opened 2015-03-29
Fix  #1674: Custom MyCodes not accepted by editor.

### LINKED ISSUE #1674: Custom MyCodes not accepted by editor.

martec:
Fix  #1674


Sama34:
I honestly doubt this will make it to 1.8.5 but can you explain what your change does?

martec:
@Sama34 
changes made in this file https://github.com/samclarke/SCEditor/blob/master/src/plugins/bbcode.js
after merged with https://github.com/mybb/mybb/blob/feature/jscripts/sceditor/jquery.sceditor.min.js
and replaced with jquery.sceditor.bbcode.min.js

and removed duplicated tags from bbcodes_sceditor.js

Sama34:
I will take a look later but I really doubt this makes it to 1.8.5

martec:
no problem if it is to 1.8.6

Destroy666x:
So this is still the same solution (modify all BBCodes) without duplicating the code?

martec:
@Destroy666x 
Yes, exactly.

install/resources/mybb_theme.xml
```diff
@@ -13793,7 +13793,7 @@ if(use_xmlhttprequest == "1")
 </tr>]]></template>
 		<template name="codebuttons" version="1805"><![CDATA[<link rel="stylesheet" href="{$mybb->asset_url}/jscripts/sceditor/editor_themes/{$theme['editortheme']}" type="text/css" media="all" />
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/sceditor/jquery.sceditor.bbcode.min.js?ver=1805"></script>
-<script type="text/javascript" src="{$mybb->asset_url}/jscripts/bbcodes_sceditor.js?ver=1804"></script>
+<script type="text/javascript" src="{$mybb->asset_url}/jscripts/bbcodes_sceditor.js?ver=1805"></script>
```
JN-Jones: Should be `1806` now. Also the template version needs to be bumped.

martec:
@JN-Jones 
done

JN-Jones:
Produced conflicts with one of the other PRs @martec

Destroy666x:
Note: the bbcode plugin file should be minified like it was before.

Also, keep in mind that merging this will most likely make SCEditor much harder.

martec:
Then better to first decide whether to use this solution or look for another. Because solve concflict issue is pretty boring.

Sama34:
+1

Destroy666x:
We should probably at least generate some kind of a readable diff or something for easier comparison at a later stage.

martec:
i found one problem in this...

not possible set if will become block level element or inline element...

explain better....
if use something this [code=xxx][/code] will become block level element... because code is block level element.
 if use something this [b=xxx][/b] will become inline element... because bold is inline element...

same analogy of div element (block element) and span (inline element)

dequeues:

martec:
not good idea use this...
so discard it.
better recreate all tags that mybb using even causing increase file size.

dequeues:
Closing this then.

PR #1940, opened 2015-03-30
Fixes #1939 format_avatar caching issue

### LINKED ISSUE #1939: format_avatar caching issue
The format_avatar function caches the avatar on the first call. However the cache doesn't consider the max_dimensions parameter. Eg I use the function sometimes with a max dimension of "17x17" to display the avatar in a text, however on a page where the function is called before without (or with different) max dimensions the cache will contain the bigger image and return it which breaks my complete text.

Original thread: [format_avatar caching issue](http://community.mybb.com/thread-164490.html)

JN-Jones:
#1939 @Destroy666x @euantorano

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3011,6 +3011,11 @@ function format_avatar($avatar, $dimensions = '', $max_dimensions = '')
 		$dimensions = $mybb->settings['useravatardims'];
 	}
 
+	if(!$max_dimensions)
+	{
+		$max_dimensions = $mybb->settings['maxavatardims'];
+	}
+
 	// An empty key wouldn't work so we need to add a fall back
 	$key = $dimensions;
 	if(empty($key))
@@ -3018,14 +3023,9 @@ function format_avatar($avatar, $dimensions = '', $max_dimensions = '')
 		$key = 'default';
 	}
 
-	if(isset($avatars[$avatar][$key]))
+	if(isset($avatars[$avatar][$key][$max_dimensions]))
 	{
-		return $avatars[$avatar][$key];
-	}
-
-	if(!$max_dimensions)
-	{
-		$max_dimensions = $mybb->settings['maxavatardims'];
+		return $avatars[$avatar][$key][$max_dimensions];
 	}
 
 	$avatar_width_height = '';
@@ -3051,12 +3051,12 @@ function format_avatar($avatar, $dimensions = '', $max_dimensions = '')
 		}
 	}
 
-	$avatars[$avatar][$key] = array(
+	$avatars[$avatar][$key][$max_dimensions] = array(
 		'image' => htmlspecialchars_uni($mybb->get_asset_url($avatar)),
 		'width_height' => $avatar_width_height
 	);
 
-	return $avatars[$avatar][$key];
+	return $avatars[$avatar][$key][$max_dimensions];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


euantorano:
Looks good to me :+1:

Destroy666x:
A nice solution, but it doesn't work because there is a typo - `$key = $dimensions;` should be `$key = $max_dimensions;`

JN-Jones:
Why should it use the max dimensions and not the actual dimensions? `$dimensions` is used to resize the avatar.

Destroy666x:
Because the dimensions provided in funciton parameter are always the same for the same avatar image, so the fix doesn't work. Just test it - this is after applying `$key = $dimensions`: http://nooo.hol.es/mybb16/showthread.php?tid=12 After changing what I said the log avatars have correct 40x40.

JN-Jones:
That forum is password protected.

Anyways, your change doesn't make any sense and instead would reintroduce the error again. Eg calling the function without a max dimension or with the same max dimension but with different dimensions wouldn't work. It sounds more like the showthread calls the function in a wring way. I'll take a look at showthread tomorrow but caching the max dimension is definitely the wrong way.

Destroy666x:
Password is test

Destroy666x:
> It sounds more like the showthread calls the function in a wring way.

Nope, same thing happens everywhere. For example profile: http://nooo.hol.es/mybb16/member.php?action=profile&uid=4 The 4 avatars should be max 30x30.

> calling the function without a max dimension

The default max dimension setting mentioned above would be used then.

> or with the same max dimension but with different dimensions 

This is indeed a problem, but I don't see a case where the same avatar image (so image with the same path) would have different dimensions. Is that even technically possible? Of course different

JN-Jones:
Hmm, I see where the issue is. Need to check something tomorrow to see what the best fix is.

JN-Jones:
The issue is that you can call the function in two ways:
- `format_avatar($user['avatar'], $user['avatardimensions'], '16x16');`
- `format_avatar($user['avatar'], '16x16');`
While the first is how the function is designed and how it's called in the core the second is used in some plugins (including @euantorano's and mine). Though that disables the scalling we should avoid breaking plugins and therefor I've changed the cache to `$avatars['avatar']['dimensions']['max_dimensions']` which should work in all cases.

Destroy666x:
Looks good now.

Destroy666x:
But now that `$key` is still reserved for `$dimensions`, I think another fallback should be added for `$max_dimensions` in case both the setting and parameter are empty too.

JN-Jones:
`$max_dimensions` should never be empty:
```php
	if(!$max_dimensions)
 	{
		$max_dimensions = $mybb->settings['maxavatardims'];
 	}
```

So using `$max_dimensions` as key shouldn't be a problem.

`$dimensions` on the other hand can be empty (eg if you only want the correct avatar url without any dimensions) and therefor can't be used as key directly.

Destroy666x:
It can be empty. Just look at the setting's description: https://github.com/mybb/mybb/blob/feature/install/resources/settings.xml#L1128 `If this is left blank then there will be no dimension restriction.`

Destroy666x:
But as I see that either the description is misleading for admins or `format_avatar()` lacks checks for that case: https://github.com/JN-Jones/mybb/blob/feature/inc/functions.php#L3041
Should be:
```php
if($max_dimensions && $dimensions[0] > $max_width || $dimensions[1] > $max_height)
```
Otherwise if the 3rd argument is blank and the setting is blank, avatars aren't displayed at all.

JN-Jones:
Missed that comment and only looked at the source. Added the check and `$key2` to avoid the empty key issues.

PR #1943, opened 2015-04-02
New settings options for admin

crazycatdevs:
Proposal as an answer to http://community.mybb.com/thread-169202.html : Addition of *forumalone* and *groupalone* settings

The task.php file **must be ignored** (I revert to original, forced to use the web interface)

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -1327,6 +1327,16 @@ if($mybb->input['action'] == "change")
 					checkAction('{$element_id}');
 				</script>";
 			}
+			else if($type[0] == "forumalone")
+			{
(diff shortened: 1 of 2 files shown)


Destroy666x:
@crazycatdevs, you're only removing some empty lines in task.php, why? Makes it only harder to read.

> The task.php file <b>must be ignored</b>

We can't ignore a file which is going to be merged into our software.. You could just create a different branch in your repository. Also, what about the fix for that issue?

crazycatdevs:
Sorry, I'm not used to work with github, and forced to use the web interface.
I'll redo that tonight.

BTW, the fix in admin/modules/config/settings.php is ok.

crazycatdevs:
#1944 is the good patch, no more "playing" with task.php

PR #1944, opened 2015-04-03
Adding new setting types

crazycatdevs:
Following http://community.mybb.com/thread-169202.html , addition of forumalone and groupalone types of settings


Destroy666x:
Saving the setting doesn't work with this code. `'select['.$setting['name'].'][]'` doesn't look correct for me. It should rather be `$element_name` - then it works. Please correct it.

Also, both new settings should be added here with a matching lang strings (for example `Single Forum/Group Selection Box`) too:
https://github.com/crazycatdevs/mybb/blob/patch-1/admin/modules/config/settings.php#L413
https://github.com/crazycatdevs/mybb/blob/patch-1/admin/modules/config/settings.php#L629
so that you can add/edit settings to use these optioncodes.

Sama34:
- Please use `singleforumselect` and `singlegroupselect` or something of the sort @Destroy666x 
- Make use of the `main_option` option.
- Remove the `size` option.

`$lang->none` should be available for the `main_option` option IIRC.

crazycatdevs:
I think it's now finished.
forumalone and groupalone are in the two needed arrays and well placed, I also moved the elseif clause for forumalone just after forumselect.
And the main_option ($lang->none) is now present.

admin/modules/config/settings.php
```diff
@@ -634,7 +636,9 @@
 		"onoff" => $lang->onoff,
 		"select" => $lang->select,
 		"forumselect" => $lang->forum_selection_box,
+		"forumselectsingle" => $lang->forumselectsingle,
 		"groupselect" => $lang->group_selection_box,
+		"groupselectsingle" => $lang->groupselectsingle
```
Destroy666x: Comma missing here.

crazycatdevs:
Woops, sorry, corrected.
And I've removed the extra commas after "cpstyle" => $lang->cpstyle (last line of the arrays)

Sama34:
There are some settings that require now updating.
- Purge Spammer > Ban Usergroup

You need to update the `./install/resources/settings.xml` and `./install/resources/upgrade33.php` files.

crazycatdevs:
./install/resources/settings.xml modified, not sure that any modification has to be done in ./install/resources/upgrade33.php : the setting is a numeric field, so no transformation is needed imho.

Destroy666x:
I think @Sama34 meant opdating the optionscode there too. Not sure, but shouldn't the `sync_settings()` function take care of it? I remember that someone said somehwere that it's not needed, I'll test it later just to be sure.

Sama34:
Seems like you are correct @Destroy666x

crazycatdevs:
I've corrected the variable names in language file, to follow the naming rules.

What about `sync_settings()` ? Have I to use it in the `./install/resources/upgrade33.php` file ?

Destroy666x:
No, it's done on every upgrade.

crazycatdevs:
Or can I send an array ? seems less dangerous for retro-compatiblity.
in settings.php, line 1342 :
`$setting_code = $form->generate_group_select($element_name, array($selected_value), array('id' => $element_id, 'main_option' => $lang->none));`

Sama34:
```PHP
	function generate_group_select($name, $selected=array(), $options=array())
	{
		// *** //

		if(!is_array($selected))
		{
			$selected = array($selected);
		}

		// *** //

		foreach($groups_cache as $group)
		{
			$selected_add = "";
			if(in_array($group['gid'], $selected))
			{
				$selected_add = " selected=\"selected\"";
			}

			// *** //
		}

		// *** //
	}
```

crazycatdevs:
All right, it's done.
For my curiosity, can you tell me why you prefer modifying this function rather than the call in settings.php ?

Destroy666x:
> For my curiosity, can you tell me why you prefer modifying this function rather than the call in settings.php?

Because it's a lot more logical and useful to extend the function itself. Form elements can be used for non-settings too.

crazycatdevs:
Ok, I'll remember that

Sama34:
You should remove the `is_array` _if_ check in that method too since it will now always be true anyways.

Destroy666x:
And
```php
        if(!is_array($selected))
        {
            $selected = array($selected);
        }
```
should be outside of the foreach loop, there is no need to execute it for each group.

crazycatdevs:
Yes, sorry guys, I sometimes want to do too fast.

Seems good now, I'll try being more careful for the next times.

Destroy666x:
https://github.com/mybb/mybb/issues/1997

PR #1948, opened 2015-04-06
Fix #1947 checkbox optionscode doesn't work

### LINKED ISSUE #1947: checkbox setting type doesn't work as expected.
Currently a checkbox setting type wouldn't consider multiple user selections. It works pretty much just like a radio button pretending to allow multiple selections when it doesn't.

Original thread: [checkbox setting type doesn't work as expected.](http://community.mybb.com/thread-169265.html)

Destroy666x:
As the title says - it worked like radios since 1.6 - only one value was saved. Multiple selections are saved to database separated with a comma now. A bit hacky solution with a hidden field, but it works and I don't think there is anything simplier than that.
https://github.com/mybb/mybb/issues/1947

Also added some missing fields that shouldn't be used as a hidden captcha.
EDIT: and fixed 2 peeker regexes to check for a complete match rather than a partial match.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -889,6 +889,9 @@ if($mybb->input['action'] == "change")
 				'email',
 				'email2',
 				'imagestring',
+				'imagehash',
+				'answer',
+				'question_id',
 				'allownotices',
 				'hideemail',
 				'receivepms',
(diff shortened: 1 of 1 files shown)


Destroy666x:
Could anyone check? Would be nice to merge it into 1.8.7 and it worked without any problem for me.

PaulBender:
Works for me

Cu8eR:
@Destroy666x @PaulBender There are a lot of PR which should be merged. Anyway I understand that you are working on MyBB 2.0 so there is lower activity for MyBB 1.8

PR #1952, opened 2015-04-07
Fix #1373 profile field permissions

### LINKED ISSUE #1373: Profie fields not checking permissions correctly
Yet another feature that is not checked correctly for permissions - if a user has 5 posts and fills a profile field, which will be later set by admin to minimum 10 posts, it will still display in postbit/profile for that user.

EDIT: same applies to `editableby` in both profile/postbit. And the Display on profile doesn't seem to accept 0.

EDIT2: postbit doesn't check `viewableby` either.. And empty fields need to be hidden there.

Destroy666x:
Adds missing `viewableby`/`editableby`/`postnum` checks.

Hides empty fields on postbit and fields that shouldn't be displayed on
profile.

`viewableby`/`editableby` set to All for new fields by default (not sure
why they were None in the first place).

Fixes regex error when a field is not required and not filled.

EDIT: also modified the ACP/MCP/registration check in userhandler, but not sure if it's a good solution..

https://github.com/mybb/mybb/issues/1373

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -401,64 +401,81 @@ function build_postbit($post, $post_type=0)
 		}
 
 		// Display profile fields on posts - only if field is filled in
-		if(is_array($profile_fields))
+		if($post_type == 0)
 		{
-			foreach($profile_fields as $field)
+			if(!isset($profile_fields))
 			{
-				$fieldfid = "fid{$field['fid']}";
-				if(!empty($post[$fieldfid]))
+				$profile_fields = $cache->read('profilefields');
+			}
+			
+			if(is_array($profile_fields))
+			{
+				foreach($profile_fields as $field)
 				{
-					$post['fieldvalue'] = '';
-					$post['fieldname'] = htmlspecialchars_uni($field['name']);
-
-					$thing = explode("\n", $field['type'], "2");
-					$type = trim($thing[0]);
-					$useropts = explode("\n", $post[$fieldfid]);
-
-					if(is_array($useropts) && ($type == "multiselect" || $type == "checkbox"))
+					if(!$field['postbit'] || !is_member($field['viewableby']) ||
+					!is_member($field['editableby'], array('usergroup' => $post['usergroup'], 'additionalgroups' => $post['additionalgroups'])) || $post['postnum'] < $field['postnum'])
 					{
-						foreach($useropts as $val)
-						{
-							if($val != '')
-							{
-								eval("\$post['fieldvalue_option'] .= \"".$templates->get("postbit_profilefield_multiselect_value")."\";");
-							}
-						}
-						if($post['fieldvalue_option'] != '')
-						{
-							eval("\$post['fieldvalue'] .= \"".$templates->get("postbit_profilefield_multiselect")."\";");
-						}
+						continue;
 					}
-					else
+					
+					$fieldfid = "fid{$field['fid']}";
+					if(!empty($post[$fieldfid]))
 					{
-						$field_parser_options = array(
-							"allow_html" => $field['allowhtml'],
-							"allow_mycode" => $field['allowmycode'],
-							"allow_smilies" => $field['allowsmilies'],
-							"allow_imgcode" => $field['allowimgcode'],
-							"allow_videocode" => $field['allowvideocode'],
-							#"nofollow_on" => 1,
-							"filter_badwords" => 1
-						);
+						$post['fieldvalue'] = '';
 
-						if($customfield['type'] == "textarea")
+						$thing = explode("\n", $field['type'], "2");
+						$type = trim($thing[0]);
+						$useropts = explode("\n", $post[$fieldfid]);
+
+						if(is_array($useropts) && ($type == "multiselect" || $type == "checkbox"))
 						{
-							$field_parser_options['me_username'] = $post['username'];
+							foreach($useropts as $val)
+							{
+								if($val != '')
+								{
+									eval("\$post['fieldvalue_option'] .= \"".$templates->get("postbit_profilefield_multiselect_value")."\";");
+								}
+							}
+							if($post['fieldvalue_option'] != '')
+							{
+								eval("\$post['fieldvalue'] .= \"".$templates->get("postbit_profilefield_multiselect")."\";");
+							}
 						}
 						else
 						{
-							$field_parser_options['nl2br'] = 0;
+							$field_parser_options = array(
+								"allow_html" => $field['allowhtml'],
+								"allow_mycode" => $field['allowmycode'],
+								"allow_smilies" => $field['allowsmilies'],
+								"allow_imgcode" => $field['allowimgcode'],
+								"allow_videocode" => $field['allowvideocode'],
+								#"nofollow_on" => 1,
+								"filter_badwords" => 1
+							);
+
+							if($customfield['type'] == "textarea")
+							{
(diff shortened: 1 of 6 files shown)


inc/functions_post.php
```diff
@@ -401,64 +401,81 @@ function build_postbit($post, $post_type=0)
 		}
 
 		// Display profile fields on posts - only if field is filled in
-		if(is_array($profile_fields))
```
Sama34: Wouldn't this break plugins that load $profile_fields outside threads?

Destroy666x: In what way?

Sama34: If I have my plugin showing profile fields in announcements now I will need to replicate all the code.

As your code now assumes only the post bit for post should display profile fields.

JN-Jones: Without testing the PR I'd agree with @Sama34 here.

Destroy666x: Well, I can revert to the old filter in `showthread.php`, but it will be buggy/inconsistent then since posts can be shown outside of `showthread.php` too. Even in the core - if you add a post through quick reply in 1.8.4, there won't be any profile fields in it because `newreply.php` is used for AJAX.

Any other ideas?

JN-Jones: Hmm, that sounds more like another bug.  Would it work if we seperate that and fix it later? As a fix I'd suggest adding a new function or modify an existing which handles profilefields and their permissions. Not sure though as I haven't worked with them for quite some time.

Sama34: Probably adding your `if(!isset($profile_fields))`check before the current `if(is_array($profile_fields))` one would fix the quick reply. That would be another issue.

Adding a function to build the fields display and check permissions could be handy for plugins too. But that would also be another issue.

JN-Jones: > Probably adding your if(!isset($profile_fields))check before the current if(is_array($profile_fields)) one would fix the quick reply. That would be another issue.

How should that fix the issue? `is_array` returns false if the variable isn't set anyways.

Anyways, @Sama34 are you okay with merging this and fixing the rest in another PR in an upcoming version?

Sama34: @JN-Jones I'm against this:
https://github.com/mybb/mybb/pull/1952/files#diff-4139bceb91b5897c2a9cddc06430deacR416

The `editableby` field should only affect edition and not display of the field.

Also, what I noted before that hasn't been updated in the PR. The code shouldn't modify the `showthread.php` file and assume that only the thread postbit should display profile fields if using the `build_postbit` function.

New issues for discussion about those changes can be created, I'm not against discussion but instead out-of-the-box changes that can affect more that what the issue at hand is meant to take.

Destroy666x: > Probably adding your if(!isset($profile_fields))check before the current if(is_array($profile_fields)) one would fix the quick reply.

No. The `build_postbit()` would still be in newreply.php. Also, the fixes you're both suggesting as an alternative would break posssible "announcement profile fields" plugins or whatever the same way, especially a new function for it.

Sama34:
I don't want to break then later, I want to postpone the breaking enough time to avoid it.

So..
* `editableby` behavior should be discussed in a new issue, don't include changes to it in this PR.
* Change on behavior of about how profile fields should be loaded (on of your changes applie in this PR) should be discussed into another issue. My comment about was confusing.
* Bug in quick reply should be taken into a new issue.

This PR should only fix the viewable permissions of a field, the post count check of the same and...
https://github.com/mybb/mybb/issues/1373#issuecomment-83763098

Everything else should be postponed to 1.8.6

I don't mind minor changes included here like the `viewableby` default value as those don't change features behavior therefore aren't worth new issues.

Destroy666x:
> I don't want to break then later, I want to postpone the breaking enough time to avoid it.

There is no way to avoid it if you want it to rely on an externally set global variable. You can't fix buggy code without touching it. The code was supposed to work only (and always) for <b>posts</b> (when `$post_type == 0` in the function), so I disagree that the need to replicate it for announcements, PMs or previews is worse than fixing a core bug in <b>posts</b> display.

As for `editableby`, I'm still waiting for more opinions on adding a new `usableby` option, which I can do quickly in this PR. Jones agreed this would be the best solution if I understood correctly, but where is the rest? @euantorano @ATofighi @Stefan-ST @Devilshakerz

Sama34:
> which I can do quickly in this PR

It has been mentioned to postpone to 1.8.6, if that was the case.

> "the permissions should always use the current settings, if the admin changes them they should be updated for existing profilefields too"

So do I. The point is you are implementing wrong usage of a **editable***by* field. I don't think we have enough time to agree + PR + testing just on just the days we have left.

There are only two real issues that need fixing in 1.8.5 (`regex` + `viewableby`) and the rest have either never been reported before or no-one complained about.

This is exactly why we shouldn't be fixing multiple issues in single PRs.

If someone wants to jump in to the discussion now then great, otherwise just update the PR and start new issues. We need to move on.

inc/functions_post.php
Sama34: > `is_array` returns false if the variable isn't set anyways.

That is the reason you will be checking if it is set before the check, set it if false. This will make profile fields work on all postbits and not assume that it should only work on threads (ignoring quick reply at the moment anyways).

Just one potential fix to consider.

JN-Jones:
Considering that we want to release 1.8.5 ASAP I'm fine with reverting those changes and create new issues for them. However as long as nobody else posts anything we'll end with the same arguments again.

JN-Jones:
@Destroy666x ?

JN-Jones:
@Destroy666x ??? Otherwise I'll defer that so we can release 1.8.5

JN-Jones:
Postpointing...

Sama34:
Unless this is updated by the end of the week I will split this into several issues and commit the proper fixes for the ones that require no discussion. (`viewableby` in postbit, https://github.com/mybb/mybb/issues/1373#issuecomment-83763098 , etc.)

JN-Jones:
As long as there isn't a fix till the end of the day (GMT+2) this won't make it into 1.8.5.

Sama34:
Oh I remember now. My only complain is the incorrect usage of the `editableby` field here.

Field visibility shouldn't be decided based on this option, that is already decided by the `viewableby` option.

What is possible with the current system:
 - Moderators can edit user profile fields which will be visible in posts, users can't edit those fields from the UCP.
 - Guests can select a, for example, "Sex" field value during registration that will be visible around the forums but user will not be able to edit it from UCP.

This removes functionality I added in the beta stages, it shouldn't.

The solution was to add a new field IIRC.

@euantorano @mybb/developers

Destroy666x:
I'll add a new field.

JN-Jones:
@Destroy666x

dequeues:

Sama34:
Could you update this so we can merge it please?

Cu8eR:
Any progress?

JN-Jones:
@Destroy666x This is our oldest PR so would be nice to get the conflicts and issues fixed.

Destroy666x:
I am slowly coming back so I'll try to deal with it ASAP

Cu8eeeR:
@Destroy666x or others MyBB devs - can you please finish this PR?

Ben-MyBB:
@mybb/developers @mybb/sqa Anyone willing to provide the final adjustments to this PR?

Issue #1959, opened 2015-04-15
Proposition : enhanced numeric option setting


crazycatdevs:
The `numeric` setting is good, but don't allow the use of the options.
I suggest to modifiy admin/modules/config/settings.php to allow a definition like:

``` php
$settings[] = array(
    'name' => 'my_numeric',
    'title' => 'a numeric field',
    'description' => 'a description',
    'optionscode' => 'numeric
min=3
max=99
step=3',
    'value' => '24'
);
```

The modification I suggest:

``` php
else if($type[0] == "numeric")
{
    $field_options = array('id' => $element_id);
    if (count($type)>1)
    {
        for($i=1; $i < count($type); $i++)
        {
            $optionsexp = explode("=", $type[$i]);
            $opt = trim($optionsexp);
            if (in_array($opt[0], array('min', 'max', 'step')))
            {
                $field_options[$opt[0]] = intval($opt[1]);
            }
        }
    }
    $setting_code = $form->generate_numeric_field($element_name, $setting['value'], $field_options);
}
```

Supryk:
good idea
+ for you

Sama34:
https://blog.isotoma.com/2012/03/html5-input-typenumber-and-decimalsfloats-in-chrome/

Step allows `any` as its value.

Destroy666x:
Any progress @crazycatdevs?

JN-Jones:
What feedback is required @Destroy666x ? Only the `any` value?

Destroy666x:
> Note that existing settings could (should?) be update to use these new options, especially min (per page and similar settings).

+ `any` if we want it.

JN-Jones:
`any` should be supported, especially with plugins in mind.

PR #1960, opened 2015-04-15
#1959 enhanced numeric option setting

crazycatdevs:
#1959 enhanced numeric option setting

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -1183,19 +1183,19 @@ if($mybb->input['action'] == "change")
 			else if($type[0] == "numeric")
 			{
 				$field_options = array('id' => $element_id);
-                            if (count($type)>1)
-                            {
-                                for($i=1; $i < count($type); $i++)
-                                {
-                                    $optionsexp = explode("=", $type[$i]);
-                                    $opt = trim($optionsexp);
-                                    if (in_array($opt[0], array('min', 'max', 'step')))
-                                    {
-                                        $field_options[$opt[0]] = (int)$opt[1];
-                                    }
-                                }
-                            }
-                            $setting_code = $form->generate_numeric_field($element_name, $setting['value'], $field_options);
+				if (count($type)>1)
+				{
+					for($i=1; $i < count($type); $i++)
+					{
+						$optionsexp = explode("=", $type[$i]);
+						$opt = trim($optionsexp);
+						if (in_array($opt[0], array('min', 'max', 'step')))
+						{
+							$field_options[$opt[0]] = (int)$opt[1];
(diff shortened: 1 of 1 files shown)


admin/modules/config/settings.php
```diff
@@ -1182,7 +1182,20 @@
...
+				$field_options = array('id' => $element_id);
+                            if (count($type)>1)
+                            {
+                                for($i=1; $i < count($type); $i++)
+                                {
+                                    $optionsexp = explode("=", $type[$i]);
+                                    $opt = trim($optionsexp);
+                                    if (in_array($opt[0], array('min', 'max', 'step')))
+                                    {
+                                        $field_options[$opt[0]] = intval($opt[1]);
```
ATofighi: it should be:
```php
                                        $field_options[$opt[0]] = (int)$opt[1];
```
`(int)` is faster than `intval()`.

crazycatdevs: Done. I'll try to keep that in mind

admin/modules/config/settings.php
```diff
@@ -1182,7 +1182,20 @@
...
+				$field_options = array('id' => $element_id);
+                            if (count($type)>1)
+                            {
+                                for($i=1; $i < count($type); $i++)
+                                {
+                                    $optionsexp = explode("=", $type[$i]);
+                                    $opt = trim($optionsexp);
+                                    if (in_array($opt[0], array('min', 'max', 'step')))
+                                    {
+                                        $field_options[$opt[0]] = int()$opt[1];
```
ATofighi: `(int)$opt[1]` :!

crazycatdevs:
Sorry, it was a long day. Corrected.

ATofighi:
please use `tabs` instead of `spaces`....
http://docs.mybb.com/1.8/development/standards/#indenting

Destroy666x:
Note that existing settings could (should?) be update to use these new options, especially `min`

crazycatdevs:
@Destroy666x : I'll try to list settings which could have these new options, but it may take time ;)

Destroy666x:
> edit : the numeric fields in the admin settings already use the min (default: 0). Only the plugins are affected

Haven't noticed your edit, but no option uses `min` because it hasn't been implemented yet. You probably see `0`s because of a bug which is going to be fixed in another PR (all empty values are casted to 0 even though they shouldn't be).

JN-Jones:
@crazycatdevs

JN-Jones:
@crazycatdevs Can you update this PR with the comments made on the issue?

JN-Jones:
New PR for this: #2193

PR #1971, opened 2015-04-24
Fix #1838 Include 'n' in format_bdays()

### LINKED ISSUE #1838: Birthday bug
Month in the birthday on user profile is not shown properly. (MyBB 1.8.4). It should be 21.8.1987 not 21.n.1987 (date is set correctly in user profile)

Original thread: [Birthday bug](http://community.mybb.com/thread-167159.html)

EDIT: Confirmed, it only applies to years over 1970 so format_bdays() function is the culprit: https://github.com/mybb/mybb/blob/feature/inc/functions.php#L5238 - it doesn't support n parameter at all.
EDIT2: Not sure what's the purpose of that function and why `mktime()` and `my_date()` aren't used.. @mybb/developers

willhartman:


Cu8eR:
#1838

PaulBender:
Works for me

PR #1972, opened 2015-04-24
Fix #1946 Case insensitive template groups

### LINKED ISSUE #1946: Template groups are case sensitive
Apparently I can create an template group as follows:
prefix: Overview
Title: Overview

But any overview* template wouldn't be sorted in this group, however I can't create an template group with overview as the prefix until I delete the former.

Makes little sense IMO.

Original thread: [Template groups are case sensitive?](http://community.mybb.com/thread-169282.html)

willhartman:


Cu8eR:
#1946

Destroy666x:
Works fine, thanks for the fix.

PR #1973, opened 2015-04-25
Fix #1913 Database export should ignore views

### LINKED ISSUE #1913: Database export should ignore views
It shouldn't be possible to select views when exporting the database within the ACP. At the moment MyBB tries views as tables which inflates (automatic) database backups.

MyBB doesn't use views therefor this is low priority. Nevertheless it's a common database feature.

willhartman:
This only affected MySQL and MySQLi.

SQLite implementation already specifies tables only.
PostgreSQL doesn't include the views in information_schema.tables anyway.

### DIFF
--- a/inc/db_mysql.php
+++ b/inc/db_mysql.php
@@ -633,11 +633,18 @@ class DB_MySQL implements DB_Base
 	{
 		if($prefix)
 		{
-			$query = $this->query("SHOW TABLES FROM `$database` LIKE '".$this->escape_string($prefix)."%'");
+			$query = $this->query("
+				SELECT `TABLE_NAME` FROM INFORMATION_SCHEMA.TABLES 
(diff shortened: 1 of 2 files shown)


willhartman:
#1913

Destroy666x:
The MySQL solution seems to be ok (haven't tested yet though), but what about PgSQL/SQLite? Both support views AFAIK. The MyBB code seems to get only the `table` type already, am I correct?

JN-Jones:
> SQLite implementation already specifies tables only.
> PostgreSQL doesn't include the views in information_schema.tables anyway.

euantorano:
Looks good to me, thanks!

Destroy666x:
Missed the reply:

> SQLite implementation already specifies tables only.
PostgreSQL doesn't include the views in information_schema.tables anyway.

You're right, didn't notice it earlier somehow, I think github showed a blank 1st comment for some reason.

Anyways, I can confirmt it works too.

PR #1984, opened 2015-05-10
require prefix setting not used on thread update, fix #1961

### LINKED ISSUE #1961: require a thread prefix for all threads doesnt work in edit post
the option "Yes, require a thread prefix for all threads" doesnt work when the user edit the post and set "No prefix". he can make a new post with a some prefix and after edit that post and set "no prefix".

sorry for my bad english.

Original thread: [require a thread prefix for all threads doesnt work in edit post](http://community.mybb.com/thread-169723.html)

Nik101010:
#1961

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -106,6 +106,13 @@ class PostDataHandler extends DataHandler
 	public $return_values = array();
 
 	/**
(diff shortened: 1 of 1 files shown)


PaulBender:
Looks good to me.

Issue #1933, opened 2015-03-24
is_numeric check in Mysql update_query


zingaburga:
From inc/db_mysql(i).php

				if(is_numeric($value))
				{
					$query .= $comma."`".$field."`={$value}";
				}
				else
				{
					$query .= $comma."`".$field."`={$quote}{$value}{$quote}";
				}

This is only done in `update_query` and not anywhere else (even in Postgres/Sqlite drivers).  
It's problematic anyway, since is_numeric can return true for strings like "0x0", which can cause query errors.

I'm not in a convenient place to check the commit history for why the change was made, but I thought I'd just point it out anyway.

zingaburga:
Also, I'm not sure if this is exactly reliable:

			if(isset($mybb->binary_fields[$table][$field]) && $mybb->binary_fields[$table][$field])
			{
				if($value[0] != 'X') // Not escaped?
				{
					$value = $this->escape_binary($value);
				}
				
				$query .= $comma."`".$field."`={$value}";
			}

If an escaped binary string just happens to start with an X, you can get weird results.

euantorano:
The intention was to allow the use of fields such `BIT(1)` (for boolean values of 0 or 1). However, this is obviously problematic. I'm not sure what the best solution would begin this case.

zingaburga:
Actually checking the datatype may be more appropriate (eg is_int).
Of course, it'd have to be done everywhere as well.

euantorano:
Yes, checking the exact type would be far better, as I'm not sure whether floats work without being quoted either.

Destroy666x:
Wouldn't something like `ctype_digit((string)$value)` be better to catch strings that are integers too, in case a plugin relied on that code in that way? Of course considering it should only catch integers, I'm not sure what it's supposed to do.

As for the 2nd issue:
> If an escaped binary string just happens to start with an X, you can get weird results.
Why?

Sama34:
I think Zinga Burga meant that if the string happens to  start with `X` then , by looking at the code, it will not be escaped.

RageSmirk:
So are you going to remove it? I already planned on writing my plugin to be compitable with this code.
[...]

PR #1990, opened 2015-05-22
Fix #1993 is_numeric check in DB classes.

euantorano:
#1993 - Removes is_numeric, adds a function `quote_val` to handle the quoting of values and only doesn't quote if the value is an integer type. This fixes the issue and also provides the benefit the original patch was meant to.

### DIFF
--- a/inc/db_mysql.php
+++ b/inc/db_mysql.php
@@ -1292,7 +1292,7 @@ class DB_MySQL implements DB_Base
 			}
 			else
 			{
-				$values .= $comma."`".$column."`='".$value."'";
+				$values .= $comma."`".$column."`=".$this->quote_val("'", $value);
 			}
 
 			$comma = ',';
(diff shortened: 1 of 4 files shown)


Destroy666x:
https://github.com/mybb/mybb/issues/1933

JN-Jones:
IMHO it'd be better to change the function to `quote_val($value, $quote="'")` so we don't need to pass the quote everytime. Also it seems more intuitive to pass the value as first parameter.

euantorano:
Makes sense, I will change it on Monday night or Tuesday.

JN-Jones:
As we should release 1.8.5 asap I'd create a new pr later today then.

euantorano:
Ok, cheers Jones. I'm away from a computer all weekend.

JN-Jones:
New PR: #1993

PR #1991, opened 2015-05-22
Using strict equality checking for password comparisons for #1977

### LINKED ISSUE #1977: password hash checks should use === and !== instead of == and !=
Background: http://3v4l.org/tT4l8 https://news.ycombinator.com/item?id=9484757

Under extremely unlucky circumstances, MyBB could accept wrong passwords. It's a very artificial example, unlikely to occur in practice, (md5sum of 0e[0-9]+), and hard to exploit due to salt; but who knows what other pitfalls there might be in PHP's string comparison using ==...

Possible examples (there should be more locations):

./inc/datahandlers/login.php:181:		if($salted_password != $this->login_data['password'])
./inc/functions_search.php:210:		if(md5($mybb->user['uid'].$forum_cache[$fid]['password']) != $mybb->cookies['forumpass'][$fid])
./misc.php:1007:					if($forum['password'] == '' && !in_array($forum['fid'], $unexp) || $forum['password'] && isset($mybb->cookies['forumpass'][$forum['fid']]) && $mybb->cookies['forumpass'][$forum['fid']] == md5($mybb->user['uid'].$forum['password']))

euantorano:
Fixes #1977 - I think I got all of them, if my search is anything to go by.

### DIFF
--- a/inc/datahandlers/login.php
+++ b/inc/datahandlers/login.php
@@ -178,7 +178,7 @@ class LoginDataHandler extends DataHandler
 
 		$plugins->run_hooks('datahandler_login_verify_password_end', $args);
 
-		if($salted_password != $this->login_data['password'])
+		if($salted_password !== $this->login_data['password'])
 		{
 			$this->invalid_combination(true);
 			return false;
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -177,7 +177,7 @@ class UserDataHandler extends DataHandler
 		}
 
 		// Has the user tried to use their email address or username as a password?
-		if($user['email'] == $user['password'] || $user['username'] == $user['password'])
+		if($user['email'] === $user['password'] || $user['username'] === $user['password'])
 		{
 			$this->set_error('bad_password_security');
 			return false;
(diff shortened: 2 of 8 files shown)


Destroy666x:
These definitely aren't all instances, there are multiple forum password checks in inc/functions.php, for example.

euantorano:
Obviously my regex missed them then. I'll have another look.

ATofighi:
https://github.com/mybb/mybb/blob/feature/inc/functions_forumlist.php#L135

Destroy666x:
Two more forum password checks should be changed:
https://github.com/MyBBStuff/mybb/blob/fix-1977-equality-comparison-for-passwords/inc/functions.php#L4101
https://github.com/MyBBStuff/mybb/blob/fix-1977-equality-comparison-for-passwords/inc/functions_search.php#L128

And possibly these:
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L180
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L199
It's very very unlikely that a user will accidentally enter 2 different strings that are interpreted by PHP the same way, nonetheless it would be nice to make sure they're always correct.

euantorano:
Fixed. I hadn't realised how badly spread out password checks are in 1.x.

JN-Jones:
Well, some of them are forum passwords, they are always a PITA in 1.x. If noone finds any more I'm going to merge this later today.

Destroy666x:
There's also this: https://github.com/MyBBStuff/mybb/blob/fix-1977-equality-comparison-for-passwords/inc/functions.php#L1507

Except that can't find anything.

JN-Jones:
New PR is #1994 as euan hasn't access to a pc till monday. @Destroy666x

PR #1993, opened 2015-05-23
Fix #1933 is_numeric check in DB classes.

### LINKED ISSUE #1933: is_numeric check in Mysql update_query

JN-Jones:
Same as #1990 but with the changes I've suggested. @euantorano @Destroy666x

### DIFF
--- a/inc/db_mysql.php
+++ b/inc/db_mysql.php
@@ -1292,7 +1292,7 @@ class DB_MySQL implements DB_Base
 			}
 			else
 			{
-				$values .= $comma."`".$column."`='".$value."'";
+				$values .= $comma."`".$column."`=".$this->quote_val("'", $value);
 			}
 
 			$comma = ',';
--- a/inc/db_mysql.php
+++ b/inc/db_mysql.php
@@ -788,7 +788,7 @@ class DB_MySQL implements DB_Base
 			}
 			else
 			{
-				$quoted_value = $this->quote_val("'", $value);
+				$quoted_value = $this->quote_val($value);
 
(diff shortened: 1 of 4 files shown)


JN-Jones:
@Destroy666x can we merge this?

inc/db_mysql.php
```diff
@@ -788,7 +788,9 @@ function insert_query($table, $array)
 			}
 			else
 			{
-				$array[$field] = "'{$value}'";
+				$quoted_value = $this->quote_val($value);
+
+				$array[$field] = "{$quoted_value}";
```
ATofighi: why you don't use:
```php
$array[$field] = $this->quote_val($value);
```
?

JN-Jones: Guess Euan missed that and simply modified the original line. Will change that (and similar lines)

JN-Jones: Fixed (hopefully all)

PaulBender:
Works for me. If no one else has any issues, this can be merged.

PR #1994, opened 2015-05-23
Using strict equality checking for password comparisons for #1977

JN-Jones:
Same as #1991 but with the last change included. @euantorano @Destroy666x


JN-Jones:
@Destroy666 can we merge this?

Destroy666x:
Yeah, it should be ok now.

PR #1998, opened 2015-05-25
Fix #1955 wrong SCEditor smilie checks

### LINKED ISSUE #1955: Wrong SCEditor smilie check
https://github.com/mybb/mybb/blob/feature/inc/functions.php#L3196
https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L931
are wrong because unclickable smilies shouldn't be completely skipped. This makes them not parsed by SCEditor at all while they should only be hidden from the toolbar. The solution is to move the check here: https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L960 and assign hidden smilies to the `$hiddensmilies` variable.

Destroy666x:
Fixes wrong checks for clickable smilies (both individual and global)
which didn't parse smilies in SCEditor at all while they should be only
hidden from the toolbar.

https://github.com/mybb/mybb/issues/1955

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -910,10 +910,14 @@ EOF;
 		// Smilies
 		$emoticon = "";
 		$emoticons_enabled = "false";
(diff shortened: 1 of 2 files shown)


PaulBender:
Works for me

PR #2000, opened 2015-05-29
Fix #1999 wrong timezone when adding user via Admin-CP

### LINKED ISSUE #1999: Adding user in Admin-CP results in wrong timezone
Manually adding a user in the Admin-CP results in the new user having not the default timezone which was set for the forum. 

Original thread: http://community.mybb.com/thread-170794.html

psde:
Fixes wrong timezone of users which are manually added via the Admin-CP by assigning the default board timezone.

https://github.com/mybb/mybb/issues/1999

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -895,6 +895,27 @@ class UserDataHandler extends DataHandler
 	}
 
 	/**
+	 * Verifies if the user timezone is valid. 
+	 * If the timezone is invalid, the board default is used.
+	 *
+	 * @return boolean True when timezone was valid, false otherwise
+	 */
+	function verify_timezone()
+	{
+		$user = &$this->data;
+
+		$timezones = get_supported_timezones();
+
+		if(!array_key_exists($user['timezone'], $timezones))
+		{
+			$user['timezone'] = $mybb->settings['timezoneoffset'];
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	* Validate all user assets.
 	*
 	* @return boolean True when valid, false when invalid.
@@ -991,6 +1012,10 @@ class UserDataHandler extends DataHandler
 		if($this->method == "insert" || array_key_exists('language', $user))
 		{
 			$this->verify_language();
+		}
+		if($this->method == "insert" || array_key_exists('timezone', $user))
+		{
+			$this->verify_timezone();
 		}
 		if($this->method == "insert" && array_key_exists('regcheck1', $user) && array_key_exists('regcheck2', $user))
 		{
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -6638,17 +6638,13 @@ function is_banned_ip($ip_address, $update_lastuse=false)
 }
 
 /**
- * Build a time zone selection list.
+ * Returns an array of supported timezones
  *
- * @param string $name The name of the select
- * @param int $selected The selected time zone (defaults to GMT)
- * @param boolean $short True to generate a "short" list with just timezone and current time
- * @return string
+ * @return string[] Key is timezone offset, Value the language description
  */
-function build_timezone_select($name, $selected=0, $short=false)
+function get_supported_timezones()
 {
-	global $mybb, $lang, $templates;
-
+	global $lang;
 	$timezones = array(
 		"-12" => $lang->timezone_gmt_minus_1200,
 		"-11" => $lang->timezone_gmt_minus_1100,
@@ -6690,6 +6686,22 @@ function build_timezone_select($name, $selected=0, $short=false)
 		"13" => $lang->timezone_gmt_1300,
 		"14" => $lang->timezone_gmt_1400
 	);
+	return $timezones;
+}
+
+/**
+ * Build a time zone selection list.
+ *
+ * @param string $name The name of the select
+ * @param int $selected The selected time zone (defaults to GMT)
+ * @param boolean $short True to generate a "short" list with just timezone and current time
+ * @return string
+ */
+function build_timezone_select($name, $selected=0, $short=false)
+{
+	global $mybb, $lang, $templates;
+
+	$timezones = get_supported_timezones();
 
 	$selected = str_replace("+", "", $selected);
 	foreach($timezones as $timezone => $label)


psde:
I was thinking about fixing this in the datahandler itself, but I did not feel comfortable about it as it seems out of place there.

Destroy666x:
> as it seems out of place there

Why? I think that adding it to the user datahandler somewhere around here: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L1038 or even better here with its own short function: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L968 would be better and preferred @psde .

JN-Jones:
Yep, I'd also prefer it in the datahandler especially for plugins as @Destroy666x said.

Destroy666x:
@psde

psde:
I updated the PR to reflect the suggested changes and tested them thoroughly. Any more suggestions?

inc/datahandlers/user.php
```diff
@@ -993,6 +1007,10 @@ function validate_user()
 		{
 			$this->verify_birthday_privacy();
 		}
```
Sama34: New line missing after this.

inc/datahandlers/user.php
```diff
@@ -888,6 +888,20 @@ function verify_checkfields()
...
+	 *
+	 * @return boolean True when valid, false when invalid.
+	 */
+	function verify_timezone()
+	{
+		$user = &$this->data;
+
+		$timezones = get_supported_timezones();
+
+		return array_key_exists($user['timezone'], $timezones);
```
Sama34: Should set an error and return false if false. True otherwise.

inc/datahandlers/user.php
psde: I tried to imitate the coding style surrounding my addition.

inc/datahandlers/user.php
psde: `array_key_exists` returns a boolean already. As per https://github.com/mybb/mybb/pull/2000#issuecomment-108704734 the timezone is reset the the boards default timezone if it does not validate.

inc/datahandlers/user.php
```diff
@@ -993,6 +1007,10 @@ function validate_user()
 		{
 			$this->verify_birthday_privacy();
 		}
+		if(!array_key_exists('timezone', $user) || !$this->verify_timezone())
+		{
+			$user['timezone'] = $mybb->settings['timezoneoffset'];
+		}
```
Sama34: This is inconsistent with the style used above.

inc/datahandlers/user.php
Destroy666x: Yeah, error is not needed when the default value is used.

Sama34: Why would you force the default value if it is wrong? It should return an error and displayed to the user accordingly.

Just comparing your code to the above checks tells me its inconsistency. The only case where this is true is when checking the 'username'.

psde: I am happy to change the function accordingly, but a lot of `verify` functions do not set an error. For example, `verify_referrer` and `verify_lastvisit` do not set an error, but they are setting a default value. I could move the handling of the default value to the `verify_timezone`, if that would make the code more consistent.

Destroy666x: > Why would you force the default value if it is wrong?

Because we do it for other options too?
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L638-654
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L671
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L681
etc.

Destroy666x: > I could move the handling of the default value to the verify_timezone, if that would make the code more consistent.

I think this could be done, yes.

Sama34: @psde
- `verify_referrer` does set an error if the user inserted a value.
- `verify_lastvisit` value is not an user input.

@Destroy666x Then move this inside `verify_options`?

I believe something around the lines of...

```PHP
function verify_timezone()
{
	$user = &$this->data;

	$timezones = get_supported_timezones();

	if(!array_key_exists($user['timezone'], $timezones))
	{
 		$this->set_error("invalid_timezone");
 		return false;
		// or force the timezone here
		//$user['timezone'] = $mybb->settings['timezoneoffset'];
	}

	return true;
}

if($this->method == "insert" && array_key_exists('timezone', $user))
{
	$this->verify_timezone();
}
```

Should be.

@euantorano

psde: > I believe something around the lines of...

This does not force the default value if the timezone is not set at all, which is the case in the original bug report. How should that case be handled?

Destroy666x: @Sama34, we can't move it to `verify_options()` because as I said `timezone` is not an option and the function will only run only if there are options.

Destroy666x: And there are is no option array when the user is created, which is the initial issue here - a default timezone should be set in that case instead of messing with errors... Which shouldn't be printed in ACP as you said yourself in the style PR.

Sama34: I have updated my code accordingly.

Destroy666x: The code above could be used, yes.

JN-Jones:

inc/functions.php
```diff
@@ -6601,16 +6601,13 @@ function is_banned_ip($ip_address, $update_lastuse=false)
 }
 
 /**
- * Build a time zone selection list.
+ * Returns an array of supported timezones
  *
- * @param string The name of the select
- * @param int The selected time zone (defaults to GMT)
- * @param boolean True to generate a "short" list with just timezone and current time
+ * @return double[] Key is timezone offset, Value the language description
```
JN-Jones: But it's not an array of `double` values, is it?

psde:
I rebased the PR to 5cec922e7f1551518c7dbba45b66378c9eff07fe and improved the code consistency as @Sama34 suggested.

Sama34:
`verify_timezone()` should probably return `true`. Otherwise merge.

inc/functions.php
```diff
@@ -6690,6 +6686,21 @@ function build_timezone_select($name, $selected=0, $short=false)
...
 	);
+	return $timezones;
+}
+
+/**
+ * Build a time zone selection list.
+ *
+ * @param string The name of the select
+ * @param int The selected time zone (defaults to GMT)
+ * @param boolean True to generate a "short" list with just timezone and current time
```
Destroy666x: This function returned and returns a string too, so `@return` should be readded.

Except that it looks ready to merge.

JN-Jones: The whole doc block is reverted for some reason. The `@return` seems to be there but the variable names should be included again.

psde:
I think I made a mistake while merging/rebasing. Should be fixed now.

Destroy666x:
Thanks.

PR #2002, opened 2015-05-29
Fix #2001 Calendar does not show errors when adding events

### LINKED ISSUE #2001: Calendar does not show errors when adding events
When adding a new event and not filling in all details, the calendar does not show any errors.

Original thread: http://community.mybb.com/thread-168437.html

psde:
The '$event_errors' variable was cleared for no reason right before outputting the template, therefore deleting the errors.

https://github.com/mybb/mybb/issues/2001

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -542,8 +542,6 @@ if($mybb->input['action'] == "addevent")
 		eval("\$calendar_select .= \"".$templates->get("calendar_addevent_calendarselect_hidden")."\";");
 	}
 
-	$event_errors = '';
-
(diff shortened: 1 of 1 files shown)


Destroy666x:
>  for no reason

I guess someone wanted to prevent a PHP notice about using undefined variable, but it was done in a very bad way. So I'd change it to:
```php
if(!isset($event_errors))
{
	$event_errors = '';
}
```
rather than removing it completely.

psde:
I see, thanks for your explanation. I will update this PR to reflect the suggested changes.

PR #2006, opened 2015-05-29
Fix #2003 Deleting an event in 'editevent' without checking checkbox …

### LINKED ISSUE #2003: Deleting an event in 'editevent' without checking checkbox throws errors
When trying to delete an event from the edit event form without checking the checkbox, multiple errors are thrown:

    The name for the event is missing. Please enter an event name.
    The description for the event is missing. Please enter an event description.

All data fields (name, description, time) are also cleared.

Original thread: http://community.mybb.com/thread-168437.html

psde:
…throws errors

Adds a new action `do_deleteevent` for deleting events and displays a proper error message (with a new language string) when the checkbox is not checked. The JavaScript function `MyBB.deleteEvent(eid)` is also using this action now. The action `do_editevent` is not capable of deleting events anymore.

https://github.com/mybb/mybb/issues/2003

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -553,6 +553,58 @@ if($mybb->input['action'] == "addevent")
 	output_page($addevent);
 }
 
+// Delete an event
+if($mybb->input['action'] == "do_deleteevent" && $mybb->request_method == "post")
+{
+	$query = $db->simple_select("events", "*", "eid='{$mybb->input['eid']}'");
+	$event = $db->fetch_array($query);
+
+	if(!$event)
+	{
+		error($lang->error_invalidevent);
+	}
+
+	$query = $db->simple_select("calendars", "*", "cid='{$event['cid']}'");
+	$calendar = $db->fetch_array($query);
+
+	// Invalid calendar?
+	if(!$calendar)
+	{
+		error($lang->invalid_calendar);
+	}
+
+	// Do we have permission to view this calendar or post events?
+	$calendar_permissions = get_calendar_permissions($calendar['cid']);
+	if($calendar_permissions['canviewcalendar'] != 1 || $calendar_permissions['canaddevents'] != 1)
+	{
+		error_no_permission();
+	}
+
+	if(($event['uid'] != $mybb->user['uid'] || $mybb->user['uid'] == 0) && $calendar_permissions['canmoderateevents'] != 1)
+	{
+		error_no_permission();
+	}
+
+	// Verify incoming POST request
+	verify_post_check($mybb->get_input('my_post_key'));
+
+	$plugins->run_hooks("calendar_do_deleteevent_start");
+
+	// Is the checkbox set?
+	if($mybb->get_input('delete', MyBB::INPUT_INT) == 1)
+	{
+		$db->delete_query("events", "eid='{$event['eid']}'");
+		$plugins->run_hooks("calendar_do_deleteevent_end");
+
+		// Redirect back to the main calendar view.
+		redirect("calendar.php", $lang->redirect_eventdeleted);
+	}
+	else
+	{
+		error($lang->delete_no_checkbox);
+	}
+}
+
 // Edit an event
 if($mybb->input['action'] == "do_editevent" && $mybb->request_method == "post")
 {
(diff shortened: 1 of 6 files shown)


psde:
I am looking for some guidance and suggestions on how to improve the current state of `calendar.php`. Currently every action is using (almost) the same code for checking permissions.

Also, I am worried about breaking compatibility with this pull request, as `do_editevent` is not capable of deleting events anymore - which is maybe used by a plugin?

Destroy666x:
> I am looking for some guidance and suggestions on how to improve the current state of calendar.php. Currently every action is using (almost) the same code for checking permissions.

Feel free to discuss it on our forums in 1.8 suggestions & feedback category.

> Also, I am worried about breaking compatibility with this pull request, as do_editevent is not capable of deleting events anymore - which is maybe used by a plugin?

I'm quite sure there aren't any plugins it breaks (public at least) and if they are, they can be fixed easily. And I don't think there is another fix which won't break anything, but feel free to share if you have something else on your mind.

However, there are still 2 corrections that need to be made:
1) Bump the version of the modified template to `1806`
2) Bump the version of the `general.js` file in all templates (and their versions too) and all ACP files

JN-Jones:
@psde?

psde:
Sorry for the late update. I bumped the versions on the theme and general.js. Which ACP files do I have to modify?

install/resources/mybb_theme.xml
```diff
@@ -13871,7 +13871,7 @@ $(function() {
 <meta http-equiv="Content-Script-Type" content="text/javascript" />
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/jquery.js?ver=1804"></script>
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/jquery.plugins.min.js?ver=1804"></script>
-<script type="text/javascript" src="{$mybb->asset_url}/jscripts/general.js?ver=1804"></script>
+<script type="text/javascript" src="{$mybb->asset_url}/jscripts/general.js?ver=1805"></script>
```
Destroy666x: Should be `1.8.6`

Destroy666x:
https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L111
https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L389
https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L521

Could be done here too:
https://github.com/mybb/mybb/blob/feature/install/resources/output.php#L43

psde:
Thanks, I see what you meant. I updated the PR to reflect the suggested changes.

calendar.php
```diff
@@ -596,6 +596,46 @@
 		// Redirect back to the main calendar view.
 		redirect("calendar.php", $lang->redirect_eventdeleted);
 	}
+	else
+	{
+		error($lang->delete_no_checkbox);
```
Destroy666x: `delete_no_checkbox` doesn't seem to be defined anywhere, at least I can't find it in my local installation with this PR.

install/resources/mybb_theme.xml
```diff
@@ -13871,7 +13871,7 @@ $(function() {
 <meta http-equiv="Content-Script-Type" content="text/javascript" />
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/jquery.js?ver=1804"></script>
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/jquery.plugins.min.js?ver=1804"></script>
-<script type="text/javascript" src="{$mybb->asset_url}/jscripts/general.js?ver=1804"></script>
+<script type="text/javascript" src="{$mybb->asset_url}/jscripts/general.js?ver=1806"></script>
```
Destroy666x: And the last thing left (if noone notices anything else) - the template version of this modified template - `headerinclude` should be bumped too.

Destroy666x:
Oh, never mind, one more thing - I guess we should add at least 1 hook to the new action.

JN-Jones:
@psde

calendar.php
psde: I added it to the language file: https://github.com/mybb/mybb/pull/2006/files#diff-505deb1daf1b38b7bace47d542d761e5R93 - is there anything else I need to do?

JN-Jones: The `headerinclude` template @Destroy666x mentioned below and the new plugin hook.

psde:
All versions should be bumped now, and I also added the requested hooks. The question about the language file is still unanswered. Do I have to bump a version for the modified language file as well?

Should I open an issue on the documentation and submit a PR for the new hooks introduced?

JN-Jones:
> Do I have to bump a version for the modified language file as well?

No, language files are handled by us before creating a relase.

> Should I open an issue on the documentation ~~and submit a PR for the new hooks introduced~~? Seems like the hooks are generated automatically.

They were generated automatically but they're not generated "on the fly". Feel free to generate an issue, however instead of adding the new hooks manually the Hook Finder should be rerun before 1.8.6 gets released.

Destroy666x:
Looks ok now, thanks.

PR #2010, opened 2015-05-30
Fix #2009 poll options not saved in editpost.php

### LINKED ISSUE #2009: Poll options values not saved in editpost.php
If you preview or force an error during the edition of the 1st post, poll options values aren't saved.

Destroy666x:
https://github.com/mybb/mybb/issues/2009


PaulBender:
Looks good to me.

PR #2012, opened 2015-05-30
Fixes #1992 send_pm bug in ACP

### LINKED ISSUE #1992: send_pm bug in ACP
send_pm loads "messages.lang.php" which may not be required (haven't dug through the code).

That file does not exist in the admin language folder, therefore a "file does not exist" error will be generated when using send_pm from ACP.

Original thread: [send_pm bug in ACP](http://community.mybb.com/thread-169648.html)

PaulBender:
Fixes #1992 

Couldn't find any reason why the messages language file wasn't being loaded.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -50,6 +50,7 @@ if(!isset($cp_language))
 
 // Load global language phrases
 $lang->load("global");
+$lang->load("messages", true);
 
 if(function_exists('mb_internal_encoding') && !empty($lang->settings['charset']))
 {
@@ -472,6 +473,7 @@ if(!empty($mybb->user['uid']))
 		$cp_language = $admin_options['cplanguage'];
 		$lang->set_language($cp_language, "admin");
 		$lang->load("global"); // Reload global language vars
+		$lang->load("messages", true);
 	}
 
 	if(!empty($admin_options['cpstyle']) && file_exists(MYBB_ADMIN_DIR."/styles/{$admin_options['cpstyle']}/main.css"))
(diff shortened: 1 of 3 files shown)


Destroy666x:
Is it surely safe to remove? It was added initially to the function: https://github.com/mybb/mybb/commit/6706d3f6744313f7e51f0d64acc072778385744f

Destroy666x:
It's needed for several strings in the PM datahandler: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/pm.php#L344 etc. The `messages` lang file is loaded globally in front-end (so no problem here), but not in the back-end.

Solution - add these lang strings to a global ACP language file.

JN-Jones:
@PaulBender

PaulBender:
I think it's safe to remove, messages.lang is loaded in the PM datahandler: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/pm.php#L342

Destroy666x:
That code is inside a conditional's `else` which uses default `$lang` if the recipient uses the same language. And in that case lang strings are empty in ACP.

PaulBender:
Okay, I went with loading the messages.lang file in the Admin CP globally.

Destroy666x:
`$lang->mybb_engine` is still empty in ACP: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/pm.php#L652 It's in front-end's `global.lang.php`, but not back-end's `global.lang.php`.

Also note that merging this PR in 1.8.7 is obligatory for https://github.com/mybb/mybb/issues/2147, otherwise task emails sent from ACP will be empty.

PaulBender:
I've added it

PR #2014, opened 2015-06-02
Fix #1816 (Very inefficient managegroups.php code)

### LINKED ISSUE #1816: Very inefficient managegroups.php code...
https://github.com/mybb/mybb/blob/feature/managegroup.php#L376 - that's quite bad, one query is executed per every group member while the leaders could be simply added to an array in this loop: https://github.com/mybb/mybb/blob/feature/managegroup.php#L307 and then checked with `in_array()`.

Krytic:
Fix #1816 => Very inefficient managegroups.php code

### DIFF
--- a/managegroup.php
+++ b/managegroup.php
@@ -299,6 +299,9 @@ else
 		LEFT JOIN ".TABLE_PREFIX."users u ON (u.uid=g.uid)
 		WHERE g.gid = '{$gid}'
 	");
+
+	$leaders_array = array();
+
 	if($db->num_rows($query))
 	{
 		$loop = 1;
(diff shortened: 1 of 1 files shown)


Krytic:
Pushed a quick fix where the "(Leader)" string wasn't being displayed.

managegroup.php
```diff
@@ -375,11 +380,10 @@
 		{
 			$email = '';
 		}
-		$query1 = $db->simple_select("groupleaders", "uid", "uid='{$user['uid']}' AND gid='{$gid}'");
-		$isleader = $db->fetch_array($query1);
+		$isleader = in_array($user['uid'], $leaders_array);
```
Destroy666x: No need for the variable, you can simply check it in the conditional

Krytic: Thanks @Destroy666x -- pushing that commit now

Destroy666x:
Haven't noticed any problem now, thanks for the fix.

PR #2015, opened 2015-06-03
Fixes #1989 IP Search shows moderators thread titles even if they hav…

### LINKED ISSUE #1989: IP Search shows moderators thread titles even if they have no permission there.
Detailed description of your problem, including steps to reproduce if necessary

Person has mod CP access. It can get the IP address of a user that has view access to a board the moderator doesn't. They go to Mod CP, IP Search and put in the IP address of the user. They will be able to see the TITLES of THREADS the user has created or posted in, even if the mod with Mod CP access doesn't have view permission on these boards.
New installation or upgrade (from which version of MyBB)?

1.6

Thanks

P.S: I've temporarily commented out the "Search Posts" checkbox under IP Search, but that's not a very safe solution.

Original thread: [IP Search shows moderators thread titles even if they have no permission there.](http://community.mybb.com/thread-170811.html)

PaulBender:
…e no permission there.

Fixes #1989

Checks against forum permission (to see if user can view that thread/post) and moderator permission (to see if user can view that post if unapproved/deleted).

### DIFF
--- a/modcp.php
+++ b/modcp.php
@@ -3569,7 +3569,69 @@ if($mybb->input['action'] == "ipsearch")
 
 			if($post_ip_sql)
 			{
-				$query = $db->simple_select('posts', 'COUNT(pid) AS count', "$post_ip_sql AND visible >= -1");
+				$where_sql = '';
+				$unviewablefids = array();
+				$onlyusfids = array();
+
+				// Check group permissions if we can't view posts not started by us
+				$group_permissions = forum_permissions();
+				foreach($group_permissions as $fid => $forumpermissions)
+				{
+					if($forumpermissions['canview'] == 0 || $forumpermissions['canviewthreads'] == 0)
+					{
+						$unviewablefids[] = $fid;
+					}
+					if(isset($forumpermissions['canonlyviewownthreads']) && $forumpermissions['canonlyviewownthreads'] == 1)
+					{
+						$onlyusfids[] = $fid;
+					}
+				}
+
+				if(!empty($unviewablefids))
+				{
+					$where_sql .= " AND fid NOT IN(".implode(',', $unviewablefids).")";
+				}
+				if(!empty($onlyusfids))
+				{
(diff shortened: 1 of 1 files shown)


modcp.php
```diff
@@ -3569,7 +3569,69 @@
...
-				$query = $db->simple_select('posts', 'COUNT(pid) AS count', "$post_ip_sql AND visible >= -1");
+				$where_sql = '';
+				$unviewablefids = array();
+				$onlyusfids = array();
+
+				// Check group permissions if we can't view posts not started by us
+				$group_permissions = forum_permissions();
+				foreach($group_permissions as $fid => $forumpermissions)
+				{
+					if($forumpermissions['canview'] == 0 || $forumpermissions['canviewthreads'] == 0)
```
Destroy666x: Filtering by `get_unviewable_forums()` results would be better than this conditional since passwords are also checked there. `get_inactive_forums()` should also be used.

Destroy666x:
The `canonlyviewownthreads` conditional should have stayed though, `get_unviewable_forums(true)` doesn't check that. However, the `$onlyusfids` part of the query was wrong - it checked the `uid` column in the `posts` table, which is the author of the post and not the thread. `LEFT JOIN` to the `threads` table will probably be needed in both queries.

PaulBender:
That should do it.

modcp.php
```diff
@@ -3569,7 +3569,77 @@
...
+
+						if($moderator['canviewdeleted'] == 1)
+						{
+							$deleted_forums[] = $moderator['fid'];
+						}
+					}
+
+					if(!empty($unapprove_forums))
+					{
+						$visible_sql .= " OR (p.visible = 0 AND p.fid IN(".implode(',', $unapprove_forums)."))";
```
Destroy666x: Doesn't `$visible_sql = ' AND p.visible > 0 AND t.visible > 0';` + this skip unapproved (and soft deleted below) posts?

If I'm not mistaken, `p.visible > 0` is always required by `AND` and the OR applies to `t.visible > 0` and the other 2 conditions.

PaulBender:
Okay, I think I've fixed the thread visibility (at least it works for me).

euantorano:
Works for me too, merging.

PR #2017, opened 2015-06-04
Fix #1964 ||~|~|| sequence shouldn't be used in poll options

### LINKED ISSUE #1964: Using ||~|~|| in polls breaks poll
Of some odd reason MyBB uses ||~|~|| to separate poll answers. Therefore using ||~|~|| in poll answers can corrupt polls.

Destroy666x:
Throws an eror if someone tries to enter `||~|~||` in one of the poll
options.

https://github.com/mybb/mybb/issues/1964

### DIFF
--- a/polls.php
+++ b/polls.php
@@ -255,21 +255,32 @@ if($mybb->input['action'] == "do_newpoll" && $mybb->request_method == "post")
 			$options[$i] = '';
 		}
 
-		if(trim($options[$i]) != "")
-		{
-			$optioncount++;
-		}
-
-		if(my_strlen($options[$i]) > $mybb->settings['polloptionlimit'] && $mybb->settings['polloptionlimit'] != 0)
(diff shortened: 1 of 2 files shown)


polls.php
```diff
@@ -255,16 +255,22 @@
...
-			$optioncount++;
+			$lengtherror = 1;
+			break;
 		}
 
-		if(my_strlen($options[$i]) > $mybb->settings['polloptionlimit'] && $mybb->settings['polloptionlimit'] != 0)
+		if(strpos($options[$i], '||~|~||') !== false)
 		{
-			$lengtherror = 1;
+			$sequenceerror = 1;
```
euantorano: Minor nitpick, but can we rename this to `sequence_error`? IMO it's easier to read that way.

Destroy666x: Well, I can change it, but there are lots of variables with words not separated with `_` even in the nearby of this line. I don't think there should be problems with reading it.

euantorano: Ok, it's not really a problem.

PR #2019, opened 2015-06-04
Fixes #2018 Use queried id instead of input

### LINKED ISSUE #2018: Use queried id instead of input
In a few instances in the Admin CP, a $mybb->input id is used instead of the queried id. For consistency, the queried id should be used.

PaulBender:
Fixes #2018


euantorano:
Looks good to me, going to merge this.

PR #2021, opened 2015-06-04
Fix #2016 with a simple bracket.

### LINKED ISSUE #2016: Strange behaviour of "find_replace_templatesets"
With MyBB 1.8, I've noticed some strange behaviour of the find_replace_templatesets function. Using the following code:

PHP Code:
 &nbsp; find_replace_templatesets('headerinclude',&nbsp;'/&#36;/',&nbsp;'{&#36;myalerts_js}');&nbsp;

Causes all global templates to be updated. This is due to the following query being ran by the function:

PHP Code:
SELECT&nbsp;tid,&nbsp;sid,&nbsp;template&nbsp;FROM&nbsp;mybb_templates&nbsp;WHERE&nbsp;title&nbsp;=&nbsp;'headerinclude'&nbsp;AND&nbsp;sid&gt;0&nbsp;OR&nbsp;sid=-1;&nbsp;

This query results in the following (where I have one global template, with the content "this is a test global template...":

The expected result would be for only the "header include" template to be returned, not all global templates.

I believe the actual query that should be ran is as follows:

PHP Code:
SELECT&nbsp;tid,&nbsp;sid,&nbsp;template&nbsp;FROM&nbsp;mybb_templates&nbsp;WHERE&nbsp;title&nbsp;=&nbsp;'headerinclude'&nbsp;AND&nbsp;(sid&gt;0&nbsp;OR&nbsp;sid=-1);&nbsp;

This query returns only the "headerinclude" template.

This code was changed in this July 2014 commit, so I'm surprised it's taken this long to cause problems...

Original thread: [Strange behaviour of "find_replace_templatesets"](http://community.mybb.com/thread-171429.html)

euantorano:


Destroy666x:
The query is used 2 times now and this needs to be edited too: https://github.com/MyBBStuff/mybb/blob/fix-2016-find-replace-templatesets/inc/adminfunctions_templates.php#L76

euantorano:
Yep, I'll fix it tonight

euantorano:
No idea why the first query in that function was duplicated :question:, but it's all fixed now.

PR #2023, opened 2015-06-06
Fixes #1965 Calendar - problem with mini calendar

### LINKED ISSUE #1965: Calendar - problem with mini calendar
According to the mini calendar on the right side the 22. march is a Monday but it is a Sunday.
calender

Also the link (») links to a week starting with Sunday although the first day of week is set to Monday.

PaulBender:
Fixes #1965


Destroy666x:
Tested it for every weekday and few weeks and noticed no problem.

Issue #313, opened 2014-03-03
Date and time should be returned in localized form


nmalcolm:
> Right now, $lastvisit and $lang->welcome_current_time (both in global.php) return date and time in the server's localization (usually, in English). This has to be modified, such that date and time are returned in localized format.

Original report: http://dev.mybb.com/issues/2080
Reporter: Huji Lee
Status: New

Destroy666x:
Other hardcoded names (e.g. weekdays) could also be translated. `strtr()` looks to be the best solution (writing it here just so I don't forget).

JN-Jones:
@Destroy666x What's with this?

PR #2024, opened 2015-06-06
Fixes #1838 Birthday bug

### LINKED ISSUE #1838: Birthday bug

PaulBender:
Fixes #1838

Deprecated the format_bdays function and use mktime for all birthdays.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -5252,6 +5252,7 @@ function get_bdays($in)
 }
 
 /**
+ * DEPRECATED! Please use mktime()!
  * Formats a birthday appropriately
  *
  * @param string The PHP date format string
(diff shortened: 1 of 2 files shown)


Destroy666x:
Looks ok now, however, `my_date()` could be used. But I'll mention that later in a PR for this: https://github.com/mybb/mybb/issues/313 - the localization fix can't be applied to pure `date()` so all calls to that which possibly show will have to be changed to `my_date()`.

PR #2025, opened 2015-06-08
Fixes #1150 Remove hardcoded HTML v2

### LINKED ISSUE #1150: Remove hardcoded HTML v2
#756 All hardcoded HTML should be moved to templates.

- [x] global.php leftovers from the previous ticket
- [x] http://community.mybb.com/thread-157062.html
- [x] http://community.mybb.com/thread-156762.html
- [x] http://community.mybb.com/thread-157254.html
- [x] http://community.mybb.com/thread-157646.html
- [x] There is some code in modcp.php and member.php that should be moved to templates. #1421
- [x] https://github.com/mybb/mybb/blob/feature/newreply.php#L267 and a few similar inputs
- [x] https://github.com/mybb/mybb/blob/feature/stats.php#L155
- [x] $timeremaining in modcp.php

PaulBender:
Fixes #1150 and #2050

~~Don't merge yet as this is a work in progress.~~ Done.


PaulBender:
This is now done.

31 templates added, 3 templates updated and 1 template renamed.

Destroy666x:
This is still missing: https://github.com/mybb/mybb/blob/feature/global.php#L906
I'll try to check the rest next week.

PaulBender:
Fixed.

JN-Jones:
Looks good to me @Destroy666x @euantorano 

However to make this work we need an upgrade script as mentioned here: https://github.com/mybb/mybb/pull/2044#issuecomment-113466104

Stefan-ST:
Or merge one of the PRs with an upgrade script first. ;)

JN-Jones:
We can even merge this first ;) And either merge one of the others later or add it when we prepare the release. That's why I tagged others to recheck whether this can be merged. As you took time to post an answer you could also check and merge this if it's ok...

Stefan-ST:
@JN-Jones I'm not sure how you check a PR but this PR is certainly not trivial with changes to 24 files... :-/

JN-Jones:
Guess why I haven't merged it yet and asked for other feedback... So next time please don't post something everyone is aware of if you don't intend to help so we can concentrate on the work we're doing here...

Stefan-ST:
@JN-Jones Your comment can be read as a suggestion to add another upgrade script. Just wanted to avoid that...

install/resources/mybb_theme.xml
```diff
@@ -822,6 +822,14 @@ blockquote cite span.highlight {
 	font-size: 13px;
 }
 
+.mycode_me {
```
Stefan-ST: Unfortunately adding new CSS classes requires updating all existing themes. We should avoid that in a minor update to avoid support threads (although most users probably won't note it in this case). IMHO it either should have done in 1.8.0 or be deferred to 2.0 unless you have another solution.

PaulBender: @mybb/developers opinions on this?

PaulBender:
Conflicts resolved and version numbers have been updated.

install/resources/mybb_theme.xml
JN-Jones: The master file is still updated. IMHO the best solution would be one which works without the class (in case the file has been modified and isn't updated) and with the class (in case nothing has been modified and the master file is used). So if someone wants to use a different color he'll be able to add the class and change the color.

Cu8eR:
It would be nice to see this in 1.8.7

PaulBender:
Resolved the merge conflict.

PaulBender:
For the /me and /slap MyCode styling, I decided on a compromise: I kept the original red inline styling but added a css class on both codes. That way if theme authors want to change the styling of the /me and /slap they can (the color can be overrided with a !important attribute).

Sama34:
Seems like a good approach.

PaulBender:
Fixed merge conflicts

PaulBender:
Fixed merge conflicts and updated to 1808.

Can we please merge this?

install/resources/mybb_theme.xml
JordanMussi: By creating a `mycode.css` stylesheet attached to `showthread.php`, `newthread.php`, `newreply.php` etc. it would mean themes without the stylesheet will inherit from the master style. So this change would be able to be included in a minor update...

Reference: http://community.mybb.com/thread-190334-post-1215259.html#pid1215259

Destroy666x: @JordanMussi, parser is used in many places, including plugins, not only these pages. Besides, there's definitely an unfixed bug report about stylesheets not created on upgrade sometimes, so I wouldn't rely on that. PaulBender's solution (style and class at the same time) looks reasonable, but it's been applied only to /me and /slap so far.

JordanMussi: Ah, fair point. I was just trying to think of a solution so that it didn't need to be in `global.css` (hence removing the concern about theme issues). It could be that `mycode.css` is included globally? If plugins use the parser in the Admin CP that could be a problem as the stylesheet wouldn't be included there...

Destroy666x: Yeah, ACP is another reason why using both style and class is good.

PaulBender:
Fixed merge conflicts

Issue #2026, opened 2015-06-10
Wrong Doc Blocks


ATofighi:
> There are several wrong doc blocks for functions. One example is the show_fields_from which returns an array and not a string. Also the "@param" block  is often wrong:

> http://manual.phpdoc.org/HTMLSmartyConve...m.pkg.html Wrote:
>> @param datatype &#36;var desc

> But we often ignore the "&#36;var" part which throws an error in several IDEs.

Original thread: [Wrong Doc Blocks](http://community.mybb.com/thread-171748.html)

euantorano:
This has annoyed me for a long time, but fixing it as a very slow and tedious process. I tend to do it as I change other things :)

JN-Jones:
I could try to fix at least the files in `inc` and `admin/inc`.But remember that the changed files package would get bigger because of this.

PR #2027, opened 2015-06-10
#2026 Wrong Doc Blocks

JN-Jones:
#2026 

Please only comment on the files I've already edited - I haven't looked at the others so comments about others will be ignored anyways.
Also note that I've only tried to fix the doc blocks and haven't changed variables where eg 0/1 is used as boolean switch.

### DIFF
--- a/inc/class_templates.php
+++ b/inc/class_templates.php
@@ -127,7 +127,7 @@ class templates
 	 * Prepare a template for rendering to a variable.
 	 *
 	 * @param string $template The name of the template to get.
-	 * @param boolean $eslahes True if template contents must be escaped, false if not.
+	 * @param boolean $eslashes True if template contents must be escaped, false if not.
 	 * @param boolean $htmlcomments True to output HTML comments, false to not output.
 	 * @return string The eval()-ready PHP code for rendering the template
 	 */
--- a/inc/class_templates.php
+++ b/inc/class_templates.php
@@ -34,7 +34,7 @@ class templates
 	/**
 	 * Cache the templates.
 	 *
-	 * @param string A list of templates to cache.
+	 * @param string $templates A list of templates to cache.
 	 */
 	function cache($templates)
 	{
@@ -56,9 +56,9 @@ class templates
 	/**
 	 * Gets templates.
 	 *
-	 * @param string The title of the template to get.
-	 * @param boolean True if template contents must be escaped, false if not.
-	 * @param boolean True to output HTML comments, false to not output.
+	 * @param string $title The title of the template to get.
+	 * @param boolean|int $eslashes True if template contents must be escaped, false if not.
+	 * @param boolean|int $htmlcomments True to output HTML comments, false to not output.
 	 * @return string The template HTML.
 	 */
 	function get($title, $eslashes=1, $htmlcomments=1)
@@ -126,9 +126,9 @@ class templates
 	/**
 	 * Prepare a template for rendering to a variable.
 	 *
-	 * @param string The name of the template to get.
-	 * @param boolean True if template contents must be escaped, false if not.
-	 * @param boolean True to output HTML comments, false to not output.
+	 * @param string $template The name of the template to get.
+	 * @param boolean $eslahes True if template contents must be escaped, false if not.
+	 * @param boolean $htmlcomments True to output HTML comments, false to not output.
 	 * @return string The eval()-ready PHP code for rendering the template
 	 */
 	function render($template, $eslashes=true, $htmlcomments=true)
@@ -138,6 +138,9 @@ class templates
 
 	/**
 	 * Fetch a template directly from the install/resources/mybb_theme.xml directory if it exists (DEVELOPMENT MODE)
+	 *
+	 * @param string $title
+	 * @return string|bool
 	 */
 	function dev_get($title)
 	{
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -53,6 +53,8 @@ class PostDataHandler extends DataHandler
 	 * post = New post
 	 * thread = New thread
 	 * edit = Editing a thread or post
+	 *
+	 * @var string
 	 */
 	public $action;
 
@@ -119,7 +121,7 @@ class PostDataHandler extends DataHandler
 	 */
 	function verify_author()
 	{
-		global $mybb;
+		global $mybb, $lang;
 
 		$post = &$this->data;
 
@@ -172,7 +174,6 @@ class PostDataHandler extends DataHandler
 	/**
 	 * Verifies a post subject.
 	 *
-	 * @param string True if the subject is valid, false if invalid.
 	 * @return boolean True when valid, false when not valid.
 	 */
 	function verify_subject()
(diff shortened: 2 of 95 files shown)


JN-Jones:
I've changed all doc blocks in "inc" and "admin/inc". Though there are also some functions in other files those are the important ones. @euantorano

inc/db_pgsql.php
```diff
@@ -676,13 +681,13 @@ function field_exists($field, $table)
...
-	 * @param string An optional name for the query.
+	 * @param resource $query The query data.
+	 * @param string $name An optional name for the query.
 	 */
-	function shutdown_query($query, $name=0)
+	function shutdown_query($query, $name="")
 	{
 		global $shutdown_queries;
-		if($name)
+		if(!empty($name))
```
Destroy666x: Any reason you used `!empty()` here? It's a parameter so it's always set, right?

inc/functions.php
```diff
@@ -2953,12 +2961,12 @@ function random_str($length="8")
 /**
  * Formats a username based on their display group
  *
- * @param string The username
- * @param int The usergroup for the user (if not specified, will be fetched)
- * @param int The display group for the user (if not specified, will be fetched)
+ * @param string $username The username
+ * @param int $usergroup The usergroup for the user (if not specified, will be fetched)
```
Destroy666x: http://community.mybb.com/thread-171872.html - 1.

inc/datahandlers/post.php
```diff
@@ -618,8 +628,6 @@ function verify_prefix()
 		}
 		else
 		{
-			$prefix_cache = build_prefixes($prefix);
```
Destroy666x: Are you sure that removing it is safe?

inc/functions.php
JN-Jones: Will fix that on monday

inc/datahandlers/post.php
JN-Jones: Ah missed that. There should be a change though, there was at least one useless variable in this code part. Will recheck on monday too.

inc/db_pgsql.php
JN-Jones: `empty` is the same as `== ""`. Doesn't really matter, it's more a personal preference - looks nicer to me.

Destroy666x: `!empty()` negation also checks `isset()` first which is not needed here. So it's a little overkill. Doesn't almost matter here, I know, but it's not really worth changing.

inc/datahandlers/post.php
Destroy666x: Now that I look again at it, the 2nd removed `build_prefixes();` call is needed as well if the 1st is overwritten by `build_prefixes($prefix)`. The very first `build_prefixes();` (not removed) doesn't seem to be needed though.

inc/db_pgsql.php
JN-Jones: After a quick look most of these cases should be changed to `trim($var) == false`. Is that ok to you?

inc/datahandlers/post.php
JN-Jones: > Also, if I'm not mistaken, the negation here is wrong: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L696 since $num_prefixes should be set true only when a prefix can be used by the user.

That's a seperate issue which I won't fix here.

> And $user may be already set above: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L667-675 so an empty() check in the linked line(s) will possibly save a query.

`get_user` caches users so there it won't run a second query.

inc/db_pgsql.php
Sama34: I don't think it was forced for the name to be an string so I wouldn't go with `trim`, `empty` seems fine.

JN-Jones: Theoretically it could be an integer here too, yes. But that would likely cause some issues. Also `trim` should also work on integers.

JN-Jones: @Destroy666x ? Otherwise I'd like to get this merged ASAP as this PR will likely get conflicts soon and those are a PITA to fix here ;)

Sama34: Somebody is going to get a PITA fixing conflicts though..:p

JN-Jones: A lot of the other PRs don't change doc blocks, they shouldn't get a problem. But I made the experience that git has problems when you change a lot of a file and someone modifies the file in the meantime.

Destroy666x: Well, I see no reason to just stay with `if($name)` as I said. It checked what it was supposed to check - if the parameter contains any value.

Also, is this PR finished? I thought it's still WIP as the title says, if I'm not mistaken not all files were changed as you mentioned.

JN-Jones: I still think that at least the `trim` version should be used but well, changed everything. This isn't WIP anymore

install/upgrade.php
```diff
@@ -682,6 +697,14 @@ function whatsnext()
 	}
 }
 
+/**
+ * Determine the nect function we need to call
```
Destroy666x: next*

inc/class_templates.php
```diff
@@ -126,9 +126,9 @@ function get($title, $eslashes=1, $htmlcomments=1)
 	/**
 	 * Prepare a template for rendering to a variable.
 	 *
-	 * @param string The name of the template to get.
-	 * @param boolean True if template contents must be escaped, false if not.
-	 * @param boolean True to output HTML comments, false to not output.
+	 * @param string $template The name of the template to get.
+	 * @param boolean $eslahes True if template contents must be escaped, false if not.
```
Destroy666x: eslashes*

inc/functions.php
```diff
@@ -1790,10 +1796,10 @@ function get_post_icons()
...
  * MyBB setcookie() wrapper.
  *
- * @param string The cookie identifier.
- * @param string The cookie value.
- * @param int The timestamp of the expiry date.
- * @param boolean True if setting a HttpOnly cookie (supported by IE, Opera 9, Konqueror)
+ * @param string $name The cookie identifier.
+ * @param string $value The cookie value.
+ * @param int|string $expires The timestamp of the expiry date.
+ * @param boolean $httponly True if setting a HttpOnly cookie (supported by IE, Opera 9, Konqueror)
```
Destroy666x: I'd change it to `(supported by the majority of browsers)`

Destroy666x:
What was changed looks and seems to work fine to me (except the 3 small change comments). So I guess this can be merged. But keep in mind that there still may be mistakes, for example just corrected several wrong return values and missing @return here: https://github.com/mybb/mybb/pull/2052/files

euantorano:
Yep, this is a pretty major changes, so I wouldn't be surprised if there are a few small typos. If we can fix the ones noted above, we can merge this (which would be super awesome).

JN-Jones:
Well, I haven't looked at the complete functions, otherwise I would still fix issues in a few years :P Will look at the three changes tomorrow and merge this afterwards if noone else mentions anything till then.

JN-Jones:
Fixed and merged

PR #2032, opened 2015-06-14
Let user select more than three users

cronhound:
Currently, users are limited to selecting three recipients when sending a PM. This commit allows users to select up to the number defined for their usergroup.

I'm unsure if you would like this changed so that the max works across both the to and bcc fields, but right now this is a lot more convenient than what it is currently.


Destroy666x:
Thanks, but as you said your PR doesn't work fully correctly. I have a complete fix which checks both fields dynamically and I'll push it as soon as my exam period will end. Surely before 1.8.6.

PR #2034, opened 2015-06-17
Fixes #1963 Editing problem in IE11

### LINKED ISSUE #1963: Editing problem in IE11
Editing in Full Mode works incorrectly in IE11 when the post contains a hyperlinked image.

I managed to reproduce the problem in this forum (in the sandbox section).

Steps to reproduce:

1. Create a thread where the title post contains a hyperlinked image in the leading line and some text in the following lines. Sample code:

Code:
[img]http://www.metal-archives.com/images/5/4/4/544.jpg?4439[/img]

This is some basic text to edit

2. Try to edit the text line in Full Edit mode. Result = Fail (be it in plain code submode or in WYSIWYG submode)

In Quick Mode, editing works. If we delete the image (in Quick Mode), and then try to edit the remaining text in Full Mode, then it works. So the problem is somehow related to the presence of an image.

Actually, in my production forum I have the Quick Advanced Editor Plus plugin installed, in which environment the Quick Edit mode editing also becomes unsuccessful in the described scenario.

But it would be nice to at least resolve the basic problem, without the plugin interference.

TIA

Original thread: [Editing problem in IE11](http://community.mybb.com/thread-169733.html)

willhartman:
#1963

Also fixed for ACP.

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -1044,6 +1044,7 @@ opt_editor = {
 	rtl: {$lang->settings['rtl']},
 	locale: "mybblang",
 	enablePasteFiltering: true,
+	autoUpdate: true,
 	emoticonsEnabled: {$emoticons_enabled},
 	emoticons: {
 		// Emoticons to be included in the dropdown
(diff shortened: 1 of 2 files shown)


JN-Jones:
You also need to change the template version to `1806`

willhartman:
Thanks, updated.

Destroy666x:
Works for me. Thanks for the PR.

PR #2040, opened 2015-06-18
Fixes #2039 Message after registration English although different lan…

### LINKED ISSUE #2039: Message after registration English although different language pack is used
It is possible that the confirmation message after registration is in English although a different language pack is used. Actually **just** said message is in English, the rest is correct.
english

Stefan-ST:
…guage pack is used

Fixes #2039

### DIFF
--- a/member.php
+++ b/member.php
@@ -527,7 +527,7 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 				// Reset language
 				if(isset($reset_lang))
 				{
-					$lang->set_language($mybb->user['language']);
+					$lang->set_language($mybb->settings['bblanguage']);
 					$lang->load("member");
 				}
 			}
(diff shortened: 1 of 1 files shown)


JN-Jones:
This may also show the message in a wrong language if the board language is different than the selected one. Instead you should check again `$user['language']` which is the correct language for the registered user.

JN-Jones:
Also there are other lines a few lines above the ones you changed (https://github.com/mybb/mybb/pull/2040/files#diff-c6480c1a3fe8c1a339ac23fcdc05d97dR515) where it's checked whether the language of the recipient is different than the one of the registered user. That may also cause wrong messages:
- User selected language a
- Admin1 selected language b
- Admin2 selected language a
Language a is loaded, for admin1 language b is loaded but as admin2 has the same language as the user language a isn't reloaded for that admin so his message is in language b.

Stefan-ST:
> Instead you should check again `$user['language']` which is the correct language for the registered user.

@JN-Jones No, `$mybb->settings['bblanguage']` is correct. Check how the language is set in global.php:
```php
$lang->set_language($mybb->settings['bblanguage']);
```

Cu8eR:
+1, works fine

PR #2044, opened 2015-06-18
Fixes #2043 Use protocol relative URLs for video MyCode

### LINKED ISSUE #2043: Use protocol relative URLs for video MyCode
If possible video MyCodes should use protocol relative URLs (and not HTTPS) to avoid mixed content warnins on forums using HTTPS.

Stefan-ST:
Fixes #2043

The remaining services don't seam to offer HTTPS support.


Destroy666x:
Any reason to include empty an upgrade file here? There are 2 or more PRs with an existing upgrade file which may get broken due to a conflict with this.

Would be nice to take care of Gravatar in this PR too: http://community.mybb.com/thread-168764.html since only 2 lines in `usercp.php` and ACP's `users.php` module need to be changed. They should be consistent too - one appends `.jpg` extension and the other not.

Sama34:
I'm curious about the upgrade file as well. If the other PRs don't get merged (for whatever reason) there would then be not need for the file then anyways because IIRC upgrade files are not required if not changes are performed.

Stefan-ST:
@Destroy666x It's probably better to create a new ticket for Gravatar. This ticket was just meant to update the embed codes.
@Sama34 The upgrade script needs to be there for changing the templates. I actually haven't checked other PRs for adding the same file... However I'll update my PR if another PR gets merged at first.

Destroy666x:
> This ticket was just meant to update the embed codes.

Yeah, but it only requires 2 simple line changes, so I don't see why not this too. Anyways, issue created.

>  The upgrade script needs to be there for changing the templates.

Weird. I recall at least few versions in which we skipped upgrade script, yet templates were updated.

Stefan-ST:
@Destroy666x Do you have an example? There are only a few updates without an upgrade script.

JN-Jones:
There needs to be a file as that's how the upgrade script builds its select box.

Destroy666x:
@Stefan-ST, never mind, I thought 1.8.2 and 1.8.3 had some template changes but I forgot they were all bumped to 1.8.4 without being introduced in changed files

JN-Jones:
As this fixes already some of the issues I've mentioned [http://community.mybb.com/thread-167116.html?highlight=video](here) could you also include the other fixes here so we don't need to push that issue and then look what of it has already been fixed?

euantorano:
Should probably create a different PR for that @JN-Jones rather than confusing this one.

PaulBender:
Looks good to me

PR #2052, opened 2015-06-23
Fix #2051 wrong is_member() negation

### LINKED ISSUE #2051: Wrong is_member() negation
As I mentioned in another PR, this check shouldn't be negated: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L696 - `$num_prefixes` should be `true` only if the user can use a prefix. And the `-1` check is not needed anymore either.

Destroy666x:
Fixes one invalid `is_member()` negation, returns empty strings instead
of `false` for HTML code, shortens redundant code (`-1` returns always
`true` for `is_member()` and `get_user()` the current user - `$mybb->user` - if the UID
matches).

https://github.com/mybb/mybb/issues/2051

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -699,14 +699,7 @@ class PostDataHandler extends DataHandler
 						}
 					}
 
-					if($required['groups'] != "-1")
-					{
-						if(!is_member($required['groups'], array('usergroup' => $user['usergroup'], 'additionalgroups' => $user['additionalgroups'])))
-						{
-							$num_prefixes = true;
-						}
-					}
-					else
+					if(is_member($required['groups'], array('usergroup' => $user['usergroup'], 'additionalgroups' => $user['additionalgroups'])))
 					{
 						$num_prefixes = true;
 					}
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3502,7 +3502,7 @@ function build_prefixes($pid=0)
 }
 
 /**
- * Build the thread prefix selection menu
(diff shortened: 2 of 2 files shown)


inc/datahandlers/post.php
```diff
@@ -708,14 +708,7 @@ function verify_prefix()
...
 
-					if($required['groups'] != "-1")
-					{
-						if(!is_member($required['groups'], array('usergroup' => $user['usergroup'], 'additionalgroups' => $user['additionalgroups'])))
-						{
-							$num_prefixes = true;
-						}
-					}
-					else
+					if(is_member($required['groups'], array('usergroup' => $user['usergroup'], 'additionalgroups' => $user['additionalgroups'])))
```
JN-Jones: As I've already mentioned elsewhere: Why not pass `$user` directly? (IIRC I've already change it here in one of my PRs)

Destroy666x: I can change it, but I don't think it makes a big difference. This way clearly shows that groups are passed and the function only receives the values it needs.

JN-Jones: Imho it's easier to read: it checks whether `$user` is in `$required['groups']`

JN-Jones:

JN-Jones:
@Destroy666x can you fix the conflicts? ;)

JN-Jones:
@Destroy666x

Destroy666x:
Will try to fix 2-5 of my assigned issues a day starting this weekend.

Destroy666x:
Conflicts were fixed, somehow

Issue #2075, opened 2015-06-25
Dashboard quick access items lack permission checks.


Destroy666x:
The side panel quick access items in the ACP lack permission checking.

For instance, if I cannot manage plugins I will see the link nonetheless.

Original thread: [Dashboard quick access items lack permission checks.](http://community.mybb.com/thread-169260.html)

Cu8eR:
Isnt it fixed?

PaulBender:
Yes, it was fixed in 1.8.6.

PR #2082, opened 2015-06-26
Fix 2075

Naththo:
#2075

### DIFF
--- a/admin/modules/home/module_meta.php
+++ b/admin/modules/home/module_meta.php
@@ -80,6 +80,7 @@ function home_action_handler($action)
 				unset($sub_menu[$id]);
 			}
 		}
+		
 		$sub_menu = $plugins->run_hooks("admin_home_menu_quick_access", $sub_menu);
 
 		$sidebar = new SidebarItem($lang->quick_access);
--- a/admin/modules/home/module_meta.php
+++ b/admin/modules/home/module_meta.php
@@ -81,12 +81,15 @@ function home_action_handler($action)
 			}
 		}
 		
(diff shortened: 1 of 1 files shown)


Destroy666x:
Looks good, but there are 2 problems:
1. Add New Forum is not hidden, the module is wrong.
2. If someone doesn't have permission to see any quick link, the Quick Access menu is still shown. These lines: https://github.com/Naththo/mybb/blob/issue2075/admin/modules/home/module_meta.php#L86-89
should be wrapped by an `if` which should check whether the `$sub_menu` array is filled.

Naththo:
Updated.

admin/modules/home/module_meta.php
```diff
@@ -66,19 +66,30 @@ function home_action_handler($action)
...
+		{
+			if(!check_admin_permissions(array("module" => $sub['module'], "action" => $sub['action']), false))
+			{
+				unset($sub_menu[$id]);
+			}
+		}
+		
+		if(!empty($sub_menu))
+		{
+			$sub_menu = $plugins->run_hooks("admin_home_menu_quick_access", $sub_menu);
```
Destroy666x: The hook shouldn't be inside the conditional, plugins may add their own quick links and `$sub_menu` won't be empty then.

Naththo:
Ope, totally missed that.

Destroy666x:
In the last commit you moved `$page->sidebar .= $sidebar->get_markup();` which should be inside the conditional, not the hook

Destroy666x:
Thanks (wrong module is still used for Add New Forum but I'll commit this directly).

PR #2097, opened 2015-06-27
Fix #2093 - "Sending message to myself with BCC will cause an error"

### LINKED ISSUE #2093: Sending message to myself with BCC will cause an error
I (Jones) send a message to myself with another user as BCC. When I'm using the Quick Reply (not sure whether the full reply has the same error) to answer (still logged in as Jones) it will cause an error: "[...] Invalid recipients: &lt;!-- start: private_read_bcc --&gt; &lt;br /&gt;BCC: &lt;a href=". Likely a variable is used twice.

Original thread: [Sending message to myself with BCC will cause an error](http://community.mybb.com/thread-159036.html)

Naththo:
#2093

### DIFF
--- a/private.php
+++ b/private.php
@@ -1112,6 +1112,7 @@ if($mybb->input['action'] == "read")
 
 	// Fetch recipient names from the database
 	$bcc_recipients = $to_recipients = array();
+	$bcc_form_val = $to_recipients = array();
 	$query = $db->simple_select('users', 'uid, username', "uid IN ({$uid_sql})");
 	while($recipient = $db->fetch_array($query))
 	{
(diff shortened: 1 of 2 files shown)


private.php
```diff
@@ -1119,11 +1120,13 @@
 		if($show_bcc && in_array($recipient['uid'], $pm['recipients']['bcc']))
 		{
 			$bcc_recipients[] = build_profile_link($recipient['username'], $recipient['uid']);
+			$bcc_form_val[] = $recipient['username'];
 		}
 		// User is a normal recipient
 		else if(in_array($recipient['uid'], $pm['recipients']['to']))
 		{
 			$to_recipients[] = build_profile_link($recipient['username'], $recipient['uid']);
+			$bcc_form_val[] = $recipient['username'];
```
Destroy666x: Why do you add regular recipients to the BCC field? This line is incorrect.

private.php
```diff
@@ -1112,6 +1112,7 @@
 
 	// Fetch recipient names from the database
 	$bcc_recipients = $to_recipients = array();
+	$bcc_form_val = $to_recipients = array();
```
Destroy666x: Why did you add it like that? Everything is an empty array, so you could just do `$bcc_form_val = $bcc_recipients = $to_recipients = array();`

private.php
```diff
@@ -1131,6 +1134,7 @@
 	if(count($bcc_recipients) > 0)
 	{
 		$bcc_recipients = implode(', ', $bcc_recipients);
+		$bcc_form_val = implode(', ', $bcc_form_val);
```
Destroy666x: The input should be comma-separated names without spaces after commas.

JN-Jones:
Thanks!

Issue #2106, opened 2015-06-28
Report Center Bug


JN-Jones:
If you soft delete a post it doesn't remove it from the unread reports in the Mod CP.   I'm fairly certain this is not intentional.

Steps to reproduce:
1) Report any post.
2) Soft Delete the post.
3) The report doesn't get marked as read.

Original thread: [Report Center Bug](http://community.mybb.com/thread-158746.html)

PR #2109, opened 2015-06-28
#2106 - Fix reports not getting deleted when soft deleting a post

Naththo:
#2106

### DIFF
--- a/moderation.php
+++ b/moderation.php
@@ -2710,7 +2710,6 @@ switch($mybb->input['action'])
 
 	// Soft delete posts - Inline moderation
 	case "multisoftdeleteposts":
-
 		// Verify incoming POST request
 		verify_post_check($mybb->get_input('my_post_key'));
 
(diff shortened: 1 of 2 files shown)


JN-Jones:
The issue should be fixed in class_moderation.php. Also you don't delete reports for the firstpost if the thread is deleted.

Destroy666x:
Now it only marks reports for threads, what about posts?

moderation.php
```diff
@@ -2739,8 +2738,9 @@
 		}
 
 		$moderation->soft_delete_posts($pids);
-
 		log_moderator_action($modlogdata, $lang->multi_soft_delete_posts);
+
```
Destroy666x: Please delete one extra line

JN-Jones:
Also note that there isn't a report type named "threads", the first post is used instead. Also if a thread is soft deleted all posts should be checked imho.

Destroy666x:
@JN-Jones, not sure what you mean,threads is correct: https://github.com/mybb/mybb/blob/feature/inc/functions.php#L4574

Destroy666x:
Yep, I can confirm it works fine with both single and inline thread/post soft deletion. Reports for all posts in a soft deleted threads are marked.

JN-Jones:
Ah, thought `mark_reports` only accepts the types we store in the database - yeah, it'll mark all posts with that query.

PR #2120, opened 2015-07-01
Fixes #2115 Inconsistent checks in db classes

### LINKED ISSUE #2115: Inconsistent checks in db classes
Most function get the query as parameter. While MySQL(i) and PgSQL simpy call the respictive functions then, SQLite always has "is_object" checks. Either all classes/function should get a proper check (&#36;query instanceof ...) or the checks should be removed.

Original thread: [Inconsistent checks in db classes](http://community.mybb.com/thread-171782.html)

JN-Jones:
#2115


euantorano:
Looks good to me.

PR #2121, opened 2015-07-01
Fixes #2102 Cache classes and their constructors

### LINKED ISSUE #2102: Cache classes and their constructors
The "datacache" class passes a variable "&#36;silent" to the constructor and calls "connect" afterwards without any parameters. However the different classes ignore that:

APC (ok):
-&gt; function __construct(&#36;silent)
-&gt; function connect()

Disk (wrong):
-&gt; function connect(&#36;silent)

eaccelerator (partly ok; others are like this too):
-&gt; function eacceleratorCacheHandler(&#36;silent): is a valid constructor but "__construct" should be used instead
-&gt; function connect()

A new interface should be added which has all the cache functions and forces the first version imho, similar to what @Euan did with the database classes.

Original thread: [Cache classes and their constructors](http://community.mybb.com/thread-171784.html)

JN-Jones:
#2102


euantorano:
Looks good to me. I only have APC, disk and memcache to test against - can anybody verify the other caches just to be sure.

inc/cachehandlers/disk.php
```diff
@@ -33,24 +32,16 @@ function connect($silent=false)
...
 
-		if(!isset($this->cache[$name]) || $hard_refresh == true)
-		{
-			@include(MYBB_ROOT."/cache/{$name}.php");
-		}
-		else
-		{
-			@include_once(MYBB_ROOT."/cache/{$name}.php");
-		}
+		@include(MYBB_ROOT."/cache/{$name}.php");
```
Sama34: Why not `include_once`?

JN-Jones: I'm not 100% sure how the disk cache work. However it seems to save the variable as php variable. And an easy example of how to break it would be to save a cache named `name` (though that would likely break anyways). Just used that as it was always used till now (the variable in the if was never set as it doesn't exist).

Stefan-ST: include_once wouldn't include the cache when it's loaded the second time.

JN-Jones:
@Destroy666x @euantorano

JN-Jones:
@mybb/developers Can we merge this?

JN-Jones:
@Destroy666x @euantorano @PaulBender

PaulBender:
Works for me, but I don't got all the cache handlers to test against.

JN-Jones:
Anyone else?

JN-Jones:
@mybb/developers

JN-Jones:
Bump

JN-Jones:
Bump

euantorano:
I've only tested Disk, Memcache and APC. They all work fine for me and the other patches look okay. If nobody else has any other caching setups, we'll have to merge and hope I guess.

JN-Jones:
Guess that goes as merge and hope?

euantorano:
I'd say so. I've tested what I can, but don't have any other working cache systems.

JN-Jones:
I'll leave it open for another 4 hours and merge it afterwards if nobody comments

PR #2134, opened 2015-07-02
Fixes #2128 sessions - useragent column still varchar(100)

### LINKED ISSUE #2128: sessions - useragent column still varchar(100) ?
This issue is so old i though it was fixed 100 years ago allready, but i see its not.

In 1.8.x table sessions - column useragent is still varchar(100) which causes useragent string to be cut,
therefore for long strings it cannot be analysed / recognized properly when needed. 

Original thread: [sessions - useragent column still varchar(100) ?](http://community.mybb.com/thread-168339.html)

JN-Jones:
#2128 

~~Note that I hadn't time to check the upgrade script against pgsql/sqlite. Will do so later, so please wait till then before mentioning issues with that ;)~~


install/resources/upgrade34.php
```diff
@@ -0,0 +1,41 @@
...
+
+function upgrade34_dbchanges()
+{
+	global $db, $output;
+
+	$output->print_header("Updating Database");
+	echo "<p>Performing necessary upgrade queries...</p>";
+	flush();
+
+	if($db->type != 'pgsql') {
```
Stefan-ST: You don't follow the coding standards here.

JN-Jones: As said: the upgrade isn't tested or anyhow like it should. It's there to remember myself that it's needed.

install/resources/upgrade34.php
```diff
@@ -0,0 +1,44 @@
...
+		$db->modify_column('sessions', 'useragent', "varchar(200) NOT NULL default ''");
+	}
+	else
+	{
+		$db->modify_column('adminsessions', 'useragent', "varchar(200)", "set", "");
+		$db->modify_column('sessions', 'useragent', "varchar(200)", "set", "");
+	}
+
+	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
+	$output->print_footer("33_done");
```
PaulBender: This should be 34_done

JN-Jones: > Note that I hadn't time to check the upgrade script against pgsql/sqlite. Will do so later, so please wait till then before mentioning issues with that ;)

PaulBender:
I've tested this and it works fine.

Upgrade works for MySQL and SQLite. Can't test PostgreSQL.

PaulBender:
I've installed PostgreSQL and this does produce errors.

This:
**$db->modify_column('adminsessions', 'useragent', "varchar(200)", "set", "");
$db->modify_column('sessions', 'useragent', "varchar(200)", "set", "");**

needs to be:

**$db->modify_column('adminsessions', 'useragent', "varchar(200)", "set", "''");
$db->modify_column('sessions', 'useragent', "varchar(200)", "set", "''");**

JN-Jones:
As already mentioned three times: this isn't finished yet...

JN-Jones:
This is now finished.

Destroy666x:
Noticed no problem.

PR #2146, opened 2015-07-06
Fixes #2145 Add A Group Leader Doesn't Put Leader In Usergroup

### LINKED ISSUE #2145: Add A Group Leader Doesn't Put Leader In Usergroup
If you add a person as a group leader in the Admin CP, it does not put the person in the usergroup.  I believe the person should have their additional usergroups updated to be in the usergroup.

Github: https://github.com/mybb/mybb/blob/featur...s.php#L314

Original thread: [Add A Group Leader Doesn't Put Leader In Usergroup](http://community.mybb.com/thread-171694.html)

PaulBender:
Fixes #2145


admin/modules/user/groups.php
```diff
@@ -468,7 +474,8 @@
 		$mybb->input = array_merge($mybb->input, array(
 				"canmanagemembers" => 1,
 				"canmanagerequests" => 1,
-				"caninvitemembers" => 1
+				"caninvitemembers" => 1,
+				"makeleadermember" => 1
```
JN-Jones: IMHO the default should be "no" as that's how it was till now.

Destroy666x: Agreed

PaulBender: Fixed

JN-Jones:
Looks fine to me, need to test it later when I'm home properly.

Destroy666x:
I can confirm it works.

PR #2158, opened 2015-07-19
Fix #257: Forum Mail as 'from' adress

### LINKED ISSUE #257: Change to SMTP mail composition to comply with standards

Evenprime:
Referencing Issue #257 :

Using a users mail adress in the "from" part of an email will always fail if the domain of that email has a strict SPF policy, like yahoo.com, aol.com and many more in the future, as this will probably become the new standard to prevent mail spoofing. It is impossible to send such emails, as these technically do have spoofed sender adresses.

The user-to-user mails already include all important information about who actually sent the message and that it originates from the forum, therefore the "from"-adress of the mail could and should simply be the default adress of the forum that is used for all other notifications etc. The reply-to part of the mail is already correctly set to the users mail adress.

### DIFF
--- a/contact.php
+++ b/contact.php
@@ -227,7 +227,7 @@ if($mybb->request_method == "post")
 		$message = $lang->sprintf($lang->email_contact, $mybb->input['email'], $user, $session->ipaddress, $mybb->input['message']);
 
 		// Email the administrator
-		my_mail($contactemail, $subject, $message, $mybb->input['email']);
+		my_mail($contactemail, $subject, $message, "", "", "", false, "text", "", $mybb->input['email']);
 
 		$plugins->run_hooks('contact_do_end');
 
--- a/member.php
+++ b/member.php
@@ -2836,17 +2836,8 @@ if($mybb->input['action'] == "do_emailuser" && $mybb->request_method == "post")
 
 	if(count($errors) == 0)
 	{
-		if($mybb->settings['mail_handler'] == 'smtp')
-		{
-			$from = $mybb->input['fromemail'];
-		}
-		else
-		{
-			$from = "{$mybb->input['fromname']} <{$mybb->input['fromemail']}>";
-		}
-
 		$message = $lang->sprintf($lang->email_emailuser, $to_user['username'], $mybb->input['fromname'], $mybb->settings['bbname'], $mybb->settings['bburl'], $mybb->get_input('message'));
-		my_mail($to_user['email'], $mybb->get_input('subject'), $message, $from, "", "", false, "text", "", $mybb->input['fromemail']);
+		my_mail($to_user['email'], $mybb->get_input('subject'), $message, "", "", "", false, "text", "", $mybb->input['fromemail']);
 
 		if($mybb->settings['mail_logging'] > 0)
 		{
--- a/sendthread.php
+++ b/sendthread.php
@@ -198,21 +198,12 @@ if($mybb->input['action'] == "do_sendtofriend" && $mybb->request_method == "post
 	// No errors detected
 	if(count($errors) == 0)
 	{
-		if($mybb->settings['mail_handler'] == 'smtp')
-		{
-			$from = $mybb->input['fromemail'];
-		}
-		else
-		{
-			$from = "{$mybb->input['fromname']} <{$mybb->input['fromemail']}>";
-		}
-
 		$threadlink = get_thread_link($thread['tid']);
 
 		$message = $lang->sprintf($lang->email_sendtofriend, $mybb->input['fromname'], $mybb->settings['bbname'], $mybb->settings['bburl']."/".$threadlink, $mybb->input['message']);
 
 		// Send the actual message
-		my_mail($mybb->input['email'], $mybb->input['subject'], $message, $from, "", "", false, "text", "", $mybb->input['fromemail']);
+		my_mail($mybb->input['email'], $mybb->input['subject'], $message, "", "", "", false, "text", "", $mybb->input['fromemail']);
 
 		if($mybb->settings['mail_logging'] > 0)
 		{


Destroy666x:
This doesn't fix the issue, only a part of it.. This sentence is wrong:

> The only situation that I can think of where the from adress is not the forum adress, is the sending of mails from users to users (the initial reason for this issue report)

There are at least 2 more places with changed `$from` parameter - contact.php and sendthread.php. `class_error.php` Possibly more in plugins.

Also, most people are not be forced to follow SPF and the there's no reason to remove the custom from Emails in that case. 

Thus as I mentioned in another PR, the best solution would be adding a new settting to force `returnemail`/`adminemail` (depending on if the 1st is filled by using the new function) IMO.

Evenprime:
I've updated the pull request to include the other two locations you mentioned.

Please note, that the pull request doesn't change the "my_mail" function. Any plugin or forum function that sets the "$from" parameter when calling "my_mail" will still result in that as the "from"-value in the resulting mail. Therefore no plugins or other functions will be influenced by this change.

It just, for these (now) three instances, changes the "from" adress to the standard adress of the forum that is used everywhere else successfully to send emails to users, because in these three cases the from adress isn't determined by the administrator configuratively or a plugin author, but arbitrarily by the user.

A user with a yahoo.com or aol.com mail adress will never be able to use any of those three features reliably/successfully, as most recipients will simply deny delivery. Therefore I consider the current behaviour to be a bug and not something that needs to have some configuration value.

If a plugin or another feature of the forum needs to have alternative "from" adresses, it would (again) be the responsibility of the administrator or plugin author to make sure that the then used "from" adresses are actually usable, which means especially not the registration mail adress of random users. Otherwise those plugin authors should too just omit the "$from" in the call to "my_mail" in the future (which automatically falls back to the forum mail adress) and instead set the reply-to value accordingly, like in those three spots in this pull request.

Destroy666x:
> A user with a yahoo.com or aol.com registration mail adress will never be able to use any of those three features reliably/successfully, as most recipients will simply bounce back the mails that he attempts to send through the forum features. Therefore I consider the current behaviour to be a bug and not something that needs to have some configuration value.

I don't think bouncing them back is the case for most recipients (rather the other way round) and "never" is definitely an exaggeration here.

> Therefore no plugins or other functions will be influenced by this change.

Which doesn't really help admins who are having problems with that. As you said they will have to change wrong codes manually and that's definitely much harder for inexperienced users than switching on a setting to cover everything. They can possibly have more issues, for example the usual BOM bugs due to wrong text editor usage.

What do others think? @JN-Jones @euantorano @Sama34 @Devilshakerz @ATofighi

Devilshakerz:
This pull request fixes the core code that's obviously incorrect - nobody from the MyBB Team advised the erroneous usage of this function so I don't think there's a reason for adding a setting that covers for plugins that were badly coded as well.

Evenprime:
>I don't think bouncing them back is the case for most recipients (rather the other way round) and "never" is definitely an exaggeration here.

Only a few free mailing services publish such hard rules like yahoo and aol. But all major players like google, outlook/hotmail, ... started checking and respecting those security records. 

Which means that sending forged mails from your server with "from" gmail.com and "to" yahoo.com will work most of the time (gmail doesn't prevent that, intentionally), but sending forged mails from your server with "from" yahoo.com and "to" gmail.com will never work. That is the whole point of that security mechanisms that have been put in place by yahoo etc.

Destroy666x:
> nobody from the MyBB Team advised the erroneous usage of this function so I don't think there's a reason for adding a setting that covers for plugins that were badly coded as well

We did it by introducing that paramater to the function and especially by misusing it ourselves  - there's no correct usage of it in the core at all if I'm not mistaken. Many fresh (and not only fresh) MyBB devs blindly copy codes which are even more incorrect than this, such as vulnerable queries.

Evenprime:
>Fixing it globally in the function if the admin needs it has no drawbacks IMO. We could also reject only emails that in fact don't meet SPF with a simple regex match or any other basic string comparision.

Every admin that has at least one visitor with a "yahoo" or "aol" mail adress would need this setting. Unless you have an obscure forum where visitors will never use neither mail host as their registration email adress, you would need that.

The SPF status of mail adresses can be quite dynamic. Unless you want to actually do regular DNS queries and parse the results for every "from" mail adress every time a mail gets sent, it would never be reliable. Also, aol and yahoo are simply the most well known big mail services that use this protection scheme. Everybody can use this for their domain, e.g. lots of small companies and businesses will likely protect their domains the same way from email forgery. It is, after all, just an additional DNS entry for your domain.

I think even if the forum itself would no longer use it, there is a legitimate reason to keep the "from" parameter in the method signature. Plugins may want to or need to handle additional sending domains. Maybe someone writes a "newsletter"-plugin that sends mails from a "newsletter@yourdomain.com" adress instead of the otherwise used "no-reply@yourdomain.com". Forcing everything to use the same "from" adress would imho be overkill and cause too many restrictions.

Sama34:
We should not introduce a setting to allow administrators _skip_ this _fix_. If that is the case.

Destroy666x:
> Every admin that has at least one visitor with a "yahoo" or "aol" mail adress would need this setting. Unless you have an obscure forum where visitors will never use neither mail host as their registration email adress, you would need that.

Not every MyBB forum is American/international and those which aren't don't need to be obscure.

> Maybe someone writes a "newsletter"-plugin that sends mails from a "newsletter@yourdomain.com" adress instead of the otherwise used "no-reply@yourdomain.com". Forcing everything to use the same "from" adress would imho be overkill and cause too many restrictions.

And that's exactly why I suggested a simple "yourdomain.com" string check in my comment above.

> We should not introduce a setting to allow administrators skip this fix. If that is the case.

There's no mentioned skipping here, rather covering a bigger number of affected incorrect function calls, if necessary.

Evenprime:
Ah, sorry, I misunderstood your intention for what would be checked by a regex.

This pull request would fix Issue #2162 too.

Stefan-ST:
@mybb/developers @mybb/sqa These changes appear reasonable to me. Anybody against merging it?

Devilshakerz:
Good to go

Destroy666x:
For the nth time, even if we decide that this is the way to go (basically not using the parameter and blindly relying on plugins doing the same...), the PR is still not complete because the parameter is still misused in `inc/class_error.php` and `inc/functions_task.php` due to `get_from_email()`. The admin might want to use proper return mail for sending and his own Email for those notifications, but it's not possible with the parameter being there for no reason.

Another thing is that admins may have no idea that `adminemail` (or `returnemail` when it's filled) is "always" used for `From:` because:
a) the installation doesn't mention it when "Contact Email:" (the field name doesn't help either..) is about to be entered
b) the setting is in Site Details, its different name ("Admin Email" this time) doesn't help either and the description can be missed. "Return Email" isn't much better.

It would be nice to correct a) at least.

Stefan-ST:
@Destroy666x You are right about inc/class_error.php (although it is far less important than the other instances).

IMHO the description of adminemail is pretty clear:
> This will be used for outgoing emails sent via the forums.

returnemail is indeed confusing. However this shouldn't block this PR for months.

Destroy666x:
Well, removing the parameter from these 2 files can be done quickly. If not by @Evenprime then by us if we cherry-pick it or whatever.

> IMHO the description of adminemail is pretty clear

No, it's not clear, it doesn't say that return Email will be used instead if you fill it. It implies it will be always used. Also, you didn't refer to the installation part where it doesn't say anything about its usage, so even experienced admins may change it to something incorrect.

Stefan-ST:
Why does it use returnemail as From anyway? It isn't what I would expect it to do and it clearly isn't what the descriptions tell.
I would expect:
From = adminemail
Return To = returnemail (or adminemail when not set)
Reply To = email of user (contact form, sendthread)

Destroy666x:
See, you wouldn't expect that and yet we expect admins to expect it...

It's because of this PR: https://github.com/mybb/mybb/pull/1732 which mentioned admin Email exposal. Whether I agree? Probably, I think it's better the current way, but not if it's not described properly. It's not good either way if there's no understandable description.

Stefan-ST:
I don't think we should fix the issues introduced by said PR here. But it should be fixed nonetheless (possibly by adding another setting for the real admin email not being exposed anywhere).

Destroy666x:
Feel free to merge and open a new issue for the rest in that case.

Destroy666x:
As for installation, it's even worse, the contact/admin/outgoing/whatever Email isn't even required to be entered, you can leave it blank...

JN-Jones:
@Evenprime @Destroy666x @Stefan-ST what's with this now?

JN-Jones:
@mybb/developers @mybb/sqa

Ben-Conway:
again @mybb/developers @mybb/sqa ?

kawaii:
@mybb/developers @mybb/sqa has any decision been made on this yet? Fairly high priority issue which has for the most part been ignored for the best part of a year.

euantorano:
@kawaii Just needs testing, then merging. I agree with @Devilshakerz that we should open another issue/PR to fix the other issues

PR #2159, opened 2015-07-19
Fix #907: Set posthandler-action for proper "subject" validation in preview

### LINKED ISSUE #907: Subject  Max. Char. error on preview
if the thread subject has more than 85 characters, mybb will remove the "RE: " part while posting replies... but the same doesnt work if we preview the post in such threads... it returns the usual error like "Please correct the following errors before continuing: The subject is too long. Please enter a subject shorter than 85 characters (currently 89)." on preview....

Original thread: [Subject  Max. Char. error on preview](http://community.mybb.com/thread-155845.html)

Evenprime:
Fix for Issue #907 

The posthandler action is used to determine the mode of validation of post subjects. When saving a reply, it is set to "post". When previewing a reply, it is currently not set at all.

Simply setting the action to "post" fixes the validation of subjects for previews.

### DIFF
--- a/newreply.php
+++ b/newreply.php
@@ -982,6 +982,7 @@ if($mybb->input['action'] == "newreply" || $mybb->input['action'] == "editdraft"
 		// Set up posthandler.
 		require_once MYBB_ROOT."inc/datahandlers/post.php";
 		$posthandler = new PostDataHandler("insert");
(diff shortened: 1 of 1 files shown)


Destroy666x:
Noticed no problem wit this solution except a bigger bug.. Thanks for your contribution.

PR #2166, opened 2015-07-23
Fixes #2132 video codes may not be loaded correctly

### LINKED ISSUE #2132: video codes may not be loaded correctly
Yahoo has two different url schemes:

http://{locale}.screen.yahoo.com/{title}-{id}.html
http://{locale}.screen.yahoo.com/{channe...-{id}.html

The first one is loaded correctly, however the second one uses the channel name as video id which results in a broken iframe (or may even result in a wrong video).

Original thread: [video codes may not be loaded correctly](http://community.mybb.com/thread-167116.html)

Stefan-ST:
Fixes #2132

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1352,10 +1352,24 @@ class postParser
 				$title = htmlspecialchars_uni($path[3]);
 				break;
 			case "myspacetv":
-				$id = $path[4]; // http://www.myspace.com/video/fds/fds/123
+				if($path[1] == 'video')
+				{
+					$id = $path[4]; // http://www.myspace.com/video/fds/fds/123
+				}
+				else
+				{
+					$id = $path[4]; // http://www.myspace.com/fds/video/fds/123
+				}
 				break;
 			case "facebook":
-				$id = $input['v']; // http://www.facebook.com/video/video.php?v=123
+				if(isset($input['v']))
+				{
+					$id = $input['v']; // http://www.facebook.com/video/video.php?v=123
+				}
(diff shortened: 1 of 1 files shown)


inc/class_parser.php
```diff
@@ -1352,10 +1352,24 @@ function mycode_parse_video($video, $url)
 				$title = htmlspecialchars_uni($path[3]);
 				break;
 			case "myspacetv":
-				$id = $path[4]; // http://www.myspace.com/video/fds/fds/123
+				if($path[1] == 'video')
+				{
+					$id = $path[4]; // http://www.myspace.com/video/fds/fds/123
```
Destroy666x: This conditional doesn't make any sense.

Stefan-ST: Yeah, I've initially had a different code for the new format. Thanks.

PaulBender:
The Facebook code doesn't work right. $id = $path[4]; needs to be $id = $path[3];

The other two work fine.

Stefan-ST:
@PaulBender What URL have you used? I'm pretty sure I tested this successfully but there might be even more patterns used by Facebook.

PaulBender:
I used this styling:
https://www.facebook.com/viralvidss/videos/377511862456521/

Stefan-ST:
When you e.g. click on the next link you'll see the styling I used. I'll fix that later.

Stefan-ST:
It should both URL formats now.

PaulBender:
Works fine for me

PR #2175, opened 2015-08-02
Fixes #2164 output_row() $options['id']

### LINKED ISSUE #2164: output_row() $options['id']
The following code in admin/inc/class_form.php makes not much sense:
```php
		if(isset($options['id']))
		{
			$options['id'] = " id=\"{$options['id']}\"";
		}
		else
		{
			$options['id'] = '';
		}
		//...
		$row .= "<div class=\"form_row\"{$options['id']}>{$content}</div>\n";

		$this->_container->construct_cell($row, $options);
```
Either $options shouldn't be modified or $options['id'] shouldn't be passed to construct_cell. Right now it generates code like this: ```<td class="first" id=" id="foo"">```

PaulBender:
Fixes #2164

### DIFF
--- a/admin/inc/class_form.php
+++ b/admin/inc/class_form.php
@@ -966,21 +966,12 @@ class DefaultFormContainer
 			$row = "<label{$for}>{$title}</label>";
 		}
 
-		if(isset($options['id']))
-		{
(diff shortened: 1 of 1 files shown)


Destroy666x:
I didn't notice any problem.

PR #2176, opened 2015-08-05
Fix positioning of edit popup

leefish:
Edit popup does not work on some themes due to the use of relative
positioning; this fixes the issue

### DIFF
--- a/jscripts/jquery.plugins.js
+++ b/jscripts/jquery.plugins.js
@@ -504,6 +504,13 @@
 
 			// Setup popup menu
 			var offset = el.offset();
+			
+			var el_parent = el.get(0).parentNode;
+			var popup_parent = popup_menu.get(0).parentNode;
+			if (el_parent === popup_parent)
+			{
+			  offset = el.position();
+			}
 			offset.top += el.outerHeight();
 
 			// We only adjust if it goes out of the page (?)


Cu8eR:
Good catch!

Stefan-ST:
@leefish Please create a ticket before submitting a pull request. This make it much easier for us to manage bugs.

Destroy666x:
Not sure what the fix does yet and hopw to reproduce the issue, but it's definitely incomplete:
- jquery.plugins.min.js is not affected by the change
- the `ver` parameter of the plugin file is not updated to `1806` in templates/files which load it

leefish:
Here is the post that covers the isue:

http://community.mybb.com/thread-159067.html

I am unsure how to create a ticket. If the code as is is acceptable then I will also do the min.js and the template changes

Destroy666x:
The ticket is ok but you should add at least 3 labels (version, bug/enhancement, status) and set the milestone.

The code seems acceptable, but the edit popup menu is 2 pixels too high compared to the previous position when I test it. The multipage popup is still ok on the other hand.

leefish:
So removing the margin completely in the css would fix it? It seems like it is the css that is the issue rather than the js, and that is theme related (css).

Destroy666x:
It's JS related because without this commit the `margin` doesn't affect the popup position negatively. And we can't expect every element with popup menu to have no `margin` in themes/plugins, same way as we can expect to have different positioning of `#content`.

I guess it's related somehow to the lack of margins in `outerHeight()` - no `true` for the 1st argument there: http://stackoverflow.com/a/12749875 But it seems to be slightly mispositioned to the bottom when `true` is added. Should be something in the middle.

leefish:
Ok, this code in the pull is the fix I have: if it is unsuitable and instead the current situation is more acceptable then just don't accept the fix. If you want to accept what I do have, then I will make the other changes. Personally, I find fixing the margin on a popup in a theme a lesser restriction than positioning of content.

Let me know.

Sama34:
How would a CSS solution look?

leefish:
What do you mean by a css solution? 

The js as it is does not position the dropdown correctly if the parent element is relatively positioned - that relative position means that the offset can be way off the page. All my addition does is add if the dropdown is appearing by a click on the parent element then position it to the element that was just clicked on instead of some other element such as the posts container, or the content div, all of which would break the dropdown.

Of course, because of my "fix" then changing the margin on the edit popup css will of course move the popup about - which is really what you want if designing a theme - control over placement. 

In the current default theme that would mean changing the css for the margin. No doubt Destroy is worried about all the custom themes out there; indeed, so am I , and quite a lot of them are broken in this regard, ie the edit popup is all over the shop.

Sama34:
This should be fixed regardless of custom themes out there.
There are some already broken anyways as you said.

Eric-Jackson:
If the only issue is it being offset by the margin, just calculate that in.

    if (el_parent == popup_parent)
    {
    	offset = el.position();
    	offset.top += el.outerHeight(true) - parseInt($(el).css("marginBottom"), 10);
    }
    else {
    	offset.top += el.outerHeight(true);
    }

leefish:
I dont feel that adding a css margin bottom to the js is the right way to go; now we would have the issue of css in the javascript to fix an issue in the css file.....

Eric-Jackson:
I believe you're missing what the code does. There's no CSS here, it's only "disregarding" any bottom margin if there is one. Margin-top on the popup container will still adjust the position of the popup, meaning the code doesn't limit it.

Eric-Jackson:
This would be a CSS issue if popups were being rewritten and fixed entirely (a pure CSS solution), however without making that leap I think this solution is the correct one.

leefish:
Yea, but the "issue" is the postbit button has a margin applied; the popup is positioned with the unstyled element as base. I think it would be a oneliner in the css to position the margin of the edit pop-up container to the postbit edit button. 

I am really not keen on a solution where the js incorporates the css when the css can be adjusted to make it work, ie I would rather add to the global stylesheet than add css in the js.

Eric-Jackson:
The elements aren't always related in the HTML (Sometimes they're siblings, but sometimes the popup is appended to the body), so a CSS solution would probably be convoluted. That's why I mentioned that only if they were being rewritten correctly would CSS be the answer. Of course `margin-bottom: 0;` might be a fix in CSS, but you start to run into issues when you stack buttons and things like that.

And example fix that you're thinking might look something like this:

```css
.postbit_buttons > a + .popup_menu {
    margin-top: -5px;
}
```

Which only targets one element, so any new buttons theme developers add they will have the same issue (And again they can only correct it when the popup_menu is a sibling of the toggle).

leefish:
Yea, but Destroy did say in his testing that it was only the edit button that was the issue - that with the other button (the one with the thread options) the margin issue isnt there. Thus it is the margin css being inherited from the postbit button?

Eric-Jackson:
But an issue regardless, and only 2px when that's the margin applied to a popup trigger. It's not an issue with thread options because your code doesn't target it (popup is appended to the body). So any popups developers add will have the same issue because they'll likely be sibling elements.

leefish:
Exactly, they add a new element as part of the theme and so can target it in the theme files ie the css

Eric-Jackson:
But making them add redundant code is sloppy, especially considering that you're not allowing them to have a bottom margin on their popup triggers.

Destroy666x:
> But an issue regardless, and only 2px when that's the margin applied to a popup trigger. 

Not if there's an element with bigger margin as I mentioned. You can't expect everyone to have popup openers with 2px or less margin. And we shouldn't fix one issue by breaking another thing.

Everything that can be done with CSS can be automatically and reliably done with JS without expecting developers to add their own CSS each time. The fix here would be something simple like:
```js
			popup_menu.css('marginTop', el.css('marginTop'));
```
so always adding the same margin to the popup menu based on the popup caller. Done, no problems anymore (at least I can't find any).

Also, please use tabulation rather than spaces next to `offset` and either wrap the added code with comments (check for reference: https://github.com/mybb/mybb/issues/2199#issuecomment-141424586) or move it to a MyBB JS file by extending `popupMenu`.

JN-Jones:
Whats with this now? @Eric-Jackson apparently started to do some additional work but this PR is still open without any discussion for more than a year. Also with @leefish not active currently I'm not sure how we can handle this without keeping it open until 1.8 reaches EOL

Eric-Jackson:
My PR can't be accepted because it changes CSS. You'd probably just have to replicate what @leefish did with her PR.

Ben-Conway:
There's no point in having this open, there has been no movement on this for ages.

PR #2185, opened 2015-09-06
Fixes an issue I've introduced with #2026

JN-Jones:
Apparently neither "0" nor "false" get "my_date" to calculate the offset correctly so reverted it to an empty string and edited the doc block to reflect that behavior

As it's a leftover of #2026 I won't create a new issue for this.


euantorano:
Looks good to me. I'll confirm it works locally in a minute or two and then merge.

euantorano:
Seems to work, merging.

PR #2186, opened 2015-09-06
Initialise $bcc_form_val to an empty string.

euantorano:
If the `$bcc_recipients` array is empty, the `$bcc_form_val` should be an empty string.

Signed-off-by: Euan Torano <euantorano@gmail.com>


JN-Jones:
Haven't found any more issues. Merging

PR #2187, opened 2015-09-06
Use `validate_user()` when updating theme

euantorano:
Fixing the issue caused by not checking the user before updating. This should fix the immediate issue.

Signed-off-by: Euan Torano <euantorano@gmail.com>


JN-Jones:
Works

Issue #2184, opened 2015-09-05
Quick edit not working when thread prefix is required


Devilshakerz:
The `There was an error editing your reply: This forum requires a thread prefix to be set. Please select a valid prefix.` error shows up when trying to save post content in quick edit mode and thread prefixes are required in the forum the thread belongs to.
Most likely caused by https://github.com/mybb/mybb/pull/1984 (#1961)

PR #2188, opened 2015-09-06
Fixes an issue where the quickedit throws an issue if the forum requi…

JN-Jones:
…res prefixes

Since 1.8.6 the posthandler checks the prefix for firstposts (to avoid that you're able to remove the prefix later again), however as the quickedit doesn't set the prefix the post handler throws an error. To avoid rewriting the post handler (which doesn't know whether or not it's a quickedit or a full edit request) I've simply added the current prefix for the post handler

Caused by: https://github.com/mybb/mybb/pull/1984 #2184


euantorano:
Seems to work for me, merging.

PR #2192, opened 2015-09-10
Fixes #2191 Issue with user awaiting activation

### LINKED ISSUE #2191: Issue with user awaiting activation
When you create a new user from the ACP and select his primary group as **awaiting activation**, you can see in the forums, the red alert with the legend of **there is "x" amount of users waiting activation**, and after a link to activate them through the ACP.

If you, for any reasons, edit again the profile for this user and put his primary group as **registered**, the red alert keeps active, and if you go to the ACP through the link of the red alert, the ACP tells you **no users were find matching the specific search criteria**.

The only way to get ride of the red alert in the forums is editing one more time the user primary group to **awaiting activation** and after, activate the user directly from the link of the red alert.

I use 1.8.6 version.

Regards.

JN-Jones:
`language` was duplicated.
All other cases are already handled by the user datahandler

#2191


PaulBender:
Looks good to me

PR #2193, opened 2015-09-10
Fixes #1959 enhanced numeric option setting

### LINKED ISSUE #1959: Proposition : enhanced numeric option setting

JN-Jones:
#1959

This is just #1960 atm, however I'll fix the issues mentioned there when I've time (so hopefully tomorrow).


admin/modules/config/settings.php
```diff
@@ -1186,7 +1186,20 @@
 			}
 			else if($type[0] == "numeric")
 			{
-				$setting_code = $form->generate_numeric_field($element_name, $setting['value'], array('id' => $element_id));
+				$field_options = array('id' => $element_id);
+				if (count($type)>1)
```
Stefan-ST: Coding standards...

JN-Jones: As already said: it's just a copy of the old PR without any work and is still WIP...

euantorano:
What's the status on this, are there any other planned enhancements?

admin/modules/config/settings.php
```diff
@@ -1186,7 +1186,24 @@
...
+				{
+					for($i=1; $i < count($type); $i++)
+					{
+						$optionsexp = explode("=", $type[$i]);
+						$opt = array_map('trim', $optionsexp);
+						if(in_array($opt[0], array('min', 'max', 'step')))
+						{
+							if($opt[0] != 'step' || $opt[1] != 'any')
+							{
+								$opt[1] = (int)$opt[1];
```
Destroy666x: Floats are also numeric. https://github.com/mybb/mybb/pull/1915/files

Destroy666x:
@euantorano 

> Note that this isn't tested yet and likely an upgrade script is required for old installations.

install/resources/settings.xml
```diff
@@ -234,7 +234,8 @@ email=Sent via Email]]></optionscode>
 			<description><![CDATA[Here you can set the number of next and previous page links to show in the pagination for threads or forums with more than one page of results.]]></description>
 			<disporder>12</disporder>
 			<optionscode><![CDATA[numeric]]></optionscode>
-			<settingvalue><![CDATA[5]]></settingvalue>
+			<settingvalue><![CDATA[5
```
Destroy666x: Why did you change values rather than optioncodes?

euantorano:
Must have missed that, oops!

Cu8eR:
Guys, is there any way to merge this soon? I would like to fix my PR #2299

JN-Jones:
I still need to fix a few issues with this and I doubt this would merge without problems anyway. However as I've been quite busy for the last couple of months I need to update my working copy first and look at some other things I may have missed. I'll try to get it done rather sooner than later but can't promise as I've still a few other things to work on.

admin/modules/config/settings.php
JN-Jones: Done

install/resources/settings.xml
JN-Jones: Must've been pretty late when I've done it. Going to change it with the next commit

JN-Jones:
Ok, I've fixed the issues @Destroy666x mentioned before (as long as I haven't missed anything which I'll check after GitHub decides to show the diff again). However I still need to check whether the settings are synchronised or whether an upgrade script is needed and whether at least `min` and `max` need to be validated after saving the settings to avoid some script kid playing with the source code again.

JN-Jones:
Optionscodes seem to be updated automatically so no change there. Added a check for `min` and `max` to avoid issues with that. What we should discuss is whether we need a special check for `step`. According to W3C the step parameter not only specifies the step itself but also the values that can be entered (eg step=1 only allows integers, step=2 only 2,4,6 and so on) however not all browser mirror that. As that's the case I'd say we can ignore it too (which means this PR is ready for testing). @Destroy666x @euantorano @Stefan-ST

JN-Jones:
Anyone?

JN-Jones:
Can we get this merged so @Cu8eR can work on #2193 so we can fix both in 1.8.9? @mybb/sqa

PaulBender:
Works for me

PR #2195, opened 2015-09-13
Fixes #1567 It's impossible to send a PM to a user with a short username

### LINKED ISSUE #1567: It's impossible to send a PM to a user with a short username
If a username is shorter than 3 characters it impossible to send him a PM because Select2 doesn't accept anything that short.

PaulBender:
Fixes #1567

- Shortens minimumInputLength to ~~1~~ 2
- Removed maximumSelectionSize on all select2 that aren't multiples
- Tied maximumSelectionSize to maxpmrecipients for PMs
- Centralized select2 placeholder language strings to the global language file (pointless to keep the same string scattered across 7 language files)

### DIFF
--- a/admin/modules/forum/management.php
+++ b/admin/modules/forum/management.php
@@ -2594,7 +2594,6 @@ document.write('".str_replace("/", "\/", $field_select)."');
 		$("#username").select2({
 			placeholder: "'.$lang->search_for_a_user.'",
 			minimumInputLength: 3,
-			maximumSelectionSize: 3,
 			multiple: false,
 			ajax: { // instead of writing the function to execute the request we use Select2\'s convenient helper
 				url: "../xmlhttp.php?action=get_users",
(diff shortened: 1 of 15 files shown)


Stefan-ST:
I don't think minimumInputLength shouldn't be 1. It makes not much sense to display suggestions after one or two entered characters because there are probably too many matches (and it only displays 15?).

PaulBender:
Should minimumInputLength be 2 then?

Cu8eR:
I suggest 2

PaulBender:
Okay, I've increased it to 2.

JN-Jones:
Can anybody check this? Would be good to have this in 1.8.6

euantorano:
I'll try and test this tomorrow night if I get time.

PaulBender:
We should get this merged.

euantorano:
Tried and tested, seems to work fine for me. Merging.

PR #2200, opened 2015-09-17
Fixes #2103 Undefined variable in warnings datahandlers

### LINKED ISSUE #2103: Undefined variable in warnings datahandlers
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/warnings.php#L491 "&#36;groupscache" is never loaded/globalized in that file.

Original thread: [Undefined variable in warnings datahandlers](http://community.mybb.com/thread-171789.html)

PaulBender:
Fixes #2103

### DIFF
--- a/inc/datahandlers/warnings.php
+++ b/inc/datahandlers/warnings.php
@@ -362,7 +362,7 @@ class WarningsHandler extends DataHandler
 	*/
 	function update_user($method='insert')
 	{
-		global $db, $mybb, $lang;
+		global $db, $mybb, $lang, $groupscache;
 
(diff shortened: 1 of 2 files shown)


Destroy666x:
I don't like this solution - it's incomplete. The variable may be still undefined if the function is called by a plugin, for instance. A code like this could be added to the function:
```php
if(!is_array($groupscache))
{
	$groupscache = $cache->read("usergroups");
}
```

Same thing should be done in `warnings.php` btw: https://github.com/mybb/mybb/blob/feature/warnings.php#L370 It's not defined there either.

PaulBender:
Added

PR #2201, opened 2015-09-17
Fixes #2130 can not set permissions on another super admin

### LINKED ISSUE #2130: can not set permissions on another super admin
I was super admin on my website. when trying to change the permissions of other super admin, the following warning appears as shown in the picture. 

[attachment=32831]

[attachment=32833]

how to fix this problem in MyBB 1.8?  :S 

because I also have added a user ID into config.php  :-/

Original thread: [can not set permissions on another super admin](http://community.mybb.com/thread-161704.html)

PaulBender:
Fixes #2130

### DIFF
--- a/inc/languages/english/admin/user_admin_permissions.lang.php
+++ b/inc/languages/english/admin/user_admin_permissions.lang.php
@@ -35,7 +35,7 @@ $l['edit_group'] = "Edit Group";
 $l['default'] = "Default";
 $l['group'] = "Group";
 
-$l['error_super_admin'] = 'Sorry, but you cannot perform this action on the specified user as they are a super administrator.<br /><br />To be able to perform this action, you need to add your user ID to the list of super administrators in inc/config.php.';
+$l['error_super_admin'] = 'Sorry, but you cannot perform this action on the specified user as they are a super administrator.';
 $l['error_delete_no_uid'] = 'You did not enter a admin user/usergroup permission id';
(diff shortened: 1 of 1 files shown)


Destroy666x:
The functionality is still inconsistent across pages though:
- the users module let's you update another super admin if you're a super admin https://github.com/mybb/mybb/blob/feature/admin/modules/user/users.php#L440
https://github.com/mybb/mybb/blob/feature/admin/modules/user/users.php#L1708
https://github.com/mybb/mybb/blob/feature/admin/modules/user/users.php#L1910
https://github.com/mybb/mybb/blob/feature/admin/modules/user/users.php#L2771
- the admin permissions module doesn't let you update super admins at all, even if you want to edit your own permissions...
https://github.com/mybb/mybb/blob/feature/admin/modules/user/admin_permissions.php#L113
https://github.com/mybb/mybb/blob/feature/admin/modules/user/admin_permissions.php#L46

It doesn't make much sense. You can for example merge a super admin with another account which is not super admin and then edit  permissions.

JN-Jones:
I think we should discuss a proper solution on the issue ticket before we do any code related changes and introducing more inconsistent behavior.

Ben-Conway:
I don't see why you would want to set permissions on another super administrator, if you can't trust them to have full access to the acp they shouldn't be a super admin.

PaulBender:
Either way, this should be merged as it corrects the error message.

Ben-Conway:
Good work @PaulBender, just noticed this fixed the error message.

PR #2204, opened 2015-09-18
Fixes #2203 Click and Hold erases thread prefix in 1.8.6

### LINKED ISSUE #2203: Click and Hold erases thread prefix in 1.8.6
After upgrading mybb 1.8.5 to 1.8.6

Steps to reproduce :
 
1. Add a prefix to a thread title
2. Modify this title with click and hold functionnality
3. Refresh the page

The prefix disappears.

PaulBender:
Fixes #2203

### DIFF
--- a/xmlhttp.php
+++ b/xmlhttp.php
@@ -373,6 +373,7 @@ else if($mybb->input['action'] == "edit_subject" && $mybb->request_method == "po
 		$updatepost = array(
 			"pid" => $post['pid'],
 			"tid" => $thread['tid'],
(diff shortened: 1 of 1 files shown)


JN-Jones:
Looks good. Though as @Destroy666x mentioned there may be other places and plugins which cause the same bug. I haven't checked any code so far but that's something which needs to be checked.

Cu8eR:
+1, works fine for me

Stefan-ST:
I couldn't find other instances so merging this.

PR #2205, opened 2015-09-19
Fixes #2090 $post['button_find'] doesn't check for permissions.

### LINKED ISSUE #2090: $post['button_find'] doesn't check for permissions.
The build of the search button in post bit does not check for user permissions on search.

https://github.com/mybb/mybb/blob/feature/inc/functions_post.php#L324

PaulBender:
Fixes #2090

Also fixes the find post/thread links in profiles.

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -321,6 +321,7 @@ function build_postbit($post, $post_type=0)
 			eval("\$post['useravatar'] = \"".$templates->get("postbit_avatar")."\";");
 		}
 
+		$post['button_find'] = '';
 		if($mybb->usergroup['cansearch'] == 1)
 		{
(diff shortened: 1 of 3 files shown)


inc/functions_post.php
```diff
@@ -321,7 +321,10 @@ function build_postbit($post, $post_type=0)
 			eval("\$post['useravatar'] = \"".$templates->get("postbit_avatar")."\";");
 		}
 
-		eval("\$post['button_find'] = \"".$templates->get("postbit_find")."\";");
+		if($mybb->usergroup['cansearch'] == 1)
+		{
+			eval("\$post['button_find'] = \"".$templates->get("postbit_find")."\";");
```
Stefan-ST: $post['button_find'] is now not defined when the user doesn't have the permission.

PaulBender: Fixed

PR #2208, opened 2015-09-21
Fixes #2206 usercp line 3222 error

### LINKED ISSUE #2206: usercp line 3222 error
Hello I was testing some things today with my brother for our forum and we found a problem when going to the usercp.php?action=usergroups to join a new usergroup, right after you send your request to join, while you're being redirected, theres an error message at the top of the screen like this

http://i.imgur.com/eruMndC.jpg

I opened usercp.php from the root folder and search the line 3222 and took this secreenshot, hope it helps

http://i.imgur.com/ECb79fL.jpg

Heres my forum: akibadesu.com

And a test account.. 
Username: Test Account 
Password: test123

I hope someone can help me fix this, thanks!

Original thread: [usercp line 3222 error](http://community.mybb.com/thread-175496.html)

Devilshakerz:


usercp.php
```diff
@@ -3221,15 +3221,18 @@
 
 			$db->insert_query("joinrequests", $joinrequest);
 
-			foreach($groupleaders[$usergroup['gid']] as $leader)
+			if(in_array($usergroup['gid'], $groupleaders))
```
JN-Jones: The array has this structure: `$groupleaders[$gid][$uid] = $user;` So wouldn't `in_array($gid, $groupleaders)` also check the `$uid` part? Probably better (and definitly more intuitive) would be `is_array($groupleaders[$gid])`.

Devilshakerz: At this point there's no need to verify that on a deeper level - if `$gid` is a valid index, it's an array of `$uid`s: https://github.com/Devilshakerz/mybb/blob/fixes-2206/usercp.php#L3169-L3181

JN-Jones: A simple script on php 5.6:

```php
<?php

$array = [1 => [2 => ['uid' => 2, 'username' => 'test']]];

echo "<pre>";
var_dump(in_array(1, $array)); // False
var_dump(in_array(2, $array)); // False
var_dump(is_array($array[1])); // True
var_dump(is_array($array[2])); // False (and a notice)
echo "</pre>";
```

As you can see `in_array` shows the wrong result.

Devilshakerz: Never mind that, I was aiming for `array_key_exists()` instead; if they key exists, it's an array as defined earlier.

PR #2215, opened 2015-09-22
Fixes #2214 Wrong error message

### LINKED ISSUE #2214: Wrong error message
I just upgraded to 1.8.5 successfully, and I like it. 

However, I noticed that one of my pet peeves hasn't yet gotten fixed. That is, when I don't have permission to view a forum, the error message says, 

"Sorry, but there are currently no threads in this forum with the specified date and time limiting options."

Really it should say that I don't have permission to see it. This is an important difference in my setup, because I can't attract new higher-level members if they think there's nothing of value in that forum. 

How do I change this error message? 

Thank you for all your excellent help...it has been superb thus far!

Original thread: [Wrong error message](http://community.mybb.com/thread-171309.html)

PaulBender:
Fixes #2214

### DIFF
--- a/archive/index.php
+++ b/archive/index.php
@@ -256,10 +256,19 @@ switch($action)
 		build_forum_breadcrumb($forum['fid'], 1);
 
 		// No threads and not a category? Error!
-		if(($threadcount < 1 || $forumpermissions['canviewthreads'] != 1) && $forum['type'] != 'c')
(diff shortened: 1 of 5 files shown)


Destroy666x:
Could have just used overwritten the string instead of creating a new template, which may be incompatible with custom themes. No loss by doing that - I don't think anyone would want to style both errors differently.

Issue #2221, opened 2015-10-06
Report Bug


euantorano:
I've tested this here and at my own forum and it happens on both.

Currently, the input box to enter a custom reason only appears when the other reason is selected from the drop down menu on the report button. If you open the report box on a thread and select a reason (not other) and then click off the box by clicking anywhere on the thread so it disappears and then click the report button again the input box to enter a custom reason for other will now show no matter which reason you select.

Original thread: [Report Bug](http://community.mybb.com/thread-183013.html)

ATofighi:
Duplicated to https://github.com/mybb/mybb/issues/2199

Destroy666x:
This is the same issue - old modal not closing properly, so marking this as duplicate.

PR #2222, opened 2015-10-08
Fixes #2199 - add keepelement to jquery-modal

### LINKED ISSUE #2199: Reputation confirmation modal sometimes remains on the screen in 1.8.6
If two (or more) reputations are given to posts on the same page (without navigating away or refreshing the page), the second reputation confirmation modal does not close and stays on the screen.
I have give more details on the forums: http://community.mybb.com/thread-178564.html

This is caused because in 1.8.6 the following lines are deleted from modal.close function in jquery.plugins.js:

@@ -70,9 +71,6 @@
     close: function() {
       this.unblock();
       this.hide();
-	  // Deletes the element (multi-modal feature: e.g. when you click on multiple report buttons, you will want to see different content for each)
-	  if (!this.options.keepelement)
-		this.$elm.remove();
       $(document).off('keydown.modal');
     },

ATofighi:
#2199 #2221

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -390,7 +390,7 @@ lang.saved = \"{$lang->saved}\";
 <meta name="copyright" content="Copyright {$copy_year} MyBB Group." />
 <link rel="stylesheet" href="./styles/{$cp_style}/login.css" type="text/css" />
 <script type="text/javascript" src="../jscripts/jquery.js"></script>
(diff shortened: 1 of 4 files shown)


Destroy666x:
It works fine, but as always:
- `general.js` version should be bumped in template(s) and several files which load it
- template version should be bumped

ATofighi:
fixed

Destroy666x:
There are 2 more in `admin/inc/class_page.php`

ATofighi:
fixed

PR #2227, opened 2015-10-11
Fixes #2113

### LINKED ISSUE #2113: Attachment buttons doubled lined
I find this kind of annoying. The buttons should be displayed on one single line.

Original thread: [Attachment buttons doubled lined](http://community.mybb.com/thread-167619.html)

SvePu:
I earlier have tested it in Opera and now in FF - In Opera 63% was ok but in FF it doesnt work.

The max. witdh for this fix must be 60%

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -9270,8 +9270,8 @@ if(use_xmlhttprequest == "1")
 </tr>]]></template>
 		<template name="post_attachments_new" version="1800"><![CDATA[<tr>
 <td class="trow1" width="1"><img src="{$theme['imgdir']}/paperclip.png" alt="" /></td>
-<td class="trow1" style="white-space: nowrap"><strong>{$lang->new_attachment}</strong> <input type="file" name="attachment" size="30" class="fileupload" /></td><td class="trow1" align="center">{$attach_update_options} {$attach_add_options}
-</td>
+<td class="trow1" style="white-space: nowrap"><strong>{$lang->new_attachment}</strong> <input type="file" name="attachment" size="30" class="fileupload" /></td>
+<td class="trow1" style="white-space: nowrap" align="center">{$attach_update_options} {$attach_add_options}</td>
 </tr>]]></template>
 		<template name="post_attachments_update" version="1604"><![CDATA[<input type="submit" class="button" name="updateattachment" value="{$lang->update_attachment}" tabindex="12" />]]></template>
 		<template name="post_captcha" version="1804"><![CDATA[<tr id="captcha_trow">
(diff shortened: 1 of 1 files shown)


Destroy666x:
> The best solution is white-space: nowrap; for that specific column/cell as I mentioned in the thread - it will ensure it doesn't break in all browsers under all resolutions. The width can also be changed, but is not fully trustable.

Also, you can update the PR rather than creating a new one.

SvePu:
Your're right, white-space: nowrap; could be the best solution.

But there is no space to show both buttons side by side.
The second row define the width of both rows.

Destroy666x:
Well, you can surely leave the `width` as I mentioned, it makes it slightly better, but please add that CSS too to make sure it'll work.

SvePu:
The relevant width part is already in default theme xml
``` php
<template name="post_attachments_attachment" version="1604"><![CDATA[<tr>
<td class="trow2" width="1" align="center">{$attachment['icon']}</td>
<td class="trow2" width="75%" style="white-space: nowrap">{$attachment['filename']} ({$attachment['size']})</td>
<td class="trow2" style="white-space: nowrap; text-align: center;">{$attach_mod_options} {$attach_rem_options} {$postinsert}</td>
</tr>]]></template>
```
I only changed it to:
``` php
<template name="post_attachments_attachment" version="1604"><![CDATA[<tr>
<td class="trow2" width="1" align="center">{$attachment['icon']}</td>
<td class="trow2" width="60%" style="white-space: nowrap">{$attachment['filename']} ({$attachment['size']})</td>
<td class="trow2" style="white-space: nowrap; text-align: center;">{$attach_mod_options} {$attach_rem_options} {$postinsert}</td>
</tr>]]></template>
```
This template is only active when attachments are already uploaded.

Destroy666x:
That's not the template and cell we're speaking about though. It's the 3rd cell of this template (new attchament row): https://github.com/mybb/mybb/blob/feature/install/resources/mybb_theme.xml#L9273 which doesn't have that CSS.

The `<td>` could be separated into a new line too, BTW.

SvePu:
I know, but the second column in template "post_attachments_attachment" also defines width of the second column of template "post_attachments_new".

That's why I've modified width in that "wrong" template. ;)

Destroy666x:
But it doesn't define `white-space: nowrap` as I said above...

SvePu:
Updated it.... also set <td> on a new line :)

Destroy666x:
Thanks. Also, last thing I forgot to mention, `version="1604"` needs to be bumped to `version="1807"` so that templates will be updated correctly when upgrading.

SvePu:
Done :)

Destroy666x:
Thank you.

SvePu:
...my pleasure ;)

PR #2233, opened 2015-10-17
Fix #2232 URLs with & and () and without protocol not parsed

### LINKED ISSUE #2232: [Regex] Why "&" disallowed in url-BB-Code
Hello,
I have found the following line in class_parser.php:
#\[url=([^\r\n\"&lt;&amp;\(\)]+?)\](.+?)\[/url\]#si

Now I have the question why it is disallowed here? I know, correct urls should match the previous regex (#\[url=([a-z]+?://)([^\r\n\"&lt;]+?)\](.+?)\[/url\]#si), but still &amp; is used very often in urls and if you forget http(s)://, your URL won't be matched.

My Questiona are:

What is this Regex for if all valid urls should match the one above?

What is the security-Problem with matching &amp; in this regex?

Thanks in advance ...
Thomas

Original thread: [[Regex] Why "&" disallowed in url-BB-Code](http://community.mybb.com/thread-175426.html)

Destroy666x:
The mentioned characters were disallowed for no apparaent reason.

https://github.com/mybb/mybb/issues/2232

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -306,7 +306,7 @@ class postParser
 			$callback_mycode['url_complex']['regex'] = "#\[url=([a-z]+?://)([^\r\n\"<]+?)\](.+?)\[/url\]#si";
 			$callback_mycode['url_complex']['replacement'] = array($this, 'mycode_parse_url_callback1');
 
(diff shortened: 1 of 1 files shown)


Stefan-ST:
I don't see why not...

PR #2238, opened 2015-10-18
Fixes #2235 Smilies "Show on clickable list?" doesn't work

### LINKED ISSUE #2235: Smilies "Show on clickable list?" doesn't work
After the upgrade to 1.8.6 smiles are displayed regardless whether "Show on clickable list?" is checked or not. This is a regression of fixing #1955.

Stefan-ST:
Fixes #2235


Destroy666x:
That's why global variables are lovely..

PR #2241, opened 2015-10-18
Fixes #2196 Subjects escaped twice in newreply.php

### LINKED ISSUE #2196: Subjects escaped twice in newreply.php
```php
$thread['subject'] = htmlspecialchars_uni($thread['subject']);
//...
			$subject = $thread['subject'];
//...
	$subject = htmlspecialchars_uni($parser->parse_badwords($subject));
```
This doesn't happen when you quote a specific post.

Stefan-ST:
Fixes #2196

### DIFF
--- a/newreply.php
+++ b/newreply.php
@@ -77,7 +77,7 @@ if(!$forum)
 // Make navigation
 build_forum_breadcrumb($fid);
 $thread_subject = $thread['subject'];
-$thread['subject'] = htmlspecialchars_uni($thread['subject']);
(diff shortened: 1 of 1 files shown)


Destroy666x:
The order is still slightly off, bad words should be parsed in the breadcrumb too, like in `showthread.php`: https://github.com/mybb/mybb/blob/feature/showthread.php#L82-L83

Stefan-ST:
@Destroy666x Good catch, fixed.

PR #2242, opened 2015-10-18
Fixes #1395 SQL error in Admin Permissions -> User Permissions

### LINKED ISSUE #1395: SQL error in Admin Permissions -> User Permissions
If no group has ticked "Yes, users of this group can access the admin CP", there will be an error because IN is empty:
```
SELECT u.uid, u.username, u.lastactive, u.usergroup, u.additionalgroups, a.permissions FROM mybb_users u LEFT JOIN mybb_adminoptions a ON (a.uid=u.uid) WHERE u.usergroup IN () ORDER BY u.username ASC
```

Stefan-ST:
Fixes #1395 

Apparently Github doesn't link indenting large parts of code...

### DIFF
--- a/admin/modules/user/admin_permissions.php
+++ b/admin/modules/user/admin_permissions.php
@@ -398,118 +398,121 @@ if(!$mybb->input['action'])
 		$usergroups[$usergroup['gid']] = $usergroup;
 	}
 
-	// Get users whose primary or secondary usergroup has ACP access
-	$comma = $primary_group_list = $secondary_group_list = '';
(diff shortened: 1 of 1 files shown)


JN-Jones:
You only need to disable whitespaces as diff via `?w=0`: https://github.com/mybb/mybb/pull/2242/files?w=0

Stefan-ST:
I know, I just wanted to point out that the are actually only a few changes. ;)

PaulBender:
Works for me

PR #2243, opened 2015-10-18
Fixes #1817 managegroups.php pagination

### LINKED ISSUE #1817: managegroups.php pagination
Another weird code from that file: https://github.com/mybb/mybb/blob/feature/managegroup.php#L344-355 It tries to create pagination, but neither the queries above it support it with `LIMIT`, nor it's output in any template..

Stefan-ST:
Fixes #1817


Destroy666x:
IMO the PR should rather fix the pagination than removing it. If a group has over 1000 members, it won't really be optimal this way.

Stefan-ST:
You are right. Done...

PaulBender:
Works fine for me

PR #2245, opened 2015-10-18
Fixes #1974 Weird default user pruning task value

### LINKED ISSUE #1974: Weird default user pruning task value
It runs every minute at 2 o'clock for some reason: http://community.mybb.com/thread-170168-post-1155669.html#pid1155669

EDIT: also times should be removed from task description strings: http://community.mybb.com/thread-170168-post-1155678.html#pid1155678

Stefan-ST:
Fixes #1974

### DIFF
--- a/install/resources/tasks.xml
+++ b/install/resources/tasks.xml
@@ -112,7 +112,7 @@
 		<title>User Pruning</title>
 		<description><![CDATA[Automatically prunes users based on criteria set in the settings every day.]]></description>
 		<file>userpruning</file>
-		<minute>*</minute>
+		<minute>0</minute>
 		<hour>2</hour>
 		<day>*</day>
 		<weekday>*</weekday>


Destroy666x:
> also times should be removed from task description strings: http://community.mybb.com/thread-170168-post-1155678.html#pid1155678

Stefan-ST:
I think the description is actually useful if you have changed the values and don't know what the default was.

Destroy666x:
But then it should say something like `every day by default` rather than just `every day`, otherwise it's misleading. And it should be consistent across all tasks.

JN-Jones:
I agree that those descriptions should be changed, however I'd split this issue if we need to discuss that. The task value is imho pretty important and should be in 1.8.7.
Also I'd change the value on upgrade **if** the minute is still set to `*`. I honestly doubt anyone would want to change the task to run every minute.

euantorano:
I agree with splitting the description discussion. Otherwise this patch fixes the issue, so I'm merging it. Adding an upgrade script would be nice, I'll do that tomorrow.

PR #2246, opened 2015-10-18
Fixes #1980 Nonexistent usernames cause invalid SQL on Search

### LINKED ISSUE #1980: Nonexistent usernames cause invalid SQL on Search
Search will produce a SQL error if the _match exact username_ option has been selected and no matching user has been found. The Select2 selector has to be bypassed first as it does not allow nonexistent usernames.

A result check needs to be added:
https://github.com/mybb/mybb/blob/feature/inc/functions_search.php#L1094-L1095
https://github.com/mybb/mybb/blob/feature/inc/functions_search.php#L1508-L1509

Stefan-ST:
Fixes #1980

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -1092,7 +1092,10 @@ function perform_search_mysql($search)
 		if($search['matchusername'])
 		{
 			$user = get_user_by_username($search['author']);
-			$userids[] = $user['uid'];
+			if($user)
+			{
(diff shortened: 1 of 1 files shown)


PaulBender:
Doesn't work for me, still produces the same SQL error as before.

Stefan-ST:
Thanks, it should work now.

PR #2247, opened 2015-10-18
Fixes #1981 $this->delete_uids can be empty and throw SQL error

### LINKED ISSUE #1981: $this->delete_uids can be empty and throw SQL error
As the title says, in some cases it can be emptied: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/user.php#L1533 and there is no check for that, which results in this: http://community.mybb.com/thread-170462.html

Stefan-ST:
Fixes #1981

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -1517,7 +1517,17 @@ class UserDataHandler extends DataHandler
 		$plugins->run_hooks('datahandler_user_delete_start', $this);
 
 		$this->delete_uids = implode(',', $this->delete_uids);
-		
+
+		if(empty($this->delete_uids))
+		{
+			$this->deleted_users = 0;
+			$this->return_values = array(
+				"deleted_users" => $this->deleted_users
+			);
+
+			return $this->return_values;
+		}
+
 		$this->delete_content();
 
 		// Delete the user
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -1621,6 +1621,11 @@ class UserDataHandler extends DataHandler
 
 		$plugins->run_hooks('datahandler_user_delete_content', $this);
 
+		if(empty($this->delete_uids))
+		{
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


Destroy666x:
What about `delete_posts` and `clear_profile`? Same issue.

Stefan-ST:
@Destroy666x Fixed, thanks.

Destroy666x:
One more thing - IMO, this should be above hooks, because there are plugins which are using `delete_uids`. Adding it below hooks has no pros - devs can always call the function with their UID list.

Stefan-ST:
I've added the code below to allow modifications of delete_uids. I'm not sure whether there is a use case but I didn't want to make that impossible. At least delete_content() has another hook.

Sama34:
Seems correct to me.

Destroy666x:
Well, if you're ok with leaving some plugins that use `delete_uids` broken, feel free to merge, the rest is ok.

Sama34:
It does not makes much sense to add/remove UIDs there yes but it is possible and it causes no issues.

Destroy666x:
I'm not sure what you're trying to say, but adding/removing UIDs was never problematic.. The problem here is that plugins that relied on the previous method (and called their own queries) will stay broken after merging this PR, so we'll have to rely on them fixing it in hooks by checking emptiness there as well. 

If we switch the order as I mentioned, removing UIDs will be still doable, e.g. by checking emptiness again in the hook and assigning 0 to the list if it's empty. It won't be optimal - useless queries for 0 will be called, but I prefer a non-optimal solution over a solution which possibly breaks something.

And adding UIDs every time any delete function is called doesn't make any sense because they won't exist anymore after the 1st call anyways (since UID is auto-increment), so you'd force the db to pointlessly look for them every other time delete functions are called. Calling delete functions once for those UIDs would make much more sense and is the way to go.

However, I'll just merge the PR with hope that plugins have already been fixed in the meantime, because noone else is going to share their opinion anytime soon anyways.

PR #2249, opened 2015-10-18
Fixes #2237 Round corners in child tables inconsistent

### LINKED ISSUE #2237: Round corners in child tables inconsistent
nopermission_css3_ori
As you can see the corners of the parent tables are correct. But both lower corners of thead in the child tables are rounded although they shouldn't be rounded.

SvePu:
* Defined the desired CSS rule only for first child of tborder class.
* Changed the version number of css3.css to 1807

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -2172,7 +2172,7 @@ ul.thread_tools li.poll {
 .thread_status.newlockfolder {
 	background-position: 0 -320px;
 }]]></stylesheet>
-	<stylesheet name="css3.css" version="1800" disporder="7"><![CDATA[tr td.trow1:first-child,
+	<stylesheet name="css3.css" version="1807" disporder="7"><![CDATA[tr td.trow1:first-child,
 tr td.trow2:first-child,
 tr td.trow_shaded:first-child {
 	border-left: 0;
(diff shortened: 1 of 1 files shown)


Cu8eR:
#2237

Destroy666x:
It works in this case, but it's unfortunately not a proper fix. I'm sure this CSS rule was added for a reason and changing it to affect only tables which are first children may break something else.

The main issue here is something else - `.tborder tbody tr:last-child td:first-child` affects all descendant, including all first/last cells from table(s) in the last row, while it should affect only the closest children. The `>` selector should be used in between elements so that only the closest ones will be affected: `.tborder > tbody > tr:last-child > td:first-child`

SvePu:
You're right @Destroy666x 

Now I've tested it with different trow colors for a better view.

unbenannt

The new solution is:
```css
.tborder tbody tr:last-child > td {
	border-bottom: 0;
}

.tborder tbody tr:last-child > td:first-child {
	-moz-border-radius-bottomleft: 6px;
	-webkit-border-bottom-left-radius: 6px;
	border-bottom-left-radius: 6px;
}

.tborder tbody tr:last-child > td:last-child {
	-moz-border-radius-bottomright: 6px;
	-webkit-border-bottom-right-radius: 6px;
	border-bottom-right-radius: 6px;
}
```

Could you test it please, if it's OK than I'll rebuild the PR

Destroy666x:
Should be ok as well.

SvePu:
Modified PR in afa0aaf8dc8eb57f9004e20150c1f919f3078902

Destroy666x:
Ok, thanks.

PR #2261, opened 2015-10-20
Fixes #2260 Stylesheets should be loaded before scripts

### LINKED ISSUE #2260: Stylesheets should be loaded before scripts
For a better SEO performance, stylesheets should be loaded before scripts.

SvePu:


Destroy666x:
I'm not sure whether we should include this. It is said to decrease loading time, but it's a myth in most browsers nowadays: http://stackoverflow.com/questions/9271276/is-the-recommendation-to-include-css-before-javascript-invalid?answertab=votes#tab-top

But I also have nothing against, except that it may "break" some plugins which append JS files that require jQuery before/after `{$stylesheets}` variable. There definitely are few of them (because it's easy to append it there) but not sure how many.

What do others think @Stefan-ST @euantorano @ATofighi  @JN-Jones?

SvePu:
A lot of performance tools say that it is better to load stylesheets first.
Google itself has two opinions:
https://www.youtube.com/watch?v=AWF7ZzPyagE
* external CSS before external JS
* internal JS before external CSS  

But you're right - in cases of plugin it would be better to keep it as it is.

Cu8eR:
I agree with @SvePu - many services recommend this approach

JN-Jones:
I'd first check how most plugins handle it. If it breaks a lot of plugins I wouldn't change it, otherwise I've nothing against it.

Destroy666x:
> A lot of performance tools say that it is better to load stylesheets first.

Those tools are often either outdated or they purposely provide worthless information to persuade people to try their paid features. E.g. they test HTML validation, even though Google clearly stated they don't care about it: http://webmasters.stackexchange.com/a/56488

Stefan-ST:
I'm against changing this. The cons (breaking plugins) overweight the pros (a mostly unnoticeable performance gain).

Devilshakerz:
While it would be a logical change, I agree that we shouldn't lean towards small modifications that might possibly introduce compatibility issues at this point (the focus will begin to shift from 1.8 to 2.0 not long from now).

SvePu:
Ok guys, I understand the concerns. Let's keep it, as it is. Please reject the PR.

PR #2267, opened 2015-10-29
Fixes #2266 Editing user in ACP resets select profile fields with amp…

### LINKED ISSUE #2266: Editing user in ACP resets select profile fields with ampersands
Set up a profile field as select box with an option containing an ampersand. The user needs to select that option. Now when you edit the user in the ACP that option isn't preselected like it is used to be. Without manually fixing the profile field it gets reset to the first option.

Stefan-ST:
…ersands

Fixes #2266


Destroy666x:
It works, but checkobex are broken as well and can't be even saved.

Stefan-ST:
@Destroy666x Good catch. Fixed.

PaulBender:
Works for me

PR #2268, opened 2015-10-29
Fixes #2257 Default search doesn't find word with apostrophes

### LINKED ISSUE #2257: Default search doesn't find word with apostrophes
Apostrophes are replaced with spaces in inc/functions_search.php. Afterwards the search string simply doesn't match words with apostrophes anymore. Fulltext search is not affected.
```php
$keywords = preg_replace("#([\[\]\|\.\,:'])#s", " ", $keywords);
```

Stefan-ST:
Fixes #2257

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -229,11 +229,12 @@ function get_password_protected_forums($fids=array())
  */
 function clean_keywords($keywords)
 {
+	global $db;
+
 	$keywords = my_strtolower($keywords);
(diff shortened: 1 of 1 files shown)


Destroy666x:
It fixes that issue but introduces a much bigger one - `"something"` searches completely stop to work for an obvious reason.

Stefan-ST:
@Destroy666x Thanks, I didn't think about that.

euantorano:
Seems to work for me.

PR #2272, opened 2015-10-29
Fixes #2074 Rating an unapproved post

### LINKED ISSUE #2074: Rating an unapproved post
I just noticed we can give reputation to an user from an unapproved post, and probably an soft deleted one as well.

IMO this shouldn't be the case. Even if I'm allowed to see the post.

Original thread: [Rating an unapproved post](http://community.mybb.com/thread-167600.html)

Stefan-ST:
Fixes #2074

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -333,7 +333,7 @@ function build_postbit($post, $post_type=0)
 		}
 
 		$post['button_rep'] = '';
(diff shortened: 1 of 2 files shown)


Destroy666x:
This fixes the button, but not the actual reputation code. You can still rate unapproved posts if you're a mod.
https://github.com/mybb/mybb/blob/feature/reputation.php#L164

Stefan-ST:
Fixed.

PR #2280, opened 2015-11-24
Fixes #1578 - PM subscriptions wrong username

### LINKED ISSUE #1578: PM subscriptions wrong username
When multiple PM subscriptions are sent, all users receive a PM addressed to the first user.
The reason is probably that the language string is overridden in sent_pm() in inc/functions.php:
```php
$lang->{$pm[$key][0]} = str_replace('{'.$i.'}', $pm[$key][$i], $lang->{$pm[$key][0]});
```

ATofighi:
#1578


PaulBender:
Works for me.

PR #2286, opened 2015-12-17
Fixes #2285 Ambiguous indirect variable access breaks PHP 7 compatibi…

### LINKED ISSUE #2285: Ambiguous indirect variable access breaks PHP 7 compatibility
https://secure.php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.indirect:
> Indirect access to variables, properties, and methods will now be evaluated strictly in left-to-right order, as opposed to the previous mix of special cases.

This makes the [`MyLanguage::parse_replace` method](https://github.com/mybb/mybb/blob/feature/inc/class_language.php#L240) parsing ambiguous because the code is interpreted differently on PHP 5 and 7:

`$this->$matches[1]`
PHP 5 interpretation (intended):
`$this->{$matches[1]}`
PHP 7 interpretation:
`($this->$matches)[1]`

[admin/modules/style/templates.php](https://github.com/mybb/mybb/blob/feature/admin/modules/style/templates.php#L1717) is the only location where said method is used, and this error causes the template group names not to display:
http://community.mybb.com/thread-187206.html

Devilshakerz:
…lity

### DIFF
--- a/inc/class_language.php
+++ b/inc/class_language.php
@@ -237,6 +237,6 @@ class MyLanguage
 	 */
 	function parse_replace($matches)
 	{
-		return $this->$matches[1];
+		return $this->{$matches[1]};
 	}
 }


Destroy666x:
https://github.com/mybb/mybb/issues/2285

PR #2295, opened 2015-12-29
Fixes #2069 - Search container width

### LINKED ISSUE #2069: Search by Username container width
Goto search.php, under search by username first enter any username which has less characters... now click on the arrow again to search another user, the drop down and the input box has width equal to the already selected username, making it difficult to pick a new user from the drop down....

its because of the foll. css rule in select2.css... imho, it would be better to define a fixed width, say 100px rather than auto...

Code:
.select2-container .select2-choice &gt; .select2-chosen {width: auto;}

Original thread: [Search by Username container width](http://community.mybb.com/thread-167249.html)

Cu8eR:
Fixes #2069 - Search container width

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -4308,7 +4308,7 @@ $(".forumjump").change(function() {
   <td class="trow1"><input type="text" id="username" class="textbox" name="username" size="40" maxlength="{$mybb->settings['maxnamelength']}" style="width: 210px" /></td>
 </tr>
 </table>
-<link rel="stylesheet" href="{$mybb->asset_url}/jscripts/select2/select2.css?ver=1806">
+<link rel="stylesheet" href="{$mybb->asset_url}/jscripts/select2/select2.css?ver=1807">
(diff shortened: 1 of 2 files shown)


Cu8eR:
@mmadhankumar

Destroy666x:
I haven't tested it yet, but it's definitely not finished due to no `ver` change.

Cu8eR:
@Destroy666x - ver number should be fixed now

Destroy666x:
The solution seems to be fine, but the new code should be separated somehow (check for reference: https://github.com/mybb/mybb/issues/2199#issuecomment-141424586).

Cu8eR:
@Destroy666x - fixed, please could you check my other PRs (NOT #2303 and #2304) in your free time? I need to clean up my patches in my fork before I start with new ones. Thanks, have a nice day. If anything, tell me.

Destroy666x:
I meant wrapping the modification itself (`min-width: 120px`), not the whole selector. The point is to set it apart from the default 3rd part code.

Cu8eR:
@Destroy666x - OK, you are right. I changed my PR. Is it ok now or I missunderstood you again?

Destroy666x:
Yes, thanks.

PR #2297, opened 2015-12-29
Fixes #2071 - Inline moderation on classic mode

### LINKED ISSUE #2071: Inline moderation on classic mode.
When the postbit is set to use the classic postbit template, selecting a post doesn't change the background color of the entire post element.

Code:
var post = element.parents('.post_content');

Should be:

Code:
var post = element.parents('.post');

Or a similar solution.

Original thread: [Inline moderation on classic mode.](http://community.mybb.com/thread-159925.html)

Cu8eR:
Fixes #2071 - Inline moderation on classic mode


Destroy666x:
As always, the version of the file and templates that include the file should be bumped.

Cu8eR:
@Destroy666x should be ok now, if not tell me...

Issue #2292, opened 2015-12-29
Float number passed to mt_srand() on 32-bit systems


Devilshakerz:
The [`secure_seed_rng()`](https://github.com/mybb/mybb/blob/mybb_1806/inc/functions.php#L7659) function can generate values outside of signed integer range on 32-bit systems, returning a float value, which is then [passed to `mt_srand()`](https://github.com/mybb/mybb/blob/mybb_1806/inc/functions.php#L7748). This results in a PHP Warning being emitted.

Community discussion:
http://community.mybb.com/thread-187491.html

Changes to both `my_rand()` and `secure_seed_rng()` are recommended as in the discussion thread.

frostschutz:
How about `return unpack('l', $output);` instead of `hexdec(substr(dechex(crc32(base64_encode(...`?

Destroy666x:
`unpack('i', $output);` seems reasonable.

frostschutz:
Doesn't work as well as I hoped, it only produces 32bit numbers even on 64bit machines after all. It seems I misunderstood the documentation...

Stefan-ST:
Fixed with #2300

Issue #1815, opened 2015-02-06
openssl_random_pseudo_bytes slow on Windows


Destroy666x:
It may even exceed the execution time: http://community.mybb.com/thread-165567-post-1137862.html#pid1137862 so not sure if we should use it as the first choice on Windows (probably we should just siwtch the order of conditionals).

euantorano:
Seems to be an issue with OpenSSL, not PHP specifically. Apparently, there's not much difference in terms of security in `mcrypt_create_iv` and `openssl_random_pseudo_bytes`. Might be better to use the `mcrypt` option as the preference...

Stefan-ST:
@euantorano Just preferring mcyrpt will cause the same issue in case mcrypt in not available. Either we have to remove it or keep it and say it's a host specific issue.

Since similar code is used by other projects ([Drupal](https://github.com/drupal/drupal/blob/84092f3d051c7d8904a63d38c81eb5f671b6e71b/includes/bootstrap.inc#L1998), [Dropbox-SDK](https://github.com/dropbox/dropbox-sdk-php/blob/728ba7ddc212657eb5fdc5c3b23f067740c95aee/lib/Dropbox/Security.php#L52)) while other don't use it on windows at all ([Typo3](https://github.com/TYPO3/TYPO3.CMS/blob/095955bea743c46ad6d6208f40e9bcdd91608915/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L1154)).

euantorano:
Good point Stefan. Not sure what the best approach is. Obviously we can't have it timing out. Maybe just have a check for Windows and skip calling the OpenSSL function if true?

Destroy666x:
We already have a check for Windows: https://github.com/mybb/mybb/blob/feature/inc/functions.php#L7550 But skipping it seems to be the most reasonable solution, yes.

Destroy666x:
There are also 2 other Windows-compatible generation methods mentioned here: http://stackoverflow.com/questions/8553644/accessing-cryptgenrandom-from-php but they don't really seem reliable either.

Destroy666x:
This was corrected (the order at least) and I don't think that more can be done, so closing.

PR #2300, opened 2015-12-31
RNG rewrite

Devilshakerz:
- added PHP7's CSPRNGs
 - checking the `$crypto_strong` parameter of `openssl_random_pseudo_bytes()`
 - `secure_seed_rng()` returns numbers in the widest possible scope (previously maximum set for 32-bit systems, now dependent on the system)
 - sources reordered, according to https://github.com/paragonie/random_compat/blob/master/ERRATA.md

`secure_seed_rng()` can be relatively slower due to restructuring (~0.8-1.2x) and `my_rand()` 6-9x slower on PHP7 and 6-14x slower on older versions.

Related: #1815, #2292

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -7773,24 +7773,24 @@ function secure_seed_rng()
 {
 	$bytes = PHP_INT_SIZE;
 
-	$output = secure_binary_seed_rng($bytes);
-
-	// convert binary data to a decimal number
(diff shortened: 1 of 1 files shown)


Devilshakerz:
@mybb/developers Feedback needed

euantorano:
Looks much improved to me :clap:

PR #2302, opened 2016-01-04
Fixes #2231 - Deprecate get_alt_bg()

### LINKED ISSUE #2231: Deprecate get_alt_bg()
As it was mentioned somehwere else a long time ago but for some reason not added here, that funciton should be deprecated since it's not used anywhere and it doesn't make any sense (one background is always set).

Either that or we can fix it, but I see no point.

https://github.com/mybb/mybb/blob/feature/admin/inc/class_page.php#L775

Cu8eR:
Fixes #2231 - Deprecate get_alt_bg()

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -768,26 +768,6 @@ EOF;
 	}
 
 	/**
-	 * Switch between two different alternating background colours.
-	 *
-	 * @return string
-	 */
-	function get_alt_bg()
-	{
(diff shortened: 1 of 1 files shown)


Destroy666x:
That's removal not depreciation, but I guess noone used this broken function anyways so removal is safe.

Cu8eR:
Yes, I know... I checked if this function is used anywhere but it is not... so I removed it completely. Thx for merging!

PR #2311, opened 2016-01-18
Fixes #2310 SQL error when posting new post with custom moderation tool

### LINKED ISSUE #2310: SQL error when posting new post with custom moderation tool
```
SQL Error:
1406 - Data too long for column 'ipaddress' at row 1
Query:
INSERT INTO posts (`tid`,`replyto`,`fid`,`subject`,`icon`,`uid`,`username`,`dateline`,`message`,`ipaddress`,`includesig`,`smilieoff`,`visible`) VALUES (1,1204167,176,'foobar',0,2824,'StefanT',1453116638,'[i]Moved to Plugin Support.[/i]',X'4827323130423034373031303136303864326635653539353933643838323573643837',1,0,1)
```

Stefan-ST:
Fixes #2310

### DIFF
--- a/inc/class_custommoderation.php
+++ b/inc/class_custommoderation.php
@@ -241,7 +241,7 @@ class CustomModeration extends Moderation
 						"uid" => $mybb->user['uid'],
 						"username" => $mybb->user['username'],
 						"message" => $post_options['splitpostsaddreply'],
-						"ipaddress" => $db->escape_binary(my_inet_pton(get_ip())),
+						"ipaddress" => my_inet_pton(get_ip()),
 					);
 					// Set up the post options from the input.
 					$post['options'] = array(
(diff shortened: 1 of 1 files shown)


Destroy666x:
Noticed no problems.

BTW, it's weird that `insert_post()` gets the IP from data (which is good):
https://github.com/Stefan-ST/mybb/blob/fix-2310/inc/datahandlers/post.php#L1083
while `insert_thread()` assumes it's always the current IP:
https://github.com/Stefan-ST/mybb/blob/fix-2310/inc/datahandlers/post.php#L1450
even though IP is still provided as data:
https://github.com/mybb/mybb/blob/fda43f3dc95421a50bf2f6eb0de0a92a10f1490b/newthread.php#L339

So developers can fake IPs of posts but not threads... IMO such inconsistency shouldn't exist.

Stefan-ST:
I've noticed that, too. But changing the datahandler is probably not worth...

PR #2322, opened 2016-01-27
Add missing group check in attachment.php

Cu8eR:
link: http://community.mybb.com/thread-188374.html

### DIFF
--- a/attachment.php
+++ b/attachment.php
@@ -140,6 +140,11 @@ if(isset($mybb->input['thumbnail']))
 }
 else
 {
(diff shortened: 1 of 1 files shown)


Destroy666x:
Why did you add it there? It should be next to other permission checks.

Destroy666x:
Actually, never mind, forum permission take care of it, I just have something broken in one of my installation. Tested it in another and it's fine. Another reason to wait with PR till the issue is pushed, not the other way round.

PR #2325, opened 2016-01-29
Fixes #2314 Wrong PM update array code

### LINKED ISSUE #2314: Wrong PM update array code
https://github.com/mybb/mybb/blob/feature/inc/tasks/dailycleanup.php#L48 - UIDs are assigned as keys, while the loop https://github.com/mybb/mybb/blob/feature/inc/tasks/dailycleanup.php#L80 assumes they're values. Thus it always updates the PM count of the 1st user.

Stefan-ST:
Fixes #2314

### DIFF
--- a/inc/tasks/dailycleanup.php
+++ b/inc/tasks/dailycleanup.php
@@ -77,7 +77,7 @@ function task_dailycleanup($task)
 
 	if(!empty($user_update))
 	{
-		foreach($user_update as $uid)
+		foreach($user_update as $uid => $data)
 		{
 			update_pm_count($uid);
 		}


Cu8eR:
Looks good

Destroy666x:
Well, I was about to push this together with some small `update_pm_counter()` corrections (the description + few things are outdated and unused in it). I understand that you take over someone's assignments that weren't touched for a long time because the person is inactive, but there's no reason to do it after 10 days when the person came back. Not a big deal here though, I'll create another issue for that.

Stefan-ST:
Sorry, due to the large number of tickets being assigned but untouched for months I sometimes forget checking the assignments.

Cu8eR:
Guys, there are many issues assigned to someone for a long time without any progress. I understand that it is a hobby and all of us have personal life, but it is better to fix it by soon by someone else than wait another XX days/months. It would be nice to fix all issues which are assigned, many of them are quite old

Destroy666x:
@Cu8eR we know that. But while it's not problematic to fix something after X months, two people working on the same thing during X days is a time waste for one.

PR #2326, opened 2016-01-29
Fixes #2306 SQL error in UserCP's Group Memberships when sql_mode=onl…

### LINKED ISSUE #2306: SQL error in UserCP's Group Memberships when sql_mode=only_full_group_by
A SQL error occurs in the Group Memberships view (User CP) when the SQL mode includes `ONLY_FULL_GROUP_BY` (default in MySQL >= 5.7.5).

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

Original thread: [SQL Error - UserCP usergroups](http://community.mybb.com/thread-187439.html)

Stefan-ST:
…y_full_group_by

Fixes #2306

### DIFF
--- a/install/resources/upgrade13.php
+++ b/install/resources/upgrade13.php
@@ -43,7 +43,7 @@ function upgrade13_dbchanges()
 			$db->write_query("ALTER TABLE ".TABLE_PREFIX."users DROP KEY username");
 		}
 
-		$query = $db->simple_select("users", "username, uid", "1=1 GROUP BY username HAVING count(*) > 1");
+		$query = $db->simple_select("users", "username, uid", "1=1 GROUP BY uid, username HAVING count(*) > 1");
 		while($user = $db->fetch_array($query))
 		{
 			$db->update_query("users", array('username' => $user['username']."_dup".$user['uid']), "uid='{$user['uid']}'", 1);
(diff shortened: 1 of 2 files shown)


Destroy666x:
What about other GROUP BY queries e.g.: 
https://github.com/mybb/mybb/blob/feature/inc/class_datacache.php#L659-L676 (and 3-4 similar codes)
https://github.com/mybb/mybb/blob/feature/install/resources/upgrade13.php#L46

Stefan-ST:
If you have any idea how to find all affected queries I'll fix them, too. Unfortunately there are a lot of GROUP BY queries.

Destroy666x:
Well, I found those by simply grepping `GROUP BY`. I haven't noticed any instances other than the mentioned 3-4 queries that use `$db->build_fields_string()` and the one in upgrade. THere weren't many results, but I may have missed something.

Stefan-ST:
Okay, I'll have a look later. Thanks, @Destroy666x.

Stefan-ST:
@Destroy666x I've checked both queries you mentioned and both appear to work fine. Searching for "GROUP BY" returns several more queries that seam fine, too.

Destroy666x:
I don't think they work fine since the same mistake is made - selected columns aren't in the `GROUP BY` clause. I can't access settings, for instance, which use a similar `$db->build_fields_string()` query.

Stefan-ST:
With MySQL 5.7.10 it works fine. I've just fixed the old upgrade script because it doesn't work without the unique key on username (with the key the whole query makes no sense...)

Stefan-ST:
Merging this as it fixes the reported issue. If there are other issues please just create a new ticket.

PR #2329, opened 2016-01-30
Fix #2224 no inline moderation checkbox in quick reply post

### LINKED ISSUE #2224: Can't moderate quick reply posts - missing checkbox
`$ismod` is not defined for quick reply's AJAX in newreply.php: https://github.com/mybb/mybb/blob/feature/inc/functions_post.php#L614

Destroy666x:
Also added several other `!isset()` checks just in case a plugin or
whatever will miss definition of these globals.. And shortened some
conditionals to simple assignments.

https://github.com/mybb/mybb/issues/2224

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -265,7 +265,7 @@ function build_postbit($post, $post_type=0)
 				}
 			}
 		}
-		
+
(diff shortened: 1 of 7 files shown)


PaulBender:
Works fine for me.

PR #2331, opened 2016-01-31
Fix #2330 port not stripped from the generated cookie domain

### LINKED ISSUE #2330: Port is not stripped from the generated cookie domain
While it should be.

Destroy666x:
As the title says, thus admins my end up with wrong setting if they trust it. Also added a check whether the generated Email is indeed an Email (it may be anything)..

https://github.com/mybb/mybb/issues/2330

### DIFF
--- a/install/index.php
+++ b/install/index.php
@@ -1772,9 +1772,9 @@ function configure()
 	global $output, $mybb, $errors, $lang;
 
 	$output->print_header($lang->board_config, 'config');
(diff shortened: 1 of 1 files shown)


Stefan-ST:
This (still) adds port 443 when using https which is probably not intended.

Destroy666x:
If you meant the board URL, I fixed it.

PR #2333, opened 2016-01-31
Fixes #2074 Rating an unapproved post

### LINKED ISSUE #2074: Rating an unapproved post

Stefan-ST:
Fixes #2074

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -333,7 +333,7 @@ function build_postbit($post, $post_type=0)
 		}
 
 		$post['button_rep'] = '';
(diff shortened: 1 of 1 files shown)


Destroy666x:
Note that the button is also shown in PMs and previews, so `$thread` is undefined, not sure what's the value of `$post['visible']` in that case.

Stefan-ST:
You are right, fixed.

PR #2336, opened 2016-01-31
Change protocol-relative embeds to HTTPS-default

JoshHarmon:
Changes jscripts/bbcodes_sceditor.js to use https URLs instead of
protocol-relative URLs. There is no technical reason to not use
secure versions of resources, as unless one is in the middle of a
man-in-the-middle attack, their request is sent much more securely.

If the user is in a MITM scenario, they're already out of luck with
the protocol-relative URLs. That's not the point of this change.
They're out of luck even with this change -- I get that.

Most users are not under active MITM attacks, and most forums are not
running a TLS configuration. As a result, users' resource requests
effectively get downgraded to HTTP for no reason, exposing request
metadata that may pose privacy risks.

Signed-off-by: Josh Harmon <me@joshharmon.me>

EDIT: I don't know if there's an issue existing for this, and was unable to turn anything up in searches. I know we've talked about this before though.

### DIFF
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -459,7 +459,7 @@ $(document).ready(function($) {
 				{
 					case 'dailymotion':
 						matches = content.match(/dailymotion\.com\/video\/([^_]+)/);
-						url     = matches ? 'http://www.dailymotion.com/embed/video/' + matches[1] : false;
+						url     = matches ? 'https://www.dailymotion.com/embed/video/' + matches[1] : false;
 						break;
 					case 'facebook':
 						matches = content.match(/facebook\.com\/(?:photo.php\?v=|video\/video.php\?v=|video\/embed\?video_id=|v\/?)(\d+)/);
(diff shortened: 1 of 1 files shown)


Destroy666x:
This changes it only in SCEditor, there are also video templates + the `ver` parameter of this file needs to be bumped in all templates and ACP file(s) + the versions of edited templates need to be bumped as well.

Stefan-ST:
@JoshHarmon Can you please create a ticket with appropriate labels and milestone before submitting a pull request. Also please do not create branches on our main repository but on your own fork (our workflow is described here: http://community.mybb.com/announcement-24.html)

JoshHarmon:
@Destroy666x I was worried that I was missing something.

@Stefan-ST I figured someone who browses GitHub more actively would know exactly what the issue number was or would have a vague idea of what the title was, and I didn't want to pollute the tracker with the dupe. As for the repo part, yes. You're absolutely right - I had the wrong remote linked and GitHub Desktop sure doesn't help me there.

Stefan-ST:
The dailymotion MyCode should be changed from http to protocol-relative though to replicate the functionality from our MyCode parser.

PR #2337, opened 2016-02-01
Fix #2312

### LINKED ISSUE #2312: Registration confirmation emails not being sent
Hi All - 

My private forum is set to force users to log in to view the board, and the registration method is Administrator activation.

I am receiving email notifications that there are new users pending admin activation, and I am able to activate new registrations with no problem. However, after activating the new user, an email is not being sent out to the user confirming that they have been activated. I can't seem to find that setting anywhere.

Any suggestions? I am running MyBB 1.8.6.

Original thread: [Registration confirmation emails not being sent](http://community.mybb.com/thread-188008.html)

dequeues:
#2312


admin/modules/user/users.php
```diff
@@ -2426,6 +2431,13 @@ function toggleBox(action)
 
 				$cache->update_awaitingactivation();
 
+				// send activation mail
+				for($i = 0; $i < $counter; $i++)
```
Stefan-ST: $counter can be omitted if you use foreach instead. ;)

admin/modules/user/users.php
```diff
@@ -2426,6 +2428,13 @@ function toggleBox(action)
 
 				$cache->update_awaitingactivation();
 
+				// send activation mail
+				foreach ($user_mail_data as $mail_data)
```
Destroy666x: There should be no space between `foreach` and `(` according to our standards. Apart from that the PR is ok

dequeues:
Oops! Fixed.

PR #2338, opened 2016-02-02
Fixes #2287 Prefix - Staff Only = Uneditable

### LINKED ISSUE #2287: Prefix - Staff Only = Uneditable
Hello! We have a set of Thread Subject Prefixes that only staff can change. "Available To Groups" we have it only staff members can edit it.

However when a person without permission to Thread Prefixes tries to edit said post they will receive an error: "The selected prefix is invalid. Please select a valid prefix." and said user cannot update their post. 

We use these prefixes to mark something that only staff can mark the thread with. This is becoming an increasingly annoying issue to us, and I feel that it is a bug and not a feature.

Original thread: [Prefix - Staff Only = Uneditable](http://community.mybb.com/thread-187132.html)

Stefan-ST:
It makes the original prefix available regardless of permissions when editing a thread.

Fixes #2287

### DIFF
--- a/editpost.php
+++ b/editpost.php
@@ -844,7 +844,7 @@ if(!$mybb->input['action'] || $mybb->input['action'] == "editpost")
 			$mybb->input['threadprefix'] = $thread['prefix'];
 		}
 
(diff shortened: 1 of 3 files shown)


PaulBender:
Works fine for me

PR #2343, opened 2016-02-06
Fixed an amusing typo in a comment in class_parser.php

skftn:
Cartridge returns, for they are the workings of the devil!

hmm


Destroy666x:
The whole sentence doesn't really make grammatical sense to me, but yeah...

PR #2347, opened 2016-02-27
Fix #2298

### LINKED ISSUE #2298: Weird signature validation conditionals
https://github.com/mybb/mybb/blob/08c3cc868f7f5043bba57ebc946d25bf67402010/usercp.php#L93 - this doesn't make any sense as `else if` since a signature can be too long and have too many images at the same time. `inline_error()` accepts arrays for errors which is more user-friendly and should be used instead of printing one error at a time...

The whitespacing in the conditional above is quite odd as well. And the signature (+ avatar) could be validated/updated by userhandler just like other stuff so that it'd be more 3rd party friendly.

dequeues:


Cu8eR:
#2298

Destroy666x:
Is the PR finished? I don't think it can work without any usercp file edits, perhaps you forgot to include the changed file.

dequeues:
I did indeed. Not sure what's happened with the lines. Maybe it's to do with Atom. I'll revert it and try again if it's an issue.

JoshHarmon:
Are you using "soft tabs"/spaces for indentation? That's what it looks like at first glance.

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
...
+		$parser_options = array(
+			'allow_html' => $mybb->settings['sightml'],
+			'filter_badwords' => 1,
+			'allow_mycode' => $mybb->settings['sigmycode'],
+			'allow_smilies' => $mybb->settings['sigsmilies'],
+			'allow_imgcode' => $mybb->settings['sigimgcode'],
+			"filter_badwords" => 1
+		);
+
+		if($mybb->user['showimages'] != 1 && $mybb->user['uid'] != 0)
```
Destroy666x: This is incorrect. Datahandler shouldn't assume these values are from the current user. You should pass them as user data.

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
...
+		{
+			$parser_options['allow_imgcode'] = 0;
+		}
+
+		$parsed_sig = $parser->parse_message($mybb->get_input('signature'), $parser_options);
```
Destroy666x: Same here, the only difference is that you pass signature as data this time, but aren't using the data.

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
 			remove_avatars($uid);
 		}
 	}
+
+	public function verify_signature()
```
Destroy666x: The function should be called in user verification if required field is set.

dequeues:
Not sure why I totally forgot about the signature being used in the datahandler. Sorted.

dequeues:
Do you mean in `validate_user()`?

Destroy666x:
Yes, I'm still commenting on several things BTW, so I'd recommend you to wait.

dequeues:
Sure.

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
...
+		$parser_options = array(
+			'allow_html' => $mybb->settings['sightml'],
+			'filter_badwords' => 1,
+			'allow_mycode' => $mybb->settings['sigmycode'],
+			'allow_smilies' => $mybb->settings['sigsmilies'],
+			'allow_imgcode' => $mybb->settings['sigimgcode'],
+			"filter_badwords" => 1
+		);
+
+		if($this->data['showimages'] != 1 && $this->data['uid'] != 0)
```
Destroy666x: I see you edited it while I was editing my comment. But now that I look at it, the conditional most likely was/is entirely wrong from the beginning, is probably another result of careless copy+paste and causes another much more serious bug... First of all, guests can't access UCP or have signature in the first place, so the 2nd part was entirely useless (it's useless here too: https://github.com/dequeues/mybb/blob/fix2298/usercp.php#L1953). Secondly, it uses current user's setting to overwrite the img code parser, which makes it possible for the user to have unlimited amount of images and ignore the group limit if they disable the `showimages` option... The conditional should be entirely deleted and `showimages` not passed.

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
...
+
+		if((($mybb->settings['sigimgcode'] == 0 && $mybb->settings['sigsmilies'] != 1) &&
+			substr_count($parsed_sig, "<img") > 0) ||
+			(($mybb->settings['sigimgcode'] == 1 || $mybb->settings['sigsmilies'] == 1) &&
+			substr_count($parsed_sig, "<img") > $mybb->settings['maxsigimages'])
+		)
+		{
+			$imgsallowed = ($mybb->settings['sigimgcode'] == 1 ? $mybb->settings['maxsigimages'] : 0);
+			$this->set_error('too_many_sig_images2', array($imgsallowed));
+			$mybb->input['preview'] = 1;
```
Destroy666x: Preview handling definitely shouldn't be here. The line is useless anyways since preview is not available if there are any errors: https://github.com/dequeues/mybb/blob/fix2298/usercp.php#L1903

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
...
+		}
+
+		$parsed_sig = ($mybb->settings['sigcountmycode'] == 0 ? $parser->text_parse_message($this->data['signature']) : $this->data['signature']);
```
Destroy666x: Ternary isn't unfortunately allowed by our standards.

usercp.php
```diff
@@ -57,67 +57,20 @@
...
-			}
-			else
-			{
-				$lang->sig_too_long .= $lang->sig_remove_chars_singular;
-			}
-			$error = inline_error($lang->sig_too_long);
-		}
-	}
-	if(isset($error) || !empty($mybb->input['preview']))
+	if (!$userhandler->verify_signature())
```
Destroy666x: No space before `(`. There also might be problem with this conditional (I recall userhandler error from another PRs), I'll test it later. For now that's all I found by reading the code.

inc/datahandlers/user.php
```diff
@@ -1783,4 +1783,61 @@ function clear_profile($delete_uids=false, $gid=0)
...
+
+		$parsed_sig = preg_replace("#\s#", "", $parsed_sig);
+		$sig_length = my_strlen($parsed_sig);
+
+		if($sig_length > $mybb->settings['siglength'])
+		{
+			$this->set_error('sig_too_long', array($mybb->settings['siglength']));
+
+			if($sig_length - $mybb->settings['siglength'] > 1)
```
Destroy666x: `$sig_length - $mybb->settings['siglength']` could be assigned to a variable

inc/datahandlers/user.php
dequeues: I don't really know how you can call it "careless copy+paste", because as far as I was aware there was no issue with the current code to be moved over regarding this comment or for me to look into this further. If there was an issue brought to attention, considering this has been here for a while, I could've looked into it and changed it appropriately.

Destroy666x: If you read my comment again, I said it was entirely wrong from the beginning... Someone copy pasted it earlier, probably when the setting was added - it shouldn't have been there in the first place.

dequeues: Oh, sorry. I took it the wrong way. I thought the part that I quoted was aimed at me.

dequeues:
If nobody else has an issue with this I'll go ahead and merge.

PR #2356, opened 2016-03-06
Fix fetch remote file

Devilshakerz:
- redirect support for fsockopen
- manual redirect in case CURLOPT_FOLLOWLOCATION is not accessible
(open_basedir or safe_mode enabled)
- PHP Notices fix


inc/functions.php
```diff
@@ -6765,19 +6766,56 @@ function fetch_remote_file($url, $post_data=array())
 
 	if(function_exists("curl_init"))
 	{
+		$can_followlocation = @ini_get('open_basedir') == null && (!@ini_get('safe_mode') || @ini_get('safe_mode') == 'Off');
+
+		$request_header = $max_redirects != 0 && !$can_followlocation;
```
euantorano: Semantics, but I'd prefer `$can_follow_location` - not a big deal though.

inc/functions.php
```diff
@@ -6765,19 +6766,56 @@ function fetch_remote_file($url, $post_data=array())
 
 	if(function_exists("curl_init"))
 	{
+		$can_followlocation = @ini_get('open_basedir') == null && (!@ini_get('safe_mode') || @ini_get('safe_mode') == 'Off');
```
Stefan-ST: @Devilshakerz This check should probably use $this->safemode. I'm also not entirely sure the open_basedir check works under all circumstances. If ini_get is disabled it always returns null regardless of the actual value.

inc/functions.php
Devilshakerz: I've kept it like this since it only relates to the cURL's option.

inc/functions.php
Devilshakerz: Should MyBB assume `FOLLOWLOCATION` is not allowed when `ini_get` is disabled?

Stefan-ST: I think so. Better safe than sorry, right?

inc/functions.php
```diff
@@ -6849,13 +6889,36 @@ function fetch_remote_file($url, $post_data=array())
...
 		$data = explode("\r\n\r\n", $data, 2);
-		return $data[1];
+
+		$header = $data[0];
+		$status_line = current(explode("\n\n", $header, 1));
+		$body = $data[1];
+
+		if($max_redirects != 0 && (strstr($status_line, ' 301 ') || strstr($status_line, ' 302 ')))
+		{
+			preg_match('/Location:(.*?)\n/', $header, $matches);
```
Stefan-ST: The regex `/Location:(.*?)\n/` doesn't match when the Location header is at the end.

PR #2361, opened 2016-03-13
Fixes #2282 Unlisted "maxreputationsperuser" setting

### LINKED ISSUE #2282: Unlisted "maxreputationsperuser" setting
The maxreputationsperuser setting can be changed when the corresponding POST value is added manually to the request updating group settings, but is not listed.

https://github.com/mybb/mybb/search?utf8...onsperuser

RE: http://community.mybb.com/thread-186824.html

Original thread: [Unlisted "maxreputationsperuser" setting](http://community.mybb.com/thread-186827.html)

PaulBender:
Fixes #2282


inc/class_core.php
```diff
@@ -598,7 +598,7 @@ function __destruct()
  */
 
 $grouppermignore = array("gid", "type", "title", "description", "namestyle", "usertitle", "stars", "starimage", "image");
-$groupzerogreater = array("pmquota", "maxpmrecipients", "maxreputationsday", "attachquota", "maxemails", "maxwarningsday", "maxposts", "edittimelimit", "canusesigxposts", "maxreputationsperthread", "emailfloodtime");
+$groupzerogreater = array("pmquota", "maxpmrecipients", "maxreputationsday", "attachquota", "maxemails", "maxwarningsday", "maxposts", "edittimelimit", "canusesigxposts", "maxreputationsperuser", "maxreputationsperthread", "emailfloodtime");
```
Destroy666x: The array is also in upgrade script, not sure what it does there but should be updated while it exists.

PaulBender: I've added them

Destroy666x:
I also believe these should be 0s: https://github.com/PaulBender/mybb/blob/fix-2282/admin/modules/user/groups.php#L67-L68

PaulBender:
Fixed both.

Destroy666x:
Noticed one last connected bug that affects all max rep per day permissions - if you try to rate the same post again, the error is thrown, even though you are updating the reputation, not adding a new one.

PaulBender:
Okay, that should fix it.

PR #2362, opened 2016-03-13
Fixes #2236 Buddy popup problem when user doesn't have any permissions

### LINKED ISSUE #2236: Buddy popup problem when user doesn't have any permissions
If a user that doesn't have the permission to see the forum at all and clicks on the buddy popup link it doesn't load a popup but instead appends a whole no permission page below the footer.

PaulBender:
Fixes #2236

### DIFF
--- a/global.php
+++ b/global.php
@@ -452,7 +452,7 @@ else
 $templatelist .= "headerinclude,header,footer,gobutton,htmldoctype,header_welcomeblock_member,header_welcomeblock_guest,header_welcomeblock_member_admin,global_pm_alert,global_unreadreports,error,footer_languageselect_option,footer_contactus";
 $templatelist .= ",global_pending_joinrequests,global_awaiting_activation,nav,nav_sep,nav_bit,nav_sep_active,nav_bit_active,footer_languageselect,footer_themeselect,header_welcomeblock_member_moderator,redirect,header_menu_calendar,nav_dropdown,footer_themeselector,task_image";
 $templatelist .= ",global_boardclosed_warning,global_bannedwarning,error_inline,error_nopermission_loggedin,error_nopermission,debug_summary,header_quicksearch,header_menu_search,header_menu_portal,header_menu_memberlist,usercp_themeselector_option,smilie,global_board_offline_modal";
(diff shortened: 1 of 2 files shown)


install/resources/mybb_theme.xml
```diff
@@ -4057,6 +4057,20 @@ $(".forumjump").change(function() {
 </div>
 <br />]]></template>
 		<template name="global_boardclosed_warning" version="1600"><![CDATA[<div class="red_alert"><strong>{$lang->bbclosed_warning}</strong></div>]]></template>
+		<template name="global_no_permission_model" version="1807"><![CDATA[<div class="modal">
```
Destroy666x: Why '_model'?

PaulBender: Fixed

install/resources/mybb_theme.xml
```diff
@@ -4057,6 +4057,20 @@ $(".forumjump").change(function() {
 </div>
 <br />]]></template>
 		<template name="global_boardclosed_warning" version="1600"><![CDATA[<div class="red_alert"><strong>{$lang->bbclosed_warning}</strong></div>]]></template>
+		<template name="global_no_permission_modal" version="1807"><![CDATA[<div class="modal">
```
Destroy666x: 1808

global.php
```diff
@@ -1018,19 +1019,46 @@
 			$allowable_actions = explode(',', ALLOWABLE_PAGE);
 			if(!in_array($mybb->get_input('action'), $allowable_actions))
 			{
-				error_no_permission();
+				if(!$mybb->get_input('modal'))
```
Destroy666x: This could be shortened to a temporary true/false variable and a single call instead of three. Not important, but IMO slightly cleaner. Except these 2 small changes it can be merged.

global.php
```diff
@@ -1018,19 +1020,33 @@
...
+			$notallowed = true;
 		}
 	}
 	else
 	{
-		error_no_permission();
+		$notallowed = true;
+	}
+
+	if($notallowed = true)
```
Destroy666x: Should be `==`

PaulBender: Fixed

PR #2364, opened 2016-03-13
Fix #2363 per theme default avatar

### LINKED ISSUE #2363: Per theme default avatar
Make it possible to use something like `{theme}` for the default avatar as suggested here: http://community.mybb.com/thread-172906.html

Destroy666x:
I didn't include `$change_dir` because it doesn't seem to be implemented
correctly anyways (I'll create an issue later).

https://github.com/mybb/mybb/issues/2363


Cu8eR:
Looks fine

Destroy666x:
The PR is finished after the last commit

JordanMussi:
@Destroy666x, I suggest using `<theme>` instead of `{theme}` since `<lang:foobar>` is [already used](https://github.com/mybb/mybb/blob/mybb_1807/inc/class_language.php) for replacing language variables in various things. Also using curly braces may cause confusion with `{$theme}`.

Another thing, I think the following setting description will make it completely clear what happens (you also have `use use` in your description - I hate typos too! :stuck_out_tongue:):
`If the user does not set a custom avatar this image will be used instead. To have a different default avatar for each theme, you may use <strong>&lt;theme&gt;</strong> which is replaced with the directory of the active theme.`

Which would display as:
> If the user does not set a custom avatar this image will be used instead. To have a different default avatar for each theme, you may use **&lt;theme&gt;** which is replaced with the directory of the active theme.

admin/modules/user/users.php
```diff
@@ -3587,7 +3587,7 @@ function build_users_view($view)
 			}
 			if(!$user['avatar'])
 			{
-				$user['avatar'] = "../".$mybb->settings['useravatar'];
+				$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
```
JordanMussi: Have you created an issue for not using `format_avatar()` here (and the one above) or could this be fixed with #2363?

Destroy666x:
> @Destroy666x, I suggest using <theme> instead of {theme} since <lang:foobar> is already used for replacing language variables in various things. Also using curly braces may cause confusion with {$theme}.

I don't know why `<theme>` would be better than `{theme}` which is used literally everywhere for replacement.. Consistency is important as you say and you can easily check that this is not the only place with `{theme}`, I even mentioned placed where it's used in the PR. And for language folders `{lang}` is used, I haven't seen the one you linked presented to the user anywhere. 

> you also have use use in your description - I hate typos too!

Thanks, corrected. The rest of the description is copied from another place that uses `{theme}` and IMO it's understandable.

admin/modules/user/users.php
Destroy666x: It's not that important, but I created an issue: https://github.com/mybb/mybb/issues/2367

JordanMussi:
Okies, I haven't been around for a while! Didn't remember `{theme}` was already used elsewhere...

dequeues:
This has conflicts which need to be resolved.

Destroy666x:
Well, conflict is fixed but rebase for some reason messed everything up, so I'll recreate it

PR #2366, opened 2016-03-14
Bug with search engine friendly URLs

motom001:
Missing property "RewriteBase /" 

http://hilfe-center.1und1.de/hosting/1und1-hosting-c10085285/skript--und-programmiersprachen-c10082634/htaccess-c10083883/hinweise-zur-erstellung-von-rewrite-rules-a10792317.html


Destroy666x:
If you have .htaccess in the forum root folder, it shouldn't be necessary. Also, it's just an exemplary file.

PR #2372, opened 2016-03-20
Fix #2370 moderator send report issues

### LINKED ISSUE #2370: Report notifications ignore moderator groups
This query in inc/functions_modcp.php just checks for users moderating the forum but not for groups:
```php
	if(!empty($forum['parentlist']))
	{
		$query = $db->query("
			SELECT DISTINCT u.username, u.email, u.receivepms, u.uid
			FROM ".TABLE_PREFIX."moderators m
			LEFT JOIN ".TABLE_PREFIX."users u ON (u.uid=m.id)
			WHERE m.fid IN (".$forum['parentlist'].") AND m.isgroup = '0'
		");

		$nummods = $db->num_rows($query);
	}
```
There is no code for isgroup=1.

Destroy666x:
Added moderator group handling and corrected `isgroup`/`issupermod` query logic. I haven't tested pgsql/SQLite.

https://github.com/mybb/mybb/issues/2370


PaulBender:
Works fine for me on all three SQL services.

PR #2373, opened 2016-03-20
Fixes #2371 - User activity in mod cp

### LINKED ISSUE #2371: Incorrect user activity when action is unset
At the moment if a Moderator is viewing the Mod CP homepage, on the Who's Online page their location is shown as "Viewing Reported Posts" rather than "Viewing Moderator CP".

The fix is rather simple, change [functions_online.php#228-232](https://github.com/mybb/mybb/blob/mybb_1807/inc/functions_online.php#L228L232) from:

``` php
            if(!isset($parameters['action']))
            {
                $parameters['action'] = 0;
            }
```

to:

``` php
            if(!isset($parameters['action']))
            {
                $parameters['action'] = 0;
                $user_activity['activity'] = "modcp";
                break;
            }
```

Cu8eR:
#2371

### DIFF
--- a/inc/functions_online.php
+++ b/inc/functions_online.php
@@ -229,6 +229,8 @@ function fetch_wol_activity($location, $nopermission=false)
 			if(!isset($parameters['action']))
 			{
 				$parameters['action'] = 0;
+				$user_activity['activity'] = "modcp";
+				break;
 			}
 
 			$accepted_parameters = array("modlogs", "announcements", "finduser", "warninglogs", "ipsearch");


Destroy666x:
I'm quite sure the fix is incorrect and the bug is somewhere else because this should take care of it: https://github.com/Cu8eR/mybb/blob/patch-4/inc/functions_online.php#L265-L268

Destroy666x:
The cause is setting the parameter to 0, which doesn't make any sense and should be removed (or changed to empty string depending on ->). Also, either the empty check should be removed or the newly added lines. But the current `break` way is more optimal.

Cu8eR:
OK, I understand. I will check it

dequeues:
Any progress with this @Cu8eR?

Cu8eR:
@dequeues - no, feel free to push your PR with fix.

Destroy666x:
Ok, closing this then

PR #2376, opened 2016-03-22
Fix #259 additional groups not considered in is_moderator() function with 0 as 1st argument

### LINKED ISSUE #259: User who is member of group that moderates certain (not all) furums is not recog...
> Funcion is_operator($fid,$uid) in source file inc/functions.php returns false when called for users that are member of a group that moderates certain forums. This becomes clear when such a user does a search, then the header moderater column for "checkall" is missing in the table of thread resulds, and also in the thread results lines for threads of forums that are not moderated by his group is missing. I found out that the function is called as is_operator() or is_operator(0,$uid) retuns an erroneous result.  
> I have made a fix that works, which is attached

Original report: http://dev.mybb.com/issues/1700
Reporter: Ad Bakker
Status: New

Destroy666x:
Not the most beautiful solution, but I think it's useful for `usergroup_permissions()` to return group list info regardless of that issue.

https://github.com/mybb/mybb/issues/259


Destroy666x:
Also added support for `$action` if `$fid` is 0. It returns true only if the user can use that action in any forum.

PR #2384, opened 2016-03-28
Fixes #2383 Update timezone

### LINKED ISSUE #2383: Update timezone
Update some minor things related to the timezone dropdown menus.

- [x] Add +8:30 timezone.
- [x] Marshall Islands is listed twice (at -12 and +12; they use +12).

PaulBender:
Fixes #2383


Destroy666x:
+08:45 seems to be the last one that's missing.

PaulBender:
Fixed everything. The upgrade script already updates the optioncode from settings.xml automatically, though a blank upgrade36.php is needed.

Destroy666x:
`optionscode` doesn't seem to be updated (unless every setting is forced to be updated based on `revert_all_settings`) but now I'm not sure if it should be anyways due to poor setting translation possibilities. It'd make them English even when they were translated earlier.

PR #2387, opened 2016-04-02
Fix #2378

### LINKED ISSUE #2378: Report spam possible with PM/E-mail report medium
You can send countless reports about the same user/post/reputation, the duplicate checks only work if ModCP reporting is enabled (proably because the report is saved to DB only then, haven't looked yet). I'm not sure why. It can be an annoyance if someone notices which medium is used and tries to abuse it.

dequeues:


Destroy666x:
https://github.com/mybb/mybb/issues/2378

The global report bar seems to be displayed, it should be disabled depending on the setting (and anything similar as well).

dequeues:
Thought I added that, my bad.

global.php
```diff
@@ -627,7 +627,7 @@
 		$can_access_moderationqueue = false;
 	}
 
-	if($can_access_moderationqueue || ($mybb->user['ismoderator'] && $mybb->usergroup['canmodcp'] == 1 && $mybb->usergroup['canmanagereportedcontent'] == 1))
+	if($can_access_moderationqueue || ($mybb->user['ismoderator'] && $mybb->usergroup['canmodcp'] == 1 && $mybb->usergroup['canmanagereportedcontent'] == 1) && $mybb->settings['reportmethod'] == "db")
```
Destroy666x: It'd be more optimal at the beginning of https://github.com/dequeues/mybb/blob/fix2378/global.php#L592

Destroy666x:
I see you changed the order but not the line - https://github.com/dequeues/mybb/blob/fix2378/global.php#L592 The whole block is for report bar.

Destroy666x:
The bar conditional is still not fully correct if you test it (doesn't work) - parentheses are missing, `&&` doesn't work as supposed to.

admin/modules/home/index.php
```diff
@@ -260,13 +260,16 @@
 	$query = $db->simple_select("posts", "COUNT(*) AS newposts", "dateline > '$timecut' AND visible='1'");
 	$newposts = my_number_format($db->fetch_field($query, "newposts"));
 
-	// Get the number of reported post
-	$query = $db->simple_select("reportedcontent", "COUNT(*) AS reported_posts", "type = 'post' OR type = ''");
-	$reported_posts = my_number_format($db->fetch_field($query, "reported_posts"));
+	// If report medium is MCP...
```
Destroy666x: No `if` here either. Besides, only the unread number is redundant due to no "mark as read" option. The other one may still be useful to see how many reports were sent.

dequeues:
I honestly have no idea why the if statements got removed.

Issue #2396, opened 2016-04-21
Smilies popup


Destroy666x:
Hi.
Calling action in misc.php "smilies" have a bug on typing custom javascript function popup window emotoicons inserter.

When we click on action for opening smilies window we typing action="onlick" editor name like this:

Code:
&lt;a href="javascript:MyBB.popupWindow('/misc.php?action=smilies&amp;amp;popup=true&amp;amp;editor=MyBBEditor&amp;amp;modal=1')"&gt;Emots&lt;/a&gt;
But code in file misc.php skip this(ignore editor name), and replace Permanently this name.
PHP Code:
&#36;onclick&nbsp;=&nbsp;"&nbsp;onclick=&#092;"MyBBEditor.insertText('&nbsp;&#36;smilie_insert&nbsp;');&#092;"";&nbsp;

Oryginal code.
PHP Code:
&#36;onclick&nbsp;=&nbsp;"&nbsp;onclick=&#092;"MyBBEditor.insertText('&nbsp;&#36;smilie_insert&nbsp;');&#092;"";&nbsp;

It should be like this
PHP Code:
&#36;onclick&nbsp;=&nbsp;"&nbsp;onclick=&#092;"".&#36;mybb-&gt;input['editor'].".insertText('&nbsp;&#36;smilie_insert&nbsp;');&#092;"";&nbsp;

Now we can get action with any html obiect to use emotoicons with some javascript code.

Code:
&lt;a href="javascript:MyBB.popupWindow('/misc.php?action=smilies&amp;amp;popup=true&amp;amp;editor=MyCustomEditor&amp;amp;modal=1')"&gt;Emots&lt;/a&gt;

And all work better and popup smilies can work with any custom text area where we want.

Original thread: [Smilies popup](http://community.mybb.com/thread-189590.html)

premiumcuyuz:
fixed?

JordanMussi:
@premiumcuyuz there is a fix for this issue in #2397 but it hasn't been tested and merged so it won't be included in 1.8.8. :smiley:

PR #2397, opened 2016-04-21
Fix #1828 video/images SCEditor permissions and #2396 editor object variable

### LINKED ISSUE #1828: Videos and images forum settings not respected by SCEditor
If you disable videos/images in a forum, they can still be used in SCEditor.

The solution is simple - pass these 2 forum settings as parameters just like it's done with smilies.

EDIT: same applies to signatures (images) and PMs (images and videos) controlled by settings.

Destroy666x:
Added proper permission checks and a hook for easy toolbar edition (also in ACP). Also fixed wrong forum conditions in editpost.php BTW.
https://github.com/mybb/mybb/issues/1828

Added a way to change the default object variable in case there are e.g. multiple editors with different permissions on one page.
https://github.com/mybb/mybb/issues/2396

Not fully tested yet.

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -890,20 +890,24 @@ EOF;
 	 * @param string $bind The ID of the textarea to bind the editor to.
 	 * @param string $editor_language The language string for the editor.
 	 * @param bool $smilies Whether or not smilies should be included
+	 * @param bool $images Whether or not images should be included
+	 * @param bool $videos Whether or not videos should be included
+	 * @param string $editor Name of the editor JS object
+	 *
 	 * @return string The build MyCode editor Javascript.
 	 */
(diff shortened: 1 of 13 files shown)


JN-Jones:
@Destroy666x What's the status of this apart from the conflicts?

JN-Jones:
These changes look good to me. If @Destroy666x doesn't reply within this week I'm going to take this over and fix the conflicts so that we can test it and hopefully include it in 1.8.9

Ben-Conway:
@JN-Jones Are you going to take it over?

JN-Jones:
Unfortunately hadn't time but considering that we plan to update SCE and other 3rd party libraries in 1.8.10 this PR probably fits better there as we need to change most SCE related files.

PR #2399, opened 2016-05-03
Fix php mail handler regression

rob006:
`$config` is global var, it should not be used in method scope.


Devilshakerz:
fixes #2402

Issue #2057, opened 2015-06-25
PM folder language problem


JN-Jones:
I just installed a fresh copy of 1.8.2 on my localhost to check if it might be an edited template problem, but it seems not.

I have actually three languages installed. Going to the User CP I realized that the language for the following items of the PM system are always remaining in english, as you can see in the included attachment.

Inbox
   Send items
   Drafts
   Trash Can

The language variables in the global.lang.php are all present.

&#36;l['folder_inbox']

&#36;l['folder_sent_items'] 
&#36;l['folder_drafts']
&#36;l['folder_trash']

Edited: Did a check. When first selecting the language the language strings are inserted into the users -&gt; pmfolders Database field, when selecting another language, those language strings are not updated, therefore they remain in the first chosen language.

In functions_user.php line 642 changing the not equal condition to equal solves the error.

from if(&#36;name != '') to if(&#36;name = '').

Original thread: [Language problem](http://community.mybb.com/thread-163027.html)

JN-Jones:
@Destroy666x posted a solution here: http://community.mybb.com/thread-163027-post-1119367.html#pid1119367 Just needs to be rechecked.

Sama34:
Should probably use a language string instead and no allowing to rename the default directories.

Just like default template groups.

Destroy666x:
Language strings are already used, but as I mentioned in my post there is useless code which insert these strings to database for the 1st viewed language. We can just remove it and default language strings will be used for non-renamed folders.

So no reason to remove this feature IMO.

Sama34:
I meant to insert something like `<lang->folder_inbox>` instead of `$lang->folder_inbox` and parse them later but if removing them altogether works too then I agree.

Destroy666x:
> I meant to insert something like <lang->folder_inbox> instead of $lang->folder_inbox and parse them later

Ah ok, that's a solution too, yes. Not sure why would we disallow directory renaming in that case either though.

On a side note - there may be a problem with the weird escape sequence just like in poll options. I haven't checked everything yet, but I don't think I saw code for disallowing/stripping the sequence.

JN-Jones:
Yeah, should be rechecked and fixed in one go then.

Sama34:
I don't think so, you can't insert `$%%$` into names.

Destroy666x:
You can insert a number and two ** though, not sure if that matters, probably not.

Sama34:
[...]

PR #2406, opened 2016-05-18
Better fix for #2057 - PM folders not translated

Destroy666x:
Used the more proper `private.php` code instead of removing it completely, which may result in new users being confused about lack of folders in menu after visiting `usercp.php` directly.

https://github.com/mybb/mybb/issues/2057


Cu8eR:
+1, I agree!

PR #2411, opened 2016-05-22
Fix #2405 [img align=X] float not cleared

### LINKED ISSUE #2405: Using [img align=X] overlaps with postbit_signature
Here's an example: http://community.mybb.com/thread-193145-post-1222394.html#pid1222394

`clear: both` is missing

Destroy666x:
Adds a `<br class="clear" />` at the end of the message. I don't think
it's the best solution, `::after` would be much better, but at least it
stops `align` from being overly annoying.

https://github.com/mybb/mybb/issues/2405

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -87,6 +87,14 @@ class postParser
 	 * @var int
 	 */
 	public $list_count;
+	
+	/**
+	 * Whether or not should a <br /> with clear: both be added at the end of the parsed message
+	 *
+	 * @access public
+	 * @var boolean
+	 */
+	public $clear_needed = false;
 
 	/**
 	 * Parses a message with the specified options.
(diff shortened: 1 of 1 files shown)


euantorano:
Seems to work fine for me. We already (mis)use `.clear` in plenty of places, so one more shouldn't be a problem.

Sama34:
Superfluous, adding `clear: both;` to `#signature` and `#post_meta` was enough.

euantorano:
It was so long as theme authors didn't change those elements in their templates. This fix ensures the clear will be added.

Sama34:
This fix ensures to insert a new tag that theme authors will have poor control over and will force those who already applied a _proper_ fix to update their themes.

I thought this was one reason the "Find updated templates" tool was implemented for, anyways.

euantorano:
@Sama34 the problem is that the vast majority of people likely don't use that tool and just expect things to work after an upgrade. This fix has very little chance of breaking layouts as far as I can see. 

Also, remember that stylesheets aren't updated on upgrade as far as I remember - this has been discussed regarding @Eric-Jackson's recent patch.

Destroy666x:
> Superfluous, adding clear: both; to #signature and #post_meta was enough.

And I thought parser was used not only in posts.

Sama34:
This was a theme issue, not a parser one. But if you consider this to be a _proper_ fix then fine.

Destroy666x:
Theme issue - yes. Parser issue - yes as well since it adds floats. But how would you affect all parsed messages without editing CSS as it was mentioned?

Besides, I checked how it looks and I don't see any theme influence, so not sure what you mean in the 1st place. If `<br /.>` is the last element without any preceding `<br />`s, it doesn't break the line.

PR #2415, opened 2016-06-02
Fix #2022: Enable opening/closing threads again by replying

### LINKED ISSUE #2022: Close Thread doesn't work via reply
Close Thread doesn't work via reply/quick reply...

https://github.com/mybb/mybb/blob/featur...#L889-L901

Original thread: [Close Thread doesn't work via reply](http://community.mybb.com/thread-171504.html)

EDIT: also: http://community.mybb.com/thread-171504-post-1161753.html#pid1161753

beverloo:
This bug was introduced by #1922, and is obvious on lines 876 and 883 of that PR.

Also switch $thread['closed'] with $closed in the conditionals because (a) it's been cached for a reason, and (b) the $closed variable is being modified within them. It reads better.

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -894,16 +894,16 @@ class PostDataHandler extends DataHandler
 				$modoptions_update = array();
 
 				// Close the thread.
-				if($modoptions['closethread'] == 1 && $thread['closed'] != 1)
+				if($modoptions['closethread'] == 1 && $closed != 1)
 				{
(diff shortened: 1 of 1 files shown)


dequeues:
This still doesn't open the thread. If you uncheck `Close Thread` (so it doesn't post `modoptions[closethread]`), `$modoptions['closethread']` gets set to what the thread already has.

https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L889-L892

Stefan-ST:
This fix is incorrect.

PR #2416, opened 2016-06-05
Fixes #2050 Remove HTML in parser

### LINKED ISSUE #2050: Remove HTML in parser
I think the HTML inside the parse would be better in templates.

PaulBender:
Fixes #2050

- Change the strike MyCode ([s]) HTML replacement from `<del>` to `<span style="text-decoration: line-through;">`
- Added classes to all default MyCodes
- Moved HTML for the URL, email, size and image MyCodes to templates
- Fixed issue with size tag taking negative values
- Added a MyCode template group


Sama34:
`<s>` is a HTML5 valid tag though.

Surprised to see we went for a "lets make everything a `<span>`" approach.

Destroy666x:
Being a valid tag doesn't mean being properly semantic. `<del>` and `<s>` have different meanings, so  preferring one over another in content provided by users will often end up with incorrect usage. Basically the same thing as: http://stackoverflow.com/a/4939839 which a lot of users complain about in MyBB (`<strong>` used in an incorrect manner, you don't insert it everywhere just because it boldens text...).

PaulBender:
I've made a few alterations to the size tag with int values.

- I moved the HTML to template (meant to do it but overlooked it)
- I removed a plus 10 that was being added to the size for no apparent reason
- Added a minimum size check (it won't accept any value smaller than 1)

These changes don't effect the normal size tag.

Sama34:
I would expect the `s` MyCode to insert a `s` HTML tag though, I'm just surprised most people expect a `span` tag for `s` , `b`, and `u` MyCode instead but choice to keep the `hr` one as a `hr`tag, shouldn't it be a `div` formatted to behave like a `hr` instead?

Why aren't yo making use of the `$templates->render()` method though? Was it decided not to be used in the core at all?

The `blockquote` class seems useful.

Destroy666x:
`<hr>` has a clear semantic meaning, so I'm not sure why a `<div>` would be a better replacement. `<s>` and `<del>` have both different meanings so why would you use `<s>` over `<del>` in all usages in posts? Just because we named the MyCode with the same letter? That's not how semantics works.

PaulBender:
Fixed merge conflicts

Issue #1729, opened 2014-12-29
Outdated CHMOD wiki link (and more docs.mybb.com links)


Destroy666x:
http://community.mybb.com/thread-164820-post-1128055.html#pid1128055

The link here is wrong:
https://github.com/mybb/mybb/blob/feature/admin/modules/tools/system_health.php#L915

EDIT: actually, there are many more links to update if you grep `docs.mybb.com`. And some docs are not even available (they need to be worked on but noone has decided to take care of them yet)... So moving to 1.8.5.

EDIT2: I listem them in https://github.com/mybb/docs.mybb.com/issues/61

andrewjs18:
some links have been updated per this PR: https://github.com/mybb/mybb/pull/2453

andrewjs18:
this should be all fixed now.

closing!

PR #2421, opened 2016-06-14
Send users who click "chmod" somewhere helpful

GeorgeLVP:
http://docs.mybb.com/CHMOD_Files.html is dead


Destroy666x:
https://github.com/mybb/mybb/issues/1729

There's also a link in ACP, would be nice if you fixed it too.

euantorano:
@GeorgeLVP could you possibly change the link in the ACP? If not, I'll do it and merge this.

PR #2423, opened 2016-06-14
Fix #2363 per theme default avatar

### LINKED ISSUE #2363: Per theme default avatar

Destroy666x:
I didn't include `$change_dir` because it doesn't seem to be implemented
correctly anyways (I'll create an issue later).

https://github.com/mybb/mybb/issues/2363

Hope it gets merged before other conflicts arise.

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -992,7 +992,7 @@ EOF;
 	}
 	else
 	{
-		$user['avatar'] = "../".$mybb->settings['useravatar'];
+		$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
 		$scaled_dimensions = array(
 			"width" => 120,
 			"height" => 120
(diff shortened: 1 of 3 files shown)


admin/modules/user/users.php
```diff
@@ -3596,7 +3596,7 @@ function build_users_view($view)
 			}
 			if(!$user['avatar'])
 			{
-				$user['avatar'] = "../".$mybb->settings['useravatar'];
+				$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
```
dequeues: {theme} here isn't actually the theme, it's `images`. I'm sure that isn't intended as per title

admin/modules/user/users.php
```diff
@@ -992,7 +992,7 @@
 	}
 	else
 	{
-		$user['avatar'] = "../".$mybb->settings['useravatar'];
+		$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
```
dequeues: {theme} here isn't actually the theme, it's `images`. I'm sure that isn't intended as per title

Destroy666x: That's how everything works... Check smilies, post icons, etc. You need to point to a path and `$theme` is not available in ACP.

dequeues: But that's going to show a 404 in the Users in ACP. That is going to give ../images/images/default_avatar.png

dequeues: Nevermind, I took it the wrong way

PR #2425, opened 2016-06-14
Fixes #1631 username's should be htmlspecialchars_uni()'d

### LINKED ISSUE #1631: username's should be htmlspecialchars_uni()'d
#1304

Functions that need revision:
- [ ] format_name
- [ ] build_profile_link
- [ ] log_admin_action
- [ ] log_moderation_action or whatever

Special attention to:
- [ ] parser
- [ ] build_postbit
- [ ] templates
- [ ] cache class

PaulBender:
Fixes #1631

Fixed remaining issues and locations that were missed originally.

### DIFF
--- a/admin/modules/user/banning.php
+++ b/admin/modules/user/banning.php
@@ -99,7 +99,7 @@ if($mybb->input['action'] == "prune")
 		$cache->update_reportedcontent();
 
 		// Log admin action
-		log_admin_action($user['uid'], $user['username']);
+		log_admin_action($user['uid'], htmlspecialchars_uni($user['username']));
 
 		flash_message($lang->success_pruned, 'success');
 		admin_redirect("index.php?module=user-banning");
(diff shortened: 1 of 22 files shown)


admin/modules/user/banning.php
```diff
@@ -99,7 +99,7 @@
 		$cache->update_reportedcontent();
 
 		// Log admin action
-		log_admin_action($user['uid'], $user['username']);
+		log_admin_action($user['uid'], htmlspecialchars_uni($user['username']));
```
Sama34: I think the log page sanitizes this already. If not, then it should.

dequeues: It does indeed. With that removed it should be fine.

dequeues:
Something I forgot to initially fix and commit after I merged is that it needs to be removed from https://github.com/mybb/mybb/blob/feature/admin/modules/user/users.php#L3504

PR #2429, opened 2016-06-26
Fixes #2022 Close Thread doesn't work via reply

### LINKED ISSUE #2022: Close Thread doesn't work via reply

Stefan-ST:
Fixes #2022


PaulBender:
Works for me

PR #2430, opened 2016-06-26
Fixes #2427 Admin interface issues on IPv6

### LINKED ISSUE #2427: Admin interface issues on IPv6
The default settings do not appear to allow access to the admin interface via IPv6. Setting ADMIN_IP_SEGMENTS in admin/index.php to 0 fixes this.

It might make sense to either change the default value or modify the check in a way that makes it also accept IPv6 addresses.

Stefan-ST:
Fixes #2427

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -11,8 +11,9 @@
 define("IN_MYBB", 1);
 define("IN_ADMINCP", 1);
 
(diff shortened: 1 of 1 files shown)


PaulBender:
Can't test IPv6, but IPv4 works fine.

euantorano:
Appears to work for me, only been able to test IPv6 at work though.

PR #2432, opened 2016-06-26
Fixes #2431 Default avatar broken on ACP when using a full URL

### LINKED ISSUE #2431: Default avatar broken on ACP when using a full URL
The code doesn't take into account that the path to the default avatar can be a full URL, too:
```php
		$user['avatar'] = "../".$mybb->settings['useravatar'];
```

Stefan-ST:
Fixes #2431

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -994,11 +994,11 @@ EOF;
 	{
 		if(my_substr($mybb->settings['useravatar'], 0, 7) === 'http://' || my_substr($mybb->settings['useravatar'], 0, 8) === 'https://')
 		{
-			$user['avatar'] = $mybb->settings['useravatar'];
+			$user['avatar'] = str_replace('{theme}', 'images', $mybb->settings['useravatar'])
 		}
 		else
 		{
-			$user['avatar'] = "../".$mybb->settings['useravatar'];
+			$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
 		}
 		$scaled_dimensions = array(
 			"width" => 120,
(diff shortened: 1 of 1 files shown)


admin/modules/user/users.php
```diff
@@ -992,7 +992,14 @@
 	}
 	else
 	{
-		$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
+		if(my_substr($mybb->settings['useravatar'], 0, 7) === 'http://' || my_substr($mybb->settings['useravatar'], 0, 8) === 'https://')
```
Destroy666x: Good catch, but any reason to use `my_substr()` for comparison with a plain ASCII string?

Stefan-ST: I just copied the existing code a few lines above. ;)

admin/modules/user/users.php
```diff
@@ -992,7 +992,14 @@
 	}
 	else
 	{
-		$user['avatar'] = "../".str_replace('{theme}', 'images', $mybb->settings['useravatar']);
```
dequeues: Why did you remove this? see #2363

admin/modules/user/users.php
Destroy666x: Copying doesn't explain why it's correct. But it'll be changed to the new URL function soon so doesn't matter.

admin/modules/user/users.php
Stefan-ST: Sorry, fixed.

admin/modules/user/users.php
```diff
@@ -992,7 +992,14 @@
...
+		if(my_substr($mybb->settings['useravatar'], 0, 7) === 'http://' || my_substr($mybb->settings['useravatar'], 0, 8) === 'https://')
+		{
+			$user['avatar'] = str_replace('{theme}', 'images', $mybb->settings['useravatar'])
```
dequeues: Missing `;`

Stefan-ST: Fixed, thanks.

dequeues:
[my_validate_url()](https://github.com/mybb/mybb/blob/bcb8bd924ea6cb1a309803cf8ce730782bad70f3/inc/functions.php#L8429) can be used now rather than my_substr() for URL

PR #2448, opened 2016-07-15
Fixes #2256 No smilies/post icons

### LINKED ISSUE #2256: No smilies/post icons
If all smilies are removed in ACP -> Configuration -> Smilies:
- [x] `misc.php?action=smilies` displays a weird table with no rows (it could show something like `There are currently no smilies available` instead), `misc.php?action=smilies&popup=true&editor=MyBBEditor&modal=1` modal too
- [x] SCEditor emoticon button is displayed, it could be hidden
- [x] post icons also are displayed when all of them are removed

PaulBender:
Fixes #2256

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1805,6 +1805,10 @@ function get_post_icons()
 	{
 		eval("\$posticons = \"".$templates->get("posticons")."\";");
 	}
+	else
+	{
+		$posticons = '';
+	}
 
 	return $posticons;
 }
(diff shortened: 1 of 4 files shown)


inc/functions.php
```diff
@@ -1801,7 +1801,10 @@ function get_post_icons()
 		eval("\$iconlist .= \"".$templates->get("posticons_icon")."\";");
 	}
 
-	eval("\$posticons = \"".$templates->get("posticons")."\";");
+	if(!empty($iconlist))
```
dequeues: We should add an else statement to set `$posticons` to an empty string or such, otherwise it just shows `Array`.

PaulBender: I added an else statement.

PR #2457, opened 2016-08-06
Fixes #2456 Hash instead of the password at the "forgot my password"

### LINKED ISSUE #2456: Hash instead of the password at the "forgot my password"
Hash instead of the password at the "forgot my password". I did not change in any way member.php file in the email displays the password hash added to the database instead of the decrypted password to login.

zrzut ekranu z 2016-08-06 10-17-30

PaulBender:
Fixes #2456

### DIFF
--- a/member.php
+++ b/member.php
@@ -1621,13 +1621,15 @@ if($mybb->input['action'] == "resetpassword")
 		require_once MYBB_ROOT.'inc/datahandlers/user.php';
 		$userhandler = new UserDataHandler('update');
 
+		$password = random_str($password_length, $mybb->settings['requirecomplexpasswords']);
+
 		while(!$userhandler->verify_password())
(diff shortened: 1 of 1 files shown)


Sama34:
I think this reverts the purpose of the while loop. @Destroy666x

Ben-Conway:
@JN-Jones said he tested it and it works okay.

Destroy666x:
Yeah, it needs to be in the loop, otherwise it may fail because of a datahandler error.

PaulBender:
Okay, I've moved it.

Sama34:
Somebody mentioned the issue being the following line:
https://github.com/mybb/mybb/blob/feature/member.php#L1641

But I checked and apart from the misleading variable name it should work, yeah.

PR #2466, opened 2016-09-05
gender neutral pronoun

neko:
i'm not a feminist or anything, but this always annoys me whenever i ban someone.


euantorano:
Fair enough, makes sense to me.

PR #2480, opened 2016-10-22
Fixes #2065 Check files returns /install missing

### LINKED ISSUE #2065: Check files returns /install missing
When you validate files using built in tool it tells you that /install folder and related files are missing. Obviously I was told during installation process to remove whole folder for security purposes. I know this is trival, but it is always worth of letting you know.

Cheers

Original thread: [Check files returns /install missing](http://community.mybb.com/thread-166171.html)

PaulBender:
Fixes #2065; will skip the install directory and the hello world plugin file if they don't exist.

### DIFF
--- a/admin/modules/tools/file_verification.php
+++ b/admin/modules/tools/file_verification.php
@@ -76,7 +76,7 @@ if(!$mybb->input['action'])
 				continue;
 			}
 
-			if(!is_dir(MYBB_ROOT."install/") && substr($parts[1], 0, 9) == "./install")
+			if(!is_dir(MYBB_ROOT."install/") && substr($parts[1], 0, 10) == "./install/")
 			{
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


admin/modules/tools/file_verification.php
```diff
@@ -71,6 +71,16 @@
 				}
 			}
 
+			if(!file_exists(MYBB_ROOT."inc/plugins/hello.php") && $parts[1] == "./inc/plugins/hello.php")
+			{
+				continue;
+			}
+
+			if(!is_dir(MYBB_ROOT."install/") && substr($parts[1], 0, 9) == "./install")
```
Devilshakerz: Can you extend the second condition by a trailing slash to make sure it's the directory?

PaulBender: Done

PR #2486, opened 2016-10-25
Fixes #1569 "Email Verification & Administrator Activation" usless

### LINKED ISSUE #1569: "Email Verification & Administrator Activation" usless
The Registration Method setting "Email Verification & Administrator Activation" is useless right now because the administrator is unable to determine if the user has verified his email or not.

PaulBender:
Fixes #1569 

Went with a dedicated page for users awaiting activation based on my plugin [Activate Users From Mod CP](https://github.com/PaulBender/Activate-Users-From-Mod-CP)

### DIFF
--- a/admin/modules/user/awaiting_activation.php
+++ b/admin/modules/user/awaiting_activation.php
@@ -1,0 +1,205 @@
+<?php
+/**
+ * MyBB 1.8
+ * Copyright 2014 MyBB Group, All Rights Reserved
+ *
+ * Website: http://www.mybb.com
+ * License: http://www.mybb.com/about/license
+ *
+ */
+
+// Disallow direct access to this file for security reasons
+if(!defined("IN_MYBB"))
+{
+	die("Direct initialization of this file is not allowed.<br /><br />Please make sure IN_MYBB is defined.");
+}
+
+$page->add_breadcrumb_item($lang->awaiting_activation, "index.php?module=user-awaiting_activation");
+
+$sub_tabs['awaiting_activation'] = array(
+	'title' => $lang->awaiting_activation,
+	'link' => "index.php?module=user-awaiting_activation",
+	'description' => $lang->awaiting_activation_desc
+);
+
+$plugins->run_hooks("admin_user_awaiting_activation_begin");
+
+if($mybb->input['action'] == "activate" && $mybb->request_method == "post")
+{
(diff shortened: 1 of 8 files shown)


JN-Jones:
From a quick look: How do we handle it if a user has the awaiting activation group as secondary usergroup? Your page only checks for the primary usergroup so that would be one possible error (though I haven't tested it yet, just looked at the code).

PaulBender:
We don't do anything, all current coding related to activation depends on the primary usergroup.

JN-Jones:
Ok, just wanted to confirm that

admin/modules/user/awaiting_activation.php
```diff
@@ -0,0 +1,205 @@
...
+		require_once MYBB_ROOT.'inc/datahandlers/user.php';
+		$userhandler = new UserDataHandler('delete');
+
+		$query = $db->simple_select("users", "uid, usergroup", "uid IN ({$user_ids})");
+		while($user = $db->fetch_array($query))
+		{
+			if($user['usergroup'] == 5)
+			{
+				++$num_deleted;
+				$userhandler->delete_user($user['uid'], 1);
```
JN-Jones: Probably instead of deleting every user in a separate call you should save them in an array and pass that array to `delete_user`. This would reduce the number of queries here.

PaulBender: I've changed it.

JN-Jones:

JN-Jones:
What I've seen so far:

* The notification on the frontend still links to the users listing with a filter for the group. Probably should link to the new page instead
* Using the Email and Admin activation method, activating a user before he validates his email also validates his email. Not sure whether this is something that can be fixed easily but as it's nowhere documented that you need to wait until the user has activated his email before activating the user it looks like a bug to me.

PaulBender:
I've changed the Admin CP link.

As for the Email/Admin method, I'm not sure of the best way to fix that.

JN-Jones:
@mybb/developers @mybb/sqa As I would like to have this in 1.8.9 has anybody any suggestions?

JN-Jones:
@PaulBender what would you think about a simple notice somewhere on that page saying something like "if you activate a user who hasn't verified his email yet he won't need to do so anymore". Maybe combined with a js link to select all users requiring admin activation. That way the behavior is clearly documented without introducing any weird additional checks.

PR #2489, opened 2016-10-29
Fixes #1487 $lang->guest is saved as guest usernames

### LINKED ISSUE #1487: $lang->guest is saved as guests usernames
If an guests posts and not username is specified `$lang->guest` is used and saved to the DB. No username should be saved and $lang->guest used only in post build.

JordanMussi:
### Fixes #1487 $lang->guest is saved as guest usernames
- New Reply & New Thread accept empty username and insert as empty to the database
- Guests can't edit posts (even when user settings allows) so edit post doesn't need chnages
- Username outputs use translatable guest string when username is empty:
  - Post Bit, Forum Display and Forum Last Post updated
  - Portal & Search updated
  - Thread Subscriptions:
    - User CP, Email & PM updated
  - Forum Subscriptions:
    - User CP & Email updated
  - Moderation Queue:
    - Mod CP already working
    - Admin CP updated
  - Admin CP attachments already working
    - Implimented searching by guest usernames (and options to search for guests & users or users only or guests only)
  - Mod CP Logs won't ever show a username for a poster so already working
  - Mod CP IP search already working
  - Syndaction does not show username
  - Archive Mode & Print Thread already working
- When quoting a post from a guest with no username specified the username part will be empty and so won't show a username (not even the guest translation) on output

Also fixed issue where Username is empty for deleted moderators in the Mod CP Mod Logs and implemented the same "N/A - Been Deleted" functionality (for rows and the filter) in the Admin CP Mod Logs

---

**This PR:**
- changes 5 language files


premiumcuyuz:
very good job

JN-Jones:
From the basic changes this looks fine, however there are several places where normal usernames are escaped while `$lang->guest` isn't. Same for `na_deleted` as far as I can tell. I haven't checked all those places whether it's escaped later on but as some of them build a profile link for normal usernames I doubt that. @JordanMussi can you take a look at that and comment if I'm wrong? Apart from that I still need to test it live but from the changes I'd say it should work.

JN-Jones:
@JordanMussi Any update on the comment above? Also there's a conflict in the upgrade file so would be nice if you could fix that.

JordanMussi:
@JN-Jones I have ensured each is `htmlspecialchars_uni`'d and resolved the merge conflict.

euantorano:
@JordanMussi: Jones has since left the team, but somebody else will review this (I'll try to if I get time at some point).

JordanMussi:
Resolved conflicts. 😃

euantorano:
Thanks, I'll give this a look over this weekend.

JordanMussi:
Awesome! Thanks @euantorano 😀

euantorano:
This has been sitting around for a while now 😉

PR #2490, opened 2016-10-29
Fixes #2488 Report Reasons not sent in report PM/email notifications

### LINKED ISSUE #2488: Report Reasons not sent in report PM/email notifications
https://community.mybb.com/thread-205425-post-1250421.html#pid1250421

The reason `title` should be fetched from the `reportreasons` table and the `reason` field should also be included for reasons that require additional information. This should then be inserted [here](https://github.com/mybb/mybb/blob/mybb_1808/inc/functions_modcp.php#L213).

---

While confirming this issue I have also noticed that the `reportmethod` setting should probably be clarified as to what it does.
Old Title: `Reported Posts Medium`
Old Description: `Please select from the list how you want reported posts to be dealt with. Storing them in the database is probably the better of the options listed.`
Old optionscode:

```
radio
db=Stored in the Database
pms=Sent as Private Messages
email=Sent via Email
```

New Title: `Reported Content Notification Method`
New Description: `Please select from the list how you would like moderators to be notified when content is reported. Note: Moderators are only notified for posts in forums they moderate or all content if they are a super moderator.`
New optionscode:

```
radio
db=Shown in the Mod CP only
pms=Sent as Private Messages and shown in the Mod CP
email=Sent via Email and shown in the Mod CP
```
#1223

JordanMussi:

### DIFF
--- a/inc/functions_modcp.php
+++ b/inc/functions_modcp.php
@@ -145,6 +145,22 @@ function send_report($report, $report_type='post')
 {
 	global $db, $lang, $forum, $mybb, $post, $thread, $reputation, $user;
 
+	$report_reason = '';
+	if($report['reasonid'])
+	{
+		$query = $db->simple_select("reportreasons", "title", "rid = '{$report['reasonid']}'", array('limit' => 1));
+		$reason = $db->fetch_array($query);
+
+		$lang->load('report');
+
+		$report_reason = $lang->parse($reason['title']);
+	}
+
+	if($report['reason'])
+	{
+		$report_reason = $lang->sprintf($lang->email_report_comment_extra, $report_reason, $report['reason']);
+	}
+
 	$modsjoin = $modswhere = '';
 	if(!empty($forum['parentlist']))
 	{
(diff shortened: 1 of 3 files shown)


JN-Jones:
Looks fine to me, however would be nice if someone else takes a second look @mybb/sqa

inc/functions_modcp.php
```diff
@@ -145,6 +145,22 @@ function send_report($report, $report_type='post')
 {
 	global $db, $lang, $forum, $mybb, $post, $thread, $reputation, $user;
 
+	$report_reason = '';
+	if($report['reasonid'])
+	{
+		$query = $db->simple_select("reportreasons", "title", "rid = '{$report['reasonid']}'", array('limit' => 1));
```
Devilshakerz: Can you cast `$report['reasonid']` to integer here? It comes from an `int` field in the database, but it's cast nonetheless in other places.

JN-Jones: Though I'm against useless casts I'd prefer it here too as it's a function which may be used by plugins

JN-Jones: @JordanMussi Can you change this?

dequeues: If Jordan is not available we could just push this and then do another commit to cast it as int.

Ben-Conway: @JN-Jones Can we get this looked at and possibly in 1.8.9?

JN-Jones: Apart from this it looked fine before, need to retest though

JN-Jones:
Seems to work fine, creating a new PR with the fix for the issue mentioned above.

PR #2492, opened 2016-11-07
Fix #1609 forum deletion

### LINKED ISSUE #1609: Incomplete forum deletion

JN-Jones:
#1609 (old PR: #1920)

Fixed the smaller issues mentioned in the previous PR. Only pagination is still missing. @Destroy666x

### DIFF
--- a/admin/modules/forum/management.php
+++ b/admin/modules/forum/management.php
@@ -1950,9 +1950,6 @@ if($mybb->input['action'] == "delete")
 		$fid = $mybb->get_input('fid', MyBB::INPUT_INT);
 		$forum_info = get_forum($fid);
 
-		$query = $db->simple_select("forums", "posts,unapprovedposts,threads,unapprovedthreads", "fid='{$fid}'");
-		$stats = $db->fetch_array($query);
-
 		// Delete the forum
 		$db->delete_query("forums", "fid='$fid'");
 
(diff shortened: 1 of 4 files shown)


JN-Jones:
Added pagination. It always deletes 50 threads, if there are still threads to delete redirects back to the same page, otherwise deletes other stuff. @mybb/sqa

JN-Jones:

JN-Jones:
Can somebody please check this?

Ben-Conway:
This looks good to me.

Cu8eeeR:
My php knowledge is not great as your guys but I didnt find anything wrong on the first look. Still should be reviewed by someone more skilled than me.

euantorano:
Looks good to me too 👍

PR #2493, opened 2016-11-08
Fix #674 Line breaks in templates search fails

### LINKED ISSUE #674: Line breaks in templates search fails
....epically :tongue:

If you search templates using a string that contains a line break even if a template with this string exists it will not be returned. *sad admin*

For example searching for:

``` html
</head>
<body>
```

Does not return the `index` template.

1.8 acts differently. When searching with an indent it fails. The following should return `memberlist_user`

``` html
    {$usergroup['groupimage']}
    {$user['userstars']}
```

http://community.mybb.com/thread-153845.html

Devilshakerz:
#674

### DIFF
--- a/admin/modules/style/templates.php
+++ b/admin/modules/style/templates.php
@@ -879,13 +879,16 @@ if($mybb->input['action'] == "search_replace")
 					$template_sets[$set['sid']] = $set['title'];
 				}
 
+				$search_string = str_replace(array("\r\n", "\r"), "\n", $mybb->input['find']);
+				$search_string = str_replace(array("  ", "\t"), "", $search_string);
+
 				// Select all templates with that search term
 				$query = $db->query("
 					SELECT t.tid, t.title, t.sid, t.template
 					FROM ".TABLE_PREFIX."templates t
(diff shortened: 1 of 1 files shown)


JN-Jones:
As we're not forcing `\n` in templates this doesn't fix the error entirely. While most of the default templates have `\n` as line ending, it may change after editing it so instead we should search for all possible line endings. I'm not yet sure whether we need to consider mixed ones (in case of search string with multiple line endings) but simply searching for `\n` doesn't work.

Devilshakerz:
We might use a `select` query with `REPLACE()` and `CHR()` / `CHAR()` instead, so the line end characters are consistent on both sides - the performance would be decreased at the expense of compatibility (the function usually isn't used often though).

JN-Jones:
I guess the performance wouldn't matter that much in this case.

JN-Jones:
@Devilshakerz Can you change this so we can get it into 1.8.9?

JN-Jones:

JN-Jones:
Looks good, merging

PR #2494, opened 2016-11-08
Fix #511 Calendar, Week Overview: Week period calculated wrong

### LINKED ISSUE #511: Calendar, Week Overview: Week period calculated wrong
calendar

As you can see, it tells me that the week has more than 2000 years :P (Note that the fix for #346 was applied on this board)

Devilshakerz:
#511

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -20,6 +20,7 @@ $templatelist .= ",calendar_weekrow_day_birthdays,calendar_weekview_day_birthday
 require_once "./global.php";
 require_once MYBB_ROOT."inc/functions_calendar.php";
 require_once MYBB_ROOT."inc/functions_post.php";
+require_once MYBB_ROOT."inc/functions_time.php";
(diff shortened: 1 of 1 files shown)


JN-Jones:
This *should* work. However have you checked whether this (or a similar issue) may happen where we use `gmmktime`? Otherwise I'll make a short test about this tomorrow @Devilshakerz

Devilshakerz:
It is used in other places but the dates there don't reach that far back, so it should be fine.

JN-Jones:

JN-Jones:
As mentioned on slack: As we're casting the `week` parameter as int there are other issues on at least 32 bit systems. Eg my calendar only can go back till the 13.12.1901 as an integer can't go below that. However we still show links to go to the week before. Instead of simply casting it quietly we should either show a proper warning or hide the links. Note that I can't test the original error due to that limitation.

Ben-Conway:
This looks good then.

PR #2499, opened 2016-11-08
Create hookable functions for password hashing

### LINKED ISSUE #2498: No hooks available in password hashing functions
The lack of plugin hooks and hardcoded `md5` pre-hashing make it impossible to hijack the hashing process using plugins without modifying the core.

Devilshakerz:
Fixes #2498

 - removed hardcoded `md5()` pre-hashing,
 - `salt_password()` marked as deprecated,
 - `create_password_hash()` and `verify_user_password()` added with new hooks, accepting the user row and raw password,
 - complete user row fetched on login attempt,
 - `my_hash_equals()` added, performing comparison using `hash_equals()` on PHP >= 5.6 or custom implementation

Session keys comparison should be also rewritten to use `my_hash_equals()` in the future.

Proof-of-concept plugin taking advantage of those changes available at: https://github.com/Devilshakerz/mybb-dvzHash

### DIFF
--- a/inc/functions_user.php
+++ b/inc/functions_user.php
@@ -56,7 +56,7 @@ function validate_password_from_username($username, $password)
 	global $mybb;
 
 	$options = array(
-		'fields' => array('username', 'password', 'salt', 'loginkey', 'coppauser', 'usergroup'),
+		'fields' => '*',
 		'username_method' => $mybb->settings['username_method'],
 	);
 
@@ -87,14 +87,13 @@ function validate_password_from_uid($uid, $password, $user = array())
 	}
 	if(!$user['password'])
 	{
-		$query = $db->simple_select("users", "uid,username,password,salt,loginkey,usergroup", "uid='".(int)$uid."'");
-		$user = $db->fetch_array($query);
+		$user = get_user($uid);
 	}
 	if(!$user['salt'])
 	{
 		// Generate a salt for this user and assume the password stored in db is a plain md5 password
 		$user['salt'] = generate_salt();
-		$user['password'] = salt_password($user['password'], $user['salt']);
+		$user['password'] = create_password_hash($user['password'], $user['salt'], $user);
 		$sql_array = array(
 			"salt" => $user['salt'],
 			"password" => $user['password']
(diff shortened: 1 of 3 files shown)


inc/functions_user.php
```diff
@@ -173,13 +172,107 @@ function update_password($uid, $password, $salt="")
...
+	$parameters = compact('password', 'salt', 'user');
+
+	if(!defined('IN_INSTALL') && !defined('IN_UPGRADE'))
+	{
+		$plugins->run_hooks('create_password_hash', $parameters);
+	}
+
+	if(is_string($parameters))
+	{
+		return $parameters;
```
JN-Jones: I guess this is for plugins who validated the password? Why not use a normal return within the plugin call using a new variable? IMHO that would look cleaner

inc/functions_user.php
```diff
@@ -173,13 +172,107 @@ function update_password($uid, $password, $salt="")
...
+	$parameters = compact('user', 'password');
+
+	if(!defined('IN_INSTALL') && !defined('IN_UPGRADE'))
+	{
+		$plugins->run_hooks('verify_user_password', $parameters);
+	}
+
+	if(is_bool($parameters))
+	{
+		return $parameters;
```
JN-Jones: Same here

inc/functions_user.php
Devilshakerz: The plugin class doesn't handle hook return values well: https://github.com/mybb/mybb/blob/mybb_1808/inc/class_plugins.php#L142-L145
If a custom function returns `false`, the hook outputs the original `$arguments` variable - this leads to messy implementations one way or another.

JN-Jones: I still don't see why we shouldn't use the return value:

```php
$plugin_validation = false;

$plugin_validation = $plugins->run_hook(...);

if(is_string($plugin_validation))
{
    return $plugin_validation;
}
```

Even if we stay with using the parameter as return variable I'd much more prefer to have a seperate index for the return value (eg `$parameters['hash']`) as overwriting the normal parameters is something I as a plugin author would never guess without looking at the code. It simply seems wrong.

euantorano: Yeah, I'd find overwriting the parameters weird personally.

euantorano:
I understand that, but modifying it as a plugin author isn't the obvious thing to do. 

I think it'll be fine here though, as it's not like there will be hundreds of plugins using these hooks or anything (like global_start, etc.).

JN-Jones:
Ok, I see that it wouldn't work with the second hook.  However I still think that an additional index would be better than overwriting the whole parameter. @Devilshakerz @euantorano

PR #2505, opened 2016-11-10
Fixes #302 Performance Optimization for the function replace_query

### LINKED ISSUE #302: Performance Optimization for the function replace_query
> replace_query() does a loop which is not necessary...
> 
> here is an optimized version of db_pgsql.php:
> 
>  @@ -1223,15 +1230,12 @@ } else { - $query = $this->write_query("SELECT {$main_field} FROM {$this->table_prefix}{$table}"); - - while($column = $this->fetch_array($query)) + $search_bit = "{$main_field} = '".$replacements[$main_field]."'"; + $query = $this->write_query("SELECT COUNT(".$main_field.") as count FROM {$this->table_prefix}{$table} WHERE {$search_bit} LIMIT 1"); + + if($this->fetch_field($query, "count") >= 1) { - if($column[$main_field] == $replacements[$main_field]) - { - $update = true; - break; - } + $update = true; } }

Original report: http://dev.mybb.com/issues/2174
Reporter: Can Özdemir
Status: New

JN-Jones:
Replaced the function with the modification mentioned in #302. Note that I haven't been able to test the new function on a board but based on the code it should work.


euantorano:
This will definitely need some testing, but I'm away from home at the minute - can anybody else test this please?

Ben-Conway:
@mybb/sqa can we get this tested?

PaulBender:
It's not working for me, getting this SQL error:

`MyBB has experienced an internal SQL error and cannot continue.

SQL Error:
    0 - ERROR: syntax error at or near "=" LINE 2: SET fid=2, uid='1', dateline=1481233223 WHERE Array='' ^ 
Query:
    UPDATE mybb_forumsread SET fid=2, uid='1', dateline=1481233223 WHERE Array='' `

All I did was view a forum.

inc/db_pgsql.php
```diff
@@ -1321,51 +1321,33 @@ function replace_query($table, $replacements=array(), $default_field="", $insert
...
 		{
-			if(is_array($main_field))
-			{
-				return $this->update_query($table, $replacements, $search_bit);
-			}
-			else
-			{
-				return $this->update_query($table, $replacements, "{$main_field}=".$this->quote_val($replacements[$main_field]));
-			}
+			return $this->update_query($table, $replacements, "{$main_field}=".$this->quote_val($replacements[$main_field]));
```
Shade-: `$main_field` is an array. Should be:

`return $this->update_query($table, $replacements, "{$main_field[0]}=".$this->quote_val($replacements[$main_field[0]]));`

JN-Jones:
@PaulBender @Shade- Can we get this retested? Don't have a working pgsql copy myself atm.

inc/db_pgsql.php
Shade-: Yep, didn't noticed it :)

Shade-:
I don't have it either, just debugged directly looking at the code. Ball passed to @PaulBender.

PaulBender:
Looks good.

PR #2508, opened 2016-11-10
Fixes #2047 Hello World plugin fails on SQLite and most likely also p…

### LINKED ISSUE #2047: Hello World plugin fails on SQLite and most likely also pgsql
Not good for an example plugin: http://community.mybb.com/thread-172178.html It should be compatible with our minimal requirements. I guess only the table creation query needs to be corrected,

PaulBender:
…gsql

Fixes #2047

### DIFF
--- a/inc/plugins/hello.php
+++ b/inc/plugins/hello.php
@@ -354,7 +354,7 @@ function hello_install()
 {
 	global $db;
 
-	// Create our entries table
+	// Create our table collation
 	$collation = $db->build_create_table_collation();
 
 	// Create table if it doesn't exist already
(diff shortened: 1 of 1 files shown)


euantorano:
Sneaked that in whilst I was commenting there ;) Does uninstallation work correctly, there's some weird looking code relating to template/templateset deletion there.

PaulBender:
Uninstalling works fine, but the code could be simplified.

euantorano:
Would you have time to do that? If not, we can leave it as is if it's not causing problems.

PaulBender:
Yeah, I can do it.

PR #2510, opened 2016-11-11
Fixes #2509 by adding the old if statement from 1.8.7

### LINKED ISSUE #2509: Signature preview saves the signature?
Just wanted to note, that upon clicking the 'Preview Signature' button, it saves the signature already, using MyBB 1.8.8 I've checked the source code, and the button code for the preview is like this:

```
<input type="submit" class="button" name="submit" value="Update Signature"/>
<input type="submit" class="button" name="preview" value="Preview Signature"/>
```

JN-Jones:
#2509


PaulBender:
Works for me

PR #2512, opened 2016-11-11
Fixes #2066 by altering the form class to allow "0" as selected value

### LINKED ISSUE #2066: Moderation Tools prefix issue
Crap the subject.

Go to ACP -&gt; Configuration -&gt; Moderator Tools -&gt; Add New Thread Tool
Name: (Empty)
Soft delete/restore thread?: Restore
Apply thread prefix?: No prefix

Click "Save Thread Tool". Proper in-line error is shown, but the value for "Apply thread prefix?" is missing.

Original thread: [Moderation Tools prefix issue](http://community.mybb.com/thread-163254.html)

JN-Jones:
#2066 

This issue is not related to the mod tools itself but to the fact that we check `$selected` via `empty` where PHP interprets `0` as empty. This makes sure that `$selected` is really an empty array otherwise we check the proper value which allows to select rows with `$value == 0`.

### DIFF
--- a/admin/inc/class_form.php
+++ b/admin/inc/class_form.php
@@ -422,7 +422,7 @@ class DefaultForm
 		foreach($option_list as $value => $option)
 		{
 			$select_add = '';
-			if(!empty($selected) && ((string)$value == (string)$selected || (is_array($selected) && in_array((string)$value, $selected))))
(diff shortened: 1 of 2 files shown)


PaulBender:
Looks good to me

PR #2521, opened 2016-11-21
Fix #2516 - Last post not updated when approving the first post using inline mod tools

### LINKED ISSUE #2516: Last post not updated when approving the first post using inline mod tools
Last post data is not updated accordingly when you post a reply to an unapproved thread and subsequently approve the first post of the thread by using inline moderation tools.

### Steps to reproduce
- Post a thread and unapprove the thread;
- Post a reply to the unapproved thread;
- Approve the first post using inline moderation tools.

Shade-:
#2516

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -1228,9 +1228,10 @@ class PostDataHandler extends DataHandler
 			}
 
 			$thread_update = array('replies' => '+1');
-			
+
 			// Update counters
 			update_forum_counters($post['fid'], array("posts" => "+1"));
+			update_forum_lastpost($thread['fid']);
 		}
 		// Post is stuck in moderation queue
 		else if($visible == 0)
@@ -1252,11 +1253,11 @@ class PostDataHandler extends DataHandler
 			$thread_update = array('replies' => '+1');
 			update_forum_counters($post['fid'], array("deletedposts" => "+1"));
 		}
-		
+
 		// Update forum count
-		if($visible == 1) {
+		if($visible == 1)
+		{
 			update_last_post($post['tid']);
-			update_forum_lastpost($thread['fid']);
 		}
 
 		$query = $db->simple_select("attachments", "COUNT(aid) AS attachmentcount", "pid='{$this->pid}' AND visible='1'");
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -1026,7 +1026,7 @@ class PostDataHandler extends DataHandler
 			$update_array = array(
 				'lastpost' => "'{$now}'"
 			);
-			if($forum['usepostcounts'] != 0)
+			if($forum['usepostcounts'] != 0 && $thread['visible'] == 1)
 			{
 				$update_array['postnum'] = 'postnum+1';
 			}
@@ -1228,6 +1228,9 @@ class PostDataHandler extends DataHandler
 			}
 
 			$thread_update = array('replies' => '+1');
+			
+			// Update counters
+			update_forum_counters($post['fid'], array("posts" => "+1"));
 		}
 		// Post is stuck in moderation queue
 		else if($visible == 0)
@@ -1253,7 +1256,6 @@ class PostDataHandler extends DataHandler
 		// Update forum count
 		if($visible == 1) {
 			update_last_post($post['tid']);
-			update_forum_counters($post['fid'], array("posts" => "+1"));
 			update_forum_lastpost($thread['fid']);
 		}
 
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -1018,7 +1018,7 @@ class PostDataHandler extends DataHandler
 			}
 		}
 
-		if($visible == 1 && $thread['visible'] == 1)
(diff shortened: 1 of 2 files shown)


JN-Jones:
TBH I'm not sure whether this is the right way to fix this. While this will fix the issue itself I feel like we should instead handle updating the lastposter properly in the datahandler which only updates the last poster if both thread and post are visible: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L1021
This may likely also apply to deleting/restoring threads so properly updating the last poster on adding a new post (or approving/unapproving/deleting/restoring) feels more like the proper solution to me @mybb/developers

Shade-:
I thought the same thing in #2521, but I didn't get any feedback. We shall update the logic here also: https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L1095 given `update_last_post()` runs inside this conditional.

JN-Jones:
@mybb/developers any other opinion here? Would be nice to have it in 1.8.9

Shade-:
What. The. Heck. Still learning git, but I guess I messed up again by rebasing my branch in an attempt to update my branch to the latest commit in the feature branch.

Guess I'm going to close this PR and open a new one with the proper changes for #2521.

Cu8eeeR:
Yep, would be nice to see this in 1.8.9

Ben-Conway:
@Shade- any update on this one? Being a high priority issue we should try to get this in 1.8.9.

Shade-:
I've removed the visibility check to ensure threads' lastposter is updated when a post is made even if the thread is not visible.

Also removed get_thread() as it seems redundant to me (already initialized in [L847](https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L847)).

inc/datahandlers/post.php
```diff
@@ -1018,7 +1018,7 @@ function insert_post()
 			}
 		}
 
-		if($visible == 1 && $thread['visible'] == 1)
+		if($visible == 1)
```
Stefan-ST: This appears to be incorrect. If the thread isn't visible the post count shouldn't be changed.

inc/datahandlers/post.php
```diff
@@ -1255,6 +1249,13 @@ function insert_post()
 			$thread_update = array('replies' => '+1');
 			update_forum_counters($post['fid'], array("deletedposts" => "+1"));
 		}
+		
+		// Update forum count
+		if($visible == 1) {
+			update_last_post($post['tid']);
+			update_forum_counters($post['fid'], array("posts" => "+1"));
```
Stefan-ST: Same here... The forum counters shouldn't be changes if the thread isn't visible. Updating the thread's last post is correct though.

inc/datahandlers/post.php
```diff
@@ -1229,11 +1228,9 @@ function insert_post()
 			}
 
 			$thread_update = array('replies' => '+1');
-
-			// Update forum count
-			update_last_post($post['tid']);
+
```
Stefan-ST: Can you remove the tabs from the empty line, please?

inc/datahandlers/post.php
```diff
@@ -1255,6 +1252,12 @@ function insert_post()
 			$thread_update = array('replies' => '+1');
 			update_forum_counters($post['fid'], array("deletedposts" => "+1"));
 		}
+		
+		// Update forum count
+		if($visible == 1) {
+			update_last_post($post['tid']);
+			update_forum_lastpost($thread['fid']);
```
Stefan-ST: It's not necessary to move update_forum_lastpost. It doesn't need to be updated when the thread isn't visible.

inc/datahandlers/post.php
```diff
@@ -1255,6 +1252,12 @@ function insert_post()
 			$thread_update = array('replies' => '+1');
 			update_forum_counters($post['fid'], array("deletedposts" => "+1"));
 		}
+		
+		// Update forum count
+		if($visible == 1) {
```
Stefan-ST: Braces should be in it's own line.
https://docs.mybb.com/1.6/Coding-Standards/

Ben-Conway:
@Stefan-ST Are you happy with the changes?

Ben-Conway:
@Stefan-ST Are these changes okay?

Cu8eeeR:
It would be nice to see this in 1.8.10 release

WildcardSearch:
I've tested everything seems fine.

@Stefan-ST can you approve the changes?

Ben-Conway:
@WildcardSearch as long as you are 100% sure this works correctly we can approve the changes and merge this.

WildcardSearch:
@Ben-Conway I'd feel more comfortable waiting on @Stefan-ST or at least have another @mybb/developers or @mybb/sqa check it.

Ben-Conway:
@andrewjs18 can you check this?

euantorano:
@Stefan-ST Can you please double check the comments regarding your review and either mark it as resolved or expand on changes you wish to see?

Ben-Conway:
Thanks @Stefan-ST. @euantorano merging since @WildcardSearch tested it already

PR #2526, opened 2016-11-23
Fixes #2299 - Missing setting descriptions in ACP

### LINKED ISSUE #2299: Missing setting descriptions
Some settings, for example Custom User Title Maximum Length and Length limit in Signatures, don't explain what happens if the value is 0. It causes misunderstandings like this: http://community.mybb.com/thread-187524.html where users may expect different behavior - e.g. max length equal 0 in this case, which would basically mean disabling usertitles/signatures.

"0 to disable the limitation.", "0 for unlimited length." or anything similar could be added to their decription.

Cu8eeeR:
Fixes #2299 - Missing setting desctriptions in ACP

@JN-Jones @PaulBender @Destroy666x @dequeues 

I am not sure about these: setting in L around n. 217 and 1411

Please review it carefully, I will try my best to find them all and fix the missing descriptions.

### DIFF
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -1408,7 +1408,7 @@ min=0]]></optionscode>
 		</setting>
 		<setting name="maxattachments">
 			<title>Maximum Attachments Per Post</title>
-			<description><![CDATA[The maximum number of attachments a user is allowed to upload per post. Set to 0 disable.]]></description>
+			<description><![CDATA[The maximum number of attachments a user is allowed to upload per post. Set to 0 to disable.]]></description>
 			<disporder>2</disporder>
 			<optionscode><![CDATA[numeric
 min=1]]></optionscode>
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -198,7 +198,7 @@ email=Sent via Email]]></optionscode>
 		</setting>
 		<setting name="showlanguageselect">
 			<title>Show Language Selector in Footer</title>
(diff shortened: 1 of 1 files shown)


JN-Jones:
@Ben-Conway can you take a first look at this?

install/resources/settings.xml
```diff
@@ -1400,15 +1400,15 @@ min=0]]></optionscode>
...
+			<description><![CDATA[If you wish to disable attachments on your board, set this option to No.]]></description>
 			<disporder>1</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
 		<setting name="maxattachments">
 			<title>Maximum Attachments Per Post</title>
-			<description><![CDATA[The maximum number of attachments a user is allowed to upload per post.]]></description>
+			<description><![CDATA[The maximum number of attachments a user is allowed to upload per post. Set to 0 disable.]]></description>
```
Shade-: "Set to 0 disable." --> "Set to 0 to disable."

Shade-:
This looks good, except for the aforementioned part for which I requested a change. Note that I've only checked the diff report, some other descriptions may require editing with the "Set to 0 to disable" notice.

Btw, +1 to @Cu8eeeR for standardizing the Yes/No labels.

Cu8eeeR:
@Shade- thank you! I fixed the typo you mentioned above. It would be really nice if someone can look at this and find other places where the change is needed.

euantorano:
Looks good to me. I had a quick glance and didn't notice anywhere else needing this change.

Ben-Conway:
Looks good to me.

PR #2527, opened 2016-11-23
Fixes #2497 - Missing lang string for "time" in moderation

### LINKED ISSUE #2497: Delayed moderation - missing lang string for time
https://community.mybb.com/thread-203431.html

Anyway. the lang string is here... am I right? 
https://github.com/mybb/mybb/blob/feature/install/resources/mybb_theme.xml#L8010

Cu8eeeR:
Fixes #2497 - Missing lang string for "time" in moderation


JN-Jones:
This would fix the issue. Need to take a look whether it's probably better to put it in the global file but this should fix the issue.

Cu8eeeR:
I think it is enought, anyway no problem to edit my PR

JN-Jones:

Ben-Conway:
Looks like I got beat to it.

PR #2529, opened 2016-11-24
Fix #2522 - Links executing JS functions should return false and point to javascript.void(0) instead of hashtags

### LINKED ISSUE #2522: Links executing JS functions should return false and point to javascript.void(0) instead of hashtags
As discussed [here](https://community.mybb.com/thread-163831.html), links which trigger JS functions should point their href attribute to `javascript.void(0)` instead of a mere hashtag `#` and `return false` to normalize browser behavior.

Shade-:
#2522 

I've also standardized onclick calls (some are called with an old-fashioned javascript: prefix, the majority don't so I've deleted the javascript: prefix where present).

Despite fixing the issue, I do believe onclick should be avoided and added to inline scripts where possible.

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3922,7 +3922,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 {$forums}
 </table>
 <br />]]></template>
-		<template name="forumdisplay_thread" version="1801"><![CDATA[<tr class="inline_row">
+		<template name="forumdisplay_thread" version="1809"><![CDATA[<tr class="inline_row">
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%"><span class="thread_status {$folder}" title="{$folder_label}">&nbsp;</span></td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%">{$icon}</td>
 	<td class="{$bgcolor}{$thread_type_class}">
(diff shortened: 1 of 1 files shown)


Destroy666x:
Template versions should be updated

Shade-:
My bad, I will update the PR ASAP.

JN-Jones:
I'm not sure whether eg radiobutton listeners needs the return false as we don't need/want to stop the default handling there (we only  want to do additional stuff). However I haven't checked how browsers even handle that. Will report back as soon as  I've done that.

Shade-:
Darn, you're right. Returning false on checkboxes should prevent their usage in favor of the function executed, so they shouldn't return false at all. I'm going to update the PR.

Shade-:
@Destroy666x how does template versioning works? In the latest commit I've just added +1 to the existing values, but I'm not sure it'll work.

JN-Jones:
The values should represent the next version so `1809` for those you're going to change with this PR

Shade-:
Done. Sorry for the multiple commits, I'm still familiarizing with command-line git.

WildcardSearch:
This is a good thing, but I agree with Shade, I'd like to see event handling done dynamically.

JN-Jones:
PR looks good to me. @mybb/sqa If noone else has any objections I'm fine with merging this later after testing a few more places

JN-Jones:
Can't find anything wrong, if @mybb/sqa doesn't report anything wrong within the next few hours I'm merging this.

JN-Jones:
Merging this then

PR #2532, opened 2016-11-26
Disallow remote avatars, fixes #2413

### LINKED ISSUE #2413: Ability to disallow remote avatars
It should be possible to disable adding remote avatars in User CP (there also could be an option to let MyBB download and store them locally).

http://community.mybb.com/thread-171303-post-1160778.html

euantorano:
This prevents users from using remote avatars, fixing #2413 

It hides the input in the User CP, and replaces all existing avatars with the default avatar when `format_avatar()` is used to generate the avatar (which I believe should be almost everywhere these days?). It also adds a notice to the header if the user is currently using a remote avatar, but remote avatars have been disabled, like the following:

screen shot 2016-11-26 at 14 59 39

Note that the check for a remote avatar in `format_avatar` is fairly simple, as it just checks for `://` in the avatar value. This should be sufficient, as uploaded avatars are always relative.


JN-Jones:
Looking through the code it seems fine. Want to test it first though, especially the ACP and the Gravatar parts.

global.php
```diff
@@ -760,6 +760,12 @@
 	eval('$pm_notice = "'.$templates->get('global_pm_alert').'";');
 }
 
+$remote_avatar_notice = '';
+if($mybb->user['avatartype'] === 'remote' && !$mybb->settings['allowremoteavatars'])
```
JN-Jones: `$mybb->user['avatartype']` equals `gravatar` for that so it doesn't show this warning if you were using a gravatar.

euantorano:
Ah, yes. Good catch.

euantorano:
@JN-Jones Fixed.

Ben-Conway:
Good work Euan.

JN-Jones:
@WildcardSearch commenting on the pr itself instead of a commit is beyter as it shows here and is more unlikely to be missed ;)

About keeping this as usergroup setting: While I see the benefit I'd say a normal setting is better here. First because of what euan said already on slack. And secondly disallowing remote avatars is only related to hiding the server ip which should be done independently to the user. It isn't a permission per se as the user can always download the image and reupload it on the board. So it's not something like can have an avatar at all.

Ben-Conway:
@euantorano can we get the conflicts fixed?

euantorano:
Yes, will do so tonight. Only became conflicted after another PR merge this afternoon.

euantorano:
Fixed.

Ben-Conway:
@JN-Jones Could you review so we can get this merged?

JN-Jones:
Yeah, will do so later today

JN-Jones:
Merging

PR #2548, opened 2016-12-04
Fix 2547 - Added fix_javascript() to URLs to prevent XSS

### LINKED ISSUE #2547: XSS in URL BBCode
The URL BBCode is not sanitized properly, leading posts/PMs open to XSS attacks. Internal thread: https://community.mybb.com/thread-206010.html

Shade-:
Fixes #2547. Tested and should work fine. Further testing might be needed @mybb/sqa

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1060,6 +1060,12 @@ class postParser
 		{
 			$nofollow = " rel=\"nofollow\"";
 		}
+		
+		// Fix eventual XSS attempts
+		if(strpos($url, 'javascript:') !== false)
+		{
+			$url = $this->fix_javascript($url);
+		}
 
 		// Fix some entities in URLs
 		$entities = array('$' => '%24', '&#36;' => '%24', '^' => '%5E', '`' => '%60', '[' => '%5B', ']' => '%5D', '{' => '%7B', '}' => '%7D', '"' => '%22', '<' => '%3C', '>' => '%3E', ' ' => '%20');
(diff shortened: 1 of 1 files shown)


inc/class_parser.php
```diff
@@ -1060,6 +1060,12 @@ function mycode_parse_url($url, $name="")
 		{
 			$nofollow = " rel=\"nofollow\"";
 		}
+		
+		// Fix eventual XSS attempts
+		if(strpos($url, 'javascript:') !== false)
```
JN-Jones: Shouldn't this check for position 0? Otherwise URLs like `http://example.com/javascript:something` might be changed too.

Shade-:
Would it be better to strip out javascript:// occurrencies then?

Devilshakerz:
Either the `[url]` regexes should be changed (it would be the only disallowed protocol) or their callback functions should ignore the part if `^javascript:` is matched.

inc/class_parser.php
Ben-Conway: @Shade- ?

Ben-Conway:
Can we get this fixed and merged? @Shade-

Shade-:
Yep sorry Ben, I'm studying at the moment but tomorrow it's vacation time here so I should push the fixes I've already made for this on my localhost.

Ben-Conway:
@Stefan-ST @JN-Jones Are we happy with the new changes?

Devilshakerz:
Submitted https://github.com/mybb/mybb/pull/2559 with more & modified lookaheads to prevent false negatives (`[url]javascript://h4x()[/url]`) and positives (`[url=javascript.com]test[/url]`)

PR #2549, opened 2016-12-05
Update SFS registration link to get an API key

MyBB-Andro:
SFS registration link was changed since they stopped anonymous registrations and requires to register in their forum first to get an API key.


Cu8eeeR:
#2550 

Please, do not create PRs without an issue.

MyBB-Andro:
Okay, will keep in mind now on. :)

PR #2559, opened 2016-12-07
Fix #2547 XSS in URL BBCode

### LINKED ISSUE #2547: XSS in URL BBCode

Devilshakerz:
#2547

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -300,16 +300,16 @@ class postParser
 
 		if($mybb->settings['allowlinkmycode'] == 1)
 		{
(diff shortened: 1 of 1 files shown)


JN-Jones:
This works fine for our parser. However SCEditor still treats `javascript:` links as a valid and parses them in WYSIWYG mode. While that doesn't produce any issue we may want to change the regex for that too (CC @samclarke )

samclarke:
@JN-Jones It shouldn't do. If you try to load a BBCode URL that starts with `javascript:` SCE should treat it as a relative URL and prefix it with the current page URL.

PR #2561, opened 2016-12-08
Fixes #1751 Splitting thread decrease the thread count of user

### LINKED ISSUE #1751: Splitting thread decrease the thread count of user
when a thread is formed by splitting a post from another thread, this split thread doesn't increase the thread count of that user... it however, results in decrease of the thread count...

To check: http://www.mybbtest.tk/member.php?action=profile&amp;uid=5

before splitting, Posts: 3, Threads: 2

after splitting, Posts: 3, Threads: 1 (instead of 3)

Original thread: [Splitting thread decrease the thread count of user](http://community.mybb.com/thread-165331.html)

Stefan-ST:
Fixes #1751

### DIFF
--- a/inc/class_moderation.php
+++ b/inc/class_moderation.php
@@ -1568,16 +1568,6 @@ class Moderation
 			update_first_post($thread['tid']);
 		}
 
-		// Subtract merged thread from user counter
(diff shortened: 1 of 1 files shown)


Ben-Conway:
Is this finished @Stefan-ST ?

Stefan-ST:
Yes, why do you ask @Ben-Conway?

PR #2565, opened 2016-12-08
Fix#2560 Empty index operator error on php 7.1

### LINKED ISSUE #2560: Empty index operator throws fatal error on PHP 7.1 when used on strings
https://github.com/mybb/mybb/blob/mybb_1808/newreply.php#L732
https://github.com/mybb/mybb/blob/mybb_1808/newreply.php#L828

https://secure.php.net/manual/en/migration71.incompatible.php#migration71.incompatible.empty-string-index-operator

Matslom:
Fixes #2560

### DIFF
--- a/newreply.php
+++ b/newreply.php
@@ -277,7 +277,8 @@ if($mybb->settings['enableattachments'] == 1 && $mybb->get_input('attachmentaid'
 	}
 }
 
-$reply_errors = $quoted_ids = array();
+$reply_errors = '';
+$quoted_ids = array();
 $hide_captcha = false;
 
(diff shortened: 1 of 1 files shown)


newreply.php
```diff
@@ -277,7 +277,7 @@
 	}
 }
 
-$reply_errors = $quoted_ids = '';
+$reply_errors = $quoted_ids = array();
```
Devilshakerz: `$reply_errors` should be kept as string - otherwise https://github.com/mybb/mybb/blob/mybb_1808/newreply.php#L882 incorrectly evaluates as true

PR #2567, opened 2016-12-08
Fix#2551 Delete invalid PIN error message on admin login

### LINKED ISSUE #2551: You have entered an invalid secret PIN.
if you go to your admin cp, leave all of the form fields blank and hit 'login', it'll throw an error about an invalid secret pin.  if you only type in the secret pin and hit 'login', it'll still tell you that you enter an invalid secret pin.

it seems that the error is defaulted to a message about an invalid secret pin.  shouldn't that error be more ambiguous for security purposes like it is when you type in an incorrect username/password when trying to sign in to the front end?

"You have entered an invalid username/password combination."

mybb 1.8.4, clean install, no plugins.

Original thread: [You have entered an invalid secret PIN.](https://community.mybb.com/thread-169663.html)

Matslom:
Fix #2551

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -162,6 +162,22 @@ elseif($mybb->input['do'] == "login")
 	require_once MYBB_ROOT."inc/datahandlers/login.php";
 	$loginhandler = new LoginDataHandler("get");
 
+	// Determine login method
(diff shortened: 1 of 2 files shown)


Ben-Conway:
@JN-Jones Looks fine to me.

PR #2568, opened 2016-12-08
Fixes #2558 - Disable edit by timestamp... - change description

### LINKED ISSUE #2558: Disable edit by timestamp for admins also disables for moderators
As the title says,

Under posting settings on the ACP:

Show 'edited by' Message for Administrators

Do you want to show edited by messages for posts created by administrators?

But disabling this option also disables the same for moderators. It should be either configurable which Group IDs can be disabled or disable only for administrators as the description says.

Thanks

Original thread: [Disable edit by timestamp for admins also disables for moderators]

EDIT: This is so that the setting description can be updated as the above is intentional behaviour.

(https://community.mybb.com/thread-206484.html)

Cu8eeeR:
Fixes #2558 - Disable edit by timestamp for admins also disables for moderators

### DIFF
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -1318,8 +1318,8 @@ show=Show to all Users]]></optionscode>
 			<isdefault>1</isdefault>
 		</setting>
 		<setting name="showeditedbyadmin">
-			<title>Show 'edited by' Message for Administrators</title>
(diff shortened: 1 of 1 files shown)


Ben-Conway:
Thanks for this @Cu8eeeR

Cu8eeeR:
Glad to help make MyBB better ;)

PR #2569, opened 2016-12-08
Fixes #2491 New reply mod options do not check specific moderator per…  

### LINKED ISSUE #2491: New reply mod options do not check specific moderator permissions
[At the moment](https://github.com/mybb/mybb/blob/mybb_1808/newreply.php#L1395) in new reply the mod options (close & stick checkboxes) are only shown to moderators of the forum (including those who can't close/stick threads).
The close/stick options should be separated out with the `canstickunstickthreads` and `canopenclosethreads` forum moderator permissions checked.
eg `is_moderator($fid, "canstickunstickthreads")`

In the post handler, each of these should also check the relevant permission before applying the change: https://github.com/mybb/mybb/blob/mybb_1808/inc/datahandlers/post.php#L891L917

PaulBender:
…missions

Fixes #2491

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -889,28 +889,28 @@ class PostDataHandler extends DataHandler
 				$modoptions_update = array();
 
 				// Close the thread.
-				if(!empty($modoptions['closethread']) && $thread['closed'] != 1)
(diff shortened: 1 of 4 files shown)


Ben-Conway:
@euantorano does this look okay?

PaulBender:
Updated template version code to 1810

euantorano:
Looks fine to me, but haven't actually tested yet.

PR #2570, opened 2016-12-09
Fix #2428 Remove <script> and <style> tags with content on help docs.

### LINKED ISSUE #2428: Help Documents HTML not parsed in misc.php?action=helpresults
I have some help documents which use HTML and inline CSS to make them look pretty.

If you use the search feature located within misc.php?action=help the search results will display the raw HTML/CSS like so:

Original thread: [Help Documents HTML not parsed in misc.php?action=helpresults](http://community.mybb.com/thread-192089.html)

Matslom:
Fix #2428

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -8479,3 +8479,20 @@ function my_validate_url($url, $relative_path=false)
 	return false;
 }
 
+/**
+ * Strip html tags from string, also removes <script> and <style> contents.
+ * 
+ * @param  string $string         String to stripe
+ * @param  string $allowable_tags Allowed html tags
+ * 
+ * @return string                 Striped string
+ */
+function my_strip_tags($string, $allowable_tags = '')
+{
+	$pattern = array(
+		'@(&lt;)style[^(&gt;)]*?(&gt;).*?(&lt;)/style(&gt;)@siu',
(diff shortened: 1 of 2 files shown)


Cu8eeeR:
@Matslom - really good job! Thank you for your contributions :)

inc/functions.php
```diff
@@ -8479,3 +8479,20 @@ function my_validate_url($url, $relative_path=false)
...
+ * @param  string $string         String to stripe
+ * @param  string $allowable_tags Allowed html tags
+ * 
+ * @return string                 Striped string
+ */
+function my_strip_tags($string, $allowable_tags = '')
+{
+	$pattern = array(
+		'@(&lt;)style[^(&gt;)]*?(&gt;).*?(&lt;)/style(&gt;)@siu',
+        '@(&lt;)script[^(&gt;)]*?.*?(&lt;)/script(&gt;)@siu',
```
Shade-: Can you replace the spaces with tabs, so the indentation aligns with the previous line? Thank you :) Also, I'm not sure if other places might need this function, but the corresponding entities for &lt; and &gt; should be added too in order to cover HTML markup which has not been `htmlspecialchars_uni()`'ed already.

euantorano:
Works for me. It might be more performant to use one long regex, but we'll leave it as is for now.

PR #2587, opened 2017-01-04
Wrong MySQL version detected

mattiabasone:
If you are using MariaDB 10.x.x the server version returned is always 5.5.5 with mysqli_get_server_info().
With this issue, you cannot use FULLTEXT search with MariaDB.
More info: http://php.net/manual/en/mysqli.get-server-info.php#118822

### DIFF
--- a/inc/db_mysql.php
+++ b/inc/db_mysql.php
@@ -1057,13 +1057,9 @@ class DB_MySQL implements DB_Base
 			return $this->version;
 		}
 
-		$version = @mysql_get_server_info();
-		if(!$version)
-		{
-			$query = $this->query("SELECT VERSION() as version");
-			$ver = $this->fetch_array($query);
-			$version = $ver['version'];
-		}
+		$query = $this->query("SELECT VERSION() as version");
+		$ver = $this->fetch_array($query);
+		$version = $ver['version'];
 
 		if($version)
 		{
--- a/inc/db_mysqli.php
+++ b/inc/db_mysqli.php
@@ -1050,13 +1050,9 @@ class DB_MySQLi implements DB_Base
 			return $this->version;
 		}
 
-		$version = @mysqli_get_server_info($this->read_link);
-		if(!$version)
-		{
-			$query = $this->query("SELECT VERSION() as version");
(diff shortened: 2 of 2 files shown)


Ben-Conway:
@Devilshakerz @euantorano Is this a valid issue?

Ben-Conway:
@mybb/developers @mybb/sqa Can anyone confirm this?

PaulBender:
I can't confirm any issue with MariaDB on my site. It's the right version in both my Admin CP and my cPanel (5.5.53).

Ben-Conway:
@PaulBender The bug requires you to be using MariaDB 10.x.x

euantorano:
I can confirm on MariaDB 10.1.18 that `mysqli_get_server_info` returns `5.5.5-10.1.18-MariaDB`, but the query `SELECT VERSION() as version` returns `10.1.18-MariaDB`.

I haven't tested this patch yet, but will do so later.

andrewjs18:
I can confirm that mybb reports the wrong version of mysql when using MariaDB.

on my live site in the admin cp:
SQL Engine 	MySQLi 5.5.5

my mysql version:
Distrib 10.1.19-MariaDB

Ben-Conway:
Thanks for checking. Could you test this PR? @andrewjs18

euantorano:
Can you also perform this change to `inc/db_mysql.php`? Otherwise, this looks good to me.

Ben-Conway:
@mattiabasone we are looking to get the 1.8.10 packages created this weekend so can you make the extra changes ASAP please.

Cu8eeeR:
@euantorano @Ben-Conway - if not, can you please a separate PR for mysql?

Ben-Conway:
@Cu8eeeR We've decided we don't want to do that and this is looking unlikely to make it into 1.8.10.

Ben-Conway:
Perfect, @mattiabasone. Thanks for contributing!

Ben-Conway:
@euantorano happy with this?

premiumcuyuz:
good jobb man!

euantorano:
I'll test this today, thanks @mattiabasone!

PR #2601, opened 2017-01-07
Fixes #2590

### LINKED ISSUE #2590: Who Posted Sorting Links Redirect To New Page
In forumdisplay.php when viewing the thread list and clicking the reply count link for the AJAX Who Posted modal...

...when clicking either sort link, the user is redirected to a page, empty except for the modal content:

Apologies if this is a duplicate, but I did search.

Original thread: [Who Posted Sorting Links Redirect To New Page](https://community.mybb.com/thread-206871.html)

WildcardSearch:
...where in the Who Posted modal dialog, sorting links redirected to a
new blank page. #2590

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -6213,15 +6213,15 @@ if(use_xmlhttprequest == "1")
 {$forumlistbits}
 </select>]]></template>
 		<template name="misc_syndication_forumlist_forum" version="1808"><![CDATA[<option value="{$forum['fid']}" {$optionselected}>{$depth} {$forum['name']}</option>]]></template>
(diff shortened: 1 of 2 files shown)


euantorano:
Personally I'd rather we fixed the main misc page, but that would be a much bigger patch which wouldn't get into 1.8.10.

The current patch wouldn't work for users with JS disabled.

WildcardSearch:
> Personally I'd rather we fixed the main misc page, but that would be a much bigger patch which wouldn't get into 1.8.10.

Yes, I thought of that, but the issue is based on a problem with the modal so I just stuck to handling the JavaScript issue. We can create a separate issue for the bigger problems.

euantorano:
Sounds fair to me, can you please create the new issue?

WildcardSearch:
Done.

PR #2607, opened 2017-01-08
Fixes #2606 Deleting a forum and the recount link while using custom …

### LINKED ISSUE #2606: Deleting a forum and the recount link while using custom admin url,
So I've discovered an issue present in MyBB 1.8.9. I moved all threads from one part of my site to another and was deleting the empty section. It gave me an alert telling me that ideally I should run the recount threads. It had a link to do so. That link was /index.php?module=tools-recount_rebuild rather than /mysecretadminurl//index.php?module=tools-recount_rebuild. This should be adjusted for the next release of MyBB 1.8.

PaulBender:
…admin url

Fixes #2606


Ben-Conway:
@euantorano is this going into 1.8.10?

PR #2608, opened 2017-01-08
Patch for issue #2606

nsuchy:
This should fix issue #2606 :)


euantorano:
Fixed in #2607. The other PR also changed `&` for `&amp;` so it was merged since it killed two birds with one stone.

Thanks for the contribution though!

nsuchy:
@euantorano No problem :)

PR #2613, opened 2017-01-12
Fixes #2612 last post info is not updating properly

### LINKED ISSUE #2612: last post info is not updating properly
after the upgrade from 1.8.9 to 1.8.10, the last post info does not appear to be updating correctly on the forum index as well as when you click inside a forum category.  if you go inside the forum section where the post exists, it seems like it's updating correctly.

here's a few screenshots...

forum index view:

215 jan 11

category view:

216 jan 11

forum view:

217 jan 11

this issue is also present on these community forums as well..

Original thread: [last post info is not updating properly](https://community.mybb.com/thread-207323.html)

Stefan-ST:
Fixes #2612

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -1230,6 +1230,7 @@ class PostDataHandler extends DataHandler
 			$thread_update = array('replies' => '+1');
 
 			// Update counters
+			update_last_post($post['tid']);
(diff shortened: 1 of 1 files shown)


Cu8eeeR:
@Stefan-ST - Thank you very much!

PR #2623, opened 2017-01-16
Fixes #276

### LINKED ISSUE #276: Find orphaned attachments shows success when failed
> I couldn't think of a better title.
> 
> More info:
> 
> I restricted access to the uploads folder so Apache couldn't modify the contents. I then went to ACP > Forums & Posts > Attachments > Find Orphaned Attachments and it shows two files where we not found in the attachments table.
> 
> I hit Delete Checked Orphans and it gave the error:
> 
>  Only some orphaned attachments were successfully deleted, others could not be removed from the uploads directory.
> 
> Which is false. No orphaned attachments were successfully deleted. I ran it again and it still showed two files.
> 
> I think there needs to be some more verification for this module.

Original report: http://dev.mybb.com/issues/1866
Reporter: Nathan Malcolm
Status: New

WildcardSearch:
Fixes #276


Shade-:
Tested, working fine. Merging.

PR #2625, opened 2017-01-16
Fixes #139

### LINKED ISSUE #139: Stats Performance w/ Today's Top Poster
It runs like crap on a large forum.  I changed the default query to something much faster then added some php to sort. Query went from 40 seconds to under 1. 

```
default:
    $query = $db->query("
        SELECT u.uid, u.username, COUNT(*) AS poststoday
        FROM ".TABLE_PREFIX."posts p
        LEFT JOIN ".TABLE_PREFIX."users u ON (p.uid=u.uid)
        WHERE p.dateline > $timesearch
        GROUP BY p.uid ORDER BY NULL DESC
    ");
```

//Custom for fix of max poster
while($user = $db->fetch_array($query))
{
    $topposter[] = array('poststoday' =>$user['poststoday'],'uid' => $user['uid'], 'username' => $user['username'] );
}

$user = max($topposter);

WildcardSearch:
Fixes #139 where an attempt was made to improve performance for caching today's top poster.


WildcardSearch:
@mybb/sqa can anyone give this a test?

You need to have a reasonable-sized database (the bigger the better) to test this on.

katosdev:
Didn't @andrewjs18 already test this?

andrewjs18:
@katosdev, I did, yes, and didn't really see any page speed improvements.

my board has over 112k posts with almost 2,000 registered users so definitely not a small forum by mybb standards.

Stefan-ST:
100k is probably not enough to have noticable impact. Even on the community forums the original query is pretty fast.

WildcardSearch:
Someone (can't remember who) suggested a setting. So BBOs can adjust the PHP:SQL ratio.

JordanMussi:
Any feedback for this PR? As @WildcardSearch says a setting to choose which query is used could be implemented defaulting to the current one.

euantorano:
I've not done any testing at all of the changes, and have never ran any large boards myself so can't comment much on the changes. Rather than a full blown setting maybe a config flag would make sense?

WildcardSearch:
That doesn't seem intuitive to me. This is a specific function of one particular page of the forum. I don't think it warrants being relegated to `config.php`.

As before, I still feel that a setting added to `Server and Optimization Options` makes the most sense.

euantorano:
Anybody got any more thoughts on this, also it's currently conflicting so that will need cleaned up if we want to merge this.

WildcardSearch:
I still haven't completed my testing. The plan is to create a test board, randomly create 1 million users, random assign to them 1 million posts, and time stamp them so that they stack second-to-second-- placing the maximum amount of users posting into one 24 hour period to be checked.

Then of course, run the query both ways and benchmark using the ADMIN debug info.

I'll try my best to get that done this weekend and then if we choose to merge this, I'll get it the conflicts sorted.

WildcardSearch:
I've resolved conflicts with this again. Can we get this in 1.8.15 or make a decision at least? @mybb/developers @mybb/management

euantorano:
I need to give this a test and make sure nothing is broken, I’ll try and do that this week.

kawaii:
@euantorano I've just tested this after @Ben-MyBB asked me to a few days ago in the Discord. Whilst I'm unsure of the performance benefit on anything other than the absolute largest of boards, I can confirm that it certainly didn't break anything when I tried it out.

euantorano:
Great, thanks @kawaii. We've had multiple people confirm it works now, so I'm happy to merge.

euantorano:
Just noticed that the ` AND p.visible=1` condition is missing from the new query. Unless I'm missing something, isn't that a bug? If so, can you create a quick patch to add that @WildcardSearch?

PR #2643, opened 2017-01-27
Fixes #2642 - Stats page refreshing

### LINKED ISSUE #2642: Stats page refreshing issue
Reproduce

Admin Control Panel &gt; Configurations &gt; Statistiscs Page &gt; Stats Limit

See Stats Limit is 15 by default.

Change it to 30.

See in Stats Page it's still 15.

Go back and change it to 45.

See in Stats Page it's 30 now.

Go back and change it to 60.

See in Stats Page it's 45 now.

It always shows the limit saved one before than actual setting.

If you save the settings twice, it shows the actual value since one before is same as actual one.

Original thread: [Stats page refreshing issue](https://community.mybb.com/thread-207771.html)

Cu8eeeR:
#2642
All credits for fix goes to #doylecc

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -1088,13 +1088,19 @@ if($mybb->input['action'] == "change")
 			$cache->update_statistics();
 		}
 
-		if(isset($mybb->input['upsetting']['statslimit']) && $mybb->input['upsetting']['statslimit'] != $mybb->settings['statstopreferrer'])
+		$statslimit = 0;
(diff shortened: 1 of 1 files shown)


Cu8eeeR:
@Shade- - you are right, thank you! I will fix this ASAP.

Cu8eeeR:
I created a new PR with your proposed changes - can you please check it? Thank you very much for your feedback! Keep good work!

PR #2653, opened 2017-02-05
Fix #2619

### LINKED ISSUE #2619: Almost all numeric values greater 9 are set to 9 in "Show All Settings"
Since MyBB 1.8.9 there's a bug in /admin/modules/config/settings.php: If you click on "Configuration-&gt;Settings-&gt;Show All Settings" and then on "Save Settings", all numeric values, which are greater then 9 are set to 9 which renders a forum almost unuseable because for example even the "Maximum Message Length" is set to only 9 characters:

[attachment=38186]

On the other hand, if you use only a single "Settings Group", the bug oddly enough is not hit.

I've verified the bug, I've installed a brand new MyBB 1.8.8 where all config changes are working as expected, a new MyBB 1.8.9 or 1.8.10 shows the bug described above.

Original thread: [Almost all numeric values greater 9 are set to 9 in "Show All Settings"](https://community.mybb.com/thread-207397.html)

Shade-:
Fixes #2619 by adding the `$options` reset. Already tested, but needs confirmation.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -944,6 +944,8 @@ if($mybb->input['action'] == "change")
 
 		while($multisetting = $db->fetch_array($query))
 		{
+			$options = array();
+
 			if(substr($multisetting['optionscode'], 0, 8) == 'checkbox')
 			{
 				$checkbox_settings[] = $multisetting['name'];


PaulBender:
Works fine for me.

PR #2654, opened 2017-02-05
Fix #2618

### LINKED ISSUE #2618: Theme selector leaves GET data in the URL
The footer theme selector passes data, including the anti-CSRF token using GET and doesn't redirect the page after the request is handled, leaving it in the URL.

The data should be passed via POST.

Shade-:
Fixes #2618 by changing method for the theme selector to POST.


PaulBender:
Works for me however the change should also be made to the language selector (same issue exists).

Shade-:
Done.

PR #2656, opened 2017-02-05
Fix #2496

### LINKED ISSUE #2496: Editing of announcements in ModCP doesn't work
A few minutes ago I've tried to update an old announcement (starting date in 2007) in my forums via the ModCP. After submitting I've got the following error message:

> The starting date for the announcement is invalid.

If I replace the date by another one in present or future the error doesn't occur. But in this case the starting date will not be saved. Instead the timestamp seems to contain only the difference from now till the given date (e.g. if I set the date in 5 minutes it shows 1.1.1970 at 1:00 GMT).

Note, that this error only occurs in ModCP. Editing in AdminCP seems to work perfectly.
I could reproduce this error on both my live system and my local test environment. In both cases the forums run on an Apache server with PHP 7.0 and MyBB 1.8.8.

Shade-:
Fixes #2496 by removing `TIME_NOW` from `$localized_time_offset` which led to `$startdate` resulting negative and thus not allowing to edit an announcement. Also fixed a typo.

### DIFF
--- a/modcp.php
+++ b/modcp.php
@@ -1121,7 +1121,7 @@ if($mybb->input['action'] == "do_new_announcement")
 
 	$localized_time_offset = (float)$mybb->user['timezone']*3600 + $mybb->user['dst']*3600;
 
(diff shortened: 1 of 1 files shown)


PaulBender:
Works for me.

PR #2670, opened 2017-02-18
Fix #2662 - Add hook for posts multimove in moderation tools

### LINKED ISSUE #2662: New hook in moderation.php to work with moved posts
As described [here](https://github.com/frostschutz/MyBB-Google-SEO/issues/56#issuecomment-274126300) a new hook added to `moderation.php` would be useful.

Matslom:
Fix #2662

### DIFF
--- a/moderation.php
+++ b/moderation.php
@@ -2477,6 +2477,8 @@ switch($mybb->input['action'])
 		// Verify incoming POST request
 		verify_post_check($mybb->get_input('my_post_key'));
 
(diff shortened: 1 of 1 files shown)


MyBB-Andro:
Thank you very much @Matslom . Good work.

euantorano:
Looks good to me

@frostschutz Is this the hook that you need for Google SEO? Whilst we're here, are there any others that you might need?

PR #2689, opened 2017-02-25
Fixes #2634

### LINKED ISSUE #2634: Blocked user's soft deleted message issue
Reproduce

Block any user.

Soft delete that user's message.

See instead of 1 you see 2 buttons and nicknames one with explanation one usual

If you click both of them user's message disappears entirely.

Original thread: [Blocked user's soft deleted message issue](https://community.mybb.com/thread-207637.html)

Shade-:
#2634

I switched the two snippets just to prioritize deleted over ignored post; it seemed more logical to me.

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -838,19 +838,19 @@ function build_postbit($post, $post_type=0)
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


PaulBender:
Works for me

Issue #2691, opened 2017-02-26
Can create multiple usergroups with the same name


SvePu:
I'd consider this a bug so I'll post it here and see if it effects anyone else. I can create many user groups with the exact same group name. There should be some sort of check.

Original thread: https://community.mybb.com/thread-207529.html

SvePu:
Please review #2692

Shade-:
The discussion is not over. Next time, please wait for a staff member to push the issue to GitHub once there is an agreement on how to proceed with the issue, thank you.

Ben-MyBB:
@mybb/developers @mybb/management @mybb/sqa @mybb/community-team What does everyone think about this? I think it's a bug IMO.

kawaii:
Personally, I wouldn't say it crosses the threshold to be classed as a bug. It's not like usernames where there's the importance of not having a direct collision, since there's nowhere in MyBB where you actually need to type a group name to perform an action. Ultimately it's a question of standards - since I can't see this 'bug' actually breaking anything. :smile_cat:

Ben-MyBB:
Well users could have multiple of the same groups and not realise, it's odd to allow people to have multiple groups with the same name. That's like allowing multiple users with the same username.

WildcardSearch:
This isn't a bug. The **id**entity is in tact. Creating two groups with the same name might be desired if the admin wants to create two different groups with slightly different permission, but are indistinguishable to regular users.

Cu8eeeR:
There should be a notice that the usergroup will have same name,,, at least. I think it is a bug but many admins use this as a feature

kawaii:
I agree with @Cu8eeeR in this case. We should simply notify the user that a group with the same name already exists and ask if they wish to continue. Perhaps @SvePu can modify his pull request to accommodate this? :)

Its-Nasyr:
I don't believe this to be a bug, I agree with Wildcard on this matter. Some communities have two variants of Administrators which leads them to creating multiple groups with slightly different names.

wagey:
Don't see the benefit in changing this, period. You can always distinguish a set of groups that share the same name by just looking at the gid.

SvePu:
@wagey -  In groupselect settings options you can't see the gids of groups.

kawaii:
How about we simply display GIDs in the ACP then so that novice users can quickly check if they don't understand that the URI contains the GID?

WildcardSearch:
[...]

PR #2692, opened 2017-02-26
Fix for #2691

SvePu:
This could be a fix for issue #2691 - please review

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -843,7 +843,7 @@ if($mybb->input['action'] == "edit")
 					$grouptitles[] = $value['title'];
 				}
 
-				if(in_array(trim($mybb->input['title']), $grouptitles))
+				if(in_array(trim($mybb->input['title']), $grouptitles) && trim($mybb->input['title']) != $usergroup['title'])
 				{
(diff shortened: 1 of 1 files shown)


admin/modules/user/groups.php
```diff
@@ -673,6 +673,23 @@
 		{
 			$errors[] = $lang->error_missing_title;
 		}
+
```
Shade-: Please remove the hidden tab chars.

admin/modules/user/groups.php
```diff
@@ -815,6 +832,23 @@
...
```
Shade-: Also here.

admin/modules/user/groups.php
```diff
@@ -673,6 +673,23 @@
 		{
 			$errors[] = $lang->error_missing_title;
 		}
+		
+		if ($mybb->input['title'])
```
Shade-: Conditionals and round brackets should not be separated with a space. Applies to the whole PR.

SvePu:
@Shade- Thx for reviewing .... I removed the hidden tab chars & extra space

Shade-:
As discussed within #2691, this PR is invalid.

Issue #2610, opened 2017-01-09
When viewing own profile via `uid=0`, the username isn't shown


euantorano:
If you access `member.php?Action=profile` (with no UID) or `user-0.html` (when using search engine friendly URLs), you are taken to your own profile page. However, the current location will be `Viewing Profile`, without the name of the profile being viewed.

Original thread: [user-0 redirects to own profile](https://community.mybb.com/thread-207240.html)

Broatcast:
Fixed the problem https://github.com/mybb/mybb/pull/2694

PR #2694, opened 2017-03-01
Fix for #2610

Broatcast:
Just simply fix the issue


Shade-:
This turns out to be wrong. The user is watching is own profile, but in Who's Online it shows that he is watching the profile you are currently logged in. $mybb->user['uid'] refers to the currently logged in user.

PR #2707, opened 2017-03-26
Fixes #2694

Shade-:
#2694

### DIFF
--- a/inc/functions_online.php
+++ b/inc/functions_online.php
@@ -161,10 +161,26 @@ function fetch_wol_activity($location, $nopermission=false)
 			elseif($parameters['action'] == "profile")
 			{
 				$user_activity['activity'] = "member_profile";
-				if(!isset($parameters['uid']))
+
+				if($parameters['uid'] == 0)
+				{
+					global $memprofile;
+
+					// $user is available in Who's Online but not in Member Profile, use $memprofile instead
+					if(!empty($user['uid']))
+					{
+						$parameters['uid'] = $user['uid'];
+					}
+					elseif(!empty($memprofile['uid']))
+					{
+						$parameters['uid'] = $memprofile['uid'];
+					}
+				}
+				else if(!isset($parameters['uid']))
 				{
 					$parameters['uid'] = 0;
 				}
+
 				$parameters['uid'] = (int)$parameters['uid'];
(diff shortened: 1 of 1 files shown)


inc/functions_online.php
```diff
@@ -161,10 +161,26 @@ function fetch_wol_activity($location, $nopermission=false)
...
+					if(!empty($user['uid']))
+					{
+						$parameters['uid'] = $user['uid'];
+					}
+					elseif(!empty($memprofile['uid']))
+					{
+						$parameters['uid'] = $memprofile['uid'];
+					}
+				}
+				elseif(!isset($parameters['uid']))
```
Devilshakerz: This condition should be executed independently, before `if($parameters['uid'] == 0)`

inc/functions_online.php
```diff
@@ -161,10 +161,26 @@ function fetch_wol_activity($location, $nopermission=false)
...
 				{
 					$parameters['uid'] = 0;
 				}
+
 				$parameters['uid'] = (int)$parameters['uid'];
```
Devilshakerz: The casting should be done before checking `if($parameters['uid'] == 0)`, e.g.:
```php
if(isset($parameters['uid']))
{
	$parameters['uid'] = (int)$parameters['uid'];
}
else
{
	$parameters['uid'] = 0;
}

if($parameters['uid'] == 0)
{
```

inc/functions_online.php
Shade-: Thank you @Devilshakerz, updated

inc/functions_online.php
```diff
@@ -176,13 +182,7 @@ function fetch_wol_activity($location, $nopermission=false)
...
 					}
 				}
-				elseif(!isset($parameters['uid']))
-				{
-					$parameters['uid'] = 0;
-				}
-
-				$parameters['uid'] = (int)$parameters['uid'];
-				if($parameters['uid'] > 0)
+				elseif($parameters['uid'] > 0)
```
Devilshakerz: this should be just `if` as `$uid_list` is not filled correctly now

Shade-: Holy moly. Did a blind fix without testing it. Thank you again mate! Fixed.

PR #2714, opened 2017-03-31
Fixes #2160 - RSS limit doesn't work...

### LINKED ISSUE #2160: RSS limit doesn't work in the generated URL - always 15
No idea why noone noticed this obvious bug yet, but the last highlighted line doesn't make any sense:
https://github.com/mybb/mybb/blob/feature/misc.php#L842-44
It was added when fixing PHP warnings as I see, but I don't see any purpose of it..

And a bit nitpicky, but the limit desciptions says "Not recommended to be beyond 50." while the limit is hardcoded to be 50: https://github.com/mybb/mybb/blob/feature/syndication.php#L43 The description should make clear that you can't use higher numbers and this conditional: https://github.com/mybb/mybb/blob/feature/misc.php#L905-908 should also be corrected (+ `(int)` casting and `is_numeric()` check are redundant).

EDIT: and lastly, RSS is said to be default but Atom is used instead: https://github.com/mybb/mybb/blob/feature/misc.php#L890

SvePu:


SvePu:
#2160 RSS limit doesn't work

misc.php
```diff
@@ -902,7 +906,7 @@
 				$all = true;
 				break;
 			}
-			elseif(is_numeric($fid))
+			elseif((int)$fid)
```
ATofighi: It cause XSS vulnerability in misc.php.
example: `misc.php?action=syndication&forums[]=1&forums[]=2<script>alert(%27hi%27)</script>&version=rss2.0&limit=15`.

misc.php
```diff
@@ -935,9 +939,9 @@
 		}
 		if((int)$limit > 0)
```
ATofighi: `(int)` can be removed from this line.

SvePu: Yes, it's the same double check like [here](https://github.com/SvePu/mybb/blob/26775de517299b33a28e74ae5e84ea39aa779289/misc.php#L909)

ATofighi:
There is an XSS Injection in this pull request. for example: `misc.php?action=syndication&forums[]=1&forums[]=2<script>alert(%27hi%27)</script>&version=rss2.0&limit=15`.

PR #2742, opened 2017-05-15
Fix #2741 Missing plugin hooks in postParser

### LINKED ISSUE #2741: Missing plugin hooks in postParser
The `postParser::parse_message` method could use more plugin hooks to allow external parsing without the interference of core matches.

dvz:
Fix #2741


inc/class_parser.php
```diff
@@ -175,6 +177,8 @@ function parse_message($message, $options=array())
 			$message = preg_replace('#(>|^|\r|\n)/slap ([^\r\n<]*)#i', "\\1<span style=\"color: red;\" class=\"mycode_slap\">* {$this->options['me_username']} {$lang->slaps} \\2 {$lang->with_trout}</span>", $message);
 		}
 
+		$message = $plugins->run_hooks("parse_message_memycode", $message);
```
euantorano: Personally I'd prefer a name of `parse_message_me_mycode` as I find it slightly easier to read with the extra `_`. Otherwise, I'm happy for these to be added.

PR #2746, opened 2017-05-19
Preview usergroup's username style via user-group settings

Pyragon:
This was useful for me when I was editing or adding a usergroup. 
Before I was adding a test account to the groups to double check the username styles every time.


Cu8eeeR:
Why not, I like it! Useful .)

Shade-:
Seems useful. If no one opposes the PR I will merge this. @mybb/developers @mybb/sqa.

Pyragon:
Done. Thank you

Shade-:
Thank you. I'm going to wait a reply from other members and eventually merge it.

euantorano:
Looks good to me, quite a useful addition IMO.

PR #2752, opened 2017-05-25
Fix #2751

### LINKED ISSUE #2751: Broken attachment name
screenshot

The problem is in global.lang.php:
`$l['rel_time'] = "<span title=\"{5}{6}\">{1}{2} {3} {4}</span>";`

This is used in the my_date function which is used by functions_post.php to get $attachdate which is then placed in the title property of the hyperlink in template `postbit_attachments_attachment`

PirataNervo:


Shade-:
This has been fixed with https://github.com/mybb/mybb/pull/2753.

PR #2753, opened 2017-05-26
Fixes #2382 my_date - no time for non-relative $format

### LINKED ISSUE #2382: my_date - no time for non-relative $format
Not sure if this needs to be considered a bug, but posting it here, because to me it looks like one. 

This bug only occurs when not the relative time format, but the time format from settings is used. It might be possible, that this is not the case anywhere in the code, then it would be no bug, but dead code ^^

https://github.com/mybb/mybb/blob/feature/inc/functions.php#L482 Only today / yesterday is rendered, but no exact time. 

For relative time the exact time is added in case it is more than 12h ago https://github.com/mybb/mybb/blob/feature/inc/functions.php#L474

I traced it down to the commit, where the my_date function was adapted to support relative time at to me it looks like adding the time has been forgotten for non relative timestamps ... 

https://github.com/mybb/mybb/commit/508f7808085565ed77cc0aea3ad743fb60a4e192#diff-7fe6e55397c77ab9a0f5d57bc4cbe5b9L391

Original thread: [my_date - no relative time](http://community.mybb.com/thread-190208.html)

### LINKED ISSUE #2751: Broken attachment name

### LINKED ISSUE #2802: If there's a quote in excerpt for pm_subscription, new date code causes an issue
If user gets private message notification for a message, starting with quote, they see something like this:

Quote:...
Here is an excerpt of the message:
------------------------------------------
Username (who get quoted): (&lt;span title="19-07-2017, 15:21"&gt;1 minute ago&lt;/span&gt;)
--
Quoted message
------------------------------------------

To view the thread, you can go to the following URL:
...

Original thread: [If there's a quote in excerpt for pm_subscription, new date code causes an issue](https://community.mybb.com/thread-211828.html)

PaulBender:
Fixes #2382 

- Added a 'normal' format to the my_date function. This will fully format the date and time without using relative dating.
- Changed a few instances of date() to my_date().
- Changed attachment hover from relative to normal (fixes #2751).
- Changed a few dates from older 1.6 style to 1.8 style.
- Fixed issue with relative time in quoted post (fix from [here](https://community.mybb.com/thread-210659-post-1274718.html#pid1274718)).
- Changed PM subscription excerpt to use normal date format (fixes #2802).

### DIFF
--- a/admin/modules/tools/mailerrors.php
+++ b/admin/modules/tools/mailerrors.php
@@ -71,7 +71,7 @@ if($mybb->input['action'] == "view")
 	$log['subject'] = htmlspecialchars_uni($log['subject']);
 	$log['error'] = htmlspecialchars_uni($log['error']);
 	$log['smtperror'] = htmlspecialchars_uni($log['smtpcode']);
-	$log['dateline'] = date($mybb->settings['dateformat'], $log['dateline']).", ".date($mybb->settings['timeformat'], $log['dateline']);
+	$log['dateline'] = my_date('relative', $log['dateline']);
 	$log['message'] = nl2br(htmlspecialchars_uni($log['message']));
 
 	?>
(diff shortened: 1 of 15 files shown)


SvePu:
Nice work @PaulBender

Cu8eeeR:
@PaulBender - please can you look at this issue? It is probably related to date issues https://community.mybb.com/thread-211828-post-1279291.html#pid1279291 Thx!

SvePu:
@Cu8eeeR  [This](https://community.mybb.com/thread-210659-post-1274718.html#pid1274718) should fit it too.

euantorano:
Is this complete as far as we know now then? I'll try and review it this week.

PaulBender:
It's complete unless somebody finds any new date issue.

euantorano:
Awesome, thanks.

Shade-:
Had some time to test this out, looks like it's working good and does not break any existing functionality.

euantorano:
Nice one, thanks @Shade-!

PR #2500, opened 2016-11-08
Fixes #1565 Add Deletion Notice

PaulBender:
Fixes #1565 

Adds a deletion notice that's visible with permissions

PR #2757, opened 2017-05-28
Fixes #2756 $post['poststatus'] missing from postbit

### LINKED ISSUE #2756: $post['poststatus'] missing from postbit
The variable $post['poststatus'] is missing from the postbit templates. This variable is suppose to inform the moderator of the post's status (unapproved/soft deleted) and was missed when #1565 was added.

PaulBender:
Fixes #2756


JordanMussi:
Works as expected but I'm wondering why you hardcoded the `<em>` & `<strong>` in #2500 rather than giving the element a class to control the display?

Cu8eeeR:
@JordanMussi - good catch, maybe we can create a separate PR for this

euantorano:
This PR looks good to me, another PR to fix the hardcoded HTML as mentioned by @JordanMussi would be greta it anybody gets a chance.

JordanMussi:
@euantorano fixed in #2784

PR #2771, opened 2017-06-27
Fixes #2371 Incorrect user activity when action is unset

### LINKED ISSUE #2371: Incorrect user activity when action is unset

JordanMussi:
Fixes #2371 
The problem was initially setting the action to `0`. When checked by the first `in_array`, the loop was exited as the non-strict comparison evaluated to true.

### DIFF
--- a/inc/functions_online.php
+++ b/inc/functions_online.php
@@ -245,7 +245,7 @@ function fetch_wol_activity($location, $nopermission=false)
 		case "modcp":
 			if(!isset($parameters['action']))
 			{
-				$parameters['action'] = 0;
+				$parameters['action'] = '';
 			}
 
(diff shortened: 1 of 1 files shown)


Shade-:
Tested, works.

PR #2772, opened 2017-06-27
Fixes #2603 ForumDisplay: Who Posted Inoperable Without JavaScript

### LINKED ISSUE #2603: ForumDisplay: Who Posted Inoperable Without JavaScript
If JavaScript is disabled, when viewing `forumdisplay.php`, the linked Replies column value is dead. The `href` attribute of the link points to `MyBB.whoPosted`.

A fallback should be introduced for a standard redirect.

See: #2590

JordanMussi:
Fixes #2603


euantorano:
Looks good to me, just need to test it.

PR #2774, opened 2017-06-29
Fixes #154 Select All for Reports

### LINKED ISSUE #154: Select All for Reports
In the modcp when marking selected reports as read the select all checkbox will not mark all reports on all pages as read.  

That's not how it behaved before and not in line with how other pages of MyBB work for selection of all on a page.  

I can have as many as 300 open reports and now when moderators go through a single page they can't use the mark all anymore and have to manually click each box.  That's time consuming and like I said...wasn't that way before.  

Please fix.

JordanMussi:
Fixes #154

Implementation the same as inline moderation.

Additional changes:
- Use strict comparison for `my_strpos` as it ignores the first item in the list without it
- Change `error_noselected_reports` to be the correct description - never used when "the selected posts have already been marked read by another user"
- Change "posts" to "content" in other language strings

---

**This PR:**
- changes 1 language file
- changes 2 templates
- adds 1 new template
- adds 1 new `.js` file

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1160,7 +1160,7 @@ if(!empty($threadcache) && is_array($threadcache))
 
 		if($ismod)
 		{
-			if(isset($mybb->cookies[$inlinecookie]) && my_strpos($mybb->cookies[$inlinecookie], "|{$thread['tid']}|"))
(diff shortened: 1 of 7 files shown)


Cu8eeeR:
Works fine, @euantorano @JordanMussi

euantorano:
Just a note for myself: there's a new tmeplate in this PR that needs to be mentioned in the release post and docs.

PR #2775, opened 2017-07-02
use isset to check if it's set and not null

f0nt4:
Not using isset() will return true if the user_cache exists even though the given uid might not be set.
This can result in an undefined offset error.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1193,13 +1193,13 @@ function user_permissions($uid=0)
 	if($uid != $mybb->user['uid'])
 	{
 		// We've already cached permissions for this user, return them.
-		if($user_cache[$uid]['permissions'])
+		if(isset($user_cache[$uid]['permissions']))
 		{
 			return $user_cache[$uid]['permissions'];
 		}
 
 		// This user was not already cached, fetch their user information.
-		if(!$user_cache[$uid])
+		if(!isset($user_cache[$uid]))
 		{
 			$user_cache[$uid] = get_user($uid);
 		}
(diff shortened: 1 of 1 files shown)


Shade-:
I'd go with `!empty()` rather than `isset()`, as `!empty()` also checks if the variable is not false which is the way this function was designed before. Can you edit the PR @f0nt4? Thank you.

f0nt4:
Using empty() seems to work as well. Changed the PR.

inc/functions.php
```diff
@@ -1193,13 +1193,13 @@ function user_permissions($uid=0)
 	if($uid != $mybb->user['uid'])
 	{
 		// We've already cached permissions for this user, return them.
-		if($user_cache[$uid]['permissions'])
+		if(empty($user_cache[$uid]['permissions']))
```
Shade-: This must be !empty, otherwise it will always fail.

inc/functions.php
```diff
@@ -1193,13 +1193,13 @@ function user_permissions($uid=0)
...
 		{
 			return $user_cache[$uid]['permissions'];
 		}
 
 		// This user was not already cached, fetch their user information.
-		if(!$user_cache[$uid])
+		if(!empty($user_cache[$uid]))
```
Shade-: And this should be `empty()`.

f0nt4:
Yup, you're right. Still a little sleepy here,

Shade-:
Not a problem. Thank you for contributing. I shall remind you that, as stated in the [guidelines](https://github.com/mybb/mybb/blob/feature/CONTRIBUTING.md), issues must be opened in the 1.8 bugs and issues community forum, discussed there and approved by a team member.

f0nt4:
Ah ok, didn't know that. Will keep it in mind for the future! Thanks for merging :)

PR #2779, opened 2017-07-04
Fixes #2767 Registered group bugs out when they have 'can view deletion notices' and cant see new posts.

### LINKED ISSUE #2767: Registered group bugs out when they have "can view deletion notices" and cant see new posts.
As I said above, the registered group bugs out when they have "can view deletion notices" and cant see new posts because I think the deleted posts count and will only count towards the max post count in that thread. It is a bit hard to explain. But if "can view deletion notices" is off, the count of posts is correct because they cannot see deleted posts anymore and can see the latest post.

JordanMussi:
Fixes #2767

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1121,10 +1121,20 @@ if(!empty($threadcache) && is_array($threadcache))
 		$threadpages = '';
 		$morelink = '';
 		$thread['posts'] = $thread['replies'] + 1;
-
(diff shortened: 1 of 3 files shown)


Ben-MyBB:
This would be good to get merged if anyone has the time to check it

Shade-:
I'm on my way to do some tests for it.

Ben-MyBB:
Thanks @Shade-

PR #2780, opened 2017-07-04
Fixes #2777 Can't remove edit reason from posts

### LINKED ISSUE #2777: Can't remove edit reason from posts
Version: 1.8.12

How to reproduce:

1. Post something 
2. Edit that post and add a reason to it, why you're editing it
3. Quick edit doesn't show current edit reason (not sure if it's supposed to be like that) but Full edit shows the current edit reason
4. Remove that reason, change the post if you want then save/submit
5. Notice the edit reason near the timestamp of the post isn't removed.

Pretty much if the edit reason is empty, it should completely take off the edit reason from the post else otherwise.

I've tried this on my site and I've even gone on to @katosdev site to tried it and got the same result. I'm sure @katosdev is running MyBB 1.8.10.

JordanMussi:
Fixes #2777


euantorano:
Just to remind myself: This adds a new template that needs to be documented in the release post and docs.

PR #2781, opened 2017-07-04
Fixes #2763 Random password at registration doesn't work

### LINKED ISSUE #2763: Random password at registration doesn't work
The password randomly generated at registration doesn't work because its hash is inserted into the mail instead of the actually password.

JordanMussi:
Fixes #2763

### DIFF
--- a/member.php
+++ b/member.php
@@ -443,16 +443,16 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 			switch($mybb->settings['username_method'])
 			{
 				case 0:
-					$emailmessage = $lang->sprintf($lang->email_randompassword, $user['username'], $mybb->settings['bbname'], $user_info['username'], $mybb->input['password']);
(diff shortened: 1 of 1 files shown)


member.php
```diff
@@ -443,16 +443,16 @@
 			switch($mybb->settings['username_method'])
 			{
 				case 0:
-					$emailmessage = $lang->sprintf($lang->email_randompassword, $user['username'], $mybb->settings['bbname'], $user_info['username'], $user_info['password']);
+					$emailmessage = $lang->sprintf($lang->email_randompassword, $user['username'], $mybb->settings['bbname'], $user_info['username'], $mybb->input['password']);
```
euantorano: Should probably use `$mybb->get_input()` here just for the sake of following what's used elsewhere.

JordanMussi: Sure - done in 6a8025a6938ac1ef656b7cc0228574b31dec8ea6

PR #2783, opened 2017-07-05
Fixes #2700 Custom Admin Theme Set Before Completed Login with 2FA

### LINKED ISSUE #2700: Custom Admin Theme Set Before Completed Login with 2FA
Hi there. I realized a quick issue earlier today. If I login to my admin account where a custom theme is set, the CSS for that theme will load at the time the 2FA Prompt loads. If I enter an invalid code, my session is reset and I have to retype my username and password, then retry 2FA. However the difference is the login screen now shows my custom admin theme as it was never switched back to default. Was this intentional or is it a bug?

Original thread: [Custom Admin Theme Set Before Completed Login with 2FA](https://community.mybb.com/thread-208939.html)

JordanMussi:
Fixes #2700 

Now the default/global admin theme and language is used for login and the user's preference is only used once they are fully authenticated.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -462,9 +462,9 @@ else
 				else if(ADMIN_IPV6_SEGMENTS > 0 && strpos($ip_address, ':') !== false)
 				{
 					// Expand IPv6 addresses
-					$hex = unpack("H*hex", my_inet_pton($ip_address));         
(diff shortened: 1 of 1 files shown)


Shade-:
Double checked, works fine.

PR #2784, opened 2017-07-05
More flexible styling for postbit_status

JordanMussi:
Fixes comment here: https://github.com/mybb/mybb/pull/2757#issuecomment-312932548


euantorano:
Not yet tested this, but have looked at the changes. Does this retain the same styling as it previously had?

JordanMussi:
Um, it has the same style as the IP logged text. Do you want strong italics added?

euantorano:
Probably for the best to have it match how it currently looks if possible :smile:

JordanMussi:
This wasn't being displayed in previous releases so there isn't a currently 😛 - I'm just not sure about having such heavy emphasis on the text.

euantorano:
Ah, fair enough. Might as well leave it as is then.

PR #2790, opened 2017-07-06
Fixes #1653 xmlhttp improvements

### LINKED ISSUE #1653: xmlhttp improvements
- [ ] as mentioned here: https://github.com/mybb/mybb/pull/1649 I'm nearly sure `username_exists` isn't used anywhere so it can be removed. Correct me if I'm wrong. @mybb/developers 
- [ ] there is username AJAX validation during registration. but duplicate/invalid Email validation doesn't exist in xmlhttp.php, so an inline error is printed instead after submitting the form. Would be nice to add it.
- [ ] passwords are checked against username/Email for being the same (insecure), not by AJAX though - another thing that would be nice to add
- [ ] `get_users` action lists all users that begin with a search term, but would be nice to check all users that contain it instead. http://community.mybb.com/thread-168070-post-1145266.html#pid1145266 + possibly return UIDs (which are already selected in queries) if noone disagrees (?): http://community.mybb.com/thread-171952.html

JordanMussi:
Fixes #1653 

---

**This PR:**
- changes 1 template
- changes 1 language file

### DIFF
--- a/member.php
+++ b/member.php
@@ -783,7 +783,64 @@ if($mybb->input['action'] == "register")
 	{
 		$plugins->run_hooks("member_register_start");
 
-		$validator_extra = '';
+		// JS validator extra
+		if($mybb->settings['maxnamelength'] > 0 && $mybb->settings['minnamelength'] > 0)
+		{
+			$lang->js_validator_username_length = $lang->sprintf($lang->js_validator_username_length, $mybb->settings['minnamelength'], $mybb->settings['maxnamelength']);
+		}
+
+		$validator_javascript = "<script type=\"text/javascript\">
+$(document).ready(function() {
+	$('#registration_form').validate({
+		rules: {
+			username: {
+				required: true,
+				minlength: {$mybb->settings['minnamelength']},
+				maxlength: {$mybb->settings['maxnamelength']},
+				remote: {
+					url: 'xmlhttp.php?action=username_availability',
+					type: 'post',
+					dataType: 'json',
+					data:
+					{
+						my_post_key: my_post_key
+					},
+				},
+			},
+			email: {
+				required: true,
+				email: true,
+				remote: {
+					url: 'xmlhttp.php?action=email_availability',
+					type: 'post',
+					dataType: 'json',
+					data:
+					{
+						my_post_key: my_post_key
+					},
+				},
+			},
+			email2: {
+				required: true,
+				email: true,
+				equalTo: '#email'
+			},
+		},
+		messages: {
+			username: {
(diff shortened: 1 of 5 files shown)


member.php
```diff
@@ -783,7 +783,64 @@
...
 		$plugins->run_hooks("member_register_start");
 
-		$validator_extra = '';
+		// JS validator extra
+		if($mybb->settings['maxnamelength'] > 0 && $mybb->settings['minnamelength'] > 0)
+		{
+			$lang->js_validator_username_length = $lang->sprintf($lang->js_validator_username_length, $mybb->settings['minnamelength'], $mybb->settings['maxnamelength']);
+		}
+
+		$validator_javascript = "<script type=\"text/javascript\">
```
Shade-: Personally, I'd like to reduce hardcoded HTML rather than adding some.

euantorano: Yeah, I agree with @Shade-. If there's any way to not hardcode this, it would be preferred.

JordanMussi: I've had a think and there's 3 ways of going about this:
1. completely hardcoded (as in the PR currently)
2. main JS in template with conditional snippets hardcoded (as before the PR)
3. completely in templates: main JS in template with conditional snippets in new templates

Before this PR the javascript was fragmented with some hardcoded and the bit that's used all the time in the template. I moved the template code to the PHP file to bring it together. I don't think number 2 is preferable at all so it's a case of 1 vs 3.

Personally I'd say number 1 as I believe this type of javascript shouldn't be in templates because the theme authors shouldn't need to change it as it's not part of the style. Although there is nothing to stop them extending it by adding `<script>` tags.

euantorano: We can certainly live with `1` for now and can always change it later if there is a sensible request to change it for a good reason.

member.php
```diff
@@ -1208,63 +1265,66 @@
 			// JS validator extra
 			$lang->js_validator_password_length = $lang->sprintf($lang->js_validator_password_length, $mybb->settings['minpasswordlength']);
 
+			$validator_javascript .= "
+	$.validator.addMethod('passwordSecurity', function(value, element, param) {
+		return (value != $('#email').val() && value != $('#username').val()
+			&& value.indexOf($('#email').val()) === false && value.indexOf($('#username').val()) === false
+			&& $('#email').val().indexOf(value) === false && $('#username').val().indexOf(value) === false);
+	}, '{$lang->userdata_bad_password_security}');\n";
```
JordanMussi: This language string is not available in this position. Preparing patch...

euantorano: Oops, good catch!

JordanMussi: Fixed in #2878.

PR #2794, opened 2017-07-11
CRAM-MD5 authentication

uzbugg:
For users who are unable to authenticate against SMTP.


Stefan-ST:
Thank you for your pull request. Unfortunately it doesn't follow our [Development standards](https://docs.mybb.com/1.8/development/standards/) and it's missing error message (it just returns false).
I've created a new PR to fix these issues: #2799

PR #2796, opened 2017-07-11
CRAM-MD5 and DIGEST-MD5 ordered

uzbugg:
For users who are unable to authenticate against SMTP. This makes sure that the safer options will be executed before  LOGIN or PLAIN methods. Of Course only if the server supplies these options.


euantorano:
Closing due to #2799 which was merged earlier.

PR #2799, opened 2017-07-11
Fixes #2798 SMTP authentication problem with DIGEST-MD5 or CRAM-MD5

### LINKED ISSUE #2798: SMTP authentication problem with DIGEST-MD5 or CRAM-MD5
hey all, 

As I am trying to set up a new SMTP server and  I've encoutered this weird problem where I can not authetificate.
I get this error message saying "The SMTP server does not support any of the AUTH methods that MyBB supports". Doesn't matter if I use ssl, tls or plain authetication, it just won't work.

So I poked around in the auth() function of smtp.php and I found out, that my SMTP server is returning these authentication methods DIGEST-MD5, CRAM-MD5 but the function is only checking for LOGIN or PLAIN methods. (see lines 357 and 381)

If I change these lines: 

Quote:if(in_array("LOGIN", &#36;auth_methods)) 

to:

Quote:if(in_array("DIGEST-MD5", &#36;auth_methods)) 

,doesn't matter which line I change to DIGEST-MD5 or CRAM-MD5, the email gets sent.

Now that is cool that I can send emails by changing these values, but:

I might won't be the only once facing this issue

I am not sure, if the code bellow the "if" is the right way to use CRAM-MD5 or DIGEST-MD5

Thank you for the time devs! I think this issue should be solved in the next release, or just point me into the right direction :rolleyes: .

Have a nice day!

Original thread: [SMTP authentication problem with DIGEST-MD5 or CRAM-MD5](https://community.mybb.com/thread-211585.html)

Stefan-ST:
Fixes #2798

This PR doesn't add DIGEST-MD5.

### DIFF
--- a/inc/mailhandlers/smtp.php
+++ b/inc/mailhandlers/smtp.php
@@ -204,7 +204,7 @@ class SmtpMail extends MailHandler
 
 		if($this->connected())
 		{
-			if(!$this->send_data('MAIL FROM:<'.$this->from.'>', '250'))
+			if(!$this->send_data('MAIL FROM:<'.$this->from.'>', 250))
 			{
 				$this->fatal_error("The mail server does not understand the MAIL FROM command. Reason: ".$this->get_error());
 				return false;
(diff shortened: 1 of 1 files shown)


euantorano:
This needs testing, but to do so requires finding or setting up an SMTP server with MD5 or CRAM-MD5 authentication. I'll try to do so if I get time.

Issue #2743, opened 2017-05-15
Replace features deprecated in PHP 7.2


dvz:
- `create_function()`
- `each()` (3rd party only)
- `assert()` with string argument (3rd party only)

https://wiki.php.net/rfc/deprecations_php_7_2

skmedix:
* ```create_function()``` - 14
  * ```/admin/inc/functions_themes.php```
    * [functions_themes.php#L423](https://github.com/mybb/mybb/blob/feature/admin/inc/functions_themes.php#L423)
  * ```/admin/modules/config/languages.php```
    * [languages.php#L348](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L348)
    * [languages.php#L352](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L352)
    * [languages.php#L561](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L561)
    * [languages.php#L562](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L562)
    * [languages.php#L566](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L566)
    * [languages.php#L567](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L567)
    * [languages.php#L598](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L598)
    * [languages.php#L602](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L602)
    * [languages.php#L621](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L621)
    * [languages.php#L625](https://github.com/mybb/mybb/blob/feature/admin/modules/config/languages.php#L625)
  * ```/inc/functions.php```
    * [functions.php#L5686](https://github.com/mybb/mybb/blob/feature/inc/functions.php#L5686)
    * [functions.php#L5687](https://github.com/mybb/mybb/blob/feature/inc/functions.php#L5687)
    * [functions.php#L6479](https://github.com/mybb/mybb/blob/feature/inc/functions.php#L6479)
---
* ```each()``` - 4
  * ```/inc/3rdparty/diff/Diff/Engine/Native.php```
    * [Native.php#L200](https://github.com/mybb/mybb/blob/feature/inc/3rdparty/diff/Diff/Engine/Native.php#L200)
[...]

PR #2805, opened 2017-07-28
Fixes #2804

### LINKED ISSUE #2804: Stylesheets processing regex break stylesheets
I stumbled upon a nasty issue while building a theme: stylesheets are saved fixing url()s to the main folder path, given they are saved in the cache/themes/themeX folder. However, the regex seems to fail under certain circumstances, giving a faulty output which makes every CSS rule after the url() unusable in Firefox, Safari, Opera and IE. Chrome is unaffected as it succeeds into reading the erroneous file.

[The faulty regex](https://github.com/mybb/mybb/blob/21edc2063c624f20d740c8866166914bbe8a3b94/admin/inc/functions_themes.php#L423). I am not a regex master, but is there a reason why it matches the literal `1)` chars combination? If I have reverse-engineered this properly, it should match the url between single/double quotes and a round bracket (`')` or `")`).

A possible fix:
`$stylesheet = preg_replace_callback("#url\((\"|'|)(.*?)(\"|')\)#", create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);`

Shade-:
Fixes #2804.


JordanMussi:
Testing on PHP 7.1.9, `\1` isn't interpreted as a backreference and `\\1` works.
`\1` matches the literal `1` as the `\` escapes the character following it in the quoted string.

Using the given regex (with the backreference correction)
```php
$stylesheet = preg_replace_callback("#url\((\"|'|)(.*)\\1\)#", create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
```
All of the following match:
```
url(foo)
url("foo")
url('foo')
url("foo)
url(foo")
url('foo)
url(foo')
```
As the quotes are captured by the main group: `(.*)`

This is resolved by using `([^\"']+)` as the main group as it matches any character except for quotes.

To conclude, either of the following are suitable fixes:
```php
$stylesheet = preg_replace_callback("#url\((\"|'|)([^\"']+)\\1\)#", create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
$stylesheet = preg_replace_callback('#url\(("|\'|)([^"\']+)\1\)#', create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
```

euantorano:
Uh oh, something else we've merged is causing conflicts here - any chance you'll have time to take a look @Shade-?

Shade-:
Yes, but as @JordanMussi pointed out PHP 5 and 7 do not behave the same way, so this might need further processing.

euantorano:
I'll look at it further later and see if there's an easier solution.

JordanMussi:
@Shade- I'm getting the same result on 5.6.

With double quotes `"\1"` is parsed by PHP to be the literal `1` character. So only `1` is passed to the regular expression. Whereas when `"\\1"` is parsed the double backslash is the literal `\`. So `\1` is passed to the regular expression.

Hence the following lines operate identically:
```php
$stylesheet = preg_replace_callback("#url\((\"|'|)([^\"'\(]+)\\1\)#", create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
$stylesheet = preg_replace_callback('#url\(("|\'|)([^"\'\(]+)\1\)#', create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
```

---

@euantorano #2797 has caused the conflict by removing the `create_function`.

admin/inc/functions_themes.php
```diff
@@ -420,7 +420,7 @@ function cache_stylesheet($tid, $filename, $stylesheet)
 		"theme" => $theme_directory
 	);
 	$stylesheet = parse_theme_variables($stylesheet, $theme_vars);
-	$stylesheet = preg_replace_callback("#url\((\"|'|)(.*)\\1\)#", create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
+	$stylesheet = preg_replace_callback("#url\((\"|'|)(.*)\1\)#", create_function('$matches', 'return fix_css_urls($matches[2]);'), $stylesheet);
```
dvz: [`create_function()`](https://secure.php.net/create_function) cannot be used as it will be deprecated with PHP 7.2. See https://github.com/mybb/mybb/issues/2743

Shade-: Yep, got to sort out the conflicts.

Shade-:
@JordanMussi odd enough, as for me on 5.6 the single slash'ed one behaves differently (patching the original issue). I agree it should be refactored though as it seems I get inconsistent results and feedbacks.

Shade-:
Solved merge conflicts. Also, I have edited @JordanMussi's suggested fix which:
- does not alter the backreference;
- matches the following patterns:
`url("foo")
url('foo')
url('foo(bar)')
url("foo(bar)")`

Instead of `[^\"'\(]` I have replaced the open round bracket with a newline (`[^\"'\n]`). This should work most of the times; the following pattern is still matched wrongly:

`url('foo);url(bar')`

However:
- it's not a valid CSS rule;
- valid CSS rules are matched and processed correctly, including urls with brackets.

Ben-MyBB:
Anyone had a chance to check through this?

PR #2835, opened 2017-09-04
Fixes #2543 Hide invalid Calendar links

### LINKED ISSUE #2543: Hide invalid Calendar links
- the Calendar cannot go more than 5 years into the future
- it's not possible to go further back than year 1901 on 32-bit systems

In both cases the Calendar shows links to the preceding/succeeding periods even though they result in the same time frame.

PaulBender:
Fixes #2543 

Will hide all month and week links outside of the valid date range (January 1901 - December 5 years from now).

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -11,11 +11,11 @@
 define("IN_MYBB", 1);
 define('THIS_SCRIPT', 'calendar.php');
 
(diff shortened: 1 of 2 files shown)


Cu8eeeR:
Works fine, thank you!

PR #2838, opened 2017-09-12
Fixes #2691 Can create multiple usergroups with the same name

### LINKED ISSUE #2691: Can create multiple usergroups with the same name

JordanMussi:
Fixes #2691 Can create multiple usergroups with the same name

---

**This PR:**
- changes 1 language file

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -747,7 +747,8 @@ if($mybb->input['action'] == "add")
 			$grouptitles = array_column($groups, 'title');
 
 			$message = $lang->success_group_created;
-			if(in_array($mybb->input['title'], $grouptitles) && count(array_keys($grouptitles, $mybb->input['title'])) > 1) {
+			if(in_array($mybb->input['title'], $grouptitles) && count(array_keys($grouptitles, $mybb->input['title'])) > 1)
+			{
 				$message = $lang->sprintf($lang->success_group_created_duplicate_title, htmlspecialchars_uni($mybb->input['title']));
 			}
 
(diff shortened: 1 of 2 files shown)


admin/modules/user/groups.php
```diff
@@ -743,7 +743,15 @@
 			// Log admin action
 			log_admin_action($gid, htmlspecialchars_uni($mybb->input['title']));
 
-			flash_message($lang->success_group_created, 'success');
+			$groups = $cache->read('usergroups');
+			$grouptitles = array_column($groups, 'title');
+
+			$message = $lang->success_group_created;
+			if(in_array($mybb->input['title'], $grouptitles) && count(array_keys($grouptitles, $mybb->input['title'])) > 1) {
```
Shade-: The bracket should be in a new line

admin/modules/user/groups.php
```diff
@@ -964,7 +972,15 @@
 			// Log admin action
 			log_admin_action($usergroup['gid'], htmlspecialchars_uni($mybb->input['title']));
 
-			flash_message($lang->success_group_updated, 'success');
+			$groups = $cache->read('usergroups');
+			$grouptitles = array_column($groups, 'title');
+
+			$message = $lang->success_group_updated;
+			if(in_array($mybb->input['title'], $grouptitles) && count(array_keys($grouptitles, $mybb->input['title'])) > 1) {
```
Shade-: Same here

admin/modules/user/groups.php
JordanMussi: Fixed in 85e640af00f8be38b5ac5e50fcfea3d7850c5727

admin/modules/user/groups.php
JordanMussi: Fixed in 85e640af00f8be38b5ac5e50fcfea3d7850c5727

Cu8eeeR:
Works fine, thank you!

PR #2828, opened 2017-08-10
Fixes #2073

Shade-:
#2073

PR #2839, opened 2017-09-13
Added stripslashes()  to the action listing on modcp.

nsuchy:
In some cases a thread title with a single quote in it may display with a backslash in front of the single quote. This fixes that issue.

Edit: I also ammended this commit so this will also fix the issue in moderator logs.

### DIFF
--- a/modcp.php
+++ b/modcp.php
@@ -904,7 +904,7 @@ if($mybb->input['action'] == "modlogs")
 	while($logitem = $db->fetch_array($query))
 	{
 		$information = '';
-		$logitem['action'] = htmlspecialchars_uni($logitem['action']);
+		$logitem['action'] = stripslashes(htmlspecialchars_uni($logitem['action']));
 		$log_date = my_date('relative', $logitem['dateline']);
 		$trow = alt_trow();
 		$logitem['username'] = htmlspecialchars_uni($logitem['username']);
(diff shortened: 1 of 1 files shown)


euantorano:
This seems to be a replica of #2828, right? I'd prefer the data in the DB to be correct rather than correcting it on output so I prefer the other PR.

nsuchy:
@euantorano This PR would be beneficial as it corrects the the previous logs as well, not just new ones (unless PR #2828 also backports the changes to log files), alternatively rather than merging this PR, I could add an action to the upgrade script that goes through previous logs and removes the slashes. Thoughts?

dvz:
> alternatively rather than merging this PR, I could add an action to the upgrade script that goes through previous logs and removes the slashes.

Sounds good.

nsuchy:
@dvz @euantorano 

> alternatively rather than merging this PR, I could add an action to the upgrade script that goes through previous logs and removes the slashes.

Are you guys in agreement with doing that?

euantorano:
Yeah, a corrective step in the upgrade script would be the prefered way IMO otherwise a while down the line we'll wonder why it does a `stripslashes` before display 😄

nsuchy:
@euantorano I'll work on that right away.

euantorano:
There's one here for 1.8.13: https://github.com/mybb/mybb/blob/feature/install/resources/upgrade41.php

nsuchy:
@euantorano Closing this PR in favor of https://github.com/mybb/mybb/pull/2840 please provide feedback and code review at your convenience.

PR #2840, opened 2017-09-14
Remove slashes from moderator log during upgrade

nsuchy:
This PR Iterates through the moderator log and removes slashes. As decided in PR #2839.

### DIFF
--- a/install/resources/upgrade41.php
+++ b/install/resources/upgrade41.php
@@ -45,11 +45,11 @@ function upgrade41_dbchanges()
 	while($row = $db->fetch_array($query))
 	{
 		$original = $row['action'];
-		$stripped = $db->escape_string(stripslashes($original));
+		$stripped = stripslashes($original);
 
 		if($stripped !== $original)
 		{
-			$db->update_query("moderatorlog", "action" => $stripped, "WHERE tid = '".$row['tid']."'");
+			$db->update_query("moderatorlog", "action" => $db->escape_string($stripped), "WHERE tid = '".$row['tid']."'");
 		}
 	}
 
--- a/install/resources/upgrade41.php
+++ b/install/resources/upgrade41.php
@@ -36,20 +36,23 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
 	// Remove backslashes from last 1,000 log files.
-	$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\\\\\%", "limit" => 1000);
+	$query = $db->simple_select('moderatorlog', 'tid, action', "action LIKE '%\\\\\\%", array(
+		"order_by" => 'tid',
+		"order_dir" => 'DESC',
+		"limit" => 1000
+	));
 
 	while($row = $db->fetch_array($query))
 	{
-    $original = $row['action'];
-    $stripped = stripslashes($original);
+		$original = $row['action'];
+		$stripped = $db->escape_string(stripslashes($original));
 
-    if($stripped !== $original)
-    {
-        $db->update_query("moderatorlog", "action" => stripped, "WHERE tid = '".$row['tid']."'");
-    }
-
+		if($stripped !== $original)
+		{
+			$db->update_query("moderatorlog", "action" => $stripped, "WHERE tid = '".$row['tid']."'");
+		}
 	}
 
-  $output->print_contents("<p>Click next to continue with the upgrade process.</p>");
+	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
 	$output->print_footer("41_done");
 }
--- a/install/resources/upgrade41.php
+++ b/install/resources/upgrade41.php
@@ -36,7 +36,8 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
 	// Iterate through moderator log and remove slashes
-	for ($x = 1; $x < $db->num_rows('moderatorlog'); $x++) {
+	for ($x = 1; $x < $db->num_rows('moderatorlog'); $x++)
+	{
 
 		$moderatorlogitemarray = $db->simple_select("moderatorlog", "*", "tid='".$x."'", array(
 			"limit" => 1
--- a/install/resources/upgrade41.php
+++ b/install/resources/upgrade41.php
@@ -35,28 +35,21 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "DuckDuckGo", "useragent" => "DuckDuckBot/1.1; (+http://duckduckgo.com/duckduckbot.html)"));
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
-	// Iterate through moderator log and remove slashes
-	for ($x = 1; $x < $db->num_rows('moderatorlog'); $x++)
+	// Remove backslashes from last 1,000 log files.
+	$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\\\\\%", "limit" => 1000);
+
+	while($row = $db->fetch_array($query))
 	{
+    $original = $row['action'];
+    $stripped = stripslashes($original);
 
-		$moderatorlogitemarray = $db->simple_select("moderatorlog", "*", "tid='".$x."'", array(
-			"limit" => 1
-		));
+    if($stripped !== $original)
+    {
+        $db->update_query("moderatorlog", "action" => stripped, "WHERE tid = '".$row['tid']."'");
+    }
 
-		$update_log_array = array(
-			"uid" => $moderatorlogitemarray[0],
-			"dateline" => $moderatorlogitemarray[1],
-			"fid" => $moderatorlogitemarray[2],
-			"tid" => $moderatorlogitemarray[3],
-			"pid" => $moderatorlogitemarray[4],
-			"action" => stripslashes($moderatorlogitemarray[5]),
-			"data" => $moderatorlogitemarray[6],
-			"ipaddress" => $moderatorlogitemarray[7]
-		);
-
-		$db->update_query("moderatorlog", $update_log_array, "WHERE tid = '".$x."'");
 	}
-	
-  	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
+
+  $output->print_contents("<p>Click next to continue with the upgrade process.</p>");
 	$output->print_footer("41_done");
 }
--- a/install/resources/upgrade41.php
+++ b/install/resources/upgrade41.php
@@ -36,7 +36,7 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
 	// Iterate through moderator log and remove slashes
-	for ($x = 0; $x < $db->num_rows('moderatorlog'); $x++) {
+	for ($x = 1; $x < $db->num_rows('moderatorlog'); $x++) {
 
 		$moderatorlogitemarray = $db->simple_select("moderatorlog", "*", "tid='".$x."'", array(
 			"limit" => 1
--- a/install/resources/upgrade41.php
+++ b/install/resources/upgrade41.php
@@ -35,6 +35,27 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "DuckDuckGo", "useragent" => "DuckDuckBot/1.1; (+http://duckduckgo.com/duckduckbot.html)"));
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
+	// Iterate through moderator log and remove slashes
+	for ($x = 0; $x < $db->num_rows('moderatorlog'); $x++) {
+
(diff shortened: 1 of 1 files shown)


install/resources/upgrade41.php
```diff
@@ -35,6 +35,27 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "DuckDuckGo", "useragent" => "DuckDuckBot/1.1; (+http://duckduckgo.com/duckduckbot.html)"));
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
+	// Iterate through moderator log and remove slashes
+	for ($x = 1; $x < $db->num_rows('moderatorlog'); $x++) {
```
Shade-: Can you move the curly bracket on a new line please?

nsuchy:
@Shade- Done.

install/resources/upgrade41.php
```diff
@@ -35,6 +35,28 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "DuckDuckGo", "useragent" => "DuckDuckBot/1.1; (+http://duckduckgo.com/duckduckbot.html)"));
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
+	// Iterate through moderator log and remove slashes
+	for ($x = 1; $x < $db->num_rows('moderatorlog'); $x++)
+	{
+
+		$moderatorlogitemarray = $db->simple_select("moderatorlog", "*", "tid='".$x."'", array(
```
Shade-: This is not efficient as it queries the db for each row, which may be problematic for tables with thousands rows.

```php
$query = $db->simple_select("moderatorlog", "*");
while ($log = $db->fetch_array($query))
{
    // Do stuff
}
```

This would be better. However, there's still room for improvement in the update_query() part, which is also ran for each row. I'm a bit busy at the moment so I can't write down the complete function I'm afraid.

euantorano:
I wonder if this can be simplified to do the replacement in the DB engine itself, using (for example) the MySQL REPLACE function.

install/resources/upgrade41.php
nsuchy: To confirm you'd like me to switch to a while loop and update every row that way? Keep in mind that regardless of how we do it every row will still be read and if applicable updated. However it could be adjusted to only update if the row contains a slash which might be the code slightly more efficient although then that ends up with another check on each row which could slow things down even further. Thoughts?

Shade-: See @euantorano's reply below :)

nsuchy:
@euantorano This is currently written with MyBB's default DB Methods so all databases are supported. It's possible to use a single engine's replace function however users not on MySQL could have compatibility issues unless every engine has a separate query which again its possible but it could make the code less efficient and harm users with 1,000s of rows.

euantorano:
Postgres and SQLite both have REPLACE functions too :) Only problem I can see is working out all of the necessary replacements. 

Even if we don't use the replacement in the DB, it's best to use one query to load the rows that contain any \ character and then only do the update if the stripped version is different

nsuchy:
@euantorano To confirm are the only supported databases MySQL, Postgres, and SQLLite? That sounds do-able.

Selecting with a `%\%' query would return ALL rows with a backslash in it which could again have a better or worse performance drawback depending on how many 1000 rows have a \.

The applicable Postgres and SQLLite queries also need to be written if we are taking this route. Is there an easy way for the upgrade script to check which DB Engine is in use to determine which query to run?

euantorano:
Yep, they're the only supported engines. You can easily tell which engine is in use using something like $db->db_type (I think, off the top of my head).

nsuchy:
@euantorano I'm not seeing the method in https://github.com/nsuchy/mybb/blob/feature/inc/db_base.php however assuming it's hidden away elsewhere (See: https://github.com/mybb/mybb/search?utf8=%E2%9C%93&q=db_type&type= where it also doesn't appear to exist) would something like this work? If not we could always check the config and grab (`$mybb->config['database']['type']`). Thoughts?

```
if ($db->db_type == mysql)
{
// do stuff
}
elseif ($db->db_type == mysqli)
{
// do stuff
}
elseif ($db->db_type == pgsql)
{
// do stuff
}
elseif ($db->db_type == sqlite) {

}
else
{
/* Display an error that the database type couldn't be determined */
}
```

euantorano:
Sorry, it's `$db->type`. For an example of its usage, see here: https://github.com/mybb/mybb/blob/141bd8565e922470132cff87d3885c815a158937/install/resources/upgrade14.php#L40

If `REPLACE` doesn't work or doesn't work well, try something like:

```php
$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\%");

while($row = $db->fetch_array($query))
{
    $original = $row['action'];
    $stripped = stripslashes($original);

    if($stripped !== $original)
    {
        // Do the update
    }
}
```

nsuchy:
@euantorano The replace syntax on Postgres is kinda obscure. As you wrote:
```
$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\%");

while($row = $db->fetch_array($query))
{
    $original = $row['action'];
    $stripped = stripslashes($original);

    if($stripped !== $original)
    {
        // Do the update
    }
}
```
That is a more readable solution. I'd prefer to use that method rather than a separate method for every database type (also it takes into account if mybb ever decides to support another database type, or someone adds their own through core edits on every update) If you and @Shade- are ok using that method I can update this PR to use that method.

euantorano:
Sounds fine to me, though perhaps we should limit the number of actions operated on. I believe moderator logs are cleaned using the task system, but we probably shouldn't process everything that's in the DB to save some time - maybe just do the first 100 or so? Any thoughts @mybb/developers @mybb/sqa?

nsuchy:
@euantorano I would say if we limit the moderator logs it should be at least 1000 for people who have more log files and heavily audit them, beyond that in most cases you won't read past last 1000 log entries, that that do probably have a large forum and competent enough admins to run this query manually without the limit option if the backslashes bother them and they're reading 10,000s of log entries. Thoughts?

nsuchy:
@euantorano Also for `$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\%");` should there be two or one backslash in the action LIKE statement? (not sure if that was a typo or intentional)

euantorano:
I think two, off the top of my head otherwise a single `\` would just be an escape sequence.

nsuchy:
@euantorano Oh that makes sense. Thanks for clarifying, once MyBB Developers and SQA teams respond with any needed information. I can adjust the code as needed then update the PR. I'll be waiting :)

WildcardSearch:
@nsuchy https://dev.mysql.com/doc/refman/5.6/en/string-comparison-functions.html

Because MySQL uses C escape syntax in strings (for example, “\n” to represent a newline character), you must double any “\” that you use in LIKE strings. For example, to search for “\n”, specify it as “\\n”. To search for “\”, specify it as “\\\\”; this is because the backslashes are stripped once by the parser and again when the pattern match is made, leaving a single backslash to be matched against.

nsuchy:
@WildcardSearch Thank you for clarifying. Now I just need to know what Developers and SQA team think about the limit and what to set it to.

WildcardSearch:
@nsuchy reread that-- I had problems with GitHub stripping slashes from my post #irony

nsuchy:
@WildcardSearch From what I understand you need two backslashes to search for one backslash. And yes the irony was pretty funny.

WildcardSearch:
@euantorano in my tests (using threads titles) this is what actually works:

```php
$query = $db->simple_select('threads', 'subject', "subject LIKE '%\\\\\\%'");
$threadList = 'Thread List:';
while ($subject = $db->fetch_field($query, 'subject')) {
	$threadList .= "<br />{$subject}";
}

echo($threadList);
```

Using one slash, two, or four yielded no results. This is weird...

nsuchy:
@WildcardSearch That's interesting, let's find out why that behavior occurs. Which database type was used in that test? (or does the same happen with every database type?)

nsuchy:
Taking into account the feedback I've received here, I've pushed a new version of the code for further review and discussion.

euantorano:
@WildcardSearch You need so many backslashes because they get parsed by PHP as escape sequences, so you have to escape both for PHP and for MySQL

install/resources/upgrade41.php
```diff
@@ -35,6 +35,21 @@ function upgrade41_dbchanges()
...
+	$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\\\\\%", "limit" => 1000);
+
+	while($row = $db->fetch_array($query))
+	{
+    $original = $row['action'];
+    $stripped = stripslashes($original);
+
+    if($stripped !== $original)
+    {
+        $db->update_query("moderatorlog", "action" => stripped, "WHERE tid = '".$row['tid']."'");
```
euantorano: Missing `$` before `$stripped`. Should maybe cast `$row['tid']` to `(int)` just to be sure.

`$stripped` should also have `$db->escape_string()` ran on it.

install/resources/upgrade41.php
```diff
@@ -35,6 +35,21 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "DuckDuckGo", "useragent" => "DuckDuckBot/1.1; (+http://duckduckgo.com/duckduckbot.html)"));
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
-  	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
+	// Remove backslashes from last 1,000 log files.
+	$query = $db->simple_select('moderatorlog', '*', "action LIKE '%\\\\\\%", "limit" => 1000);
```
euantorano: If all we're using is `tid` and `action`, no need to `SELECT *`, use `$db->simple_select('moderatorlog', 'tid, action', "action LIKE '%\\\\\\%", "limit" => 1000);` instead.

Also should order by `tid` descending.

install/resources/upgrade41.php
```diff
@@ -35,6 +35,21 @@ function upgrade41_dbchanges()
...
+
+	while($row = $db->fetch_array($query))
+	{
+    $original = $row['action'];
```
euantorano: Indentation here looks slightly off. My example code used 4 spaces, the core uses tabs.

Azareal:
Have you considered using a subquery instead of a loop? It would be faster, would avoid loading a thousand rows into PHP, would use less I/O, and would generally be safer.

`$db->update_query("moderatorlog", "action" => stripped, "WHERE tid = '".$row['tid']."'");`
Also, you forgot a dollar sign.

kanadaj:
Wouldn't it be better to simply check if the database supports UPDATE queries, and if the type is not on the list of known DBs fall back to the loop? In fact, even a simply try-catch should do the job I believe. For the databases that support it, this would be much faster.

WildcardSearch:
> Which database type was used in that test?

@nsuchy MySQLi

@euantorano I just wanted someone else to confirm this odd query I was telling nscuchy to code lol

install/resources/upgrade41.php
nsuchy: There's no way for a user to modify 'tid' so it'd just add extra processing time. If you a certain you want it (int)ed I'll make the change, however good catch on $stripped.

install/resources/upgrade41.php
nsuchy: I'll fix that.

install/resources/upgrade41.php
nsuchy: I'm making those adjustments now.

nsuchy:
@euantorano I just pushed a commit with the requested changes.

install/resources/upgrade41.php
```diff
@@ -35,6 +35,24 @@ function upgrade41_dbchanges()
...
+	$query = $db->simple_select('moderatorlog', 'tid, action', "action LIKE '%\\\\\\%", array(
+		"order_by" => 'tid',
+		"order_dir" => 'DESC',
+		"limit" => 1000
+	));
+
+	while($row = $db->fetch_array($query))
+	{
+		$original = $row['action'];
+		$stripped = $db->escape_string(stripslashes($original));
```
Shade-: `escape_string()` should be ran afterwards, possibly inside `update_query()`. Otherwise, `$stripped` might be different from `$original` even though it might have not been touched by `stripslashes`.

nsuchy: Resolved.

euantorano:
Looks good to me. Any more comments before merging @mybb/developers?

install/resources/upgrade41.php
```diff
@@ -35,6 +35,24 @@ function upgrade41_dbchanges()
 	$db->insert_query('spiders', array("name" => "DuckDuckGo", "useragent" => "DuckDuckBot/1.1; (+http://duckduckgo.com/duckduckbot.html)"));
 	$db->insert_query('spiders', array("name" => "UptimeRobot", "useragent" => "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)"));
 
-  	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
+	// Remove backslashes from last 1,000 log files.
+	$query = $db->simple_select('moderatorlog', 'tid, action', "action LIKE '%\\\\\\%", array(
```
JordanMussi: Closing single quote for LIKE statement?

euantorano:
Oops, good catch! That needs fixing.

install/resources/upgrade41.php
```diff
@@ -35,6 +35,24 @@ function upgrade41_dbchanges()
...
+	));
+
+	while($row = $db->fetch_array($query))
+	{
+		$original = $row['action'];
+		$stripped = stripslashes($original);
+
+		if($stripped !== $original)
+		{
+			$db->update_query("moderatorlog", "action" => $db->escape_string($stripped), "WHERE tid = '".$row['tid']."'");
```
JordanMussi: Invalid syntax here... needs to be an array.
`array('action' => $db->escape_string($stripped))`

JordanMussi: @dvz do you want to fix this in #2850?

dvz: Will do

PR #2851, opened 2017-09-23
allow users to access contact.php when board status is set to closed

kawaii:
This just seemed logical to me. Perhaps it could do with having a setting created but I don't see why this shouldn't just be allowed.

### DIFF
--- a/global.php
+++ b/global.php
@@ -966,6 +966,7 @@ $closed_bypass = array(
 		'logout',
 	),
 	'captcha.php',
+	'contact.php',
 );
 
 // If the board is closed, the user is not an administrator and they're not trying to login, show the board closed message


kawaii:
Reasonable suggestions. :slightly_smiling_face: If I get the time I'll amend this PR, but I'm not a native PHP developer so it's probably best that someone else picks this up from inside the team.

JordanMussi:
It's the same as what you've already done, just a few lines further down with the `$force_bypass` variable.

Ben-MyBB:
@kawaii Any update on this?

kawaii:
@Ben-MyBB I think this PR needs to be closed and I need to remake it with the suggested modifications since I deleted the original fork and can't figure out the correct git commands to extend this PR itself.

PR #2857, opened 2017-10-02
Fix #2614 - Hooks with same class method of different objects not being added

### LINKED ISSUE #2614: Hooks with same class method of different objects not being added
Hooks that call the same method, although using different objects, are not being registered.

&#36;method_representation is identical for different objects:
[https://github.com/mybb/mybb/blob/featur...ns.php#L68](https://github.com/mybb/mybb/blob/feature/inc/class_plugins.php#L68)

which causes the new hook to be rejected here:
[https://github.com/mybb/mybb/blob/featur...hp#L76-L79](https://github.com/mybb/mybb/blob/feature/inc/class_plugins.php#L76-L79)

(a plugin developer's solution would be to assign them at different priorities)

Original thread: [Hooks with same class method of different objects not being added](https://community.mybb.com/thread-173799.html)

/fxt links -wc

euantorano:
Fixes #2614 

Rather than using `get_class`, use [`spl_object_hash`](http://php.net/manual/en/function.spl-object-hash.php):

> This function returns a unique identifier for the object. This id can be used as a hash key for storing objects, or for identifying an object, as long as the object is not destroyed. Once the object is destroyed, its hash may be reused for other objects.

### DIFF
--- a/inc/class_plugins.php
+++ b/inc/class_plugins.php
@@ -60,15 +60,18 @@ class pluginSystem
 			}
 
 			if(is_string($function[0]))
-			{ // Static class method
+			{
+				// Static class method
 				$method_representation = sprintf('%s::%s', $function[0], $function[1]);
 			}
 			elseif(is_object($function[0]))
-			{ // Instance class method
-				$method_representation = sprintf('%s->%s', get_class($function[0]), $function[1]);
+			{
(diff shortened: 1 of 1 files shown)


Shade-:
Tested, works as intended.

PR #2174, opened 2015-07-31
Fixes #2172 - Stopforumspam IP ranges

Cu8eR:
Fixes #2172  - Stopforumspam Reserved and Private IP ranges

PR #2858, opened 2017-10-02
Fix #2172 - Stopforumspam Reserved and Private IP ranges

### LINKED ISSUE #2172: Stopforumspam Reserved and Private IP ranges
This is in relation to the file of inc/stopforumspam.php

Currently a spammer ip isn't tested again Private or Reserved IP ranges which means if the IP is spoofed that either of those could attempt to be used to check the stopforumspam.com database.
This likely doesn't cause any errors but it will decrease available resources at the site for checking spammer IP,s and  will also unnecessarily use up a little bandwidth.

adding two filter inclusions will remove this problem, those filters are:
 FILTER_FLAG_NO_PRIV_RANGE & FILTER_FLAG_NO_RES_RANGE **

These filters Fail validation for the following:
Private IPv4 ranges: 10.0.0.0/8, 172.16.0.0/12 and 192.168.0.0/16.
Reserved IPv4 ranges: 0.0.0.0/8, 169.254.0.0/16, 192.0.2.0/24 and 224.0.0.0/4. 
*\* This flag does not apply to IPv6 addresses. 

Information on filters from: http://php.net/manual/en/filter.filters.flags.php 

The current conditional looks like this:

```
if(filter_var($email, FILTER_VALIDATE_EMAIL) && filter_var($ip_address, FILTER_VALIDATE_IP)) // Calls to the API with invalid email/ip formats cause issues
        {
```

It just needs editing to this:

```
if(filter_var($email, FILTER_VALIDATE_EMAIL) && filter_var($ip_address, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE)) // Calls to the API with invalid email/ip formats cause issues
        {
```

Credited Source: 
The Originally mention of the filter additions is in the comments of: 
http://php.net/manual/en/function.filter-var.php

euantorano:
Fixes #2172

This adds the `FILTER_FLAG_NO_PRIV_RANGE` and `FILTER_FLAG_NO_RES_RANGE` to `filter_var` when validating that the given IP address is a valid IP. It also adds a setting to determine whether a user is marked as a spammer in case the email or IP address validation check fails (`stopforumspam_block_on_invalid_email_or_ip`).


dvz:
Does it make sense to add a setting when the e-mail address values are validated on change (by `UserDataHandler`)? Stored IP addresses are treated as trusted information, so users with private ones would indicate a misconfiguration (or a security issue), but shouldn't be classified as spammers.

euantorano:
Do you mean to add another additional setting (not quite following I don't think)?

I added a new setting based upon the conversation in #2174 trying to fix this issue:

> I don't think the user should be validated when that if fails BTW, an else with $is_spammer set to true would be nice.
>
> But then I'm not sure if that's a good idea together with the weak email validation and/or testing in localhost. Could probably make it optional with a new SFS setting.

dvz:
> Do you mean to add another additional setting (not quite following I don't think)?

IMO no setting should be added - invalid e-mail and IP addresses make it impossible to validate on SFS, but that shouldn't affect the validation (ideally an exception would be thrown).
Incorrect e-mail addresses should be caught earlier and private IP ranges shouldn't occur in a properly configured environment.

euantorano:
So if the email is invalid or the IP is invalid, throw an exception and block registration/posting/whatever is being checked? At the moment if either is invalid, then they are not blocked and the check isn't executed.

dvz:
Exceptions are already handled according to the `stopforumspam_block_on_error` setting value and same fallback can work for incorrect values that should only be a result of internal (post-validation) inconsistencies.

euantorano:
Fair enough. I'll amend this PR to throw an error when the email or IP is invalid. I'll remove the check for internal IPs too.

dvz:
> I'll remove the check for internal IPs too

If you intend to skip the SFS validation for private ranges, sounds good - there is no point in submitting those.

euantorano:
Yes, that's what I intend.

euantorano:
Closing this, as my working copy is a little broken...

PR #2859, opened 2017-10-02
Fix #2367 - ACP not using format_avatar() for formatting avatars

### LINKED ISSUE #2367: ACP not using format_avatar() for formatting avatars
As the title says and as mentioned here: https://github.com/mybb/mybb/pull/2364 Shouldn't be too hard too change, just the extraordinary scaling needs to be considered. Also `../` before avatars that use relative path, which could be added in `format_avatar()` if `IN_ADMINCP` is defined.

euantorano:
Fixes #2367

Use `format_avatar` in the ACP user list. I couldn't find any other obvious occurrences of avatars being used in the ACP, but please point them out if I've missed any.

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -3614,30 +3614,18 @@ function build_users_view($view)
 				$user['view']['warninglevel'] = get_colored_warning_level($warning_level);
 			}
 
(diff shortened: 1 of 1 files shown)


Shade-:
Works fine.

PR #2862, opened 2017-10-03
Fix #2861 External anchors do not use rel="noopener"

### LINKED ISSUE #2861: External anchors do not use rel="noopener"
https://developers.google.com/web/tools/lighthouse/audits/noopener

h/t [Clickbait](https://community.mybb.com/user-117364.html)

dvz:
#2861


inc/class_parser.php
```diff
@@ -1064,7 +1064,11 @@ function mycode_parse_url($url, $name="")
 		$nofollow = '';
 		if(!empty($this->options['nofollow_on']))
 		{
-			$nofollow = " rel=\"nofollow\"";
+			$nofollow = " rel=\"noopener nofollow\"";
+		}
+		else
+		{
+			$nofollow = " rel=\"noopener\"";
```
JordanMussi: Better to have this in [the template](https://github.com/dvz/mybb/blob/23405ae39351af0a7212c860f496857da9f37f53/install/resources/mybb_theme.xml#L8718) where the `target="_blank"` is to avoid confusion?

admin/inc/class_page.php
```diff
@@ -829,11 +829,16 @@ function output_nav_tabs($tabs=array(), $active='')
 			{
 				$target = " target=\"{$tab['link_target']}\"";
 			}
+			$rel = '';
```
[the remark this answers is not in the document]
dvz: That might break some core/extension JS behavior (in case of internal URLs access to `window.opener` can be needed)

inc/class_parser.php
dvz: If we chose to edit the template a separate variable (like `{$rel}`, with hardcoded value) would be a better solution so it doesn't possibly have to be modified again - I didn't go this way yet to make sure want that given recent problems with themes that haven't been updated (and compatibility with e.g. http://community.mybb.com/mods.php?action=view&pid=450 that parses those).

admin/inc/class_page.php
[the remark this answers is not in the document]
JordanMussi: Ah yes of course!

inc/class_parser.php
JordanMussi: Yep as you have done in 03de79e1f99c58857683b6fa531a9e584b0b9674 is fine then :smiley:

PR #2867, opened 2017-10-07
Fix #2865 - SQL error with PostgreSQL when running reputation recount

### LINKED ISSUE #2865: SQL error with PostgreSQL when running reputation recount
```
SQL Error:

0 - ERROR: operator does not exist: ` integer LINE 3: WHERE `uid`='1' ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.
Query:
SELECT SUM(reputation) as total_rep FROM mybb_reputation WHERE `uid`='1'
```

Steps to reproduce:

- MyBB 1.8.12
- PostgreSQL backend

Run the 'recount reputation' feature in the ACP, not sure if it's specific to this issue but no users on my board had any reputation points at all.

Original thread: [SQL error with PostgreSQL when running reputation recount](https://community.mybb.com/thread-213180.html)

euantorano:
Fixes #2865.

PostgreSQL doesn't use backtick characters to quote identifiers.

### DIFF
--- a/admin/modules/tools/recount_rebuild.php
+++ b/admin/modules/tools/recount_rebuild.php
@@ -236,7 +236,7 @@ function acp_recount_reputation()
 		$query2 = $db->query("
 			SELECT SUM(reputation) as total_rep
 			FROM ".TABLE_PREFIX."reputation
-			WHERE `uid`='{$user['uid']}'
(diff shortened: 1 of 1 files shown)


PaulBender:
Works for me

PR #2868, opened 2017-10-07
Fix #2863 - Improve Twitch Video MyCode

### LINKED ISSUE #2863: Improve Twitch Video MyCode
The default MyCode for embedding Twitch is not only somewhat useless but misleading. One might think that you can embed a livestream using it, but no - you can only embed a previous broadcast video due to how the regex works.

Original thread: https://community.mybb.com/thread-211032.html

euantorano:
Fixes #2863 

This improves the Twitch video MyCode in the following ways:

- Supports new video URLs of the form `https://www.twitch.tv/videos/179723472`
- Supports old video URLs of the form `https://www.twitch.tv/waypoint/v/179723472`
- Support Twitch channel embeds using URLs of the form `https://www.twitch.tv/waypoint`

This modifies the `video_twitch_embed` template to differentiate between video and channel embeds.


Shade-:
Works fine.

PR #2871, opened 2017-10-14
trow1 should be trow2 - member_resetpassword

Eldenroot:
trow1 should be trow2 - member_resetpassword


WildcardSearch:
Looks like a straight-forward fix. Good work @Eldenroot

PR #2872, opened 2017-10-14
extension of PR #2851

kawaii:
#2851

@JordanMussi asked me to extend my PR allowing access to the contact page when the board is closed, but since I'd deleted the original forked repository I couldn't figure out the git magic to extend the original PR, so I've created this new one. #2851 can be closed.

### DIFF
--- a/global.php
+++ b/global.php
@@ -966,7 +966,7 @@ $closed_bypass = array(
 		'logout',
 	),
 	'captcha.php',
-	'contact.php,
+	'contact.php',
 );
 
 // If the board is closed, the user is not an administrator and they're not trying to login, show the board closed message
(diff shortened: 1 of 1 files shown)


global.php
```diff
@@ -966,6 +966,7 @@
 		'logout',
 	),
 	'captcha.php',
+	'contact.php,
```
Shade-: Missing a single quote at the end of the string, just before the coma.

global.php
```diff
@@ -1008,6 +1009,7 @@
 		'resetpassword',
 	),
 	'captcha.php',
+	'contact.php,
```
Shade-: Same here.

kawaii:
@Shade- fixed, sorry I was tired. ;)

JordanMussi:
Thanks @kawaii! 😀

PR #2873, opened 2017-10-15
Fix emails from contact.php being rejected by correctly configured mail servers

### LINKED ISSUE #257: Change to SMTP mail composition to comply with standards

kawaii:
This fixes #257 and another few issues I've seen related to correctly configured SMTP servers rejecting mail from MyBB's contact form due to it trying to spoof an address that the SMTP server has no control over.

### DIFF
--- a/contact.php
+++ b/contact.php
@@ -223,7 +223,7 @@ if($mybb->request_method == "post")
 		$message = $lang->sprintf($lang->email_contact, $mybb->input['email'], $user, $session->ipaddress, $mybb->input['message']);
 
 		// Email the administrator
-                my_mail($contactemail, $subject, $message, $mybb->settings['adminemail']);
+		my_mail($contactemail, $subject, $message, $mybb->settings['adminemail']);
 
 		$plugins->run_hooks('contact_do_end');
 


euantorano:
This removes the use for the `email` input on the contact page. I see a few options:

- Remove that input field (meaning that admins might not be able to email the user back
- Add a setting to allow spoofing of from address
- Append the `email` field to the body of the email

Thoughts @mybb/developers @mybb/sqa

kawaii:
Appending the email to the body of the message seems like the sanest solution to me.

contact.php
```diff
@@ -223,7 +223,7 @@
 		$message = $lang->sprintf($lang->email_contact, $mybb->input['email'], $user, $session->ipaddress, $mybb->input['message']);
 
 		// Email the administrator
-		my_mail($contactemail, $subject, $message, $mybb->input['email']);
+                my_mail($contactemail, $subject, $message, $mybb->settings['adminemail']);
```
Shade-: The indentation seems to be incorrect. We use tabs, not spaces.

euantorano:
Actually the best option should be to set the `Reply-To` mail header using the `$return_email` parameter of `my_mail`: https://github.com/mybb/mybb/blob/3902684d7c974ad4a9ececdd049abdc216ff7f3d/inc/functions.php#L566

euantorano:
Updated PR in #2877 using the `$return_email` parameter.

PR #2876, opened 2017-10-16
Fixes #2875

### LINKED ISSUE #2875: Registration impossible with Custom Profile Field using regex
Every now and then we got informed by users not being able to register because of a some Custom Profile Fields on our registration page (we use 2 additional fields). Error message: 'You did not enter a valid value for the "Shop Code" field. Please enter a valid value before continuing.'

None of our custom fields were ticked 'required' so we finally decided to pull them from our registration page just in case.
You can imagine our disbelief when we realized, when we got yet another email that registration would be impossible because of said custom field still being 'required', yet not even being displayed.

Since this wasn't true for the other field, I tried to find a pattern and I think I found it:

- Custom profile field 1 ('required' not ticked): uses regex with `[a-zA-Z]` -> unintentionally 'required'
- Custom profile field 2 ('required' not ticked): does not use regex -> works as advertised as not being 'required'

Immediately after deleting the regex of Custom Profile Field 1, everything was working fine again!
My guess would be, that the system checks for letters in regex Custom Field and therefore rejects an empty field?
I mean it mages sense regarding the regex, but why is a NULL entry being parset at all?
And if the field isnt enabled on the registration page, this is even more dangerous imho.

Original thread: [Registration impossible with Custom Profile Field using regex](https://community.mybb.com/thread-209338.html)

Eldenroot:
Fixes #2875 

All credits go to @Shade, thank you!

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -574,7 +574,7 @@ class UserDataHandler extends DataHandler
 						$this->set_error('max_limit_reached', array($profilefield['name'], $profilefield['maxlength']));
 					}
 
-					if(!empty($profilefield['regex']) && !preg_match("#".$profilefield['regex']."#i", $profile_fields[$field]))
+					if(!empty($profilefield['regex']) && !empty($profile_fields[$field]) && !preg_match("#".$profilefield['regex']."#i", $profile_fields[$field]))
 					{
 						$this->set_error('bad_profile_field_value', array($profilefield['name']));
 					}


euantorano:
I'm not 100% sure this is the best approach, as I said in the linked issue:

> I'd perhaps not use that solution from and would instead check if the field is set to be displayed, and if not skip all checking for the field's value. If it's hidden, it's obviously not going to be filled in.

It should work this way, but it's not entirely obvious what the condition is doing if you come back to look at it later.

Shade-:
The thing is, if the profile field is visible, a regex is set but there is not an input from the user, the check is performed anyway and it returns the error. That's why I added a check for the field's emptiness.

euantorano:
What if the field is visible, and required but not filled in by the user - will it still return an error (as it should) with this patch? I haven’t tested this patch at all or looked at the code near it at all yet.

Shade-:
Yes, thanks to [this snippet](https://github.com/Eldenroot/mybb/blob/d16b8345c5d2643d25608a973a856f546c2f8eb5/inc/datahandlers/user.php#L528-L539) which is ran before.

euantorano:
Ah, fair enough. Should be good to go then :)

Ben-MyBB:
Has anyone done any proper testing on this? @Eldenroot

PR #2877, opened 2017-10-18
Fix #257 - Change to SMTP mail composition to comply with standards

### LINKED ISSUE #257: Change to SMTP mail composition to comply with standards

euantorano:
This partially fixes #257 by using the `$return_email` parameter of `my_mail` when sending an email from the `Contact Us` page.

This replaces #2873

### DIFF
--- a/contact.php
+++ b/contact.php
@@ -223,7 +223,7 @@ if($mybb->request_method == "post")
 		$message = $lang->sprintf($lang->email_contact, $mybb->input['email'], $user, $session->ipaddress, $mybb->input['message']);
 
 		// Email the administrator
-		my_mail($contactemail, $subject, $message, $mybb->input['email']);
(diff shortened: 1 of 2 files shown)


dvz:
Same changes could be made to the `emailuser` action: https://github.com/mybb/mybb/blob/mybb_1812/member.php#L2894

euantorano:
Yup, will add that.

euantorano:
Fixed. The changes in `member.php` should work, but I haven't had time to test them.

PR #2880, opened 2017-10-19
Fixes #2879 Password security validation incorrect from #2790

### LINKED ISSUE #2879: Password security validation incorrect from #2790
Password security validation was not properly implemented in #2790.

```javascript
return (value != $('#email').val() && value != $('#username').val()
	&& value.indexOf($('#email').val()) === false && value.indexOf($('#username').val()) === false
	&& $('#email').val().indexOf(value) === false && $('#username').val().indexOf(value) === false);
```

`value.indexOf($('#email').val())` for example will never return `false`, it returns `-1` when the needle is not found so always marks as invalid.

Even with that fixed, if the user hasn't filled out either the username or email field the password will be marked invalid as `value.indexOf($('#email').val())` would give `-1` because empty string (`''`) will always be found in string at position `0`.

JordanMussi:
Fixes #2879 Password security validation incorrect from #2790

### DIFF
--- a/member.php
+++ b/member.php
@@ -1267,9 +1267,14 @@ $(document).ready(function() {
 
 			$validator_javascript .= "
 	$.validator.addMethod('passwordSecurity', function(value, element, param) {
-		return (value != $('#email').val() && value != $('#username').val()
-			&& value.indexOf($('#email').val()) === false && value.indexOf($('#username').val()) === false
-			&& $('#email').val().indexOf(value) === false && $('#username').val().indexOf(value) === false);
+		return !(
+				$('#email').val() != '' && value == $('#email').val()
+				|| $('#username').val() != '' && value == $('#username').val()
+				|| $('#email').val() != '' && value.indexOf($('#email').val())
+				|| $('#username').val() != '' && value.indexOf($('#username').val())
+				|| $('#email').val() != '' && $('#email').val().indexOf(value)
+				|| $('#username').val() != '' && $('#username').val().indexOf(value)
+		);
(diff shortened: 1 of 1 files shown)


member.php
```diff
@@ -1267,9 +1267,14 @@
 
 			$validator_javascript .= "
 	$.validator.addMethod('passwordSecurity', function(value, element, param) {
-		return (value != $('#email').val() && value != $('#username').val()
-			&& value.indexOf($('#email').val()) === false && value.indexOf($('#username').val()) === false
-			&& $('#email').val().indexOf(value) === false && $('#username').val().indexOf(value) === false);
+		return !(
+				$('#email').val() != '' && value == $('#email').val()
```
euantorano: I wonder if it makes sense to break each of these branches out into individual variables to make it easier to read and follow (and prevent similar errors in the future)...

JordanMussi: Quite possibly, although it would end up looking pretty much the same here so comments might be a more practical idea?

euantorano: Sure, comments work too.

euantorano:
Testing this again, it doesn't seem quite right. As you try to sign up, fill in a username (no email), then start writing a password. It'll say:

> The password you entered is similar to either your username or email address. Please enter a stronger password.

This seems to happen no matter what I put in the password field.

JordanMussi:
Hmm - I'll have to look at this again later in the week. In the meantime perhaps remove the offending part of #2790 so that it cannot be included in the 1.8.13 release by accident?

euantorano:
I’ll have a look during this week too (hopefully today) and see if we can get it fixed quickly.

PR #2886, opened 2017-10-22
Fixes 2885

### LINKED ISSUE #2885: PGSQL, SQLite Database Classes Do Not Abide By DB_Base Interface
In `session::create_session()` `DB_Base::replace_query()` is used with two extra parameters that only affect the PGSQL database class.

https://github.com/mybb/mybb/blob/bf6e33fdfd2d93c4c696023f1f69cb9ae8024177/inc/class_session.php#L526

The definition in `db_base.php`:

https://github.com/mybb/mybb/blob/bf6e33fdfd2d93c4c696023f1f69cb9ae8024177/inc/db_base.php#L352

There are many other places in which `replace_query` is used with three to four arguments rather than the two parameters specific in `DB_Base`.

The best course of action, imo would be to edit the existing definition to include the optional parameters and edit all of the database class wrappers to include the optional arguments in their definitions.

If that course of action is agreed upon, I am happy to create the pull request.

WildcardSearch:
Fixes #2885 
Updated database wrapper interface definition for `::replace_query` and updated database classes accordingly.


inc/db_base.php
```diff
@@ -349,7 +349,7 @@ function rename_table($old_table, $new_table, $table_prefix=true);
 	 * @param string $table The table
 	 * @param array $replacements The replacements
```
euantorano: Doc comments need updating too to reflect the new parameters (applies to every class, not just this one).

WildcardSearch:
Documentation added.

PR #2889, opened 2017-10-23
Fixes #2879 - Password security validation incorrect from #2790

### LINKED ISSUE #2879: Password security validation incorrect from #2790

euantorano:
This is based on #2880. The only change is to check the return value of `indexOf` against `-1` as `-1` is `true` in JS. Pinging @JordanMussi.

This seems to work on a local board.


PaulBender:
Works for me

PR #2891, opened 2017-10-26
Use INSTALL_ROOT instead of MYBB_ROOT

chack1172:


Ben-MyBB:
@JordanMussi No issues with this one?

PR #2894, opened 2017-11-02
Add modal CSS for ACP

Eric-Jackson:


admin/styles/default/modal.css
```diff
@@ -1,9 +1,39 @@
+.blocker {
+	position: fixed;
+	top: 0; right: 0; bottom: 0; left: 0;
```
kisphp: I would put each declaration on it's own line.

admin/styles/default/modal.css
```diff
@@ -1,9 +1,39 @@
...
+	width: 100%; height: 100%;
```
kisphp: same here

Shade-:
Works.

PR #2896, opened 2017-11-03
Update Horde_Text_Diff_Engine_Native, fix array pointer issues

dvz:
fix #2797
https://github.com/mybb/mybb/issues/2743#issuecomment-341786421

based on patches shared on https://core.trac.wordpress.org/ticket/41526

### DIFF
--- a/inc/3rdparty/diff/Diff/Engine/Native.php
+++ b/inc/3rdparty/diff/Diff/Engine/Native.php
@@ -20,7 +20,7 @@
  *
  * Copyright 2004-2017 Horde LLC (http://www.horde.org/)
  *
- * See the enclosed file COPYING for license information (LGPL). If you did
+ * See the enclosed file LICENSE for license information (LGPL). If you did
  * not receive this file, see http://www.horde.org/licenses/lgpl21.
  *
  * @author  Geoffrey T. Dairiki <dairiki@dairiki.org>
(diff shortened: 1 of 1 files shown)


inc/3rdparty/diff/Diff/Engine/Native.php
```diff
@@ -196,18 +196,30 @@ protected function _diag ($xoff, $xlim, $yoff, $ylim, $nchunks)
                     continue;
                 }
                 $matches = $ymatches[$line];
-                foreach ($matches as $y) {
+                reset($matches);
+                while ($y = current($matches)) {
```
euantorano: Reading [the docs for `current`](http://php.net/manual/en/function.current.php) and the return value, shouldn't we explicitly check for `$y !== false` here?

inc/3rdparty/diff/Diff/Engine/Native.php
```diff
@@ -196,18 +196,30 @@ protected function _diag ($xoff, $xlim, $yoff, $ylim, $nchunks)
...
                     if (empty($this->in_seq[$y])) {
                         $k = $this->_lcsPos($y);
                         assert($k > 0);
                         $ymids[$k] = $ymids[$k - 1];
                         break;
-                    } elseif ($y > $this->seq[$k - 1]) {
+                    }
+                    next($matches);
+                }
+                while ($y = current($matches)) {
```
euantorano: Again, shouldn't we explicitly check `$y !== false`?

inc/3rdparty/diff/Diff/Engine/Native.php
dvz: Indeed; cc @johnbillion @SergeyBiryukov

euantorano:
Looks good to me, but haven't tested yet.

Stefan-ST:
Works for me.

PR #2900, opened 2017-11-08
Fixes #309 Memberlist searching case sensitive with PostrgreSQL

### LINKED ISSUE #309: Memberlist searching case sensitive with PostrgreSQL
> In PostgreSQL the LIKE keyword is case sensitive. Because of this the letters or search function becomes case sensitive. As far as I know the only way to fix this is to wrap both sides of the WHERE/LIKE clause with upper/lower.
> 
> On a side note a category for starting with a symbol might be handy.

Original report: http://dev.mybb.com/issues/2104
Reporter: Braden Obrzut
Status: New

PaulBender:
Fixes #309

### DIFF
--- a/memberlist.php
+++ b/memberlist.php
@@ -158,6 +158,16 @@ else
 	$search_query = '1=1';
 	$search_url = "";
 
+	switch($db->type)
+	{
+		// PostgreSQL's LIKE is case sensitive
+		case "pgsql":
+			$like = "ILIKE";
+			break;
+		default:
+			$like = "LIKE";
+	}
+
 	// Limiting results to a certain letter
 	if(isset($mybb->input['letter']))
 	{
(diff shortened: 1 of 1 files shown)


kawaii:
I had some time this evening (and happen to run a board using PostgreSQL backend myself) so I've tested this PR. I downloaded the provided `patch` file [here](https://github.com/mybb/mybb/commit/6ab8717a7559e0b9b1538510d2e03ebc08e1b298.patch) and ran the following command:
```
patch < memberlist.patch
```
I verified that the changes applied correctly however in MyBB 1.8.13 it does not seem to have resolved the issue. Perhaps someone else can confirm that my testing methodology isn't flawed?

PaulBender:
It might be your testing methodology as the fix is working fine for me. This fix is supported by [Postgres's own documentation](https://www.postgresql.org/docs/current/static/functions-matching.html).

PR #2908, opened 2017-11-18
#2907 - Add Composer configuration

### LINKED ISSUE #2907: Add composer to manage third party libraries and autoloading.
We currently manage third party dependencies by downloading the sources and copying them into the [`inc/3rdparty`](https://github.com/mybb/mybb/tree/develop/1.10/inc/3rdparty) directory. This is a time consuming effort and causes libraries to only be updated when a contributor takes the time to visit the site for each dependency and realise that there is an updated version available.

The [composer](https://getcomposer.org/) project provides a dependency manager for PHP projects. This allows us to specify any 3rd party dependencies in a simple JSON file, and install and update them easily using the `composer` command line utility. It also generates autoload files to automatically load classes and functions from these third party libraries, as well as our own core classes.

euantorano:
Fixes #2907 by adding a Composer configuration file and ammends the `.gitignore` to ignore the vendor directory.

Dependencies are set to be installed to the `inc/vendor` folder, with classes autoloaded following the PSR-4 standard from `inc/src`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ Thumbs.db
 
 # Generic git files
 *.orig
+
+inc/vendor
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,42 @@
+{
+    "name": "mybb/mybb",
+    "description": "MyBB is a free and open source, community-based forum software project.",
+    "type": "project",
+    "keywords": [
+        "forum",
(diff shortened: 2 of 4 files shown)


kawaii:
Confirmed that this is functioning correctly! :smile:

PR #2917, opened 2017-11-24
Changed default video http prefix to https

nsuchy:
This pull request changes the default video prefix from http to https.


Shade-:
Sorry, SCEditor will be updated to the latest version (2.0.0 or newer) which contains the fix you applied. Updating third party libraries on our own is not the best way to keep up with updates coming from their developers.

PR #2933, opened 2017-12-10
Fix #2932 Change validate_email_format to use filter_var

### LINKED ISSUE #2932: Change `validate_email_format` to use `filter_var`
The [`validate_email_format`](https://github.com/mybb/mybb/blob/bffd4ebde400a0da30d3cf509d3b0e64667a676d/inc/functions.php#L6428) function currently uses the following regular expression to check if an email address looks valid:

```
^[a-zA-Z0-9&*+\-_.{}~^\?=\/]+@[a-zA-Z0-9-]+\.([a-zA-Z0-9-]+\.)*[a-zA-Z0-9-]{2,}$
```

Using a regex for email validation is notoriously tricky, so it would make sense to instead use the built in [`filter_var`](http://php.net/manual/en/function.filter-var.php) function with `FILTER_VALIDATE_EMAIL`. We already use this combination elsewhere, such as [when changing a users avatar to a Gravatar in the ACP](https://github.com/mybb/mybb/blob/c022a16ff1b5d63ffb5123c501320948c48a8043/admin/modules/user/users.php#L637).

JoshHarmon:


JordanMussi:
#2932

euantorano:
Thanks. I did read a comment whilst looking at this somewhere that `FITLER_VALIDATE_EMAIL` may not be available in odler versions of PHP 5.2 (`< 5.2.29` if I remember correctly...). We need to check whether that's the case, and if so provide some kind of fallback (which is also needed anywhere else `FILTER_VALIDATE_EMAIL` is used.

JordanMussi:
Seems to me that `FILTER_VALIDATE_EMAIL` was added in PHP 5.2? I can't find any references to it not being in 5.2...

Looks like it was introduced in PHP 5.2 here:
https://github.com/php/php-src/blame/cb9d81ef4f07f82835273800b0cb3d6a67816050/ext/filter/filter.c#L49

From the [5.2 changelog](http://php.net/ChangeLog-5.php):
It was definitely around before PHP 5.2.9:
```
Version 5.2.9
...
PHP Fixed bug #47282 (FILTER_VALIDATE_EMAIL is marking valid email addresses as invalid). (Ilia)
```
And before PHP 5.2.1:
```
Version 5.2.1
...
Fixed the validate email filter so that the letter "v" can also be used in the user part of the email address. (Derick)
```
So if it was around before 5.2.1 it must have been in 5.2.0 surely?

euantorano:
Yep, sounds good to me. Will merge this.

PR #2934, opened 2017-12-11
Fixes #2920 - adds the Laravel Container

### LINKED ISSUE #2920: Add a Dependency Injection Container
The current MyBB code uses tightly coupled code and global variables for instances of classes. This leads to repetitive `global $mybb` lines and makes it difficult to mock and test individual components.

With a Dependency Injection Container, all services would be defined in the DI, allowing them to be easily swapped out and mocked. It would also make it easy to extend instances of classes (such as replacing the core `$mybb` instance with an extended `MyCustomMyBBCore` instance that adds additional methods) from within plugins.

## Choosing a DI

There are multiple DI libraries for PHP, all with similar features and approaches. As far as I see it, there are the following well known libraries to choose from that meet our requirements:

- [Pimple](https://pimple.symfony.com/) - a simple DI container, no support for automatic resolution of dependencies. Does not conform to PSR-11.
- [PHP-DI](http://php-di.org/) - a slightly more complex DI container with support for automatic resolution of dependencies based upon constructor type hints. Conforms to PSR-11.
- [Container by the League of Extraordinary packages](http://container.thephpleague.com/) - a simple but powerful DI container that conforms with PSR-11.
- [Laravel's Container](https://github.com/illuminate/container) - the core container from the Laravel framework. Documentation is only available as part of the Laravel documentation, [found here](https://laravel.com/docs/5.5/container). Conforms to PSR-11.
- [Symfony's DependencyInjection Component](https://github.com/symfony/dependency-injection) - the core container from the Symfony framework. Documentation is only available as part of the Symfony documentation, [found here](https://symfony.com/doc/current/components/dependency_injection.html). Conforms to PSR-11.

## Choice of library

Ideally we'd choose a library sooner rather than later, and preferrably it would be one that conforms to PSR-11 (which would make it easier to swap out in the future due to it following a defined interface).

euantorano:
Fixe #2920. This adds the `Illuminate\Container` package as a dependency, and also adds a new `inc/src/functions.php` file that contains funcitons under the `\MyBB\` namespace. These functions are incldued by the autolaoder automatically.

It defines a single function, which can be used to resolve classes or get the container instance:

```php
// Get an instance of the container
$container = \MyBB\app();

// Get an instance of a class with no contructor parameters
$instance = \MyBB\app('SomeClass');

// Get an instance of a class with a set of constructor parameters
$instance = \MyBB\app('SomeOtherClass', ['foo']);
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,14 +21,15 @@
         "issues": "https://github.com/mybb/mybb/issues"
     },
     "require": {
-        "php": "^5.6 || ^7.0"
+        "php": "^5.6 || ^7.0",
+        "illuminate/container": "^5.5"
     },
     "autoload": {
         "psr-4": {
             "MyBB\\": "inc/src/"
         },
         "files": [
-            "inc/functions.php"
+            "inc/src/functions.php"
         ]
     },
     "minimum-stability": "stable",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,194 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "ab509c7218859e6745963ef4457859f5",
-    "packages": [],
+    "content-hash": "ce198f53defdf572a82353710ba981ab",
+    "packages": [
+        {
+            "name": "illuminate/container",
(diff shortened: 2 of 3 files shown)


036:
Looks good Euan.

PR #2945, opened 2017-12-15
Implement laravels query builder

036:


euantorano:
This contains the changes for the config loader PR too, which will make his unmergeable when the config PR is merged.

PR #2946, opened 2017-12-16
Enhance Admin CP Plugins List

JordanMussi:
- Converts links to buttons
- Adds configure button (link set by the `_configure` plugin function) shown when plugin is installed and returns a configure link
- Documentation added to `hello` plugin

### DIFF
--- a/admin/modules/config/plugins.php
+++ b/admin/modules/config/plugins.php
@@ -542,11 +542,11 @@ if(!$mybb->input['action'])
 
 		$table = new Table;
 		$table->construct_header($lang->plugin);
-		$table->construct_header($lang->controls, array("colspan" => 2, "class" => "align_center", "width" => 300));
+		$table->construct_header($lang->controls, array("colspan" => 3, "class" => "align_center", "width" => 300));
 
 		if(empty($a_plugins))
 		{
-			$table->construct_cell($lang->no_active_plugins, array('colspan' => 3));
+			$table->construct_cell($lang->no_active_plugins, array('colspan' => 4));
 			$table->construct_row();
 		}
 		else
@@ -558,11 +558,11 @@ if(!$mybb->input['action'])
 
 		$table = new Table;
 		$table->construct_header($lang->plugin);
-		$table->construct_header($lang->controls, array("colspan" => 2, "class" => "align_center", "width" => 300));
+		$table->construct_header($lang->controls, array("colspan" => 3, "class" => "align_center", "width" => 300));
 
 		if(empty($i_plugins))
 		{
-			$table->construct_cell($lang->no_inactive_plugins, array('colspan' => 3));
+			$table->construct_cell($lang->no_inactive_plugins, array('colspan' => 4));
 			$table->construct_row();
 		}
 		else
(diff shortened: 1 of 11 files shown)


admin/modules/config/plugins.php
```diff
@@ -663,31 +665,47 @@ function build_plugin_list($plugin_list)
...
 			{
-				$table->construct_cell("<a href=\"index.php?module=config-plugins&amp;action=activate&amp;plugin={$plugininfo['codename']}&amp;my_post_key={$mybb->post_code}\">{$lang->install_and_activate}</a>", array("class" => "align_center", "colspan" => 2));
+				$table->construct_cell("<a href=\"index.php?module=config-plugins&amp;action=activate&amp;plugin={$plugininfo['codename']}&amp;my_post_key={$mybb->post_code}\" class=\"button\">{$lang->install_and_activate}</a>", array("class" => "align_center", "colspan" => 3));
 			}
 		}
 		// Plugin is activated and installed
 		else if($plugininfo['is_active'])
 		{
-			$table->construct_cell("<a href=\"index.php?module=config-plugins&amp;action=deactivate&amp;plugin={$plugininfo['codename']}&amp;my_post_key={$mybb->post_code}\">{$lang->deactivate}</a>", array("class" => "align_center", "width" => 150));
+			if ($configure_link !== false) {
```
euantorano: Code style - this should have no space between `if` and the `(` and the `{` should be on a new line.

JordanMussi: Done in e3d47839eb1e6e066bdddc614b24e27f90f7816b 👻

Ben-MyBB:
@euantorano I assume this is something we would be looking to add into 1.10?

euantorano:
Probably 1.9.0 I’d say, since the work is already done :)

inc/plugins/hello.php
```diff
@@ -83,6 +83,27 @@ function hello_info()
...
+{
+	global $db;
+
+	$query = $db->simple_select('settinggroups', 'gid', "name='hello'");
+
+	$gid = (int)$db->fetch_field($query, 'gid');
+
+	if($gid > 0)
+	{
+		return "index.php?module=config-settings&amp;action=change&amp;gid={$gid}";
```
Matslom: Is returning the url the best option? Future routing improvements may change url structure. 
IMHO returned `GID` gives us more freedom.

euantorano: I'd agree with that @Matslom. Returning GID seems much more useful to me.

JordanMussi: I went with URL as many plugins use a new module to hold their configuration. Only accepting a GID as the return would prevent the plugin from linking to the module.

Ben-MyBB:
I personally think it would be good to get this in 1.9, ping @JordanMussi

JordanMussi:
@Ben-MyBB I've rebased the changes onto 1.9 https://github.com/mybb/mybb/compare/develop/1.9...JordanMussi:enhance/acp-plugins-list. I can't seem to reopen this PR to update the changes maybe that's something for a maintainer to do? Not too sure why it hasn't shown the updates to the branch here as well (should be showing force pushed to 2f8bb4cce5437bb51cd32ab3349550cacaae043d)

euantorano:
I can’t re-open it either for whatever weird reason - the button is disabled for me too.

Ben-MyBB:
When we hover over the re-open button it says that we can't reopen since it has been force pushed or recreated?

JordanMussi:
If you’re happy with the changes can just create a new PR I suppose :-)

euantorano:
Yeah, I'd like to see a new PR for sure. The only issue I originally had was returning a link to the settings, but that's not. Abigails deal and we can change it in the future when we make ACP changes.

PR #2947, opened 2017-12-16
Fixes #2198 Optimize thread redirect database structure

### LINKED ISSUE #2198: Optimize thread redirect database structure
Currently, threads that are left behind as a redirect store their redirect information in the closed column on the threads table in the style as **moved|$tid**. This can slow a site down as a full table scan is needed for queries that need to avoid moved threads.

- Add a moved column to the thread table that stores a moved thread's tid.
- Change the closed column to a tinyint and use 2 for moved threads.

PaulBender:
Fixes #2198 

I'm pretty sure I found all the moved bits in the code.


Ben-MyBB:
@euantorano wonder if this would be as well being merged into 1.9?

euantorano:
Only reason it hasn’t been is the possible conflict/issue with the upgrade script in the future. I haven’t really thought about how to handle that.

Ben-MyBB:
Seems like this is possibly a duplicate of https://github.com/mybb/mybb/pull/3143 and already merged into 1.9?

euantorano:
Yes, looks like it. This is also targeted at `develop/1.10` rather than `develop/1.9`.

PR #2950, opened 2017-12-19
Develop/1.10 feature caching improvements

tijdmachine:
- Added a rewritten cache interface based on the current `/inc/cachehandlers/interface.php`
- Added a FileStore class which makes use of [Illuminate/Filesystem](https://github.com/laravel/framework/tree/5.5/src/Illuminate/Filesystem) & implements the `/inc/src/Contracts/Cache/Store.php` interface
 
Usage
```
// init a new FileStore instance
$cache = new FileStore();

// store an item based on key & value
$cache->put(‘key’, ‘value’));

// select an item based on key
$cache->fetch(‘key’));

// delete an item based on key
$cache->delete('key');

// flush cache
$cache->flush();

// get size of a cached item
$cache->size_of(‘key’);
```


036:
Looking good.

tijdmachine:
Saw I missed a few scalar type hints already, will add them tomorrow

euantorano:
Sorry, I might not have explained my thoughts too clearly 😉 

I was meaning to add a dependency upon `[Illuminate\Cache](https://github.com/illuminate/cache)`, which would provide all of the different store types (including `FileStore` which you've implemented).

I was then going to inject the `Illuminate\Contracts\Cache\Store` instance into the core MyBB `[datacache](https://github.com/mybb/mybb/blob/feature/inc/class_datacache.php)` class.

Also note that `inc/settings.php` shouldn't be checked in, `cache/themes/index.html` shouldn't be deleted and `config.default.php` seems to have ome weird changes being shown here (0 changes).

If need be, I can provide some pseudocode of what I was meaning, but won't get a chance until tomorrow after work.

PR #2953, opened 2017-12-21
Update to class_mailhandler.php (fixes #2883)

### LINKED ISSUE #2883: SMTP mailhandler should use the from name
The SMTP mailhandler doesn't currently use the `FROM` command to set the from name for emails sent (though the PHP mail handler does).

Originally reported here: https://community.mybb.com/thread-213661.html

fluxtion:
Due to my general GitHub noobishness, I ended up fixing 3 issues in one PR so this is a carbon copy without the other fixes.

### DIFF
--- a/inc/class_mailhandler.php
+++ b/inc/class_mailhandler.php
@@ -7,7 +7,6 @@
  * License: http://www.mybb.com/about/license
  *
  */
-
 /**
  * Base mail handler class.
  */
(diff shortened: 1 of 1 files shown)


fluxtion:
@kawaii Can you be my tester again please? This works so I'd like to see it merged. :)

kawaii:
Confirmed this one at the start of the year but forgot to mark it as such. @euantorano will the new SwiftMailer stuff we're sneaking into 1.9 have any effect on or conflict with this at all? If not provided you're happy I guess we can merge.

fluxtion:
Thanks :)

euantorano:
@kawaii I might have to modify the SwiftMailer branch slightly, but I still need to finish work on that anyway so I'm happy to merge this now you've confirmed it works.

PR #2954, opened 2017-12-21
Fix functions_themes.php for PHP 7.2 (fixes #2927)

### LINKED ISSUE #2927: Possible counting of non-countable objects in build_new_theme()
https://github.com/mybb/mybb/blob/mybb_1814/admin/inc/functions_themes.php#L578

Related: #2824

fluxtion:
Simple but effective.


euantorano:
Looks good to me!

PR #2955, opened 2017-12-21
Escaping keyword "groups" for MySQL 8 (#2921)

### LINKED ISSUE #2921: SQL error on install due to reserved word in MySQL 8.0.2
When installing MyBB 1.8.X using MySQL 8.0.2 or above they will receive the following SQL error:
image
This is due to `groups` now being a reserved word. This can be remedied by ensuring that column names are always quoted in queries.

fluxtion:
I used a regex search to find instances of "groups" in query string parameters and backtick them.

Unfortunately, I can't get MySQL 8 working on my WAMP stack for some reason so I'd appreciate it if someone else could test to see if this does indeed fix #2921 

Thanks :)

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1378,17 +1378,17 @@ if(!empty($threadcache) && is_array($threadcache))
 					foreach($gids as $gid)
 					{
 						$gid = (int)$gid;
-						$gidswhere .= " OR ','||groups||',' LIKE '%,{$gid},%'";
+						$gidswhere .= " OR ','||`groups`||',' LIKE '%,{$gid},%'";
 					}
-					$query = $db->simple_select("modtools", 'tid, name', "(','||forums||',' LIKE '%,$fid,%' OR ','||forums||',' LIKE '%,-1,%' OR forums='') AND (groups='' OR ','||groups||',' LIKE '%,-1,%'{$gidswhere}) AND type = 't'");
+					$query = $db->simple_select("modtools", 'tid, name', "(','||forums||',' LIKE '%,$fid,%' OR ','||forums||',' LIKE '%,-1,%' OR forums='') AND (`groups`='' OR ','||`groups`||',' LIKE '%,-1,%'{$gidswhere}) AND type = 't'");
 					break;
 				default:
 					foreach($gids as $gid)
 					{
 						$gid = (int)$gid;
-						$gidswhere .= " OR CONCAT(',',groups,',') LIKE '%,{$gid},%'";
+						$gidswhere .= " OR CONCAT(',',`groups`,',') LIKE '%,{$gid},%'";
 					}
-					$query = $db->simple_select("modtools", 'tid, name', "(CONCAT(',',forums,',') LIKE '%,$fid,%' OR CONCAT(',',forums,',') LIKE '%,-1,%' OR forums='') AND (groups='' OR CONCAT(',',groups,',') LIKE '%,-1,%'{$gidswhere}) AND type = 't'");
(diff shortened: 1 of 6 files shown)


kawaii:
Good start, but these fixes need extending to the following file too:
https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_tables.php

fluxtion:
It makes me nervous that `mysql_db_tables.php` makes creates `groups` columns in multiple tables that I found no references to in database queries using my regex search. Regex was always going to be an inexact science, which is why I missed the entire file that constructs these tables in the first place.

The brute force alternative is to go through every single call to the database class API - many hundreds of calls - and manually escape them all. Given that we're hoping to change the the entire database class and its API, which will probably involve rewriting every query call anyway, it seems like a massive waste of time for anyone to do this now.

My favoured solution would still be to redirect every parameter in the various API methods that generate a query through a function designed to specifically identify instances of `"groups"` and escape them programmatically. This is by far the most time-efficient stopgap and unless somehow can provide a solid PoC to demonstrate that the following is insecure or prone to bugs, I won't implement anything else:

`function escape_groups($param_str) {
    $param_arr = explode(",", "$param_str);
    foreach($param_arr as &$param) {
         if(trim($param) == "groups")) $param = "`groups`";
    }
    $res = implode(",", $param_arr);
    return $res;
}`

euantorano:
@fluxtion I'm not opposed to the idea, but some things to consider:

- Would the above patch apply only to the functions like `insert_query`, `update_query`, etc (eg: exclude `write_query` and `query`)?
- If it was included in `write_query` and `query` and similar where arbitrary SQL is passed, consider the following query:

```SQL
INSERT INTO posts (pid, data) VALUES (1, 'This is a post that happens to include some words like join, groups, and more');
```

The above patch would seemingly change the post `data` into:

> This is a post that happens to include some words like join, `groups`, and more

This would mean the original content is altered in a way that doesn't make any sense.

In order for the automated wy to work correctly in that scenario would be to properly parse the SQL statement.

----------

Note that I'm not sure if there are any scenarios when the `groups` keyword is used in a raw query as above, but it still may happen within plugin code and other 3rd party code.

Obviously automating it is far easier, but there are some cases to consider where it may cause invalid results.

dvz:
> [...] columns in multiple tables that I found no references to in database queries using my regex search. Regex was always going to be an inexact science [...]

What regex were you using?

Processing all queries to escape the column will create unnecessary overhead - IMO the codebase is not large enough to warrant such operation. If something similar was to be implemented, all columns should be be quoted with backticks. This shouldn't be an issue once MyBB start using a query builder (so it's probably not worth it now).

fluxtion:
> If something similar was to be implemented, all columns should be be quoted with backticks.
This is what I originally proposed but Euan didn't want it.

I'm well aware that it's not a computationally efficient solution but it is time efficient and would only be a stopgap. I've hit my limit as far as I'm prepared to scan files for every single query. Someone can either finish what I started by looking for other instances of db class calls and escape them manually or I can propose a programmatic method.

I suggest someone in the future tries `/simple_select.*groups/` and friends instead of using variable names like I did. But this still won't find dynamically generated queries, query functions with array arguments or multiline function calls.

It seems to me that it would be easy to "fix" the issue at hand to the extent that myBB installs on MySQL 8 but far harder to prevent more insidious, less predictable bugs during use.

euantorano:
Unfortunately `query` and `write_query` are still used in a bunch of places (eg: anywhere where you need to do a `JOIN` or anything), so you'd still need to do the find/replace for all isntances of those two calls.

This PR is a good first step, but there are probably still more to find so I'll leave this open for now (until the very last minute).

euantorano:
After further review, that patch breaks PostgreSQL support, as the ``` character is not an escape character for column names and causes syntax errors.

This means we need to either check the DB engine in use, or we need to do this at the DB class level.

Either way, this issue will be pushed back to 1.8.16 rather than 1.8.16.

PR #2957, opened 2017-12-21
Added more duct-tape to search.php (fixes #2915)

### LINKED ISSUE #2915: Inline Moderation Select All
While using Inline Moderation (version 1.8.12) the "select all results" doesn't actually work.  Tried on 3 installs including a near default version.

Can someone else confirm.  Example use would be to view more than one page of results for threads then try to close them all by either clicking the "Select all XXX results in this search." option.  Then open or close the threads.  For me it's not working at all on any site I tested. May require further testing but I can't bother with it now.

Bug does exist.

fluxtion:
The entire latter half of `search.php` needs to be rewritten. Almost identical code is replicated in at least 6 places and the interaction between the code and the database is inconsistent as hell. One of these broken promises is what caused the distant failure of moderation "select all" on user thread lists (#2915). `getallids()` in `moderation.php` expects `search.php` to behave in a sane way wrt interacting with the `searchlog` table but it doesn't and therefore `search.php` is at fault for the issue.

Notably, although there aren't issues related to #2915 yet, there should be and will be eventually. Trying to mass moderate "recent threads" as well as "today's threads" will exhibit exactly the same failure until someone fixes `search.php` more thoroughly. I'm happy to do that since I already understand exactly what the problems are. And I get great pleasure out of refactoring massively duplicated code (unironically).

In the meantime, this is a band-aid over a bullet wound. But it fixes #2915 and that'll do for now.

### DIFF
--- a/search.php
+++ b/search.php
@@ -1303,17 +1303,17 @@ elseif($mybb->input['action'] == "finduser")
 }
 elseif($mybb->input['action'] == "finduserthreads")
 {
-	$where_sql = "t.uid='".$mybb->get_input('uid', MyBB::INPUT_INT)."'";
+	$where_sql = "uid='".$mybb->get_input('uid', MyBB::INPUT_INT)."'";
 
 	$unsearchforums = get_unsearchable_forums();
 	if($unsearchforums)
 	{
-		$where_sql .= " AND t.fid NOT IN ($unsearchforums)";
+		$where_sql .= " AND fid NOT IN ($unsearchforums)";
 	}
 	$inactiveforums = get_inactive_forums();
 	if($inactiveforums)
 	{
-		$where_sql .= " AND t.fid NOT IN ($inactiveforums)";
+		$where_sql .= " AND fid NOT IN ($inactiveforums)";
 	}
 
 	$permsql = "";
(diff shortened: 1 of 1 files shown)


kawaii:
Tested and this PR does indeed fix the linked issue, I wonder if we should merge this as-is or if it's worth just making all of the suggested improvements in a single larger patch? @mybb/developers @mybb/sqa

fluxtion:
Obviously it's not my decision but I would prefer to take my time on `search.php` and get it right. There's a lot of stuff to be done there.

I would prefer to see this PR merged and then work on #2958 as a separate issue.

tommm:
If more widespread changes are required, or a rewrite/refactor of code, I would recommend either waiting for 1.10 or creating a new branch for the fixes (as apposed to creating multiple PRs or commits that may make it harder to spot a regression).

fluxtion:
This PR is a clean standalone fix for #2915 so I think it should be merged now.

As for the wider project (#2958), I think it might be a good idea to do it in two phases. Firstly, I'll modify the code so that "select all" inline moderation works everywhere. This will only be a ~50 line change. I'll submit a PR for that. And then for the refactoring part, which is where the regressions would emerge if there were to be any, I'll do whatever you consider best practice.

kawaii:
@euantorano LGTM, ready for merging.

PR #2963, opened 2017-12-23
Start implementing Twig for the front-end

euantorano:
This is **very** much a work in progress, but opening this PR now will allow people to follow the progress of my translating templates across to Twig, and to see how things are going to be structured according to my plan.

Any feedback on the structure/layout of Twig files, naming of extensions and functions and such would be appreciated.

I can't decide whether to have separate extension classes with a couple of functions each, or one large core `MyBB` extension that registers several key Twig functions...


036:
I personally think having separate extension classes is the way to go. That way everything can be found according to the name rather than muddling though one large file.

euantorano:
@036 My current thinking is that I might add a `ThemeExtension`, which registers the `theme` global variable and provides some utility functions (such as a function to iterate through stylesheets for the current page, the `get_asset_url` function, and perhaps a few more as I discover them). I think that strikes the right balance between separation and making it easy to reason where things can be found.

PaulBender:
Looking good.

As for structure/layout of Twig files, I think they should be in a **resources** folders in the root of the board.

JordanMussi:
> As for structure/layout of Twig files, I think they should be in a resources folders in the root of the board.

I think the `inc` directory serves the purpose well enough for 1.x. Administrators will have already restricted access to the `inc` directory and adding a `resources` directory gains no benefit.

@euantorano I think creating a `develop/1.9-twig` branch and making that the target for this PR would be a good idea. Then others can continue to work on this and make PRs to that branch and all the Twig changes can be merged as one into `develop/1.9` when they are ready.

euantorano:
@JordanMussi That's the plan, once I have this working 😄

usercp.php
```diff
@@ -2983,44 +2982,31 @@
...
-	{
-		$disable_delete_drafts = 'disabled="disabled"';
-		eval("\$drafts = \"".$templates->get("usercp_drafts_none")."\";");
-	}
 
 	$plugins->run_hooks("usercp_drafts_end");
 
 	output_page(\MyBB\template('usercp/drafts.twig', [
 	    'drafts' => $drafts,
+        'deleteDraftsEnabled' => $deleteDraftsEnabled,
```
JordanMussi: Indentation a bit iffy here 🔍

euantorano:
Yeah, got some other things to commit too. Should get it done tonight hopefully.

euantorano:
Ok, we now have a working drafts page using Twig 😄 

screen shot 2018-01-01 at 22 18 16

There are still several `TODO` comments, which I can patch up in the future (I want to add an extension with miscellaneous Twig functions like `my_date` and `my_number_format`, etc. which don't really need an extension each).

I'll merge this in and create a new `develop/1.9` branch tomorrow if there aren't any comments on this PR by the time I get home from work tomorrow night.

justinsoltesz:
Not a huge issue as the new theme won't use alternating row classes, but based on your screenshot it looks like the alternating row function doesn't work the way it should. The columns have different background colors instead of rows.

euantorano:
@justinsoltesz Ah, yes. Good catch. I can fix that by assigning the value to a variable, or setting the class on the parent `<tr>`.

justinsoltesz:
Honestly, I would say just remove the function entirely and set everything to trow1. It won't be needed with the new theme.

euantorano:
Alright, I won't complain about that 😄

PR #2965, opened 2017-12-27
modcp.php: use `l.fid <> 0` instead of `!l.fid`

mytbk:
PostgreSQL doesn't use the `!` operator


euantorano:
Thanks for the patch!

PR #2966, opened 2017-12-28
Fixes inline moderation "select all" in all scenarios (#2958 extends #2957)

fluxtion:
This is the first step in fixing the inline moderation "select all" feature for thread lists. By changing the interaction between `search.php` and the database (`searchlog`) to behave consistently and as expected elsewhere in the codebase, "select all" moderation now works everywhere it can be used (check "View New Posts" and "View Today's Posts" for example -- both fixed).

At this point, I'm comfortable that inline "select all" works as expected everywhere that it is supposed to. I would like to refactor the code in `search.php` to reduce the horrible replication and general atomic wreckage but this is an enhancement and will happen later and with more care (possibly using an entirely new branch as suggested by @tommm).

As it is, all the bugs in inline thread moderation are fixed and therefore I think this PR should be merged. This fix supersedes #2957 and is a complete resolution to the bugs I predicted in #2958.

### DIFF
--- a/search.php
+++ b/search.php
@@ -1303,17 +1303,17 @@ elseif($mybb->input['action'] == "finduser")
 }
 elseif($mybb->input['action'] == "finduserthreads")
 {
-	$where_sql = "t.uid='".$mybb->get_input('uid', MyBB::INPUT_INT)."'";
+	$where_sql = "uid='".$mybb->get_input('uid', MyBB::INPUT_INT)."'";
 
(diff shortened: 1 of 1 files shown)


kawaii:
I'll be reviewing and testing this one later today, looks good from a cursory glance though.

Ben-MyBB:
@kawaii Did you get a chance to test this one?

PR #2968, opened 2018-01-02
Fix mycode_auto_url to detect links when adjacent to square bracket

JordanMussi:
Before this change all of these URLs would not be parsed:
```
[list]
[*]https://mybb.com
[/list]

[b]https://mybb.com[/b]

MyBB [https://mybb.com]
```

The following function as they did before this change:
```
[url=https://mybb.com]MyBB[/url]

[url=https://mybb.com/about/]https://mybb.com[/url]

[url]https://mybb.com[/url]
```

My only concern is the efficiency of the new regex. Would it be better to only do this change?
```diff
-[\>\s\(\)]
+[\>\s\(\)\[]|(?:\[\*\])
```
Even though this would result in the `[b]https://mybb.com[/b]` continue to not be detected?

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1512,8 +1512,8 @@ class postParser
 	{
 		$message = " ".$message;
 		// Links should end with slashes, numbers, characters and braces but not with dots, commas or question marks
-		$message = preg_replace_callback("#([\>\s\(\)])(http|https|ftp|news|irc|ircs|irc6){1}://([^\/\"\s\<\[\.]+\.([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu", array($this, 'mycode_auto_url_callback'), $message);
-		$message = preg_replace_callback("#([\>\s\(\)])(www|ftp)\.(([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu", array($this, 'mycode_auto_url_callback'), $message);
+		$message = preg_replace_callback("#([\>\s\(\)\[]|(?:\[(?!url)(?:.*?)\]))(http|https|ftp|news|irc|ircs|irc6){1}://([^\/\"\s\<\[\.]+\.([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu", array($this, 'mycode_auto_url_callback'), $message);
+		$message = preg_replace_callback("#([\>\s\(\)\[]|(?:\[(?!url)(?:.*?)\]))(www|ftp)\.(([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu", array($this, 'mycode_auto_url_callback'), $message);
 		$message = my_substr($message, 1);
 
 		return $message;


kawaii:
Tested working on MyBB 1.8.14, LGTM. @euantorano any comments on @JordanMussi's concerns regarding the efficiency of the regex?

euantorano:
Can't comment much on the efficiency personally. If it works though, and there's no noticeable slowdown in page generation time we can merge this.

kawaii:
I created a thread with a full page of 25 posts and tested load times before and after patching `inc/class_parser.php`, @dvz do you have anything to add to this or are these benchmarks sufficient?

Before patch:
```
Generated in 27 ms (28.59% PHP / 71.41% PostgreSQL) 
SQL Queries: 20 / Server Load: 0.37 / Memory Usage: 2 MB
```
After patch:
```
Generated in 30 ms (37.78% PHP / 62.22% PostgreSQL) 
SQL Queries: 20 / Server Load: 0.77 / Memory Usage: 2 MB
```

dvz:
It would need to be run separately with many iterations to give some meaningful results (the matching can get 2x slower or more, depending on the message, but that still can be acceptable).

The `(?:\[\*\])` pattern is slower than the one in the PR currently.

lairdshaw:
Hi. I was the reporter on the community forums of issue #3066 - thank you for the fix. Another problem introduced by this PR is that the middle URL of these three when included in a post will not be automatically converted into a link:

`[url]http://example.com/[/url] http://example.com/ [url]http://example.com/[/url]`

This can be resolved by adjusting the regexes from:

`#([\>\s\(\)\[]|(?:\[(?!url)(?:.*?)\]))(http|https|ftp|news|irc|ircs|irc6){1}://([^\/\"\s\<\[\.]+\.([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu`

and

`#([\>\s\(\)\[]|(?:\[(?!url)(?:.*?)\]))(www|ftp)\.(([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu`

to:

`#([\>\s\(\)\[]|(?:\[(?!url)(?:[^\]]*?)\]))(http|https|ftp|news|irc|ircs|irc6){1}://([^\/\"\s\<\[\.]+\.([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu
`
and

`#([\>\s\(\)\[]|(?:\[(?!url)(?:[^\]]*?)\]))(www|ftp)\.(([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#iu`

That is, to replace the `.*` in each with `[^\]]*`

mohammadg:
Yes this is breaking custom mycode.

As the regexp has been edited to batch square brackets next to URL's it is matching things like:
`\[custommycode\]blah\[custommycode\]`

If you want the behaviour where both URLs next to square brackets parse correctly then you need to avoid matching custom mycode or change how mycode uses `mycode_auto_url`.

Right now I'm using:
`([\>\s\(\)\[]|(?:\[(?:b|\*)\]))`

Which doesn't match anything in square brackets before a url except `[*]` and `[b]` as hardcoded defaults which is not ideal. Imo the change should be reverted and explicit URL mycode uses for the cases mentioned above unless there is an elegant way to not match custom mycode (or for custom mycode to not use `mycode_auto_url`).

euantorano:
I suggested pretty much the same thing a while back actually. The auto URL code should be one of the last things to be ran.

Sama34:
Shouldn't the parsing order be taken into consideration as well ? Why is a code parsed before my custom one with parsing order of `-999` ?

lairdshaw:
@euantorano I remember reading you writing that. The current problem with that is that given `my_auto_url` as it is currently implemented (i.e. including the merged PR on which we're commenting), if it is moved further down than I suggested above, then `[url]http://example.com/[/url]` will be turned into a link to `[url]http://example.com/[/url]` instead of to, as it ought to be, a link to `http://example.com/`.

lairdshaw:
In that case, my PR #3106 is redundant. Hopefully your parser tests would flush out any problems with my suggestion.

lairdshaw:
A minor(?) amendment to my suggestion above is that some `>`s would still be valid, so selective detection of opening HTML tags in the regexes that are the subject of this PR would be required, e.g. `<p>` and `<li>` tags could be detected with `<(?:p|li)[^>]*>`, with the main objective being to *not* detect opening `<a>` tags so as not to double up these tags.

labrocca:
Can I suggest parsing order for custom mycodes allow a -1 sort order to not parse with default mycode?  Just a thought.

euantorano:
The idea is that some custom MyCodes might interfere with others so you may want those to run before other custom MyCode. 

The way that the parser currently works has all core MyCode run before any custom codes, so a -1 sort order causing core codes to be ignored would be difficult as it currently works unfortunately. I don’t want to introduce other bigger changes to the parser for 1.8.16 (we broke it enough in .15...), but can you open a new issue for that suggestion?

labrocca:
"The way that the parser currently works has all core MyCode run before any custom codes"

Correct, as I thought. Which is why a -1 sort order could run it before the default mycode and ignore further parsing.

You would know if this is easy enough to implement or if you should wait.

I truly hope that MyBB puts all MyCode into the database alongside custom mycode where it can be altered, disabled, sorted or reverted very easily.

Sama34:
MyCodes already have a parsing order. It should only require updating the code and see if it works.

I also think all MyCodes should be inserted to the DB and manageable from the ACP.

We could implement some import/export feature and [Yumi's code](http://mybbhacks.zingaburga.com/showthread.php?tid=269) to reduce issues and the flawed MyCodes shared in the community.

Since not all codes would be that easy to migrate, we could also open the possibility of allowing more complex codes.

The parser is frail to enhance but I'm sure there is space for improvement.

PR #2969, opened 2018-01-05
Fixed jgrowl messagge error

chack1172:
When a user try to go to the last post of a thread, if the last post is soft deleted a jgrowl error message will popup if usergroup can see deletion notice.


euantorano:
Hi, @kawaii has been testing this and can't seem to replicate the issue this fixes. Is there a thread or issue where this problem was reported?

PR #2971, opened 2018-01-05
Starting work on translating subscribed threads to Twig

euantorano:
This is part of the ongoing effort to translate templates to twig. This PR is to translate the subscribed threads page in the user CP to Twig. The translation has only just begun, but I'm making this PR to make it clear that it's what I'm currently working on should anybody else start working on it.


euantorano:
Ok, the `usercp.php?action=subscriptions` page is done.

There are quite a few functions we should look at turning into functions, which we can look at at a later time.

I'me going to create a usercp layout next, with the user CP menu as not having it in these pages makes navigation fun...

Issue #2972, opened 2018-01-05
Template Conversion to Twig Format


kawaii:
This issue is to track the progress of which templates have been converted to the new format for the Twig engine. Please reference this issue in your relevant PRs.

- [ ] announcement
- [ ] announcement_edit
- [ ] announcement_quickdelete
- [ ] attachment_icon
- [ ] calendar
- [ ] calendar_addevent
- [ ] calendar_addeventlink
- [ ] calendar_addevent_calendarselect
- [ ] calendar_addevent_calendarselect_hidden
- [ ] calendar_addprivateevent
- [ ] calendar_addpublicevent
- [ ] calendar_day
- [ ] calendar_dayview
- [ ] calendar_dayview_birthdays
- [ ] calendar_dayview_birthdays_bday
- [ ] calendar_dayview_event
- [ ] calendar_dayview_noevents
- [ ] calendar_editevent
- [ ] calendar_event
- [ ] calendar_eventbit
- [ ] calendar_event_editbutton
- [ ] calendar_event_modoptions
- [ ] calendar_event_userstar
- [ ] calendar_jump
- [ ] calendar_jump_option
- [ ] calendar_mini
- [ ] calendar_mini_weekdayheader
- [ ] calendar_mini_weekrow
- [ ] calendar_mini_weekrow_day
- [ ] calendar_mini_weekrow_day_link
- [ ] calendar_move
- [ ] calendar_nextlink
- [ ] calendar_prevlink
- [ ] calendar_repeats
- [ ] calendar_select
- [ ] calendar_weekdayheader
- [ ] calendar_weekrow
- [ ] calendar_weekrow_day
- [ ] calendar_weekrow_day_birthdays
- [ ] calendar_weekrow_day_events
- [ ] calendar_weekrow_day_external
- [ ] calendar_weekrow_day_today
- [ ] calendar_weekview
- [ ] calendar_weekview_day
- [ ] calendar_weekview_day_birthdays
- [ ] calendar_weekview_day_event
- [ ] calendar_weekview_day_event_time
- [ ] calendar_weekview_month
- [ ] calendar_weekview_nextlink
- [ ] calendar_weekview_prevlink
- [ ] calendar_year
- [ ] calendar_year_sel
- [ ] changeuserbox
- [ ] codebuttons
- [ ] contact
- [ ] debug_summary
- [ ] editpost
- [ ] editpost_delete
- [ ] editpost_disablesmilies
- [ ] editpost_postoptions
[...]

PR #2975, opened 2018-01-06
Working on #2972 Template Conversion to Twig Format (warnings.php)

PaulBender:
Working on #2972

This PR is for the warning system (warnings.php).

### DIFF
--- a/inc/views/base/warnings/warning_row.twig
+++ b/inc/views/base/warnings/warning_row.twig
@@ -1,0 +1,31 @@
+{% set row = alt_trow() %}
+{% if warning.active == 1 %}
+    <tr>
+        <td class="trow_sep" colspan="5">{{ lang.active_warnings }}</td>
+    </tr>
+{% endif %}
+{% if warning.active == 2 %}
+    <tr>
+        <td class="trow_sep" colspan="5">{{ lang.expired_warnings }}</td>
+    </tr>
+{% endif %}
+<tr>
+    <td class="{{ row }}">
+        {{ warning.warning_type }} {{ trans('warning_points', warning.points) }}
+        {% if warning.post_subject %}
+            <br /><small>{{ lang.warning_for_post }} <a href="{{ warning.post_link|raw }}#pid{{ warning.pid }}">{{ warning.post_subject }}</a></small>
+        {% endif %}
+    </td>
+    <td class="{{ row }}" style="text-align: center;">
+        {{ warning.date_issued|raw }}
+    </td>
+    <td class="{{ row }}" style="text-align: center;">
+        {{ warning.expires }}
+    </td>
+    <td class="{{ row }}" style="text-align: center;">
+        {{ warning.issuedby|raw }}
+    </td>
+    <td class="{{ row }}" style="text-align: center;">
+        <a href="warnings.php?action=view&amp;wid={{ warning.wid }}">{{ lang.view }}</a>
+    </td>
+</tr>
--- a/inc/views/base/warnings/warnings.twig
+++ b/inc/views/base/warnings/warnings.twig
@@ -1,0 +1,38 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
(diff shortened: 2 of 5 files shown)


euantorano:
Nice!

inc/views/base/warnings/warning_row.twig
```diff
@@ -0,0 +1,31 @@
+{% set row = alt_trow() %}
```
euantorano: @justinsoltesz was talking about completely removing alternating row classes. I've left them in the templates I've done too, but we might end up removing them (and the function) completely...

inc/views/base/warnings/warnings.twig
```diff
@@ -0,0 +1,38 @@
...
+{% endblock head %}
+
+{% block body %}
+    {{ multipage|raw }}
+    <table border="0" cellspacing="{{ theme.borderwidth }}" cellpadding="{{ theme.tablespace }}" class="tborder">
+        <thead>
+            <tr>
+                <td class="thead" colspan="5">
+                    {% if warning_level > 0 %}
+                        <div class="float_right">{{ trans('current_warning_level', warning_level, user.warningpoints, mybb.settings.maxwarningpoints)|raw }}</div>
```
euantorano: I wonder if it's worth marking the `trans` function as HTML safe... There are quite a few translations that have HTML in them unfortunately.

warnings.php
```diff
@@ -752,121 +752,92 @@
...
+        SELECT w.*, t.title AS type_title, u.username, p.subject AS post_subject
+        FROM " . TABLE_PREFIX . "warnings w
+        LEFT JOIN " . TABLE_PREFIX . "warningtypes t ON (t.tid=w.tid)
+        LEFT JOIN " . TABLE_PREFIX . "users u ON (u.uid=w.issuedby)
+        LEFT JOIN " . TABLE_PREFIX . "posts p ON (p.pid=w.pid)
+        WHERE w.uid='{$user['uid']}'
+        ORDER BY w.expired ASC, w.dateline DESC
+        LIMIT {$start}, {$perpage}
+    ");
+    while($warning = $db->fetch_array($query))
```
euantorano: Since you've fixed codestyle elsewhere in this code, here's another one 😄

warnings.php
```diff
@@ -752,121 +752,92 @@
...
+        $last_expired = $warning['expired'];
+
+        if ($warning['post_subject']) {
+            $warning['post_subject'] = $parser->parse_badwords($warning['post_subject']);
+            $warning['post_subject'] = htmlspecialchars_uni($warning['post_subject']);
+            $warning['post_link'] = get_post_link($warning['pid']);
+        }
+
+        $warning['username'] = htmlspecialchars_uni($warning['username']);
+        $warning['issuedby'] = build_profile_link($warning['username'], $warning['issuedby']);
```
euantorano: We should probably add this as a Twig function which is HTML safe. It's used fairly often afterall.

We should maybe add a `CoreExtension` (or `MybbExtension` or some other name) with some utilities like this, `my_date`, and probably  afew others.

warnings.php
```diff
@@ -752,121 +752,92 @@
...
+
+        $warning['date_issued'] = my_date('relative', $warning['dateline']);
+
+        if ($warning['type_title']) {
+            $warning_type = $warning['type_title'];
+        } else {
+            $warning_type = $warning['title'];
+        }
+
+        $warning['warning_type'] = htmlspecialchars_uni($warning_type);
```
euantorano: This will be escaped in the template too, so you can remove the `htmlspecialchars_uni` here or else it will be double escaped and look weird.

PaulBender:
Done. All templates in warnings.php have been converted

PaulBender:
We should consider adding some Twig functions for the following:
- inline errors
- codebuttons and smilieinserter

euantorano:
@PaulBender Mind starting an issue for any Twig functions you find that would be useful? We can build a list in a single issue and cross them off as we implement them.

PR #2978, opened 2018-01-09
Fix #2977 Add plugin hook to format_name()

### LINKED ISSUE #2977: Add plugin hook to format_name()
Add a new plugin hook at https://github.com/mybb/mybb/blob/mybb_1814/inc/functions.php#L3075 so the username format can be further modified basing on available information.

dvz:
#2977

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3050,7 +3050,7 @@ function random_str($length=8, $complex=false)
  */
 function format_name($username, $usergroup, $displaygroup=0)
 {
-	global $groupscache, $cache;
+	global $groupscache, $cache, $plugins;
 
 	if(!is_array($groupscache))
 	{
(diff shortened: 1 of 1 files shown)


euantorano:
If we're making changes here, it might make sense to change `substr_count` to `strpos` here: https://github.com/dvz/mybb/blob/38b07864201ac497e75d1aec937dcc870fa2ee57/inc/functions.php#L3067 - I don't know why it uses `substr_count`, but `strpos` should be marginally more efficient.

We should also probably check that `$usergroup` is in the `$groupscache` array here: https://github.com/dvz/mybb/blob/38b07864201ac497e75d1aec937dcc870fa2ee57/inc/functions.php#L3065

euantorano:
Any comments on my above comment @dvz? Also, in the original issue @tommm made a suggestion about possibly caching formatting and seeing if it made much difference.

dvz:
`strpos` seems to speed it up even with the `isset($groupscache[$usergroup])` check added.

inc/functions.php
```diff
@@ -3050,30 +3050,52 @@ function random_str($length=8, $complex=false)
...
-	if($userin == 0)
-	{
-		$format = "{username}";
-	}
+		if(isset($groupscache[$usergroup]))
+		{
+			$ugroup = $groupscache[$usergroup];
 
-	$format = stripslashes($format);
+			if(strpos($ugroup['namestyle'], "{username}") === false)
```
euantorano: Shouldn't this be `!== false`, as you use the group namestyle only if it contains the `{username}` string. The current logic only uses it if it __doesn't__ contain the `{username}` string.

PR #2980, opened 2018-01-09
Fixes #2979 — No Specific Error Message For Users w/o Permission To Vote In Polls

### LINKED ISSUE #2979: No Specific Error Message For Users w/o Permission To Vote In Polls
I think there's a bug in the poll-system. 
I made a poll, only members are allowed to vote on this poll. Everything works fine.
But, if a guest (someone who's not logged in) opens the poll, it say's 'This poll is closed' in the blue bar on top. But is isn't closed at all, it's just not available for not-logged-in persons (guests). After they logged in, the message 'This poll is closed' is disappeared and they can vote.

Why the message 'This poll is closed' appears if guests open the poll page? Should'nt there be a message like 'This poll is only for logged-in members'?

Hope somebody can help me with this issue.......

Original thread: [This poll is closed?](https://community.mybb.com/thread-215244.html)

WildcardSearch:
Fixes #2979


showthread.php
```diff
@@ -482,6 +482,10 @@
 			$showresults = 1;
 		}
 
+		if ($forumpermissions['canvotepolls'] != 1) {
```
euantorano: Mixed code style here - below this section you use 1.8 style (`{` on new line).

WildcardSearch: Good catch. Man, it is so hard to write PHP like that for me now...

WildcardSearch:
Done, @euantorano

PR #2984, opened 2018-01-12
Working on #2972 Template Conversion to Twig Format (calendar.php)

PaulBender:
Working on #2972

This PR is for the calendar (calendar.php).

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -242,316 +242,230 @@ if($mybb->input['action'] == "addevent")
 	add_breadcrumb(htmlspecialchars_uni($calendar['name']), get_calendar_link($calendar['cid']));
 	add_breadcrumb($lang->nav_addevent);
 
-	$plugins->run_hooks("calendar_addevent_start");
+    $plugins->run_hooks("calendar_addevent_start");
 
-	// If MyCode is on for this forum and the MyCode editor is enabled inthe Admin CP, draw the code buttons and smilie inserter.
-	if($mybb->settings['bbcodeinserter'] != 0 && (!$mybb->user['uid'] || $mybb->user['showcodebuttons'] != 0) && $calendar['allowmycode'] == 1)
-	{
-		$codebuttons = build_mycode_inserter("message", $calendar['allowsmilies']);
-		if($calendar['allowsmilies'] == 1)
-		{
-			$smilieinserter = build_clickable_smilies();
-		}
-	}
+    // If MyCode is on for this forum and the MyCode editor is enabled inthe Admin CP, draw the code buttons and smilie inserter.
+    if ($mybb->settings['bbcodeinserter'] != 0 && (!$mybb->user['uid'] || $mybb->user['showcodebuttons'] != 0) && $calendar['allowmycode'] == 1) {
+        $codebuttons = build_mycode_inserter("message", $calendar['allowsmilies']);
+        if ($calendar['allowsmilies'] == 1) {
+            $smilieinserter = build_clickable_smilies();
+        }
+    }
 
-	// Previous selections
-	$name = $description = '';
-	if(isset($mybb->input['name']))
-	{
-		$name = htmlspecialchars_uni($mybb->get_input('name'));
-	}
+    // Previous selections
+    if (isset($mybb->input['name'])) {
+        $select['name'] = $mybb->get_input('name');
+    }
 
-	if(isset($mybb->input['description']))
-	{
-		$description = htmlspecialchars_uni($mybb->get_input('description'));
-	}
+    if (isset($mybb->input['description'])) {
+        $select['description'] = $mybb->get_input('description');
+    }
 
-	$single_month = $start_month = $end_month = $repeats_sel = $repeats_3_days = $repeats_4_occurance = $repeats_4_weekday = $repeats_5_month = $repeats_5_occurance = $repeats_5_weekday = $repeats_5_month2 = array();
-	foreach(range(1, 12) as $number)
-	{
-		$single_month[$number] = $start_month[$number] = $end_month[$number] = $repeats_5_month[$number] = $repeats_5_month2[$number] = '';
-	}
-	foreach(range(1, 5) as $number)
-	{
-		$repeats_sel[$number] = '';
-	}
-	foreach(range(0, 6) as $number)
-	{
-		$repeats_3_days[$number] = $repeats_4_weekday[$number] = $repeats_5_weekday[$number] = '';
-	}
-	foreach(range(1, 4) as $number)
-	{
-		$repeats_4_occurance[$number] = $repeats_5_occurance[$number] = '';
-	}
-	$repeats_4_occurance['last'] = $repeats_5_occurance['last'] = '';
-	$repeats_4_type = array(1 => '', 2 => '');
-	$repeats_5_type = array(1 => '', 2 => '');
+    $select['single_month'] = $select['start_month'] = $select['end_month'] = $select['repeats_sel'] = $select['repeats_3_days'] = $select['repeats_4_occurance'] = $select['repeats_4_weekday'] = $select['repeats_5_month'] = $select['repeats_5_occurance'] = $select['repeats_5_weekday'] = $select['repeats_5_month2'] = array();
+    foreach (range(1, 12) as $number) {
+        $select['single_month'][$number] = $select['start_month'][$number] = $select['end_month'][$number] = $select['repeats_5_month'][$number] = $select['repeats_5_month2'][$number] = '';
+    }
+    foreach (range(1, 5) as $number) {
+        $select['repeats_sel'][$number] = '';
+    }
+    foreach (range(0, 6) as $number) {
+        $select['repeats_3_days'][$number] = $select['repeats_4_weekday'][$number] = $select['repeats_5_weekday'][$number] = '';
+    }
+    foreach (range(1, 4) as $number) {
+        $select['repeats_4_occurance'][$number] = $select['repeats_5_occurance'][$number] = '';
+    }
 
-	if($mybb->request_method == "post")
-	{
-		$single_day = $mybb->get_input('single_day', MyBB::INPUT_INT);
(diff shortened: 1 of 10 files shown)


calendar.php
```diff
@@ -242,316 +242,229 @@
...
+        if ($year == $end_year) {
+            $select['end_year'] = "selected=\"selected\"";
+        } else {
+            $select['end_year'] = '';
+        }
+
+        $years[] = $year;
+    }
+
+    $timezones = build_timezone_select("timezone", $timezone);
```
euantorano: I wonder if this is worth making a Twig function - do we use the same function elsewhere (such as the User CP)?

calendar.php
```diff
@@ -242,316 +242,229 @@
...
+        $select['repeats_5_years2'] = 1;
+        $timezone = $mybb->user['timezone'];
+    }
+
+    $years = [];
+
+    // Construct option list for years
+    for ($year = my_date('Y'); $year < (my_date('Y') + 5); ++$year) {
+        if ($year == $single_year) {
+            $select['single_year'] = "selected=\"selected\"";
```
euantorano: Would be nice to do this logic inside the view rather than building HTML in the backend code. Same goes with the other instances of this type of code.

inc/views/base/calendar/addevent.twig
```diff
@@ -0,0 +1,404 @@
...
+                        </select>
+                    </td>
+                </tr>
+            {% else %}
+                <input type="hidden" name="calendar" value="{{ calendar.cid }}" />
+            {% endif %}
+            <tr>
+                <td class="trow2" style="vertical-align: top;"><strong>{{ lang.event_date }}</strong></td>
+                <td class="trow1">
+                    <script type="text/javascript">
```
euantorano: A note for @justinsoltesz @Eric-Jackson - would be nice to take this inline JS out. The less inline JS we have, the better in my opinion.

inc/views/base/calendar/addevent.twig
```diff
@@ -0,0 +1,404 @@
...
+                            {
+                                $("#repeats_type_"+repeats).show();
+                            }
+                        }
+                    // -->
+                    </script>
+                    <dl style="margin-top: 0; margin-bottom: 0; width: 100%;">
+                        <dt><input type="radio" name="type" value="single" {{ select.type_single }} id="type_single" onclick="toggleType();" /> <label for="type_single"><strong>{{ lang.type_single }}</strong></label></dt>
+                        <dd style="margin-top: 4px;" id="single_selects">
+                            <select name="single_day">
```
euantorano: Perhaps a good idea to try and abstract this into a loop and possibly break it into a function or partial view or something?

calendar.php
PaulBender: As far as I'm aware it's the same everywhere. So it could be a Twig function

calendar.php
PaulBender: I agree, but I'm not sure how to do that.

inc/views/base/calendar/addevent.twig
JordanMussi: Would be great if we could have none so that admins can use a CSP without `unsafe-inline`.

calendar.php
euantorano: Two options:

- Do the conditional check in Twig
- Set a variable such as `$select['single_year_selected'] = true` and in twig do something like:
```twig
<option{%if select.single_year_selected %} selected="selected"{%endif}/>
```

Option two is the easiest by far.

calendar.php
euantorano: Alright, sounds good. Not sure whether we should have a `TimeDateExtension`  for this and `my_date` (and whether we have any other time/date related functions to add to it) or if we should just create a `CoreExtension`/`MybbExtension` with these kinds of functions...

PaulBender:
I've finished the conversion work on the calendar. Once I fix the above issues (I hope to do them tomorrow) this will be ready to merge.

One thing I want to ask everyone, should we keep the $gobutton function? I've been including it so far as {{ gobutton|raw }} but it seems pointless. All it does is load a template that contains a submit button, something that can be done inline easily.

euantorano:
I just moved it inline for the subscriptions template personally.

PaulBender:
That should be everything. If there are no other issues this can be merged.

inc/views/base/calendar/addevent.twig
```diff
@@ -0,0 +1,320 @@
...
+                        <dt><input type="radio" name="type" value="single"{% if select.type_single %} checked="checked"{% endif %} id="type_single" onclick="toggleType();" /> <label for="type_single"><strong>{{ lang.type_single }}</strong></label></dt>
+                        <dd style="margin-top: 4px;" id="single_selects">
+                            <select name="single_day">
+                                {% for day in days %}
+                                    <option value="{{ day.day }}"{% if day.single_day %} selected="selected"{% endif %}>{{ day.day }}</option>
+                                {% endfor %}
+                            </select>
+                            &nbsp;
+                            <select name="single_month">
+                                <option value="1"{% if select.single_month.1 %} selected="selected"{% endif %}>{{ lang.month_1 }}</option>
```
euantorano: Not a big deal, but repeated blocks like these where it's easy to forget to update a single number make me anxious. The problem with using a loop here would be the month language variable, right? If so, you an kind of cheat:

```
{% for i in 1..12 %}
    <option value="{{ i }}"{% if select.single_month[i] %} selected="selected"{% endif %}>{{ attribute(lang, 'month_' ~ i) }}</option>
{% endfor %}
```

The trick here is using the `[]` square brackets for the condition, and the [`attribute`](https://twig.symfony.com/doc/2.x/functions/attribute.html) core Twig function to get the language variable (with string concatenation using the `~` operator).

I __believe__ that should work, but I haven't tested it at all.

inc/views/base/calendar/calendar.twig
```diff
@@ -0,0 +1,105 @@
...
+{% block body %}
+    <table border="0" cellspacing="{{ theme.borderwidth }}" cellpadding="{{ theme.tablespace }}" class="tborder">
+        <thead>
+            <tr>
+                <td class="thead" colspan="8">
+                    <div class="float_right">
+                        {% if calendar.prev_link %}
+                            <a href="{{ calendar.prev_link|raw }}">&laquo; {{ calendar.prev_month_name }} {{ calendar.prev_month_year }}</a>
+                        {% endif %}
+                        {% if calendar.sep == true %}
```
euantorano: Could simplify this to `{% if calendar.sep %}`, since you already use the short form for `next_link` below (unless this can be a string or some other type or something...).

inc/views/base/calendar/calendar.twig
```diff
@@ -0,0 +1,105 @@
...
+                                {% if calendar_permissions.canaddevents %}
+                                    <a href="calendar.php?action=addevent&amp;calendar={{ calendar.cid }}">{{ lang.add_public_event }}</a> | <a href="calendar.php?action=addevent&amp;calendar={{ calendar.cid }}&amp;private=1">{{ lang.add_private_event }}</a>
+                                {% endif %}
+                            </td>
+                            <td align="right">
+                            <span class="smalltext"><strong>{{ lang.jump_month }}</strong></span>
+                            <select name="month">
+                                <option value="{{ calendar.month }}">{{ calendar.currentmonth }}</option>
+                                <option value="{{ calendar.month }}">----------</option>
+                                <option value="1">{{ lang.alt_month_1 }}</option>
```
euantorano: Again, could use the same loop trick I mentioned in the other template file hopefully.

inc/views/base/calendar/dayview.twig
```diff
@@ -0,0 +1,84 @@
...
+                            {% if calendar_permissions.canaddevents %}
+                                <a href="calendar.php?action=addevent&amp;calendar={{ calendar.cid }}">{{ lang.add_public_event }}</a> | <a href="calendar.php?action=addevent&amp;calendar={{ calendar.cid }}&amp;private=1">{{ lang.add_private_event }}</a>
+                            {% endif %}
+                        </td>
+                        <td align="right">
+                            <span class="smalltext"><strong>{{ lang.jump_month }}</strong></span>
+                            <select name="month">
+                                <option value="{{ calendar.month }}">{{ calendar.currentmonth }}</option>
+                                <option value="{{ calendar.month }}">----------</option>
+                                <option value="1">{{ lang.alt_month_1 }}</option>
```
euantorano: Loop trick again 😉

inc/views/base/calendar/weekview.twig
```diff
@@ -0,0 +1,111 @@
...
+                            {% if calendar_permissions.canaddevents %}
+                                <a href="calendar.php?action=addevent&amp;calendar={{ calendar.cid }}">{{ lang.add_public_event }}</a> | <a href="calendar.php?action=addevent&amp;calendar={{ calendar.cid }}&amp;private=1">{{ lang.add_private_event }}</a>
+                            {% endif %}
+                        </td>
+                        <td align="right">
+                            <span class="smalltext"><strong>{{ lang.jump_month }}</strong></span>
+                            <select name="month">
+                                <option value="{{ calendar.week_from_month }}">{{ calendar.week_from_one }}</option>
+                                <option value="{{ calendar.week_from_month }}">----------</option>
+                                <option value="1">{{ lang.alt_month_1 }}</option>
```
euantorano: Loop again 😄

PaulBender:
I've converted the most months to use for loops. The ones I didn't I moved to their own file and included it where necessary (so there's only one instance instead of four).

euantorano:
Nice one, thanks. Looks much cleaner now!

euantorano:
@PaulBender Can you update the template list in #2972 with the templates you've sorted with this?

inc/views/base/calendar/addevent.twig
Eric-Jackson: Yeah same goes for inline CSS imo, which is happening after the conversion to Twig I believe. Not sure if it would be better to remove JS now, or wait until we're making the changes to the HTML for a responsive theme.

PR #2985, opened 2018-01-12
Working on #2972 Template conversion to Twig Format (breadcrumb)

chack1172:


inc/functions.php
```diff
@@ -4233,69 +4233,32 @@ function fix_mktime($format, $year)
...
-	}
-
-	eval("\$activebit = \"".$templates->get("nav_bit_active")."\";");
-	eval("\$donenav = \"".$templates->get("nav")."\";");
-
-	return $donenav;
+	return \MyBB\template('partials/breadcrumb.twig', [
+		'navbits' => $navbits,
+		'multipage' => $multipage,
+		'theme' => $theme
```
euantorano: You shouldn't need to pass in `theme` here. It's provided by the [`ThemeExtension`](https://github.com/mybb/mybb/blob/develop/1.9/inc/src/Twig/Extensions/ThemeExtension.php).

PaulBender:
Twig function is probably the best.

PR #2987, opened 2018-01-14
Announcements — For #2972

WildcardSearch:
`announcements.php` Twig conversion for #2972

### DIFF
--- a/inc/views/base/announcements/announcements.twig
+++ b/inc/views/base/announcements/announcements.twig
@@ -1,0 +1,30 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
+    <title>{{ mybb.settings.bbname }} - {{ lang.forum_announcement }}</title>
+{% endblock head %}
+
+{% block body %}
+
+<table border="0" cellspacing="{{theme.borderwidth}}" cellpadding="{{theme.tablespace}}" class="tborder tfixed clear no_bottom_border">
+	<tr>
+		<td class="thead">
+			<strong>{{lang.forum_announcement}}</strong>
+		</td>
+	</tr>
+	<tr>
+		<td id="posts_container">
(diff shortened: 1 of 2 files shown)


inc/views/base/announcements/announcements.twig
```diff
@@ -0,0 +1,30 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
+    <title>{{ mybb.settings.bbname }} - {{ lang.forum_announcement }}</title>
+{% endblock head %}
+
+{% block body %}
+
+<table border="0" cellspacing="{{theme.borderwidth}}" cellpadding="{{theme.tablespace}}" class="tborder tfixed clear no_bottom_border">
```
euantorano: Very small nitpick, but pesonally I like to have a space between the `{{` and the variable name 😄

WildcardSearch:
Done @euantorano

euantorano:
One more small thing @WildcardSearch. This line uses `htmlspecialchars_uni`: https://github.com/WildcardSearch/mybb/blob/9cf50b02a6eb56a5cf02fde0357e6114e3c7ac4c/announcements.php#L99

This will lead to the subject being escaped twice and displaying weird characters (Twig automatically escapes). The best approach is to remove the `htmlspecialchars_uni` on that line.

WildcardSearch:
Done @euantorano 

So going forward, remove all `htmlspecialchars_uni()`, correct?

euantorano:
@WildcardSearch Kind of. There are some cases where they may still be needed such as when passing user submitted content (usernames, posts, etc.) into a language variable that has HTML (some language variables have `<br />` and such in them). In that case, you need to still use `htmlspecialchars_uni` as we have to output the language variable with `|raw` (which disable's Twig's escaping).

PR #2988, opened 2018-01-14
Postbit — For #2972

WildcardSearch:
For #2972 

**Main templates:**
- [x] postbit -> removed
- [x] posbit_classic -> `postbit/postbit.twig`

**Sub templates:**
- [x] postbit_groupimage -> removed
- [x] postbit_userstar -> removed
- [x] postbit_online -> removed
- [x] postbit_away -> removed
- [x] postbit_offline -> removed
- [x] postbit_avatar -> removed
- [x] postbit_find -> removed
- [x] postbit_pm -> removed
- [x] postbit_rep_button -> removed
- [x] postbit_www -> removed
- [x] postbit_email -> removed
- [x] postbit_reputation -> removed
- [x] postbit_warn -> removed
- [x] postbit_warninglevel -> removed
- [x] postbit_purgespammer -> removed
- [x] postbit_profilefield_multiselect_value -> removed
- [x] postbit_profilefield_multiselect -> removed
- [x] postbit_profilefield -> removed
- [x] postbit_author_user -> removed
- [x] postbit_author_guest -> removed
- [x] postbit_reply_pm -> removed
- [x] postbit_forward_pm -> removed
- [x] postbit_delete_pm -> removed
- [x] postbit_replyall_pm -> removed
- [x] postbit_editedby_editreason -> removed
- [x] postbit_editedby -> removed
- [x] postbit_editreason -> removed
- [x] postbit_edit -> removed
- [x] postbit_quickdelete -> removed
- [x] postbit_quickrestore -> removed
- [x] postbit_inlinecheck -> removed
- [x] postbit_posturl -> removed
- [x] postbit_quote -> removed
- [x] postbit_multiquote -> removed
- [x] postbit_report -> removed
- [x] announcement_edit -> removed
- [x] announcement_quickdelete -> removed
- [x] postbit_iplogged_show -> removed
- [x] postbit_iplogged_hiden -> removed
- [x] postbit_status -> removed
- [x] postbit_signature -> removed
- [x] postbit_icon -> removed
- [x] postbit_deleted -> removed
- [x] postbit_ignored -> removed
- [x] postbit_deleted_member -> removed
- [x] postbit_attachments_thumbnails_thumbnail -> `postbit/postbit_attached_thumbnail`
- [x] postbit_attachments_images_image -> `postbit/postbit_attached_image`
- [x] postbit_attachments_attachment -> `postbit/postbit_attachment`
- [x] postbit_attachments_attachment_unapproved -> removed
- [x] postbit_attachments_thumbnails -> removed
- [x] postbit_attachments_images -> removed
- [x] postbit_attachments -> removed

Suggestions/reviews welcome.


euantorano:
I believe @justinsoltesz and @Eric-Jackson are talking about removing the horizontal postbit (`postbit`), so it's probably not worth translating that one - just do `postbit_classic`, but name it as `postbit` and always render that template. We'll have to remove the option from the user CP options too.

inc/views/base/postbit/postbit.twig
```diff
@@ -0,0 +1,46 @@
...
+		<div class="author_information">
+				<strong><span class="largetext">{{ post.profilelink|raw }}</span></strong> {{ post.onlinestatus|raw }}<br />
+				<span class="smalltext">
+					{{ post.usertitle|raw }}<br />
+					{{ post.userstars|raw }}
+					{{ post.groupimage|raw }}
+				</span>
+		</div>
+		<div class="author_statistics">
+			{{ post.user_details|raw }}
```
euantorano: Might be worth moving the content of the user details tmeplate inline, since the plan is to have one postbit template.

inc/views/base/postbit/postbit.twig
```diff
@@ -0,0 +1,46 @@
...
+			{{ post.iplogged|raw }}
+			<div class="float_right">
+				{{ post.poststatus|raw }}
+			</div>
+		</div>
+		{{ post.input_editreason|raw }}
+	</div>
+	<div class="post_controls">
+		<div class="postbit_buttons author_buttons float_left">
+			{{ post.button_email|raw }}{{ post.button_pm|raw }}{{ post.button_www|raw }}{{ post.button_find|raw }}{{ post.button_rep|raw }}
```
euantorano: Ouch, lots of raws here. We should probably do the conditions in the template and include other templates for each button.

WildcardSearch:
Will do.

justinsoltesz:
Correct, we will only have one postbit style in the 1.9 theme.

inc/views/base/postbit/postbit.twig
justinsoltesz: I don't think there's even a good reason to break out the buttons into separate templates, is there? They will only be used in this template when the postbits are combined.

Having the conditions in the template would be ideal.

WildcardSearch: This is a WIP. I just did that to get me up and displaying correctly on the new template. The next move is to walk back through the included templates and work them out one at a time.

inc/views/base/postbit/postbit.twig
WildcardSearch: Will do.

inc/views/base/postbit/postbit.twig
WildcardSearch: Justin ninja'd me. I was thinking of passing an array of boolean values to Twig and just `if` each button. What do you guys think?

euantorano: That sounds like the best approach to me @WildcardSearch!

WildcardSearch: Yeah, @euantorano, that sounds like a plan.

Trust me, everything I do, I am thinking about how I am going to manipulate these templates in my plugins lol

WildcardSearch:
Update PR with a list of templates to work through before postbit will be converted fully.

justinsoltesz:
Just a general comment after looking at `postbit.twig` -- the logic should be indented to match the HTML syntax. Makes it easier to read.

WildcardSearch:
I'm not sure I understand what you mean, @justinsoltesz

justinsoltesz:
Take a look at how the various `if` statements are indented in this file:

https://github.com/mybb/mybb/blob/develop/1.9/inc/views/base/calendar/calendar.twig

WildcardSearch:
I see what you mean, @justinsoltesz . But in the example you linked, the outputted HTML is not indented correctly, ie.:

https://github.com/mybb/mybb/blob/d398853cdaf789a82c3683290a08a19189a1db2a/inc/views/base/calendar/calendar.twig#L27-L30

The conditional `td` is indented too far.

justinsoltesz:
MyBB's output already isn't indented correctly... I think it's better for the template to be easier to read (and subsequently be easier to edit) than have the outputted page source be indented correctly.

WildcardSearch:
Gotcha. I will do a code cleanup when I get all the work done.

WildcardSearch:
@euantorano and @justinsoltesz, I have things working with the new template system for postbit and code cleaned up as best as I could.

euantorano:
Nice one @WildcardSearch, thanks. I'll have a look and tets of this either this week or at the weekend.

inc/views/base/postbit/postbit.twig
```diff
@@ -0,0 +1,218 @@
...
+			{% if post.isguest %}
+				&nbsp;
+			{% else %}
+				{{ lang.postbit_posts }} {{ post.postnum }}<br />
+				{{ lang.postbit_threads }} {{ post.threadnum }}<br />
+				{{ lang.postbit_joined }} {{ post.userregdate }}
+				{% if post.replink %}
+					<br />{{ lang.postbit_reputation }} {{ post.userreputation|raw }}
+				{% endif %}
+				{% for field in post.profile_fields %}
```
justinsoltesz: Does this result in all of the profile fields being shown in the postbit? I believe these fields are currently 'optional' -- you can add a variable to include them if you want, but they aren't included by default.

(That said, the 1.9 theme concepts don't have _any_ fields in the postbit...)

WildcardSearch: @justinsoltesz the profile fields that are to be displayed are loaded before `build_postbit` is called. In other words, it is a global variable `$profile_fields` that the calling code would define.

justinsoltesz: Sorry, you're right. I didn't realize they were automatically added based on a setting in the ACP in 1.8 (they definitely weren't in a previous version).

WildcardSearch: No problem. I am getting to know this section of code much better than I ever have before. I guess that is another benefit of this conversion as well as some long overdue code cleanup and convention updates.

PR #2992, opened 2018-01-16
Fixes #2948 Guest poll voting updates

### LINKED ISSUE #2948: Guest poll voting updates
Currently, if a guest votes in a poll, a cookie is stored on their computer as to what they voted in and for. As this method can be bypassed easily, a better method using IP addresses should be used.

- Add an ipaddress column to the pollvotes table and log the IP of the voter.
- Remove the cookie method of checking for guest votes and instead use the IP address of the guest to check instead.

PaulBender:
Fixes #2948

### DIFF
--- a/inc/class_core.php
+++ b/inc/class_core.php
@@ -161,6 +161,7 @@ class MyBB {
 		'adminsessions' => array('ip' => true),
 		'maillogs' => array('ipaddress' => true),
 		'moderatorlog' => array('ipaddress' => true),
+		'pollvotes' => array('ipaddress' => true),
 		'posts' => array('ipaddress' => true),
 		'privatemessages' => array('ipaddress' => true),
 		'searchlog' => array('ipaddress' => true),
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -11652,14 +11652,14 @@ if(use_xmlhttprequest == "1")
 <td class="{$optionbg}" style="text-align: center;">{$votes}</td>
 <td class="{$optionbg}" style="text-align: center;">{$percent}%</td>
 </tr>]]></template>
-		<template name="showthread_poll_results" version="1800"><![CDATA[<table border="0" cellspacing="{$theme['borderwidth']}" cellpadding="{$theme['tablespace']}" class="tborder tfixed">
+		<template name="showthread_poll_results" version="1815"><![CDATA[<table border="0" cellspacing="{$theme['borderwidth']}" cellpadding="{$theme['tablespace']}" class="tborder tfixed">
 <colgroup>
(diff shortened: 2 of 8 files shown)


kawaii:
Ran into the following error when testing this on MyBB 1.8.14 with PostgreSQL 10.1:
```
SQL Error:
0 - ERROR: invalid byte sequence for encoding "UTF8": 0xac
Query:
INSERT INTO mybb_pollvotes (pid,uid,voteoption,dateline,ipaddress) VALUES ('1','1','1','1516133517','�')
```

kawaii:
Confirmed working with MariaDB 10.2 on MyBB 1.8.14, I'm going to do another PostgreSQL test quickly just to validate that I'm not doing something wrong.

euantorano:
Encodings are a pain in the backside unfortunately. I can't remember if we set the encoding when we startup or not. I'll have a look at this tomorrow and see if I can replicate your issue.

PaulBender:
Try it now. I changed the way poll votes are inserted into the database from its old write_query method to using insert_query/insert_query_multiple method.

Ben-MyBB:
@kawaii

kawaii:
Thanks @PaulBender, your push fixed the issue I was experiencing with PostgreSQL. This is ready to be merged now. I do have one comment however: I voted in a poll using a logged in account, then when logged out I was able to undo the poll vote I had just made (due to the IPs matching of course). Would it not make sense for us to also differentiate between votes made by guests (where we can scrutinize the IP address) and votes made by registered users? In the event where I am using a shared IP address, someone else without a forum account could potentially undo my vote as a guest user without needing to be logged in to my original account. @mybb/developers @mybb/sqa

euantorano:
Good point @kawaii. I definitely think that’s unintended behaviour. If a user has voted from an account, only somebody logged in to that account should be able to reverse it.

kawaii:
Another potential issue I forsee (though personally this is getting a little out of scope) is that multiple guests using the same shared connection would be considered the same user by MyBB. Though I can't honestly see a workaround for that unless you want to start scrutinizing user agents too.

PaulBender:
I added a uid check for guests.

kawaii:
Odd, it seems that the UID check doesn't have the intended effect. As a guest it still displays as though I've already voted but now attempting to undo the vote results in the following error message instead of processing the undo request:
```
You haven't voted in this poll.
```

PaulBender:
I forgot to add the check to the showthread query.

I also changed the undo vote bit to its own variable (something I believed @justinsoltesz complained about).

kawaii:
Great, works as intended now. Thanks @PaulBender! :slightly_smiling_face:

PR #2994, opened 2018-01-16
Fixes #2993 - banning without ban reason results in blank page

### LINKED ISSUE #2993: Banning a member without filling in a ban reason results in blank page
If you accidentally forget to fill in a ban reason while trying to ban a member, you'll end up in a blank modcp.php page.

expected result: you should be thrown back where you came from with an error message printed on screen.

ronie:
When you accidentally forget to enter a ban reason when banning a user, you'll end up in a blank modcp.php page.

The issue is caused by $errors not properly being defined as an array.

 This PR fixes issue #2993

### DIFF
--- a/modcp.php
+++ b/modcp.php
@@ -4267,6 +4267,8 @@ if($mybb->input['action'] == "do_banuser" && $mybb->request_method == "post")
 		}
 	}
 
+	$errors = array();
+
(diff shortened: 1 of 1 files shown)


JordanMussi:
@kawaii Milestone and label should be set on #2993 not this PR.

PR #2995, opened 2018-01-16
[FIX] string to array conversation causing Illegal string offset warning on PHP 7.2

Matslom:
Fix for calendar.php Illegal string offset warning on PHP 7.2

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -2232,7 +2232,8 @@ if(!$mybb->input['action'])
     $in_month = 0;
 
     // Iterate weeks (each week gets a row)
-	$week = $day_bit = '';
+    $week = [];
+	$day_bit = [];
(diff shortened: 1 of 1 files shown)


calendar.php
```diff
@@ -2232,7 +2232,8 @@
     $in_month = 0;
 
     // Iterate weeks (each week gets a row)
-	$week = $day_bit = '';
+    $week = [];
```
euantorano: Needs to be `array()` since we maintian PHP 5.2 compatibility in 1.8 unfortunately.

Matslom: 1.8? but it's dev1.9 branch

Matslom: I will check it tomorrow, but I'm pretty sure that on 1.8 that bug is not present.

PR #2999, opened 2018-01-19
Consider both primary and secondary groups for Moderator CP

### LINKED ISSUE #2990: ModCP Permissions to view Reported Posts
Permissions don't work to see modcp reported posts if the permission group is secondary group.

From modcp.php:

Code:if(&#36;mybb-&gt;input['action'] == "reports")
{
    if(&#36;mybb-&gt;usergroup['canmanagereportedcontent'] == 0)
    {
        error_no_permission();
    }
So that has to be changed to reflect secondary group permissions.

Original thread: [ModCP Permissions to view Reported Posts](https://community.mybb.com/thread-215384.html)

euantorano:
Fixes #2990 

The code in the current `feature` branch runs a query globally in the Mod CP if a user isn't a super moderator to see if they have any moderator access at all and sets up various variables (such as `$numreportedposts` which was causing the problem in #2990 itself). However, it only considers the user's UID and their primary usergroup (`$mybb->user['usergroup']`) and ignores any additional groups that may be configured when running this query.

I'm not sure if it was like that for a reason, so hopefully this patch doesn't break anything else - it really needs testing with some different permission setups to make sure it doesn't cause problems.

### DIFF
--- a/modcp.php
+++ b/modcp.php
@@ -58,7 +58,14 @@ $flist_queue_attach = $wflist_reports = $tflist_reports = $flist_reports = $tfli
 $moderated_forums = array();
 if($mybb->usergroup['issupermod'] != 1)
 {
-	$query = $db->simple_select("moderators", "*", "(id='{$mybb->user['uid']}' AND isgroup = '0') OR (id='{$mybb->user['usergroup']}' AND isgroup = '1')");
+	$in = explode(',', $mybb->user['additionalgroups']);
+	$in = array_map('intval', $in);
+	$in[] = (int) $mybb->user['usergroup'];
+	$in = array_unique($in);
+
+	$in_query = implode(',', $in);
+
+	$query = $db->simple_select("moderators", "*", "(id='{$mybb->user['uid']}' AND isgroup = '0') OR (id IN ({$in_query}) AND isgroup = '1')");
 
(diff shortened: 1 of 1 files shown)


kawaii:
This indeed fixes the issue noted in #2990 but I suggest we get a few of the others from @mybb/developers and @mybb/sqa to test it too, since it's a fairly critical and delicate part of the software.

grulichw:
Why not just change the query to:

‘’’php
$query = $db->simple_select("moderators", "*", "(id='{$mybb->user['uid']}' AND isgroup = '0') OR (id IN ({$mybb->usergroup['all_usergroups']}) AND isgroup = '1')");
‘’’

Is $mybb->usergroup[‘all_usergroups’] less riable than your method? Otherwise we should simply change the one query line instead of adding all the extra checks through $mybb->user. It is a better code fix that over large loads = reduced computing power.

euantorano:
@grulichw There is no such array key index one my install - I dumped the `$mybb->user` array to double check. There is an `all_usergroups` array key in the array retrieved by calling [`usergroup_permissions`](https://github.com/mybb/mybb/blob/f7b8a16d6774a98c16f5c21acfd23fe52270c525/inc/functions.php#L1253) but I can't find any other reference to that array key in MyBB.

euantorano:
@grulichw I hadn't realised that was there actually. It's probably an easier solution, thanks.

Issue #2976, opened 2018-01-08
Useful Twig functions


PaulBender:
We should consider adding some Twig functions for the following:
- [ ] Inline Errors
- [ ] codebuttons (the editor)
- [ ] smilieinserter
- [ ] build_profile_link
- [ ] my_date

JordanMussi:
Also https://github.com/mybb/mybb/pull/2975#discussion_r160036829
> I wonder if it's worth marking the trans function as HTML safe... There are quite a few translations that have HTML in them unfortunately.

Shade-:
Would be nice to add get_thread_link too.

WildcardSearch:
Marked `$lang->parse()` replacement marked.

arcalys:
Is this still something that needs to be done for 1.9?  
Some of the unmarked items already exist, such as:
- `get_thread_link()`


It would be helpful to clarify what’s still left to implement.

Sama34:
` get_*_link()` are taken care (fully, finished) on #5171

I went ahead and struck-through _captcha_ because that seems to require some server side handling. Please correct me if wrong.

As per the rest, I think they may be useful.

Possibly related : https://discord.com/channels/215876847634743296/386812669513957386/1357804632495231086

PR #3000, opened 2018-01-19
For #2976 — $lang->parse() Replacement

WildcardSearch:
For #2976

This replaces `$lang->parse()` with a Twig filter that works like: `{{ myVar|lparse }}`


WildcardSearch:
Oops, wrong branch lol

PR #3007, opened 2018-01-22
Fix #3006 - custom moderation tool SQL error on postgresql

### LINKED ISSUE #3006: Custom moderator tools broken on PostgreSQL
If you try to execute a custom moderator tool when using PostgreSQL as the database engine, you encounter the following SQL error:

```

SQL Error:
    0 - ERROR: column "2" does not exist LINE 1: SELECT * FROM mybb_modtools WHERE tid="2" ^ 
Query:
    SELECT * FROM mybb_modtools WHERE tid="2" 
```

euantorano:
Fixes #3006 by removing `"` around moderation tool ID in query. Should still work on both postgres and mysql.

### DIFF
--- a/inc/class_custommoderation.php
+++ b/inc/class_custommoderation.php
@@ -32,7 +32,7 @@ class CustomModeration extends Moderation
 		global $db;
 
 		// Get tool info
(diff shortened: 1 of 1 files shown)


PaulBender:
Works for me as well.

PR #3009, opened 2018-01-24
Update .gitattributes

kawaii:
Removes some unnecessary files from exported releases.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,7 +2,7 @@
 .gitattributes export-ignore
 .gitignore export-ignore
 /README.md export-ignore
-/LICENCE export-ignore
+/LICENSE export-ignore
 /CONTRIBUTING.md export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1 +1,8 @@
 * text=auto eol=lf
+.gitattributes export-ignore
+.gitignore export-ignore
+/README.md export-ignore
+/LICENCE export-ignore
```
dvz: Should be _LICENSE_

dvz:
Not sure about excluding the sample `.htaccess`/configuration files - it's not best that they are distributed in this form, but the idea was that administrators have the content available for installation.

kawaii:
@dvz I personally think that those files should instead be in the MyBB documentation and not part of the source distribution. I've updated the PR with the typo fix for the license file.

PR #3011, opened 2018-01-25
#2972 (stats.php) forum statistics to twig

Matslom:
For #2972 

- [x]  stats
- [x]  stats_thread
- [x]  stats_topforum

```diff
+ stats -> stats/stats.twig
- stats_thread
- stats_topforum
```


Matslom:
Not sure about this lines. It's not very messy, but maybe splitting this lang into two parts would be better idea. 
https://github.com/Matslom/mybb/blob/fabee05ad8ff33ac807066949b94b1f73b23d4e1/inc/views/base/stats/stats.twig#L41-L42

euantorano:
@Matslom It may make sense to split that language variable into two parts, yeah. It doesn't look too bad like this though, so I'd be happy to merge this if you're happy with it.

Matslom:
@euantorano So I will split it to two parts. Tomorrow most likely.

euantorano:
Great, thank you very much.

Matslom:
@euantorano I will try fix it in this week.

skmedix:
@euantorano You can close it now.

PR #3018, opened 2018-02-04
Working on #2972 Template Conversion to Twig Format (managegroup.php)

PaulBender:
#2972 for managegroup.php

### DIFF
--- a/managegroup.php
+++ b/managegroup.php
@@ -210,10 +210,8 @@ elseif ($mybb->input['action'] == "joinrequests") {
         ORDER BY u.username ASC
     ");
     while ($user = $db->fetch_array($query)) {
-        $user['reason'] = htmlspecialchars_uni($user['reason']);
         $altbg = alt_trow();
         $regdate = my_date($mybb->settings['dateformat'], $user['regdate']);
-        $user['username'] = htmlspecialchars_uni($user['username']);
         $user['profilelink'] = build_profile_link($user['username'], $user['uid']);
 
         $users[] = $user;
@@ -272,8 +270,7 @@ else
 
     if ($usergroup['type'] == 5) {
         $usergroup['usergrouptype'] = $lang->group_public_invite;
-    }
-    elseif ($usergroup['type'] == 4) {
+    } else if ($usergroup['type'] == 4) {
         $query = $db->simple_select("joinrequests", "COUNT(*) AS req", "gid='{$gid}'");
         $numrequests = $db->fetch_array($query);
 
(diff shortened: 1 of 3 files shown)


managegroup.php
```diff
@@ -195,38 +197,42 @@
...
+
+    $query = $db->query("
+        SELECT j.*, u.uid, u.username, u.postnum, u.regdate
+        FROM " . TABLE_PREFIX . "joinrequests j
+        LEFT JOIN " . TABLE_PREFIX . "users u ON (u.uid=j.uid)
+        WHERE j.gid='{$gid}' AND j.uid != 0
+        ORDER BY u.username ASC
+    ");
+    while ($user = $db->fetch_array($query)) {
+        $user['reason'] = htmlspecialchars_uni($user['reason']);
```
euantorano: The use of `htmlspecialchars_uni()` here (and perhaps below?) will cause double escaping.

managegroup.php
```diff
@@ -257,185 +263,141 @@
...
+    $plugins->run_hooks("managegroup_start");
+
+    $lang->members_of = $lang->sprintf($lang->members_of, htmlspecialchars_uni($usergroup['title']));
+    $lang->add_member = $lang->sprintf($lang->add_member, htmlspecialchars_uni($usergroup['title']));
+    $lang->invite_member = $lang->sprintf($lang->invite_member, htmlspecialchars_uni($usergroup['title']));
+    $usergroup['pendingrequests'] = false;
+
+    if ($usergroup['type'] == 5) {
+        $usergroup['usergrouptype'] = $lang->group_public_invite;
+    }
```
euantorano: Should be `} else if (...`.

managegroup.php
```diff
@@ -257,185 +263,141 @@
...
+
+    // Display group leaders (if there is any)
+    $query = $db->query("
+        SELECT g.*, u.username, u.usergroup, u.displaygroup
+        FROM " . TABLE_PREFIX . "groupleaders g
+        LEFT JOIN " . TABLE_PREFIX . "users u ON (u.uid=g.uid)
+        WHERE g.gid = '{$gid}'
+    ");
+
+    $leaders_array = array();
```
euantorano: For consistency with the below `$leaders = []`, can you use square brackets here too to initialise the array?

managegroup.php
```diff
@@ -257,185 +263,141 @@
...
+        $leader_count = $db->num_rows($query);
+        while ($leader = $db->fetch_array($query)) {
+            $leader['username'] = htmlspecialchars_uni($leader['username']);
+            $leader_name = format_name($leader['username'], $leader['usergroup'], $leader['displaygroup']);
+            $leader['profilelink'] = build_profile_link($leader_name, $leader['uid']);
+
+            $leaders_array[] = $leader['uid'];
+
+            // Get commas...
+            if ($loop != $leader_count) {
```
euantorano: This logic could be done in the view itself, using [the `loop.last` variable](https://twig.symfony.com/doc/2.x/tags/for.html#the-loop-variable).

PaulBender:
All changes have been made

PR #3028, opened 2018-02-15
Remove my_strip_tags from affecting help documents

nsuchy:
As Euan and I previously discussed in private inquiries, help documents should allow Javascript code to execute. This pull request removes my_strip_tags in-order to allow them.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -8743,26 +8743,6 @@ function my_validate_url($url, $relative_path=false, $allow_local=false)
 }
 
 /**
- * Strip html tags from string, also removes <script> and <style> contents.
- *
- * @param  string $string         String to stripe
- * @param  string $allowable_tags Allowed html tags
- *
- * @return string                 Striped string
- */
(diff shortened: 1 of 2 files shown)


nsuchy:
The my_strip_tags function is/was only used in misc.php for help tags, as this pull request removes its use in misc.php, there's no longer a need for this function to remain in functions.php and it should be removed.

nsuchy:
After discussing with Euan, some plugins may be (very bad idea considering the function is vulnerable to an easy XSS attack) using the function my_strip_tags, I've reverted the commit which removes the function. Welp time to start reviewing every MyBB plugin and reporting the XSS issue if they are using that function xD

JordanMussi:
> I've reverted the commit which removes the function.

@nsuchy might be better to add an `@deprecated` to the phpDoc for the function and remove it in 1.9.

nsuchy:
@JordanMussi Sure thing

Shade-:
There's no reason why this should be closed @nsuchy.

PR #3032, opened 2018-02-16
Working on #2972 Template Conversion to Twig Format (member.php)

PaulBender:
#2972 for member.php

This includes all member.php pages except for the registration and profile pages (due to how complex they are; I'll do those as separate PRs).

### DIFF
--- a/inc/views/base/member/activate.twig
+++ b/inc/views/base/member/activate.twig
@@ -11,12 +11,12 @@
                 <td class="thead" colspan="2"><strong>{{ lang.activate_account }}</strong></td>
             </tr>
             <tr>
-                <td class="trow1" width="30%"><strong>{{ lang.username }}</strong></td>
-                <td class="trow1"><input type="text" class="textbox" name="username" value="{{ activate.username }}" /></td>
+                <td class="trow1" width="30%"><strong><label for="username">{{ lang.username }}</label></strong></td>
+                <td class="trow1"><input type="text" class="textbox" name="username" id="username" value="{{ activate.username }}" /></td>
             </tr>
             <tr>
-                <td class="trow1" width="30%"><strong>{{ lang.activation_code }}</strong></td>
-                <td class="trow1"><input type="text" class="textbox" name="code" value="{{ activate.code }}" /></td>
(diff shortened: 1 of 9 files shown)


inc/views/base/member/activate.twig
```diff
@@ -0,0 +1,28 @@
...
+{% endblock head %}
+
+{% block body %}
+    <form action="member.php" method="post">
+        <table border="0" cellspacing="{{ theme.borderwidth }}" cellpadding="{{ theme.tablespace }}" class="tborder">
+            <tr>
+                <td class="thead" colspan="2"><strong>{{ lang.activate_account }}</strong></td>
+            </tr>
+            <tr>
+                <td class="trow1" width="30%"><strong>{{ lang.username }}</strong></td>
```
euantorano: Can you wrap this in a `<label for="username">` and add `id="username"` to the associated input? I've noticed a lot of inputs don't have labels linked and it's useful for accessibility to add them. If they're already there before we convert to the new theme, it makes it harder to miss them 😄

inc/views/base/member/emailuser.twig
```diff
@@ -0,0 +1,42 @@
...
+            {% endif %}
+            <tr>
+                <td width="40%" class="trow2"><strong>{{ lang.email_subject }}</strong></td>
+                <td width="60%" class="trow2"><input type="text" class="textbox" size="50" name="subject" value="{{ email.subject }}" /></td>
+            </tr>
+            <tr>
+                <td valign="top" width="40%" class="trow1"><strong>{{ lang.email_message }}</strong></td>
+                <td width="60%" class="trow1"><textarea cols="50" rows="10" name="message">{{ email.message }}</textarea></td>
+            </tr>
+            {{ captcha|raw }}
```
euantorano: Captcha is fairly common, perhaps we should make a Twig function for it - thoughts?

euantorano:
Looks good, just a very minor change 👍

PaulBender:
I've added label and ids to all input fields.

inc/views/base/member/emailuser.twig
PaulBender: Definitely should be one.

euantorano:
Brilliant, thanks!

PR #3035, opened 2018-02-18
Working on #2972 Template Conversion to Twig Format (newthread.php)

PaulBender:
#2972 for posting a new thread (newthread.php)

### DIFF
--- a/inc/views/base/newthread/newthread.twig
+++ b/inc/views/base/newthread/newthread.twig
@@ -35,12 +35,39 @@
             </tr>
             {{ loginbox|raw }}
             <tr>
-                <td class="trow2" width="20%"><strong>{{ lang.thread_subject }}</strong></td>
-                <td class="trow2">{{ prefixselect|raw }}<input type="text" class="textbox" name="subject" size="40" maxlength="85" value="{{ newthread.subject }}" tabindex="1" /></td>
+                <td class="trow2" width="20%"><strong><label for="subject">{{ lang.thread_subject }}</label></strong></td>
+                <td class="trow2">
+                    {% if newthread.showprefixes %}
+                        <select name="threadprefix">
+                            <option value="0">{{ lang.no_prefix }}</option>
+                            {% for prefix in prefixes %}
+                                <option value="{{ prefix.pid }}"{% if prefix.selected %} selected="selected"{% endif %}>{{ prefix.prefix }}</option>
+                            {% endfor %}
+                        </select>
+                    {% endif %}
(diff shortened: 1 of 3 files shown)


inc/views/base/newthread/newthread.twig
```diff
@@ -0,0 +1,227 @@
...
+    {% endif %}
+    <form action="newthread.php?fid={{ forum.fid }}&amp;processed=1" method="post" enctype="multipart/form-data" name="input">
+        <input type="hidden" name="my_post_key" value="{{ mybb.post_code }}" />
+        <table border="0" cellspacing="{{ theme.borderwidth }}" cellpadding="{{ theme.tablespace }}" class="tborder">
+            <tr>
+                <td class="thead" colspan="2"><strong>{{ lang.post_new_thread }}</strong></td>
+            </tr>
+            {{ loginbox|raw }}
+            <tr>
+                <td class="trow2" width="20%"><strong>{{ lang.thread_subject }}</strong></td>
```
euantorano: As in the other PR, would you mind adding `<label>` elements where we have raw text labels for inputs? Makes them harder to miss for the new theme as it gets written that way 😄

newthread.php
```diff
@@ -510,550 +508,424 @@
...
+        $newthread['message'] = $mybb->get_input('message');
+        $newthread['subject'] = $mybb->get_input('subject');
+    }
+
+    // Generate thread prefix selector
+    if (!$mybb->get_input('threadprefix', MyBB::INPUT_INT)) {
+        $mybb->input['threadprefix'] = 0;
+    }
+
+    $prefixselect = build_prefix_select($forum['fid'], $mybb->get_input('threadprefix', MyBB::INPUT_INT));
```
euantorano: As I commented in the other PR, it might make sense to create a new version of this function that just returns an array.

newthread.php
```diff
@@ -510,550 +508,424 @@
...
+            $noshowattach = 1;
+        }
+
+        if ($mybb->usergroup['attachquota'] == 0) {
+            $newthread['friendlyquota'] = $lang->unlimited;
+        } else {
+            $newthread['friendlyquota'] = get_friendly_size($mybb->usergroup['attachquota']*1024);
+        }
+
+        $newthread['friendlyusage'] = get_friendly_size($usage['ausage']);
```
euantorano: As commented in the other PR, this may mak sense as a Twig function.

PaulBender:
I've made the changes. For thread prefixes, I changed the original function to return an array instead of templates. Did a similar thing with the post icons.

PR #3036, opened 2018-02-19
Working on #2972 Template Conversion to Twig Format (newreply.php)

PaulBender:
#2972 for posting a new reply (newreply.php)

### DIFF
--- a/inc/views/base/newreply/newreply.twig
+++ b/inc/views/base/newreply/newreply.twig
@@ -38,12 +38,29 @@
             </tr>
             {{ loginbox|raw }}
             <tr>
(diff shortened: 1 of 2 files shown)


inc/views/base/newreply/newreply.twig
```diff
@@ -0,0 +1,263 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
+    <title>{{ trans('post_reply_to', thread.subject) }}</title>
+    <script type="text/javascript" src="{{ asset_url('jscripts/post.js') }}"></script>
```
euantorano: We should probably add a new `block` to the master layout to put JS at the end, before the closing `</body>`. I'll do that at a later date and move things like this to that block.

PR #3041, opened 2018-02-26
Working on #2972 Template Conversion to Twig Format (member register)

PaulBender:
#2972 for member registration (member.php?action=register)

I wasn't sure how to handle all the inline $validator_javascript, so I left it.


inc/views/base/member/register.twig
```diff
@@ -0,0 +1,370 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
+    <title>{{ mybb.settings.bbname }} - {{ lang.registration }}</title>
+    <script type="text/javascript" src="{{ asset_url('jscripts/validate/jquery.validate.min.js') }}"></script>
+{% endblock head %}
+
+{% block body %}
+    <form action="member.php" method="post" id="registration_form">
+        <input type="text" style="visibility: hidden;" value="" name="regcheck1" />
```
euantorano: Shouldn't this just be `<input type="hidden">` or is there a particular reason for this approach?

PaulBender: I believe that's a spam bot check.

euantorano:
Ah, that would make sense.

PR #2997, opened 2018-01-18
Index page (index.php) for #2972

Matslom:
FOR #2972

**templates;**
*index*
- [x]  index
- [x]  index_birthdays
- [x]  index_birthdays_birthday
- [x]  index_boardstats
- [x]  index_logoutlink
- [x]  index_stats
- [x]  index_statspage
- [x]  index_whosonline
- [x]  index_whosonline_memberbit

```diff
+ index -> index/index.twig
- index_birthdays
- index_birthdays_birthday
+ index_boardstats -> index/boardstats.twig
- index_logoutlink
- index_stats
- index_statspage
- index_whosonline
- index_whosonline_memberbit
```

PR #3043, opened 2018-02-26
Working on #2972 Template Conversion to Twig Format (index)

Shade-:
Working on #2972, converting index templates. My first conversion, hope everything's correct.

### DIFF
--- a/inc/views/base/index/index.twig
+++ b/inc/views/base/index/index.twig
@@ -1,0 +1,95 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
+    <title>{{ mybb.settings.bbname }}</title>
+{% endblock head %}
+
+{% block body %}
+    {{ forums|raw }}
+    {% if (mybb.settings.showwol and mybb.usergroup.canviewonline) or mybb.settings.showindexstats or (mybb.settings.showbirthdays and birthdays|length > 0) %}
+    <table border="0" cellspacing="{{ theme.borderwidth }}" cellpadding="{{ theme.tablespace }}" class="tborder">
+        <thead>
+            <tr>
+                <td class="thead{{ collapsedthead.boardstats }}">
+                    <div class="expcolimage"><img src="{{ theme.imgdir }}/collapse{{ collapsedimg.boardstats }}.png" id="boardstats_img" class="expander" alt="[-]" title="[-]" /></div>
+                    <div><strong>{{ lang.boardstats }}</strong></div>
+                </td>
+            </tr>
+        </thead>
+        <tbody style="{{ collapsed.boardstats_e }}" id="boardstats_e">
+            <tr>
+                <td class="tcat"><span class="smalltext"><strong>{{ lang.whos_online }}</strong> [<a href="online.php">{{ lang.complete_list }}</a>]</span></td>
+            </tr>
+            <tr>
+                <td class="trow1">
+                    <span class="smalltext">
+                        {{ lang.online_note }}<br />
+                        {% for user in users %}
+                            {{ user.profilelink|raw }}{% if loop.last != true %}{{ lang.comma }} {% endif %}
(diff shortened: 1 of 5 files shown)


euantorano:
The PR #2997 by @Matslom also targets `index.php`. I haven't looked at this code yet, but @Matslom was using raw HTML for the forum bit templates - if this PR converts those to twig too, we'll go with this PR rather than the other.

inc/views/base/index/index.twig
```diff
@@ -0,0 +1,95 @@
...
+        <tbody style="{{ collapsed.boardstats_e }}" id="boardstats_e">
+            <tr>
+                <td class="tcat"><span class="smalltext"><strong>{{ lang.whos_online }}</strong> [<a href="online.php">{{ lang.complete_list }}</a>]</span></td>
+            </tr>
+            <tr>
+                <td class="trow1">
+                    <span class="smalltext">
+                        {{ lang.online_note }}<br />
+                        {% for user in users %}
+                        	{{ user.profilelink|raw }}{% if loop.last != true %}, {% endif %}
```
euantorano: This should use `{{ lang.comma }}` ratehr than a raw `,`.

inc/views/base/index/index.twig
```diff
@@ -0,0 +1,95 @@
...
+                <td class="tcat">
+                    <span class="smalltext"><strong>{{ lang.todays_birthdays }}</strong></span>
+                </td>
+            </tr>
+            <tr>
+            	<td class="trow1">
+                    {% spaceless %}
+                    <span class="smalltext">
+                        {% for user in birthdays %}
+                        	{{ user.profilelink }}{% if user.age %} ({{ user.age }}) {% endif %}{% if loop.last != true %}, {% endif %}
```
euantorano: Doesn't this need `|raw` for the profile link?

euantorano: Also, again use `{{ lang.comma }}`.

inc/views/base/index/index.twig
```diff
@@ -0,0 +1,95 @@
...
+                    <span class="smalltext"><strong>{{ lang.boardstats }}</strong></span>
+                </td>
+            </tr>
+            <tr>
+                <td class="trow1">
+                    <span class="smalltext">
+                        {{ lang.stats_posts_threads }}<br />
+                        {{ lang.stats_numusers }}<br />
+                        {{ lang.stats_newestuser|raw }}<br />
+                        {{ lang.stats_mostonline }}
```
euantorano: This should be `|raw` as it contains dates formatted using `my_date` which can contain HTML for relative dates.

index.php
```diff
@@ -25,343 +25,261 @@
...
+    // Build a forum cache.
+    $query = $db->simple_select('forums', '*', 'active!=0', array('order_by' => 'pid, disporder'));
 
-	$forumsread = array();
-	if(isset($mybb->cookies['mybb']['forumread']))
-	{
-		$forumsread = my_unserialize($mybb->cookies['mybb']['forumread']);
-	}
+    $forumsread = array();
+    if (isset($mybb->cookies['mybb']['forumread']))
```
euantorano: This needs reformatting to have the `{` on the same line as the `if`.

Shade-:
Ouch. Didn’t spot that @matslom already worked on it, I looked at the template list in #2972 and didn’t find index checked. Nevermind, I am going to update the list with the ones being worked on as it seems like there are some not checked yet. I am also going to fix those asap, thank you for reviewing!

Shade-:
Oh, and this PR does not convert forums yet. Do you think they should be added here or as a separate PR?

euantorano:
I'm not sure, I haven't looked at how big a job it is to convert them. Some of them are done (depth 2, which I needed for another page) but there's still a bunch to do.

Shade-:
I'll handle them in a separate PR as it looks neater.

euantorano:
Ok, fair enough. Question is whether to merge this one or the other now, as both achieve the same thing 😉

PR #3046, opened 2018-02-28
Working on #2972 Template Conversion to Twig Format (header)

Shade-:
- [x]  header_menu_calendar
- [x]  header_menu_memberlist
- [x]  header_menu_portal
- [x]  header_menu_search
- [x]  header_quicksearch
- [x]  header_welcomeblock_member
- [x]  header_welcomeblock_member_admin
- [x]  header_welcomeblock_member_moderator
- [x]  header_welcomeblock_member_pms
- [x]  header_welcomeblock_member_user
- [x]  header_welcomeblock_member_search

Twigplates edited and fixed:
- [x]  partials/header.twig (corresponding MyBB template: header)
- [x]  header/welcomeblock_guest.twig (corresponding MyBB template: header_welcomeblock_guest)

I wasn't sure about the following piece of code:
```php
// Load appropriate welcome block for the current logged in user
if ($mybb->user['uid'] != 0) {

    // Format the welcome back message
    $lang->welcome_back = $lang->sprintf($lang->welcome_back, build_profile_link(htmlspecialchars_uni($mybb->user['username']), $mybb->user['uid']), $lastvisit);

}
// Otherwise, we have a guest
else {
    switch ($mybb->settings['username_method']) {
        case 0:
            $login_username = $lang->login_username;
            break;
        case 1:
            $login_username = $lang->login_username1;
            break;
        case 2:
            $login_username = $lang->login_username2;
            break;
        default:
            $login_username = $lang->login_username;
            break;
    }
}
```

I have tried creating a new variable for `build_profile_link(htmlspecialchars_uni($mybb->user['username'])` and `$lastvisit`, but I can't manage to include them in Twig. Also, I have no idea how to handle switch blocks and/or convert them to Twig. @euantorano 

This PR also fixes an issue in the preliminary conversion of the header template in which the whole guest/member block was included into `div.upper`, causing `div#container` to close earlier than expected and messing up everything on the rest of the page.

### DIFF
--- a/global.php
+++ b/global.php
@@ -9,9 +9,8 @@
  */
 
 $working_dir = dirname(__FILE__);
-if(!$working_dir)
-{
-	$working_dir = '.';
+if (!$working_dir) {
+    $working_dir = '.';
 }
 
 // Load main MyBB core file which begins all of the magic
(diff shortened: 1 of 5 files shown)


euantorano:
To get the variables into the views, the easiest way is to add a Twig global. There are a few ways to do this:

- Create the variables inside a Twig extension. We've been discussing having a new `CoreExtension` that provides various functions like `my_date`, `my_number_format`, etc. so it might make sense to build the global variables used in the header in there, then register them like we do in the `LangExtension`: https://github.com/mybb/mybb/blob/develop/1.9/inc/src/Twig/Extensions/LangExtension.php#L39
- Register the global variable inside `global.php`, using something like:

```php
$twig = \MyBB\app('Twig_Environment');

$twig->addGlobal('lastVisit', $lastvisit);
``` 

For that switch, the closest thing is to just use `if`:

```twig
{% if mybb.settings.username_method == 1 %}
    {{ lang.login_username1 }}
{% elseif mybb.settings.username_method == 2 %}
    {{ lang.login_username2 }}
{% else %}
  {{ lang.login_username }}
{% endif %}
```

An alternative would be to use the `attribute()` function, something like:

```twig
{% if mybb.settings.username_method is in [1, 2] %}
    {{ attribute(lang, 'login_username' ~ mybb.settings.username_method) }}
{% else %}
  {{ lang.login_username }}
{% endif %}
```

Though the first approach is short enough and easy enough to read in my opinion.

Shade-:
Rebased. For some reason @PaulBender and @WildcardSearch's PRs were included too (they are the ones conflicting), I'll try to get rid of them from others of mine.

euantorano:
Any thoughts on my above comment @Shade-?

Shade-:
I have already implemented the switch as you suggested. Is there anything else I'm missing?

euantorano:
I must have missed that, sorry! This looks good to me then.

PR #3047, opened 2018-02-28
Fixes #3040 - Online Today Pagination

### LINKED ISSUE #3040: Online Today Pagination
If you have more than 1000 users online today the function my_number_format() in online.php changes the $todaycount with a comma. Then it gets passed to the multipage function and that breaks it.  

I suggest you fix this by ensuring any multipage variables have $mybb->settings['thousandssep'] and $mybb->settings['decpoint'] removed but short term fix is to correct the online.php code.  I just added a $todaycounted variable that is now passed to multipage without the comma.

This is of course a Big Board only problem as you would need 1000+ members to have been online that day.

Eldenroot:
Fixes #3040 - Online Today Pagination

### DIFF
--- a/online.php
+++ b/online.php
@@ -90,6 +90,8 @@ if($mybb->get_input('action') == "today")
 		}
 	}
 
+	$multipage = multipage($todaycount, $perpage, $page, "online.php?action=today");
+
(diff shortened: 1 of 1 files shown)


Shade-:
Works, thank you.

PR #3053, opened 2018-03-04
Working on #2972 Template Conversion to Twig Format (video)

Shade-:
- [x]  video_dailymotion_embed
- [x]  video_facebook_embed
- [x]  video_liveleak_embed
- [x]  video_metacafe_embed
- [x]  video_myspacetv_embed
- [x]  video_twitch_embed
- [x]  video_veoh_embed
- [x]  video_vimeo_embed
- [x]  video_yahoo_embed
- [x]  video_youtube_embed

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1458,8 +1458,11 @@ class postParser
 
 		$id = htmlspecialchars_uni($id);
 
-		eval("\$video_code = \"".$templates->get("video_{$video}_embed")."\";");
-		return $video_code;
+		return \MyBB\template('parser/video.twig', [
+			'id' => $id,
+			'type' => $video,
+			'local' => $local
+		]);
 	}
 
 	/**
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1456,8 +1456,6 @@ class postParser
 			return "[video={$video}]{$url}[/video]";
 		}
 
-		$id = htmlspecialchars_uni($id);
-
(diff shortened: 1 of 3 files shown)


euantorano:
We've got more than a few conflicts here due to the rebase @Shade-. @PaulBender found it easier to simply recreate his PRs, so that might be the way to go.

inc/views/base/parser/video.twig
```diff
@@ -0,0 +1,21 @@
+{% if type == 'dailymotion' %}
```
euantorano: I wonder if it's worth splitting these out into partial templates? This seems a bit messy to me personally.

inc/class_parser.php
```diff
@@ -1458,8 +1458,11 @@ function mycode_parse_video($video, $url)
 
 		$id = htmlspecialchars_uni($id);
```
euantorano: This will lead to double escaping as the `id` is also escaped in the template.

Shade-: Thank you for spotting this, will get rid of it.

inc/views/base/parser/video.twig
Shade-: Having 10+ templates for such a repetitive thing doesn't look clean personally, but I can push a PR if you want :)

euantorano: We can leave it as is for now, and if people provide feedback against it, we can always split them out easily at a later date - with twig we don't need PHP changes anymore afterall 😉

PR #3067, opened 2018-03-17
Fixed embedded videos not working after #2968

### LINKED ISSUE #3066: Embedded videos no longer play after upgrading to 1.8.15
After upgrading our forum software from MyBB 1.8.14 to 1.8.15, a couple of members noticed that embedded YouTube videos were no longer playable, and displayed an error message when the play arrow was clicked.
 
Digging around a bit, I was able to resolve this problem by moving lines 500 to 509 of inc/class_parser.php to just prior to the call to &#36;this-&gt;mycode_auto_url on line 453 of the same file. I am not very familiar with the MyBB code, so this may have unintended consequences, but so far everything seems to be OK. Have tested on a vanilla install and confirmed the same problem with same solution.

Emphasising my lack of familiarity with the code, here's (kind of roughly/vaguely and potentially imprecisely) why I think this solution works: the code to convert the video BBCode into an embedded video - the moved lines - expects a bare URL, whereas the code to auto-generate a link out of a bare URL - &#36;this-&gt;mycode_auto_url - converts bare URLs to links. Shifting the former code so that it occurs prior to the latter code makes sure a bare URL is passed as expected.

Original thread: [Embedded videos no longer play after upgrading to 1.8.15](https://community.mybb.com/thread-216380.html)

Shade-:
This PR fixes #3066 by parsing videos before URLs. I have proof-checked the whole parser and to me the only two MyCodes affected by #2968 are images and videos; all the others should not require any URL to be parsed.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -450,6 +450,16 @@ class postParser
 			$message = preg_replace_callback("#\[img=([1-9][0-9]*)x([1-9][0-9]*) align=(left|right)\](\r\n?|\n?)(https?://([^<>\"']+?))\[/img\]#is", array($this, 'mycode_parse_img_disabled_callback4'), $message);
 		}
 
+		// Convert videos when allow.
+		if(!empty($this->options['allow_videocode']))
+		{
+			$message = preg_replace_callback("#\[video=(.*?)\](.*?)\[/video\]#i", array($this, 'mycode_parse_video_callback'), $message);
+		}
+		else
+		{
+			$message = preg_replace_callback("#\[video=(.*?)\](.*?)\[/video\]#i", array($this, 'mycode_parse_video_disabled_callback'), $message);
+		}
+
 		$message = $this->mycode_auto_url($message);
 
 		$message = str_replace('$', '&#36;', $message);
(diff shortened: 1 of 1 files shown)


lairdshaw:
Just a possibly irrelevant thought here: what about custom MyCodes? Should they be considered at all here?

Right now, if I define a custom MyCode as:

`\[test\=([^\]]+)\](.*?)\[/test\]`

as replaced by:

`<a href="$1" class="some-magic-class">$2</href>`

Then the post:

`[test=http://example.com/]How can I stop this URL from being replaced with a link? http://example.com/[/test]`

is rendered as:

`<a href="http://example.com/" class="some-magic-class">How can I stop this URL from being replaced with a link? <a href="http://example.com/" target="_blank" rel="noopener" class="mycode_url">http://example.com/</a></href>`

which is invalid HTML.

Shade-:
You're right. I will look into a workaround to run custom MyCodes before the auto_url() function, or a fix to the function itself.

lairdshaw:
I'm not sure the order matters: the auto_url() function is going to replace the bare URL with a link either way, i.e., reordering wouldn't have prevented the above (edited) http://example2.com from being converted into a link. Fixing the function might be the way to go, but it seems tricky: ideally, it would be possible to specify which parts of a MyCode-regex replacement were subject to auto-linkifying, but (and I say this as a newbie to MyBB, so I could be wrong) it seems as though this would take a fair bit of reworking of the parsing code and even the custom MyCodes interface. Will be interested in any case to see what you come up with!

Shade-:
I'd go with preventing the auto_url() function to run in custom MyCodes if an appropriate setting (to be added to the MyCode panel) is enabled. Anyway, I'd wait for others to chime in. @mybb/sqa @mybb/developers

euantorano:
I'll have a look at this later on today, I don't like how fragile the parser seems to be and definitely want to add unit tests for 1.9.

euantorano:
This fixes the video issue for me. In the future we need to look at cleaning up and improving the parser - it may be best to leave that to 1.9 and see if we can make use of the work I started for 2.0 in some way.

PR #3078, opened 2018-03-24
Add basic unit test setup

euantorano:
This configures PHPUnit and PHPSpec, along with PHPCS for the `inc/src` folder only. It also adds Travis CI.

There is only one example test provided at the moment, but the plan is to add unit tests for new classes and functions as time goes on.


euantorano:
Also included is some code reformatting using `phpcbf` to ensure the PSR-2 code style is met.

euantorano:
Looks like there's an issue with the `lowest` build variant, I'll need to try and work out what the problem is - we may use PHPUnit 6 instead of 7 to resolve it...

PR #3086, opened 2018-03-26
mycode in not removed in email notifications + fix for removing neste…

Eldenroot:
…d mycodes

mycode in not removed in email notifications +fix for removing nested mycodes

Author: @azalea4va

https://community.mybb.com/thread-216388.html

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1708,7 +1708,8 @@ class postParser
 		$message = preg_replace_callback("#\[code\](.*?)\[/code\](\r\n?|\n?)#is", array($this, 'mycode_parse_code_callback'), $message);
 
 		$find = array(
-			"#\[(email|color|size|font|align|)=[^]]*\](.*?)\[/\\1\]#is",
+			"#\[(b|u|i|s|url|email|color|img)\](.*?)\[/\\1\]#is",
+			"#\[(email|color|size|font|align|video|)=[^]]*\](.*?)\[/\\1\]#is",
 			"#\[img=([1-9][0-9]*)x([1-9][0-9]*)\](\r\n?|\n?)(https?://([^<>\"']+?))\[/img\]#is",
 			"#\[url=((?!javascript)[a-z]+?://)([^\r\n\"<]+?)\](.+?)\[/url\]#si",
 			"#\[url=((?!javascript:)[^\r\n\"<&\(\)]+?)\](.+?)\[/url\]#si",
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1726,10 +1726,10 @@ class postParser
 		$messageBefore = "";
 		// The counter limit for this "for" loop is for defensive programming purpose only. It protects against infinite repetition. 
 		for($cnt = 1; $cnt < 5 && $message != $messageBefore; $cnt++)
-			{
-				$messageBefore = $message;
-				$message = preg_replace($find, $replace, $messageBefore);
-			}
+		{
+			$messageBefore = $message;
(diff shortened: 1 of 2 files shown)


inc/datahandlers/post.php
```diff
@@ -1607,11 +1607,15 @@ function insert_thread()
 
 				// Queue up any forum subscription notices to users who are subscribed to this forum.
 				$excerpt = my_substr($thread['message'], 0, $mybb->settings['subscribeexcerpt']).$lang->emailbit_viewthread;
```
JordanMussi: Isn't this line now useless as it is done below?

JordanMussi: Isn't this line now useless as it is done below?

Eldenroot: OK, you are right. Everything else should be fine.

Eldenroot:
@euantorano - please can you review this PR? Thank you!

Shade-:
The whole piece of code does not adhere to our coding standards for 1.8. Please edit your PR with proper styling.

Eldenroot:
@Shade- fixed, should be fine :)

inc/class_parser.php
```diff
@@ -1709,18 +1709,27 @@ function text_parse_message($message, $options=array())
...
 			"$2",
 			"$4",
 			"$3 ($1$2)",
 			"$2 ($1)",
 		);
-		$message = preg_replace($find, $replace, $message);
+		
+		$messageBefore = "";
+		// The counter limit for this "for" loop is for defensive programming purpose only. It protects against infinite repetition. 
+		for($cnt = 1; $cnt < 5 && $message != $messageBefore; $cnt++)
```
Shade-: If I understand this correctly, the loop has been introduced to remove nested MyCodes. However by limiting it to 5 repetitions, if you have 6+ nested MyCodes there will still be some remaining. Therefore, I do not feel this is the correct way to do it.

inc/class_parser.php
```diff
@@ -1709,18 +1709,27 @@ function text_parse_message($message, $options=array())
 
 		$find = array(
 			"#\[(b|u|i|s|url|email|color|img)\](.*?)\[/\\1\]#is",
+			"#\[(email|color|size|font|align|video|)=[^]]*\](.*?)\[/\\1\]#is",
```
Shade-: `(email|color|size|font|align|video|)` there's an extra `|` at the end of this regex which must be removed.

inc/class_parser.php
Eldenroot: Yes, you are right, so what do you suggest? The easiest way is to increase the number, but I am not sure if you will like it.

inc/class_parser.php
Eldenroot: Yes, you are right, I missed that, I will fix that ASAP, thank you for reviewing ;)

inc/class_parser.php
euantorano: I'm also not sure what the best approach would be, personally. This will only do 4 repetitions though (the loop starts at 1, and goes to less than 5 - `1, 2, 3, 4`. 4 repititions seems like an odd limitation.

euantorano:
Great, thanks. I'll have to give this a test this weekend.

effone:
@Eldenroot @Shade- Whats the status of it? Fixed? Not fixed?

Shade-:
If it was for me this could be merged, but I'd wait @euantorano for a final decision.

Sama34:
I think this is related to #3144

effone:
'nuff tested.

PR #3093, opened 2018-04-02
Fixes #3034 usercp2.php removal

### LINKED ISSUE #3034: Usercp2.php removal
In 1.9 we should remove the usercp2.php and look to merge this into one since the file doesn't do a huge amount and tidy things up a little bit.

effone:
Removal of usercp2
https://github.com/mybb/mybb/issues/3034

Unfortunately messed and mixed up with my other PR and reverted commit. Is it okay or I have to create fresh?

### DIFF
--- a/usercp.php
+++ b/usercp.php
@@ -10,6 +10,7 @@
 
 define("IN_MYBB", 1);
 define('THIS_SCRIPT', 'usercp.php');
+define("ALLOWABLE_PAGE", "removesubscription,removesubscriptions");
 
 $templatelist = "usercp,usercp_nav,usercp_profile,usercp_changename,usercp_password,usercp_subscriptions_thread,forumbit_depth2_forum_lastpost,usercp_forumsubscriptions_forum,postbit_reputation_formatted,usercp_subscriptions_thread_icon";
 $templatelist .= ",usercp_usergroups_memberof_usergroup,usercp_usergroups_memberof,usercp_usergroups_joinable_usergroup,usercp_usergroups_joinable,usercp_usergroups,usercp_nav_attachments,usercp_options_style,usercp_warnings_warning_post";


usercp.php
```diff
@@ -10,6 +10,7 @@
 
 define("IN_MYBB", 1);
 define('THIS_SCRIPT', 'usercp.php');
+define("ALLOWABLE_PAGE", "removesubscription,removesubscriptions");
```
Shade-: This likely blocks every other connection to usercp.php. Please remove it, thank you.

effone: Isn't it for allow users even when they are not allowed to view board? I thought it was there to remove subscription from a mail link or something ....

Anyway, removed.

effone: @Shade- @euantorano or anyone can you guys please confirm this? As much I have cross-checked "ALLOWABLE_PAGE" is not blocking anything... Further cross-checking made me believe again that as I said it allows user to access page while viewing the board is not allowed.

Shade-: Yep, sorry, my fault. You can include that again.

effone: Added back. Thanks.

inc/functions_online.php
```diff
@@ -516,27 +516,21 @@ function fetch_wol_activity($location, $nopermission=false)
 			{
 				$user_activity['activity'] = "usercp_subscriptions";
 			}
-			elseif($parameters['action'] == "notepad" || $parameters['action'] == "do_notepad")
-			{
-				$user_activity['activity'] = "usercp_notepad";
-			}
-			else
+			elseif($parameters['action'] == "addfavorite" || $parameters['action'] == "removefavorite" || $parameters['action'] == "removefavorites")
```
Sama34: Do we have favorites in mybb ?

PaulBender: They were removed back in 1.4.

euantorano: Might as well leave it there if it's currently in `usercp2.php` - this should just be a direct port. We can remove dead code like that at a later date.

PR #3108, opened 2018-04-09
Fixes #3088 trim input

### LINKED ISSUE #3088: A space character flagging an email as invalid
I have just started a Mybb site and with the 50 or so users who have attempted to register, I have run into this issue twice that I know of.  I do classify it as a bug although it may be a design bug and not a coding bug.  Either way it should be addressed.  It is confusing for users and took hours for me to figure out what was going on.

If one enters an email address of the form "xyz@abc.com ", what one sees on the screen is just 
xyz@abc.com".  if in response to typing that, one gets a message that the email address was invalid, the user sees what appears to be a perfectly valid email address as being rejected.  Either the code needs to be corrected to ignore leading/trailing white-space (which is the norm for this type of thing) or the less user-friendly but minimally acceptable solution of having a more descriptive error message that at least echoes the email address that was rejected (eg, "The address you entered, 'xyz@abc.com ', is not a valid email address." or even better "The address you entered, 'xyz@abc.com ', is not a valid email address. Email addresses may not contain the character ' '.")

Original thread: [A space character flagging an email as invalid](https://community.mybb.com/thread-216476.html)

effone:
Couldn't reproduce this. Javascript validation blocked me. However modifying blindly as per discussion ....
https://github.com/mybb/mybb/issues/3088

### DIFF
--- a/inc/class_core.php
+++ b/inc/class_core.php
@@ -454,7 +454,8 @@ class MyBB {
 				{
 					return array();
 				}
-				foreach ($this->input[$name] as $ikey => $ivalue) {
+				foreach($this->input[$name] as $ikey => $ivalue)
+				{
 					$this->input[$name][$ikey] = trim($ivalue);
 				}
 				return $this->input[$name];


Shade-:
I'm not against this, but if we were going this way, I think all trimmed `get_input` calls should be removed accordingly. This is a grep against the current codebase:

```
./xmlhttp.php:	$mybb->input['query'] = ltrim($mybb->get_input('query'));
./xmlhttp.php:	$password = trim($mybb->get_input('password'));
./inc/plugins/hello.php:	$message = trim($mybb->get_input('message'));
./admin/modules/style/fastyle.php:		if (!trim($mybb->get_input('title'))) {
./search.php:$mybb->input['keywords'] = trim($mybb->get_input('keywords'));
./reputation.php:	$mybb->input['comments'] = trim($mybb->get_input('comments')); // Trim whitespace to check for length
./warnings.php:	if(!trim($mybb->get_input('reason')))
./usercp.php:		$mybb->input['avatarurl'] = trim($mybb->get_input('avatarurl'));
./memberlist.php:	$search_username = htmlspecialchars_uni(trim($mybb->get_input('username')));
./memberlist.php:	$mybb->input['website'] = trim($mybb->get_input('website'));
./memberlist.php:		$mybb->input[$cfield] = trim($mybb->get_input($cfield));
./report.php:				$comment = trim($mybb->get_input('comment'));
```

euantorano:
I also have some other improvements I’d like for get_input() too, such as handling array parameters correctly (eg: if the input type is string, but the post bar is actually an array, then use the first element of the array). It might be worth making both changes at the same time.

Sama34:
But would rimming input to be a good idea ?

effone:
@Shade
I can find and correct those get_inputs, if required, but double trimming doesn't harm anything, assures on the contrary ...

So, what to do now? Any decision finalized yet or I just leave the PR as is?

Shade-:
It's just a waste of resources and should be removed. Anyway, @Sama34 and @dvz expressed concerns about trimming all inputs so I'd wait for a final decision.

euantorano:
Any more thoughts on this @dvz given discussion elsewhere ([https://github.com/mybb/mybb/issues/3088#issuecomment-381252030](https://github.com/mybb/mybb/issues/3088#issuecomment-381252030))?

dvz:
IMO fields shouldn't be trimmed if there is no clear reason.

effone:
I have made this PR, but honestly I will go with @dvz. We are still not sure we need it, so as per [YAGNI ](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) we should leave it as is. it should not be implemented by only a single user report as I have failed to reproduce it (first post) and even it is a problem the user will not be able to reach upto that stage. I don't know How OP bypassed JS validation.

Ben-MyBB:
So can this be closed? @euantorano @dvz 
regardless, thanks again for another PR @effone

effone:
Yep, this should be closed without merge @Ben-MyBB

PR #3115, opened 2018-04-14
Fixes #3109 – Maxloginattempts lockout

### LINKED ISSUE #3109: maxloginattempts Settings
This is a serious problem.

That setting is there to prevent bruteforcing into accounts.  In ACP under the Login settings there is parameters for protecting against malicious login attempts. In the mybb_users table a column exists for loginattempts.  

For the past few years I have seen occasional bruteforce attempts into accounts and my logs show an error that the loginattempt column was full because I have my type as tinyint(2) which is max limit 127.  So that number is reached and I get a mysql error logged.  I figured MyBB just did a poor job of checking the limit wasn't reached before the login ran.  

Looking into inc/functions.php I can see the login_attempt_check() function. In there you can see it uses the cookie instead of the column. 

But this is the important part....the setting "maxloginattempts" is NOWHERE TO BE FOUND IN MyBB. This is a bruteforce prevention setting and it has no function for members.  It is used in the ACP though.  Yet for members it is not.

I'm not even sure how to consider this bug.  Do you just remove the setting?  Do you make it function properly inside login_attempt_check()?  Or maybe just fix the query that gives the error.

PHP Code:&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;update_array&nbsp;=&nbsp;array(&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;'loginattempts'&nbsp;=&gt;&nbsp;1&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;);&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;db-&gt;update_query("users",&nbsp;&#36;update_array,&nbsp;"uid&nbsp;=&nbsp;'{&#36;mybb-&gt;user['uid']}'");&nbsp;

I'm aware that if it does a check to the database first and uses the setting then locks out accounts you could potential bruteforce someone to prevent them from logging in.  So there is that problem. But I do believe that is the intent of the setting we all assumed worked in that way.  It does offer a disable "0" option in the description.

I do hope this bug gets labelled important and is addressed quickly.

I was stunned by this.  I wonder if this setting ever had any code written in any version of MyBB?  I been getting this error since 1.6x.

Original thread: [maxloginattempts Settings](https://community.mybb.com/thread-216874.html)

Shade-:
This PR tries to solve #3109 by adapting the frontend code to the backend one, in which login attempts are saved in the database and a hard lockout is issued after a customizable number of failed attempts.

By doing so, a new column has been added (named `loginlockoutexpiry`, identical to the backend counterpart) to store lockout expiry dates. As an `uid` is not available before validating credentials, a soft lockout is set alongside the database one by setting a cookie with the expiry date, similarly to the past. This soft lockout has been extended to display in the modal too by adding 2 new templates (`header_welcomeblock_guest_login_modal` for the usual login modal and `header_welcomeblock_guest_login_modal_lockout` for the locked out one).

It worked fine in my tests, but as it vastly changes how the lockout system is designed, it might be worth a lot of independent tests as well.


Eldenroot:
Works fine for me, but would be nice to have someone else who can confirm to be 100 % sure.

euantorano:
@Shade- This is showing as being conflicted, any chance you can take a look?

Shade-:
Sure, will do

Shade-:
Done

inc/functions.php
```diff
@@ -6358,92 +6357,100 @@ function get_inactive_forums()
  * @param bool $fatal (Optional) Stop execution if it finds an error with the login. Default is True
  * @return bool|int Number of logins when success, false if failed.
  */
-function login_attempt_check($fatal = true)
+function login_attempt_check($uid = 0, $fatal = true)
```
euantorano: I doubt many plugins use this function, but we should make a note that the function signature has changed in the release notes.

euantorano:
This seems to work fine for me, but since it's changing a fairly critical code path, it could do with some extra testing on the test board @mybb/sqa

Shade-:
Agreed. Got an exam in some days, and will have some days off to perform some testing afterwards.

PR #3110, opened 2018-04-11
Fixes #1790 Collapse state alt / title fix

effone:
https://github.com/mybb/mybb/issues/1790

PR #3116, opened 2018-04-14
Fixes #1790 Collapse state alt / title fix - Extended - Template versions update

### LINKED ISSUE #1790: Expcolimage in Board Statistic
Basically, this is not important but slightly disturbing. In this image --&gt; http://prntscr.com/5xllhq if I click [-], Board statistic will be slide up. It's work fine.Expcolimage will show [+]. 
And when I refresh the page with board statistic still slide up, it's work fine too. 
The problem is when my cursor pointing to expcolimage after refresh (still slide up), it will show [-] should be to [+]. 
It's happened in this forum too.

Original thread: [Expcolimage in Board Statistic](http://community.mybb.com/thread-166219.html)

effone:
As per the changes committed in this PR:
https://github.com/mybb/mybb/pull/3110
All affected template's version updated.


effone:
@Shade- 
Comments in PR https://github.com/mybb/mybb/pull/3110 incorporation

PR #3117, opened 2018-04-14
Fixes #3113 – Added default language to set_language()

### LINKED ISSUE #3113: Default Language conflitc with Group Join Request
I found the following problem:
My Default board language is Portuguese - Brazil, but if a user is using the Default language in your User CP and he is the leader of a group, when I send a join request he receives the email in the English language. It only resolves when the user directly selects Portuguese - Brazil as their language.

Note: With the other emails working normally, this problem only happens when it is a Join Request.

Shade-:
This PR fixes #3113 by allowing `set_language` to use the board's default `bblanguage`.

### DIFF
--- a/inc/class_language.php
+++ b/inc/class_language.php
@@ -74,16 +74,16 @@ class MyLanguage
 	 * @param string $language The language to use.
 	 * @param string $area The area to set the language for.
 	 */
-	function set_language($language="english", $area="user")
+	function set_language($language="", $area="user")
 	{
 		global $mybb;
 
 		$language = preg_replace("#[^a-z0-9\-_]#i", "", $language);
 
-		// Default language is English.
+		// Use the board's default language
 		if($language == "")
 		{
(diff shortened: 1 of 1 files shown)


Walter-Correa:
Your changes have resolved the problem, because when a user configures their language to Default, in the database, the "language" field is blank (null).
The only problem that is happening now is the moment of the confirmation of sending the request to join the group:
The message box that says the message was sent correctly, below appears: "Click here if you do not want to wait any longer.". If the leader language is the same as mine, the phrase: "Click here if you do not want to wait any longer." appears in the correct language, but if the language of the leader is English and my Portuguese, the phrase appears in English.

euantorano:
@Shade- Do you want to look at the issue reported by @Walter-Correa in this same PR or a separate PR?

Shade-:
I have looked into it, and could not isolate @Walter-Correa's issue.

PR #3119, opened 2018-04-14
Fixes #3112 – Added hard limit on awayreason

### LINKED ISSUE #3112: Awayreason and MySQL errors
MyBB doesn't check that awayreason isn't more than the 200 character limit as the column in users table is varchar(200). This creates an error.log entry for that query and displays it on the site if error reporting is on.

I was able to replicate at MyBB.com as well.  

"1406 - Data too long for column 'awayreason' at row 1"

Should be an easy fix in the datahandlers/user.php where it's cut-off with my_substr() at 200 characters.

Shade-:
This PR fixes #3112 by adding a hard limit to `awayreason` field, and a soft limit to the user inputs.

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -378,7 +378,7 @@ class UserDataHandler extends DataHandler
 			$this->set_error("invalid_birthday");
 			return false;
 		}
-		else if($birthday['year'] == date("Y"))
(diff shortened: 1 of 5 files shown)


Eldenroot:
Works fine, thank you!

euantorano:
+1 to putting this in the user data handler.

Shade-:
Done :)

PR #3120, opened 2018-04-14
Fixes #2067 Attachment Usage Information Improvement

### LINKED ISSUE #2067: improve function get_friendly_size
I noticed that when a user with no stored attachments opens a new thread or a new reply gets the following at the bottom of his screen:

[attachment=33675]

I have located the source of N/A in this text, and why it is not 0. It appears that the query and statement below that at line 976 of newthread.php and line 1180 of newreply.php:

PHP Code:
&#36;query&nbsp;=&nbsp;&#36;db-&gt;simple_select("attachments",&nbsp;"SUM(filesize)&nbsp;AS&nbsp;ausage",&nbsp;"uid='".&#36;mybb-&gt;user['uid']."'");&#36;usage&nbsp;=&nbsp;&#36;db-&gt;fetch_array(&#36;query);&nbsp;

results in a variable &#36;usage['ausage'] equal to NULL. This variable is passed to the function get_friendly_size at line 991 of newthread.php and line 1197 of newreply.php.

PHP Code:
&#36;friendlyusage&nbsp;=&nbsp;get_friendly_size(&#36;usage['ausage']);&nbsp;

I have checked it, and the problem can be solved by changing this statement in both modules to:

PHP Code:
&#36;friendlyusage&nbsp;=&nbsp;get_friendly_size((int)&#36;usage['ausage']);&nbsp;

by which we get at the bottom of the newthread and newreply pages:

[attachment=33676]

But I think it is more a matter of principle to look more closely at the function get_friendly_size in /inc/functions.php. This function checks at line 3654 whether the passed argument is numeric:

PHP Code:
if(!is_numeric(&#36;size)){&nbsp; &nbsp; &nbsp; &nbsp;return&nbsp;&#36;lang-&gt;na;}&nbsp;

where &#36;size is the passed argument. Apparently a NULL variable is not considered to be a numerical value or a numerical string, and function is_numeric returns false for an argument with value NULL. I think this is wrong in principle, because PHP considers NULL equal to zero when compared with an integer. That's why my proposal is to change line 3654 of /inc/functions.php into:

PHP Code:
if(!is_numeric(&#36;size)&nbsp;&amp;&amp;&nbsp;!is_null(&#36;size))&nbsp;

I tried it and this yields the same result as changing newthread.php and newreply.php.

I realize that this is a very very small issue, so when the opinion is that this kind of things is not worth looking at, please let me know. I still have a few :D .

EDIT
I know that comparing NULL with negative integers goes wrong, but that is not relavant function get_friendly_size.

Original thread: [improve function get_friendly_size](http://community.mybb.com/thread-166517.html)

effone:
https://github.com/mybb/mybb/issues/2067

Didn't modify function `get_friendly_size( )` as it may be designed this way to use at somewhere else...

Changes:

- Modified newpost & newreply not to display attachment usage (only display quota) when no attachment quota is used as discussed [here](https://community.mybb.com/thread-166517-post-1136969.html#pid1136969).
- User CP Attachment usage enhanced.


install/resources/mybb_theme.xml
```diff
@@ -12035,7 +12036,7 @@ if(use_xmlhttprequest == "1")
 					</tr>
 					<tr>
 						<td class="trow2" width="40%"><strong>{$lang->attachstats_spaceused}</strong></td>
-						<td class="trow2" width="60%">{$friendlyusage} ({$percent})</td>
+						<td class="trow2" width="60%">{$friendlyusage}</td>
```
Shade-: If removed completely, you can remove `attachments_usage_percent` too, because then there's only one reference to it (`attachments_usage_quota`) ☺️

effone: Modified `attachments_usage_percent` to hold required space and brackets.
Hard coded space removed from usercp.php
Done.

inc/languages/english/usercp.lang.php
```diff
@@ -291,7 +291,8 @@
 $l['date_uploaded'] ="Posted";
 $l['no_attachments_selected'] = "You did not select any attachments to delete.";
 $l['attachments_deleted'] = "The selected attachments have been successfully deleted.<br />You will be now redirected to the attachments manager.";
-$l['attachments_usage_quota'] = "- Using {1} of {2} ({3}) in {4} Attachments";
+$l['attachments_usage_quota'] = "- Using {1} of {2} in {3} Attachments";
+$l['attachments_usage_percent'] = "( {1}%)";
```
Sama34: Is that space intentional ?

effone: Yep. While the % is being appended to the usage size ...
Just moved it out of usercp.php, as shade instructed to avoid hardcoding and placed in language.

PR #3121, opened 2018-04-15
Fixes #2910 StopForumSpam Confidence level correction

### LINKED ISSUE #2910: StopForumSpam confidence levels greater than 100%
Overall confidence when checking a user against StopForumSpam can end up being greater than 100%, as it can be a value created by adding three confidence values together. The setting only allows values between 0-100. The solution is to divide the total confidence by the numebr of checks made to get an average confidence.

Originally reported here: https://community.mybb.com/thread-214186

effone:
https://github.com/mybb/mybb/issues/2910


euantorano:
This looks good, but I'm going to open another issue now 😛

PR #3106, opened 2018-04-08
Fixes #3103 - mycode_auto_url bug

lairdshaw:
Fixes #3103 - mycode_auto_url bug

Issue #3103, opened 2018-04-08
mycode_auto_url bug


euantorano:
I mentioned this problem in a comment on GitHub pull request #2968, but here's a proper bug report.

This: 

http://example1.com/ http://example2.com/ http://example3.com/

is the result of the MyCode:

Code:[url]http://example1.com/[/url] http://example2.com/ [url]http://example3.com/[/url]

We would expect the URL for example2 to have been converted into a link, but it hasn't been (at least, not on the forum software as it is at time of posting).

This can be fixed by replacing each of the occurrences of

Code:.*
in each of the regexes in mycode_auto_url() in inc/class_parser.php with:

Code:[^&#92;]]*

Original thread: [mycode_auto_url bug](https://community.mybb.com/thread-216811.html)

effone:
Is this fixed?

lairdshaw:
Yep, was fixed by  #3127. Can be closed.

PR #3127, opened 2018-04-17
Moves automatic URL replacement to the end of parsing.

lairdshaw:
Aims to fix the issues raised in the comments of #2968.
Supersedes #3103.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1513,8 +1513,9 @@ class postParser
 		$message = " ".$message;
 
 		// Links should end with slashes, numbers, characters and braces but not with dots, commas or question marks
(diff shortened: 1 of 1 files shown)


euantorano:
This definitely seems like the better approach IMO. I assume we can close #3106 after I test and merge this?

I'm going to write some basic unit tests for the parser across today and tomorrow, though they'll only be in the 1.9 branch for now.

labrocca:
Fix #3127 fixed my custom MyCode and appears to be working well.

lairdshaw:
What's the protocol for changing PRs like this? I have a few amendments which also fix the problem I mentioned in #3067 - that of links potentially being replaced within existing links due to custom MyCodes. Should I file a separate PR noting that this one is superseded?

lairdshaw:
Nvm. Figured it out. But feel free to edit the PR if you don't think the second commit belongs here.

Sama34:
I'd let it be if it works. It would be perfect if you can share examples of what your PR fix so we can test it propertly. Currently there seems to be many discussion taking place in different PRs/issues. Thanks.

euantorano:
Been doing some testing on this, and in my (very basic) tests it seems to be working exactly as it should - thanks a bunch @lairdshaw!

Sama34:
The following also got fixed but I don't think it has an issue assigned to it.
https://community.mybb.com/thread-216948.html

euantorano:
That’s a nice coincidence :)

PR #3130, opened 2018-04-17
Feature: parser unit tests

euantorano:
This is a work in progress test suite for the `\postParser`. It also adds a trait for tests that make use of the 1.x MyBB core to initialise various global variables expected by MyBB.

Currently only the following MyCodes are tested standalone:

- `[b]`
- `[u]`
- `[i]`
- `[s]`

I will be expanding this PR over time to cover every MyCode on its own, and then nested MyCodes (eg: bold inside a quote, quotes within quotes, etc.). Hopefully this will help us catch any remaining issues in the parser.


euantorano:
This is now based on the current `feature/1.9` branch and includes some changes to parser templates due to the existing templates being incorrect based upon the existing tests I've implemented so far.

I'm going to merge this and work on more unit tests in separate PRs as I'd like to get the basic testing and CI infrastructure in place now.

PR #3134, opened 2018-04-18
Fixes #3027 This is bad parser :D

### LINKED ISSUE #3027: Code Parser
Code:[code]This is bad parser.[/code]

Above is example of a bug.  Posting it here so you can see it in action. Staff can edit my post to view the mycode used.

btw, this wasn't a bug in 1.6x.

Original thread: [Code Parser](https://community.mybb.com/thread-215901.html)

effone:
Attempt to fix https://github.com/mybb/mybb/issues/3027
Capture greedy ...


Shade-:
Works fine. Good work!

PR #3137, opened 2018-04-19
Resolves #2986 reCAPTCHA v.1 drop

### LINKED ISSUE #2986: ReCaptcha v1 support being dropped
Google is dropping support for Recaptcha v1 so we should likely remove this for .16

effone:
Attempt to comply: https://github.com/mybb/mybb/issues/2986

To Do:
- [x] Remove reCAPTCHA settings
- [x] Remove reCAPTCHA templates
- [x] Remove all template references in scripts
- [x] Remove all checks and processes related to reCAPTCHA generation / refresh
- [x] Remove abandoned language variables from `global.lang.php`
- [x] Alter `optionscode` for `captchaimage` in existing `settings` table to remove `reCAPTCHA` _(upgrade)_
- [x] Check `value` of `captchaimage` in existing `settings` table and if it is set to **2** _(upgrade)_
                    - reset value to **1**
                    - remove values of `captchaprivatekey` & `captchapublickey`


inc/class_captcha.php
```diff
@@ -104,11 +104,7 @@ function __construct($build = false, $template = "")
 		{
 			$this->captcha_template = $template;
 
-			if($this->type == 2)
-			{
-				$this->captcha_template .= "_recaptcha";
-			}
-			elseif($this->type == 4){
+			if($this->type == 4){
```
euantorano: Code style - can you move the `{` to a new line?

install/resources/upgrade44.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
euantorano: We already have an upgrade script for 1.8.16: https://github.com/mybb/mybb/blob/feature/install/resources/upgrade43.php - the naming is super confusing, I know 😞

effone: Was not familiar with upgrade script system. Sorry. Will merge to that.

inc/class_captcha.php
effone: Done.

install/resources/upgrade44.php
effone: @euantorano 44 unlinked. code transferred to 43

effone:
@euantorano does it worth digging as Google clearly stated to "Please **register a new key** and upgrade to **v2.**". Old keys will not work thats because I have removed those from db.

Source: https://developers.google.com/recaptcha/docs/faq

euantorano:
Alright, sounds good to me. Good work!

effone:
@euantorano I have cleared the conflict, however `post_captcha_recaptcha,` has to be removed from `newreply.php` and `newthread.php` top `$templatelist`.

Sama34:
Can't you get them removed here @effone ?

effone:
@Sama34 I tried (last two commits) but it has to be done after merge. If I do that now conflict will arrive again.

PR #3143, opened 2018-04-20
Fixes #2198 Optimize thread redirect database structure

### LINKED ISSUE #2198: Optimize thread redirect database structure

PaulBender:
Fixes #2198

I'm pretty sure I found all the moved bits in the code.

Resubmitted due to me needing to refork my repository.


install/resources/upgrade43.php
```diff
@@ -0,0 +1,104 @@
+<?php
+/**
+ * MyBB 1.8
+ * Copyright 2014 MyBB Group, All Rights Reserved
```
Azareal: Nitpick: Copyright header is 2014 and for 1.8

effone: Ample of files are that way in MyBB, I guess it represents what generally declared as `@since` with version number

PaulBender:
Fixed merge conflicts and updated upgrade file (upgrade43 will be 1.8.16's upgrade file).

euantorano:
I need to give this some thorough testing, the thing that’s put me off so far is the upgrade script causing problems when it comes to future versions of 1.8… I’m not sure what the best approach is for that - perhaps now is the time to look at improving the upgrade script?

PaulBender:
I fixed the merge conflicts and updated some things.

RE the upgrade; I'm not sure how to handle that, other than making sure it doesn't conflict with the 1.8 scripts (it shouldn't if we keep the numbering up to date).

PaulBender:
Fixed merge conflicts, and removed the upgrade file (we can worry about that later).

euantorano:
Cheers. I’ll try merge this this week so we don’t break it again.

PR #3146, opened 2018-04-21
Working on #2972 Template Conversion to Twig Format (search.php)

PaulBender:
#2972 for the search system (search.php & functions_search.php)

### DIFF
--- a/inc/views/base/search/results_posts_row.twig
+++ b/inc/views/base/search/results_posts_row.twig
@@ -18,19 +18,11 @@
         <span class="smalltext">
             {{ lang.post_thread }} <a href="{{ post.thread_url|raw }}">{{ post.thread_subject }}</a><br />
             {{ lang.post_subject }} <a href="{{ post.post_url|raw }}#pid{{ post.pid }}">
-                {% if post.subject|length > 50 %}
-                    {{ post.subject|slice(0, 50) }}...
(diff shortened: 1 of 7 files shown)


PaulBender:
This is finished

inc/views/base/search/results_posts_row.twig
```diff
@@ -0,0 +1,48 @@
...
+            {{ lang.post_subject }} <a href="{{ post.post_url|raw }}#pid{{ post.pid }}">
+                {% if post.subject|length > 50 %}
+                    {{ post.subject|slice(0, 50) }}...
+                {% else %}
+                    {{ post.subject }}
+                {% endif %}
+            </a>
+        </span><br />
+        <table width="100%"><tr><td><span class="smalltext"><em>
+            {% if post.message|length > 200 %}
```
euantorano: This logic can be simplified. The documentation from the `slice` filter states:

> If length is given and is positive, then the sequence will have up to that many elements in it. If the variable is shorter than the length, then only the available variable elements will be present. If length is given and is negative then the sequence will stop that many elements from the end of the variable. If it is omitted, then the sequence will have everything from offset up until the end of the variable.

PaulBender: Changed it.

PR #3151, opened 2018-04-23
Working on #2972 Template Conversion to Twig Format (misc.php)

PaulBender:
#2972 for miscellaneous actions (misc.php).


PaulBender:
This is finished.

I removed AIM from the IM pop-up code due to the messenger being discontinued (we should remove the field entirely).

PaulBender:
Shall I remove the entire IM center section then?

Also, I'd like to work on moving the contact fields to profile fields.

euantorano:
Sure, that would make sense to me. Feel free to work on moving them across to profile fields, yeah :)

PR #3157, opened 2018-04-25
Fixes #2764 Admin permission type icon

### LINKED ISSUE #2764: Admin permissions icon may be missing
When you access the admin permissions section on the ACP there is list with all admin. This list displays an icon for each user depending on the inheritance if his permissions.

This icon may be missing when the foreach loop doesn't return a result:
```php
			if($admin['permissions'] != "")
			{
				$perm_type = "user";
			}
			else
			{
				$groups = explode(",", $admin['additionalgroups'].",".$admin['usergroup']);
				foreach($groups as $group)
				{
					if($group == "") continue;
					if($group_permissions[$group] != "")
					{
						$perm_type = "group";
						break;
					}
				}

				if(!$group_permissions)
				{
					$perm_type = "default";
				}
			}
```
It should check whether $perm_type is already set or not instead.

effone:


effone:
Messed. Recreated.

PR #3159, opened 2018-04-26
Added htmlspecialchars_uni()

labrocca:
This sanitizes the user input as it did in MyBB 1.6x.  Custom pages and plugins from 1.6x may not sanitize the output and it can (and was) easily done in the past by the datahandler.  It should be placed back in unless there is a specific reason it was removed.

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -1110,12 +1110,12 @@ class UserDataHandler extends DataHandler
 			"regdate" => (int)$user['regdate'],
 			"lastactive" => (int)$user['lastactive'],
 			"lastvisit" => (int)$user['lastvisit'],
-			"website" => $db->escape_string($user['website']),
+			"website" => $db->escape_string(htmlspecialchars_uni($user['website'])),
 			"icq" => (int)$user['icq'],
-			"aim" => $db->escape_string($user['aim']),
-			"yahoo" => $db->escape_string($user['yahoo']),
-			"skype" => $db->escape_string($user['skype']),
-			"google" => $db->escape_string($user['google']),
+			"aim" => $db->escape_string(htmlspecialchars_uni($user['aim'])),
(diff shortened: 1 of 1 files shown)


euantorano:
Escaping seems to have been removed here: https://github.com/mybb/mybb/commit/26c09b6181547993346f2e1c9669e734704a4c92

Note that due to the escape in `usercp.php` the fields would be double escaped with this patch: https://github.com/mybb/mybb/blob/feature/usercp.php#L395 - this PR should be extended to prevent that from happening.

labrocca:
And what would be the negative effects of double escaping?  Seems rather small compared to the potential for persistent XSS.

euantorano:
Double escaped fields look weird (eg: `&amp;amp;` rather than `&amp;`). When it's such a trivial fix, it makes sense to fix 😃

labrocca:
Looks Weird < XSS Exploit

I asked at MyBB.com forum if this changed in core for 1.8x was ever publicly notified to the user base?  Because while it's obviously not exploited in core code.  Plugins and custom pages could greatly be effected.

euantorano:
I sent a reply on the forums. I’m not sure if it was ever documented, but I can’t be certain.

labrocca:
@Sama34 Then don't commit my changes.  I'm one guy and I have my own questions but if the community doesn't agree with my argument then so be it.  I looked at your plugins. None appear effected by my commit.

And "You are breaking 1.8 features to make outdated plugins usable", is not correct. I'm asking for the change to make 1.6x plugins secure and free from exploit, which imho DOES MAKE SENSE.  If you need to me to go through all the plugins that exist in the MyBB community which have this exploit I'll waste my time and and do it.  Is that what you need?

Sama34:
We understand your suggestion and the reasoning behind it. However, it seems most disagree and I think it would be better to accept and keep the change. We appreciate suggestions and of course, discussion is _open_ to anyone who wishes to bring up their concerns.

PR #3153, opened 2018-04-24
AIM Removal

effone:
As [AIM has been discontinued](https://help.aol.com/articles/aim-discontinued); removed all instances of AIM from MyBB ...

PR #3161, opened 2018-04-27
Fixes #3124 XSS and $mybb->input in Templates

### LINKED ISSUE #3124: XSS and $mybb->input in Templates
I'd like to point out that MyBB templates should never use &#36;mybb-&gt;input directly in the templates. This creates a situation where the data is not sanitized on the page.  This opens up XSS where links to the page could insert malicious code including javascript to execute on the site including things like login popups which may mimic site login popup and then steal the credentials.

XSS is low-risk but still a risk and it's not the best practice to use unsanitized user input anywhere.  But MyBB does this in a number of the default/master templates.  

Most of the time the &#36;mybb-&gt;input is used in case there is an error in form submission and the data remains in the form.  That's convenient for the member but it's a risk as well.  I suggest that MyBB does a template search and see which templates do this in order to decide if they want to sanitize the input or just remove the convenience.  

Please review asap.

Original thread: [XSS and $mybb->input in Templates](https://community.mybb.com/thread-216966.html)

effone:
Attempt to fix #3124 $mybb->input usage drop in template.

To Do:
- [x] Security check added for restricting input usage in template.

Template: member_register_coppa
- [x] 01. `$mybb->input['bday3']` : Was already sanitised, assigned to new variable: `$birthday_year`

Template: modcp_editprofile
- [x] 02. `$mybb->input['birthday_year']` : Was already sanitised, assigned to new variable: `$birthday_year`
- [x] 03. `$mybb->input['website']` : Was already sanitised, assigned to new variable: `$user_website`
- [x] 04. `$mybb->input['icq']` : Was already sanitised, assigned to new variable: $user_icq
- [x] 05. `$mybb->input['aim']` : Meant to be dropped by PR #3153. **SKIPPED**
- [x] 06. `$mybb->input['skype']` : Was already sanitised, assigned to new variable: `$user_skype`
- [x] 07. `$mybb->input['google']` : Was already sanitised, assigned to new variable: `$user_google`
- [x] 08. `$mybb->input['yahoo']` : Was already sanitised, assigned to new variable: `$user_yahoo`
- [x] 09. `$mybb->input['modpost_time']` : Was already sanitised, assigned to new variable: 
 `$modpost_time`
- [x] 10. `$mybb->input['suspost_time'] `: Was already sanitised, assigned to new variable:  `$suspost_time`

Template: modcp_editprofile_signature
- [x] 11. `$mybb->input['action_time'] `: Was already sanitised, assigned to new variable:  `$action_time`

Template: modcp_finduser
- [x] 12. `$mybb->input['username']`  : **Sanitised**. New variable: `$username`

Template: modcp_warninglogs
- [x] 13. `$mybb->input['filter']['username']` : Was already sanitised, assigned to new variable: `$filter_username`
- [x] 14. `$mybb->input['filter']['mod_username']` : Was already sanitised, assigned to new variable: `$filter_modusername`
- [x] 15. `$mybb->input['filter']['reason']` : Was already sanitised, assigned to new variable: `$filter_reason`

Template: moderation_delayedmoderation_merge
- [x] 16 . `$mybb->input['delayedmoderation']['subject']` : Was already sanitised, assigned to new variable: `$delayedmoderation_subject`
- [x] 17 . `$mybb->input['delayedmoderation']['threadurl']` : Was already sanitised, assigned to new variable: `$delayedmoderation_threadurl`

Template: moderation_delayedmoderation_move
- [x] 18. `$mybb->input['redirect_expire']` : Was already sanitised, assigned to new variable: `$redirect_expire`

Template: moderation_confirmation
- [x] 19.  `$mybb->input['action']` : Was already sanitised, assigned to new variable: `$action`

Template: moderation_purgespammer
- [x] 20. `$mybb->input['uid']` : **Sanitised**. Reused already available variable: `$uid`

Template: private_move
- [x] 21. `$mybb->input['fid']` : Was already sanitised, assigned to new variable: `$from_fid`

Template: reputation_add
- [x] 22 - 25. `$mybb->input['pid']` (4 instances) : Was already sanitised, assigned to new variable: `$reputation_pid`

Template: reputation_add_delete
- [x] 26. `$mybb->input['pid']` : **Sanitised**. New variable: `$reputation_pid`

Template: contact
- [x] 27. `$mybb->input['subject']` : Was already sanitised, assigned to new variable: `$contact_subject`
- [x] 28. `$mybb->input['message']` : Was already sanitised, assigned to new variable: `$contact_message`
- [x] 29. `$mybb->input['email']` : Was already sanitised, assigned to new variable: `$user_email`

### DIFF
--- a/admin/inc/functions.php
+++ b/admin/inc/functions.php
@@ -666,12 +666,6 @@ function check_template($template)
 		return true;
 	}
 
-	// Block usage of input fields : XSS protection
-	if(preg_match('#->input\[#', $template))
-	{
-		return true;
-	}
-
 	// Any other malicious acts?
 	// Courtesy of ZiNgA BuRgA
 	if(preg_match("~\\{\\$.+?\\}~s", preg_replace('~\\{\\$+[a-zA-Z_][a-zA-Z_0-9]*((?:-\\>|\\:\\:)\\$*[a-zA-Z_][a-zA-Z_0-9]*|\\[\s*\\$*([\'"]?)[a-zA-Z_ 0-9 ]+\\2\\]\s*)*\\}~', '', $template)))
--- a/admin/inc/functions.php
+++ b/admin/inc/functions.php
@@ -666,6 +666,12 @@ function check_template($template)
 		return true;
 	}
 
+	// Block usage of input fields : XSS protection
+	if(preg_match('#->input\[#', $template))
+	{
+		return true;
+	}
+
 	// Any other malicious acts?
 	// Courtesy of ZiNgA BuRgA
 	if(preg_match("~\\{\\$.+?\\}~s", preg_replace('~\\{\\$+[a-zA-Z_][a-zA-Z_0-9]*((?:-\\>|\\:\\:)\\$*[a-zA-Z_][a-zA-Z_0-9]*|\\[\s*\\$*([\'"]?)[a-zA-Z_ 0-9 ]+\\2\\]\s*)*\\}~', '', $template)))
--- a/modcp.php
+++ b/modcp.php
@@ -3277,12 +3277,12 @@ if($mybb->input['action'] == "editprofile")
 		$newtitle = '';
 	}
 
-	$birthday_year	= $mybb->input['birthday_year'];
-	$user_website	= $mybb->input['website'];
-	$user_icq		= $mybb->input['icq'];
-	$user_skype		= $mybb->input['skype'];
-	$user_google	= $mybb->input['google'];
-	$user_yahoo		= $mybb->input['yahoo'];
+	$birthday_year = $mybb->input['birthday_year'];
+	$user_website = $mybb->input['website'];
+	$user_icq = $mybb->input['icq'];
+	$user_skype = $mybb->input['skype'];
+	$user_google = $mybb->input['google'];
+	$user_yahoo = $mybb->input['yahoo'];
 
 	$plugins->run_hooks("modcp_editprofile_end");
 
--- a/modcp.php
+++ b/modcp.php
@@ -3276,6 +3276,13 @@ if($mybb->input['action'] == "editprofile")
 		$newtitle = '';
 	}
 
+	$birthday_year	= $mybb->input['birthday_year'];
+	$user_website	= $mybb->input['website'];
+	$user_icq		= $mybb->input['icq'];
+	$user_skype		= $mybb->input['skype'];
+	$user_google	= $mybb->input['google'];
+	$user_yahoo		= $mybb->input['yahoo'];
+
 	$plugins->run_hooks("modcp_editprofile_end");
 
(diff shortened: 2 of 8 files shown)


Eldenroot:
@effone you are on fire! Good job!

admin/inc/functions.php
```diff
@@ -666,6 +666,12 @@ function check_template($template)
 		return true;
 	}
 
+	// Block usage of input fields : XSS protection
+	if(preg_match('#->input\[#', $template))
```
Sama34: What is this for?

effone: To not allow `$class->input['something']` in templates by user edits.

Shade-: This likely breaks plugins in which inputs may be referenced safely (Template Conditionals for example, which is widely used). While it's perfectly fine to get rid of core references this is something I would not want to see included. My 2 cents.

effone: No it has no effect if the $class->input['some'] is already present in templates. This code validates when the user manually inserts such code in template and tries to save. If we allow users to do that then what is the point of dropping it?

If required I can remove it. I felt if we drop some support we should force it. We can't just indirectly encourage that way ...

Shade-: I know that existing templates are not affected, but using Template Conditionals I can for example manipulate the $mybb->input array and check for its values to display an active menu, which is a safe usage. This would reduce if not kill its functionality.

Eliminating the core references reduces potential XSS attacks as they were put there most likely without security in mind. That should be the point of this PR.

Shade-: It will not be as secure, but admins should know what they are doing. If security is a concern in this case – which is very unlikely to be exploitable anyway – I'd add a warning when a template includes a bare `$mybb->input` reference. However, note that `$mybb->input` gathers data from PHP's `$_POST` and `$_GET` arrays which are equally available in templates.

effone: So, what you suggest is inserting input in templates MyBB will accept the change with a flash message showing MyBB doesn't support this, use at your own risk, or something like?

Shade-: If it was for me, I'd drop this check and leave it as it always has been. But lets see what others think.

Sama34: I agree no change is necessary. Leave it as is.

euantorano:
I would agree too. So long as the core doesn’t use it, it should be fine.

effone:
Reverted. Thanks.

modcp.php
```diff
@@ -3276,6 +3277,13 @@
 		$newtitle = '';
 	}
 
+	$birthday_year	= $mybb->input['birthday_year'];
+	$user_website	= $mybb->input['website'];
+	$user_icq		= $mybb->input['icq'];
```
euantorano: There's some weird spacing going on here 😉

effone: Rectified.

PR #3177, opened 2018-04-29
Fixes #3175 PM Postbit

### LINKED ISSUE #317: User Merge bug/problem
> I just found out that the "merge users" tool does not update buddy lists, so the buddy connections of the account that is being deleted are lost.

Original report: http://dev.mybb.com/issues/2223
Reporter: jan wohlgemuth
Status: Confirmed

### LINKED ISSUE #3175: PM Postbit
Current code in functions_post.php.

```php
if($mybb->settings['enablepms'] == 1 && (($post['receivepms'] != 0 && $usergroup['canusepms'] != 0 && $mybb->usergroup['cansendpms'] == 1 && my_strpos(",".$post['ignorelist'].",", ",".$mybb->user['uid'].",") === false) || $mybb->usergroup['canoverridepm'] == 1))
```

Grabs the &#36;usergroup['canusepm'] in order to show the PM button.  However that is previously set by display group.

```php
// Get the usergroup
    if($post['userusername'])
    {
        if(!$post['displaygroup'])
        {
            $post['displaygroup'] = $post['usergroup'];
        }
        $usergroup = $groupscache[$post['displaygroup']];
    }
    else
    {
        $usergroup = $groupscache[1];
    }
```

I consider this a bug because the ability to see the PM button in postbit should depend on if the user has "canusepms" on any of a members groups and not the display group.

Original thread: [PM Postbit](https://community.mybb.com/thread-214239.html)

Sama34:
Fixes #3175 PM Postbit
I tried to update the code so for it to be more consistent.

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -1445,17 +1445,20 @@ if($mybb->input['action'] == "event")
 	$event['description'] = $parser->parse_message($event['description'], $event_parser_options);
 
 	// Get the usergroup
-	if($event['username'])
+	if($event['usergroup'])
 	{
-		if(!$event['displaygroup'])
-		{
-			$event['displaygroup'] = $event['usergroup'];
-		}
-		$user_usergroup = $groupscache[$event['displaygroup']];
+		$user_usergroup = usergroup_permissions($event['usergroup']);
 	}
 	else
 	{
-		$user_usergroup = $groupscache[1];
+		$user_usergroup = usergroup_permissions(1);
+	}
+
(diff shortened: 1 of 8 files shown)


euantorano:
This looks good to me.

As usual with permission changes, we need to make sure we do a bunch of testing on this. Pinging @kawaii: I'll be updating the test board tonight with changes I've merged, including this. If we can try to make sure we test permissions that would be great.

PR #3179, opened 2018-04-29
Fixes #3144 MyCode affects signature length Issue

### LINKED ISSUE #3144: MyCode affects signature length Issue
The "MyCode affects signature length" seems to be not working. See here:

I have the signature length limit set to 40, without BBCode, their is 31 letters and I have the setting set to not count the BBCode, however it's asking me to remove 71 letters.

Original thread: [MyCode affects signature length Issue](https://community.mybb.com/thread-217088.html)

Sama34:
Fixes #3144 MyCode affects signature length Issue

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1768,4 +1768,39 @@ class postParser
 
 		return $message;
 	}
+
+	/**
+	 * Removes any MyCode from a string
+	 *
+	 * @param string $message The message to be parsed
+	 * @return string The parsed message.
+	 */
+	function remove_mycode($message)
+	{
+		$find = array(
+			"#\[quote\](.*?)\[\/quote\](\r\n?|\n?)#si",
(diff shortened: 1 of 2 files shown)


euantorano:
Looking at this, custom MyCode will still affect signature length - do we think that's an acceptable compromise?

Sama34:
You are correct, the only way I can think for us to achieve that is by removing HTML tags instead. But nowhere it says HTML will count or not to signature length. So we will need to decide on that first.

effone:
But count should be based on only the visible test part regardless of styling. I think stripping html tags is fine whether or not html is enabled.
I just don't know what character we count against an inserted image.

effone:
I feel this PR needs to be recreated with a complete different approach (mine was just to count the visible part without any tags). I can work on it, in fact I was already. I just need a confirmation about the character count to be considered for an inserted image.

Eldenroot:
@effone - this is quite tricky - it should count only characters without images.

effone:
Replaced by this PR
https://github.com/mybb/mybb/pull/3357

PR #3185, opened 2018-05-01
Resolves #2139 : Soft deleted threads and answers

### LINKED ISSUE #2139: Soft deleted threads and answers
After soft deleting a thread the quick reply and the "New reply" button are still shown, however when I try to answer (either via Quick reply or when clicking on "New Reply") I get an error that the specified thread does not exist. I haven't checked the permissions here but either the buttons should be removed, I should be able to answer or I shouldn't see the thread. The current way is somehow wrong (I know the thread exists, I see the reply options but get an error).

Original thread: [Soft deleted threads and answers](http://community.mybb.com/thread-172302.html)

effone:
Attempt to fix: https://github.com/mybb/mybb/issues/2139

To Do:
- [x] Dynamically show / hide quickreply, new reply button, thread rating, thread tools for deleted thread
- [x] Disable functionality of above said tools for deleted threads
- [x] Dynamically show / hide post control buttons for deleted posts
- [x] Disable post control actions deleted posts
- [x] Disable certain moderator actions for deleted thread (showthread)
- [x] Disable moderation tools for above said disabled actions (showthread)
- [x] Handle custom mod tools for deleted thread
- [ ] Swap Soft Delete / Restore moderator tool on thread delete

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -11553,7 +11553,7 @@ if(use_xmlhttprequest == "1")
 </script>]]></template>
 		<template name="showthread_moderationoptions_approve" version="1816"><![CDATA[<option class="option_mirage" value="approvethread">{$lang->approve_thread}</option>]]></template>
 		<template name="showthread_moderationoptions_custom" version="120"><![CDATA[<optgroup label="{$lang->custom_mod_tools}">{$customthreadtools}</optgroup>]]></template>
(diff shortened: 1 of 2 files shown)


Sama34:
Whenever I delete a thread or post the button stops working (it doesn't toggle). Deleting the first post does removes some features but deleting a post doesn't seem to have the same effect for post buttons.

effone:
Recreated

PR #3188, opened 2018-05-02
Fixes #3173 search forum filtering

### LINKED ISSUE #3173: search forum filtering
I think I found a bug where when you filter any search (so it seems) via the forum column, it does not sort if via asc-desc or desc-asc properly.

Original thread: [search forum filtering](https://community.mybb.com/thread-214678.html)

effone:
Attempt to fix #3173

### DIFF
--- a/search.php
+++ b/search.php
@@ -90,7 +90,7 @@ if($mybb->input['action'] == "results")
 			}
 			break;
 		case "forum":
-			$sortfield = "t.fid";
+			$sortfield = "f.name";
 			break;
 		case "starter":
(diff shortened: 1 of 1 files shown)


Shade-:
This produces a SQL error due to `$sortfield` being used also [here](https://github.com/effone/mybb/blob/d52d9f54d1559d6f8fdf4ab9a4697fb0e4fa63cf/search.php#L346).

Shade-:
@effone

effone:
@Shade-

PR #3190, opened 2018-05-05
Fixes #1890 Sort threads by prefix bug

### LINKED ISSUE #1890: Sort threads by prefix bug
Okay so I'm not entirely sure if this is intended or if it these are bugs.

In the plug-ins section, when I sort the threads by the 1.8 prefix it displays all threads with the 1.8 prefix. There are two issues that I noticed. There are only 6 pages of 1.8 plug-ins,  yet on the pagination it shows 42 pages. 

Second issue is that on the 7th page it displays one thread. The thread is "Becoming an Approved Developer and Modifications Website Guide." It displays only this thread for every page after page 6. 

Page 7: http://community.mybb.com/forum-102-page-7.html?prefix=28

To see what I'm talking about just look at random pages after page 6.

Original thread: [Sort threads by prefix bug](http://community.mybb.com/thread-167723.html)

EDIT: This code looks wrong: https://github.com/mybb/mybb/blob/feature/forumdisplay.php#L613-626 The first query doesn't consider any parameters at all (why?) and the 2nd query misses prefix.

effone:
Attempt to fix #1890 
Forumdisplay Pagination Count Rectification

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -611,29 +611,9 @@ if(isset($fpermissions['canonlyviewownthreads']) && $fpermissions['canonlyviewow
 
 if($fpermissions['canviewthreads'] != 0)
 {
-	// How many posts are there?
-	if(($datecut > 0 && $datecut != 9999) || isset($fpermissions['canonlyviewownthreads']) && $fpermissions['canonlyviewownthreads'] == 1)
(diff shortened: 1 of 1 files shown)


Shade-:
Works for me.

PR #3194, opened 2018-05-07
Fixes #299 Unwanted CSS Path Modifier Removal

### LINKED ISSUE #299: MyBB CSS editor converts ../../../ (etc.) into ../
> I found this the other day, when I edited my global.css file using the MyBB advanced editor. A lot of url() statements were of the form ../../../images/night/tcat.png for example, and they were all changed to ../images/night/tcat.png upon saving. I didn't think this was a big deal, and the site still worked in all the web browsers I tested.
> 
> However, the server started generating a lot of error logs, and they were all reporting that the file "cache/themes/theme32/images/night/tcat.png" could not be found (with various images). So apparently there are some browsers out there that will try to load CSS URLs relative to the actual CSS file's position in the tree.
> 
> I've solved the problem for now by having absolute URLs, but maybe this is a bug?

Original report: http://dev.mybb.com/issues/1973
Reporter: Adrian Hayter
Status: New

effone:
Attempt to fix #299

### DIFF
--- a/admin/inc/functions_themes.php
+++ b/admin/inc/functions_themes.php
@@ -509,8 +509,8 @@ function resync_stylesheet($stylesheet)
 	return false;
 }
 
-// Depreciated since MyBB 1.8.16
 /**
+ * @deprecated
  * @param string $url
  *
  * @return string
@@ -527,8 +527,8 @@ function fix_css_urls($url)
 	}
 }
 
-// Depreciated since MyBB 1.8.16
 /**
+ * @deprecated
  * @param array $matches Matches.
  *
  * @return string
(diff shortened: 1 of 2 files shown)


euantorano:
That seems to make sense to me @Sama34.

effone:
I can't see those functions used anywhere ...

dvz:
It's good practice to avoid introducing breaking changes and assume everything is used - even if we searched through Extend plugin codes there are extensions that are published somewhere else, or not available publicly at all.

effone:
Alright I have placed the function back with a comment for all 3 as depreciated.

Sama34:
I would like to suggest the description in the `edit_stylesheet_simple_mode_desc` and `edit_stylesheet_advanced_mode_desc` language variables to be extended to add the path to where files are saves in the cache directory (both full and minimized files) as suggested before.

@mybb/developers

PR #3195, opened 2018-05-07
Fixes #2632 begin with bots

### LINKED ISSUE #2632: Robots in between online user order according to activity
Hi,

Username ascending orders robots first on index who's online section. While last activity ascending orders robots in between users which looks quite odd.

It should list robots either at first or last of the users according to activity. In between users confuses members as thinking robots as users.

Thanks

Original thread: [Robots in between online user order according to activity](https://community.mybb.com/thread-207265.html)

effone:
Attempt to fix #2632


Shade-:
Works.

PR #3196, opened 2018-05-08
Fixes #2137 Stylesheet order reset after importing a theme

### LINKED ISSUE #2137: Stylesheet order reset after importing a theme
The Stylesheet order is not being kept after a theme is imported.

https://github.com/mybb/mybb/blob/feature/admin/inc/functions_themes.php#L337

effone:
Attempt to fix #2137 
Preserve Stylesheet Disporder, remove unwanted process to reset disporder.


Shade-:
Works.

PR #3199, opened 2018-05-08
AIM Removal

effone:
As [AIM has been discontinued](https://help.aol.com/articles/aim-discontinued); removed all instances of AIM from MyBB ...

https://github.com/mybb/mybb/pull/3153 recreated.


effone:
@euantorano please check this PR before merging any other as it will surely conflict. Please. Thanks.

inc/captcha_fonts/read_me.html
```diff
@@ -59,33 +59,32 @@ <h3>
       </li>
     </ul>
     <ul>
-      <li>Q: How can I use this font in <a href="http://www.aol.com/aim/">AOL Instant
-      Messenger</a>, <a href="http://messenger.msn.com/">MSN Messenger</a>, <a href=
```
Sama34: Isn't this a third-party library file? Do we modify these?

effone: No clue ...
The quote marks were appearing broken though, so I was bound to touch that file.

install/resources/settings.xml
```diff
@@ -2452,15 +2452,6 @@ delete=Delete]]></optionscode>
 			<isdefault>1</isdefault>
 			<helpkey></helpkey>
 		</setting>
-		<setting name="allowaimfield">
-			<title>Allow AIM Screen Name To Usergroups</title>
-			<description><![CDATA[Select the usergroups which should be allowed to use the AIM Screen Name contact field.]]></description>
-			<disporder>2</disporder>
```
Sama34: Since you are removing this setting you need to downgrade the `disporder` for the rest of the settings. Thus, `allowyahoofield` `disporder` value would be `2` now an so on.

effone: Ok will do. But is there any impact?

misc.php
```diff
@@ -883,13 +883,6 @@
 
 	// Build IM navigation bar
 	$navigationbar = $navsep = $imtype = $imtype_lang = '';
-	if(!empty($user['aim']) && is_member($mybb->settings['allowaimfield'], array('usergroup' => $user['usergroup'], 'additionalgroups' => $user['additionalgroups'])))
-	{
-		$imtype = "aim";
-		$imtype_lang = $lang->aol_im;
-		eval("\$navigationbar .= \"".$templates->get("misc_imcenter_nav")."\";");
```
Sama34: You need to delete this template from `theme.xml` too.

install/resources/settings.xml
Sama34: Not really. The display order can be assigned within the install code. I suggested removing this way back but never got to do so myself.

Until we remove it we should try to be consistent.

effone: Done.

misc.php
effone: Done.

effone:
All done. @Sama34

install/resources/upgrade43.php
```diff
@@ -35,7 +35,13 @@ function upgrade43_dbchanges()
 		$db->update_query('settings', array('value' => 1), "name='captchaimage'"); // Reset CAPTCHA to MyBB Default
 		$db->update_query('settings', "value=''", 'name IN (\'captchapublickey\', \'captchaprivatekey\''); // Clean out stored credential keys
 	}
-
+	
+	if($db->field_exists('users', 'aim'))
```
dvz: The order of arguments for [`DB_Base::field_exists()`](https://github.com/mybb/mybb/blob/mybb_1815/inc/db_base.php#L156) and [`DB_Base::drop_column()`](https://github.com/mybb/mybb/blob/mybb_1815/inc/db_base.php#L362) is inverted here

effone: Oops! Corrected.
https://github.com/mybb/mybb/pull/3220

PR #3201, opened 2018-05-09
Fix #2225 wordfilter cannot replace strings with * in them

### LINKED ISSUE #2225: wordfilter cannot replace strings with * in them
If you set up a replace rule like this:
'c***_ck' => Toast
If a user then writes c**_*ck it will not be relaced as the \* will be treated as wildcard and it does not include itself.

Sama34:
Fix #2225 wordfilter cannot replace strings with * in them

### DIFF
--- a/admin/modules/config/badwords.php
+++ b/admin/modules/config/badwords.php
@@ -59,6 +59,7 @@ if($mybb->input['action'] == "add" && $mybb->request_method == "post")
 	{
 		$new_badword = array(
 			"badword" => $db->escape_string($mybb->input['badword']),
+			"regex" => $mybb->get_input('regex', 1),
 			"replacement" => $db->escape_string($mybb->input['replacement'])
 		);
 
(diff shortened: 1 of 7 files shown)


admin/modules/config/badwords.php
```diff
@@ -59,6 +59,7 @@
 	{
 		$new_badword = array(
 			"badword" => $db->escape_string($mybb->input['badword']),
+			"regex" => $mybb->get_input('regex', 1),
```
euantorano: Rather than using `1` here, can you use `MyBB::INPUT_INT` instead? it's a little bit more readable that way.

inc/class_parser.php
```diff
@@ -635,11 +635,19 @@ function parse_badwords($message, $options=array())
 					$badword['replacement'] = "*****";
 				}
 
-				// Take into account the position offset for our last replacement.
-				$badword['badword'] = str_replace('\*', '([a-zA-Z0-9_]{1})', preg_quote($badword['badword'], "#"));
+				if($badword['regex'])
+				{
+					//_dump($badword['badword']);
```
euantorano: Can you remove this comment please? Looks like it was left behind from debugging 😄

Sama34:
I have fixed the early mentioned issues.

euantorano:
Great, thanks!

Issue #2866, opened 2017-10-06
member.php "Awaiting Activation Error"


euantorano:
Sorry English.

When a member moves to "Awaiting Activation" group.

This member is in the user control panel when "Recent Activation Email" is clicked.

Will not be able to activate the member account

Because, mybb_awaitingactivation table, this user type value is null.

member.php, line 1402: 
PHP Code:&#36;query&nbsp;=&nbsp;&#36;db-&gt;simple_select("awaitingactivation",&nbsp;"*",&nbsp;"uid='".&#36;user['uid']."'&nbsp;AND&nbsp;(type='r'&nbsp;OR&nbsp;type='e'&nbsp;OR&nbsp;type='b')");&nbsp;

to correct

member.php, line 1531: 
PHP Code:&#36;uid&nbsp;=&nbsp;&#36;user['uid'];&nbsp;
this code after

PHP Code:if(!&#36;user['type'])&nbsp;{&#36;user['type']&nbsp;=&nbsp;"r";}&nbsp;

I hope I made myself clear

Original thread: [member.php "Awaiting Activation Error"](https://community.mybb.com/thread-213410.html)

Eldenroot:
OK, mentioned fix works for me, if anything I can pull a PR :)

euantorano:
Might be best to fix the admin CP to actually fill that column with the correct value I think.

ozanakkaya:
Admin cp does not have. Because, "do_resendactivation" adding a new row to mybb_awaitingactivation table.

And
$cache->update_awaitingactivation(); ???

Shade-:
I haven't looked at it, but we can add it. `$cache->update_awaitingactivation()` updates the cache by reading from the database, it does not update the database itself if that's what you asked.

ozanakkaya:
What if the wrong address is waiting for activation??

effone:
That is to be taken care by admin. If admin is worried about storing valid mails he should not set registration to 'Administrator Activation' at the first place. But thats a different issue. Our current situation is about manually moving a user to Awaiting Activation group (though I have no clue why anyone will want to move an activated member to awaiting).

Sama34:
The user activation process should be the same as the set option in settings.

Ben-MyBB:
PR wasn't linked to this issue but I can't replicate this after #3204

PR #3204, opened 2018-05-12
member.php: Add missing "mybb->" in resendactivation action

mytbk:


Shade-:
`$mybb->user` refers to a logged in user; resendactivation is meant to be used by guests, so `$mybb->user['uid']` will always resolve to `0`, invalidating its purpose.

It's still a mystery to me where `$user` is declared. @mybb/developers @mybb/sqa

effone:
Right. Its not declared for this action. So, after all there is a mistake.

mytbk:
> $mybb->user refers to a logged in user; resendactivation is meant to be used by guests, so $mybb->user['uid'] will always resolve to 0, invalidating its purpose.

Hmm, I found this when I log in as an unactivated user and click the resend activation email link in usercp, then mybb gives an SQL error:

```
SQL Error:
0 - ERROR: invalid input syntax for integer: "" LINE 1: SELECT * FROM mybb_awaitingactivation WHERE uid='' AND type=... ^
Query:
SELECT * FROM mybb_awaitingactivation WHERE uid='' AND type='b'
```

effone:
Resend activation link appears after login. So, by that time user obtains his UID.
I think this is a valid PR.

Shade-:
Ah now I get it – indeed this will fix the issue.

Shade-:
Thank you for your contribution! Next time, please open an issue before pushing a PR so it'll appear in the fixed issues list when we deploy new versions of MyBB.

effone:
Related?
https://github.com/mybb/mybb/issues/2866

Shade-:
Possibly. I'm not sure why `$user` is used if not declared.

euantorano:
> Possibly. I'm not sure why $user is used if not declared.

Goodness knows, it's probably been that way for ages.

effone:
Since a year, 1.8.12 (as per blame)
Not revealing the name though ...

PR #3205, opened 2018-05-12
Fix missing ) in the usergroups action of usercp

euantorano:


Shade-:
Darn, my bad. Sorry.

euantorano:
No problem, I should have noticed it before 😄

PR #3206, opened 2018-05-12
Fixes #3203 Move Contact fields to Custom Profile Fields

### LINKED ISSUE #3203: Move Contact fields to Custom Profile Fields
https://community.mybb.com/thread-214204.html

Move the contact fields from their current hardcoded database fields to Custom Profile Fields that Admins can Add/Edit/Delete with ease.

PaulBender:
Fixes #3203

- Added one column to the profilefields table, which when set will treat the profile field as a contact field. Contact fields will appear where the current hardcoded fields appear.
- Added Facebook, Twitter and Skype as the default fields included on install.


PaulBender:
Conflicts fixed

PR #3216, opened 2018-05-17
Resolves #2139 : Soft deleted threads and answers

### LINKED ISSUE #2139: Soft deleted threads and answers

effone:
Attempt to fix: https://github.com/mybb/mybb/issues/2139
Recreated #3185 

To Do:
- [x] Dynamically show / hide quickreply, new reply button, thread rating, thread tools for deleted thread
- [x] Disable functionality of above said tools for deleted threads
- [x] Dynamically show / hide post control buttons for deleted posts
- [x] Disable post control actions deleted posts
- [x] Disable certain moderator actions for deleted thread (showthread)
- [x] Disable moderation tools for above said disabled actions (showthread)
- [x] Handle custom mod tools for deleted thread
- [x] Swap Soft Delete / Restore moderator tool on thread delete

### DIFF
--- a/jscripts/thread.js
+++ b/jscripts/thread.js
@@ -40,7 +40,13 @@ var Thread = {
 		var new_post_ids = new Array();
 		var quoted = Cookie.get("multiquote");
 		var is_new = true;
-		if(quoted)
+		if($("#pid" + pid).next("div.post").hasClass('deleted_post'))
+		{
+			$.jGrowl(lang.post_deleted_error, {theme:'jgrowl_error'});
+			var deleted = true;
+		}
+
+		if(quoted && !deleted)
 		{
 			var post_ids = quoted.split("|");
 
@@ -57,7 +63,7 @@ var Thread = {
 		}
 
 		var mquote_a = $("#multiquote_"+pid).closest('a')
-		if(is_new == true)
+		if(is_new == true && !deleted)
 		{
 			new_post_ids[new_post_ids.length] = pid;
 			mquote_a.removeClass('postbit_multiquote').addClass('postbit_multiquote_on');
@@ -270,6 +276,11 @@ var Thread = {
 				// Take pid out of the id attribute
 				id = $(this).attr('id');
 				pid = id.replace( /[^\d.]/g, '');
+				if($("#pid" + pid).next("div.post").hasClass('deleted_post'))
+				{
+					$.jGrowl(lang.post_deleted_error, {theme:'jgrowl_error'});
+					return false;
+				}
 
 				// Create a copy of the post
 				if($('#pid_' + pid + '_temp').length == 0)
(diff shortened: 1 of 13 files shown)


euantorano:
I think you should be able to edit a post inside a soft deleted thread personally, I can think of a few cases where you might do that then un-delete the thread. 

I’ll check the update script through.

effone:
@euantorano With this PR the posts can be edited in a soft deleted thread as long as the post itself is not soft deleted.

Sama34:
Regarding of the final decision about editing posts, users can still see the quote/report/etc buttons even if they can't use them.

effone:
@Sama34 if you read the main Issue by @JN-Jones #2139  that is the thing what has been pointed out as bug. If I can't use the tool I shouldn't have seeing the button - thats the gist.

Sama34:
Why don't you use the approve/unapprove feature and not soft delete? What did you use before we implemented soft delete? We could also add a new moderation option to "Allow Editing of Soft Deleted Threads" and make it selected by default.

Eldenroot:
New option would be useful :) anyway can be added later.

jscripts/thread.js
```diff
@@ -40,7 +40,13 @@ var Thread = {
...
 		var quoted = Cookie.get("multiquote");
 		var is_new = true;
-		if(quoted)
+		if($("#pid" + pid).next("div.post").hasClass('deleted_post'))
+		{
+			$.jGrowl(lang.post_deleted_error, {theme:'jgrowl_error'});
+			var deleted = true;
+		}
+
+		if(quoted && !deleted)
```
Shade-: From a quick look, this should raise a fatal error if `deleted` is not defined. You should add `var deleted = false` after `is_new = true` and change its value within the deleted_post check above.

effone: Done!

Eldenroot:
I tested this PR as effone requested, seems to be fine. Can anybody else confirm this? I would like to see a separated pr with new setting in future

effone:
The settings thing what @Sama34 said is easy to implement. Just getting no urge as I have a feeling that this is not gonna be considered for .16 as its a lengthy PR and needs quiet a bit of testing.
I have created this PR 3 times, tested in deep, still @Shade- has caught a mistake. So, you see ... It needs multiple testing by multiple people.

euantorano:
Updated template versions to `1818`, currently testing this for 1.8.18.

effone:
**Note:**

Admins must confirm the following chunk of CSS is present in `global.css` else need to add manually, preferably at line 1340 to make this PR work as desired.

```
.postbit_qrestore,
.status_type {
	display:none!important;
}

.deleted_post .postbit_qrestore,
.deleted_post .status_type {
	display:inline!important;
}

.deleted_post .postbit_mirage {
	display:none!important;
}
```

YashvirGaming:
hello why mybb deleting all my threads after 2-3 days, how to prevent it
I don't want my threads to be deleting, please how to stop it from being auto delete

PR #3218, opened 2018-05-17
Fixes #2094 Smart bug in color name

### LINKED ISSUE #2094: smart Bug in color name
hello men,
in today i want to release a theme with 4 colors, in run its work good but after export want to test it,

and see an error that is 'theme file is not complete or is not a theme file'

after many tests find the problem. in name of color such as "Blue&amp;Red=Blue Red" should be remove '&amp;' or replace with another character such as '_' . (for this should be write an topic for select an allowed characters for these places &lt;namly: in textboxes and textareas&gt;

to be succeed

excuse me for my language,my en is not good,sorry  (A) ;) :blush:

Original thread: [smart Bug in color name](http://community.mybb.com/thread-170598.html)

effone:
Attempt to fix #2094 
Validation added to catch invalid color naming pattern.

### DIFF
--- a/admin/modules/style/themes.php
+++ b/admin/modules/style/themes.php
@@ -1094,9 +1094,16 @@ if($mybb->input['action'] == "edit")
 
 			foreach($colors as $color)
 			{
-				$color = explode("=", $color);
-
-				$properties['colors'][$color[0]] = $color[1];
+				$color = trim($color);
+				if(preg_match('/^[a-z0-9]+={1}[a-z0-9]+$/i', $color))
+				{
+					$color = explode("=", $color);
+					$properties['colors'][$color[0]] = $color[1];
+				}
+				else
+				{
+					$errors[] = $lang->sprintf($lang->error_invalid_color, $color);
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


admin/modules/style/themes.php
```diff
@@ -1094,9 +1094,16 @@ function checkAction(id)
 
 			foreach($colors as $color)
 			{
-				$color = explode("=", $color);
-
-				$properties['colors'][$color[0]] = $color[1];
+				$color = trim($color);
+				if(preg_match('/^[a-z0-9]+={1}[a-z0-9]+$/i', $color))
```
euantorano: The `{1}` here shouldn't be necessary. Also, this wouldn't allow uppercase letters, which I would expect it to allow. the pattern should probably be `a-zA-Z0-9`.

effone: {1} makes the message more user friendly. If you mistake in 3 naming at a stretch there will be exact same message thrice without indicating where is the problem.

a-zA-z not required. Thats silly, still massively used. There is case insensitive flag `/i` in place and thats all we need.

@euantorano

euantorano:
Ah I missed the flag, sorry.

PR #3226, opened 2018-05-26
Add Prague, Bratislava to the timezones

BKPepe:
Hello,
Honestly, I don't know which branch I should use as a base.

My commit adds the capital cities Prague (CZ), Bratislava (SK) to the timezones in the user interface and also in the ACP.

I fixed typo in the capital city of Bangladesh. It'is Dhaka according to the [Wikipedia](https://en.wikipedia.org/wiki/Dhaka).


Eldenroot:
I can confirm, these missing cities should be added.

euantorano:
Thanks for your contribution 😄

PR #3227, opened 2018-05-28
Resolves #3164 Closed vs Locked

### LINKED ISSUE #3164: Closed vs. Locked
Not a big issue, just something odd came to notice.
Some places it is "Closed Thread" some places it is "Locked Thread".
We should choose either of two and stick with it. Isn't it?

Note: Majority is with implementing "**Closed Thread**" everywhere.
[Original Thread](https://community.mybb.com/thread-216704.html)

AndreRl:
Fixes #3164 consistency problem.


inc/languages/english/forumdisplay.lang.php
```diff
@@ -80,7 +80,7 @@
 $l['icon_no_new'] = "No new posts.";
 $l['icon_new'] = "New posts.";
 $l['icon_hot'] = " Hot thread.";
-$l['icon_lock'] = " Locked thread.";
+$l['icon_lock'] = " Closed thread.";
```
effone: Why don't you change the var name to `icon_close` ?

effone:
Nice, but still there are '`locked` in backend codes, I don't know, we can ignore ...
For example: `get_forum_lightbulb(...)` in `functions_forumlist.php`.

AndreRl:
Fixed. Could of left it but it would still appear as "lock" in the global.css file. Missing one for subforums... Should be in a similar context to functions_forumlist.php

AndreRl:
Lang files were updated @Eldenroot. I believe I've gotten them all.

install/resources/mybb_theme.xml
```diff
@@ -323,7 +323,7 @@ a:active {
 	background-position: 0 -30px;
 }
 
-.forum_offlock {
+.forum_offclose {
```
euantorano: Note: Themes don't get upgraded by the upgrade system. We need to make a note in the release notes in **big** letters about this @mybb/developers @mybb/sqa.

PR #3235, opened 2018-06-03
Fixes #3232 SQL Error when opening the moderator panel

### LINKED ISSUE #3232: SQL Error when opening the moderator panel
The error below occurs when a moderator (primary group moderator) want to open the moderator-panel.
MyBB Version: 1.8.6
PHP Version: 5.5.9-1ubuntu4.14
SQL Engine: PostgreSQL 9.3.10
active plugins: Email Admin-Created User (1.0) Favicon (1.6.1) MyCode: [spoiler] (1.2)
inactive plugins: First Post Only (1.2)

MyBB SQL Error

SQL Error:
    0 - ERROR: operator does not exist: ! smallint LINE 7: WHERE (t.fid &lt;&gt; 0 AND t.fid IN (0,'3')) OR (!l.fid) ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. 
Query:
    SELECT l.*, u.username, u.usergroup, u.displaygroup, t.subject AS tsubject, f.name AS fname, p.subject AS psubject FROM mybb_moderatorlog l LEFT JOIN mybb_users u ON (u.uid=l.uid) LEFT JOIN mybb_threads t ON (t.tid=l.tid) LEFT JOIN mybb_forums f ON (f.fid=l.fid) LEFT JOIN mybb_posts p ON (p.pid=l.pid) WHERE (t.fid &lt;&gt; 0 AND t.fid IN (0,'3')) OR (!l.fid) ORDER BY l.dateline DESC LIMIT 5

Original thread: [SQL Error when opening the moderator panel](https://community.mybb.com/thread-187731.html)

effone:
Attempt to fix #3232 
Applied alternate operator.

### DIFF
--- a/modcp.php
+++ b/modcp.php
@@ -4743,7 +4743,7 @@ if(!$mybb->input['action'])
 		$where = '';
 		if($tflist_modlog)
 		{
-			$where = "WHERE (t.fid <> 0 {$tflist_modlog}) OR (l.fid <> 0)";
+			$where = "WHERE (t.fid != 0 {$tflist_modlog}) OR (l.fid != 0)";
 		}
 
 		$query = $db->query("


euantorano:
I think the SQL error is actually related to the part: `(!l.fid)` (`ERROR: operator does not exist: ! smallint`). `<>` is supported syntax for `not equals`, along with `!=`.

AndreRl:
@euantorano the actual bug report is from 2016. You remove the last trace of the `not equals` syntax in #2965

euantorano:
Yep, so this can be closed, along with the original report 😄 Good catch @AndreRl!

effone:
The actual part in the issue is not even exist in modcp.php. As the `<>` operator is also in question I just used the alternate.

PR #3240, opened 2018-06-06
Fixes #2445 Host check code doesn't work anymore

### LINKED ISSUE #2445: Host check code doesn't work anymore
https://github.com/Devilshakerz/mybb/blob/4b11a11b4f5c1fea22b622bf8e5f8f073e625a4f/inc/functions_serverstats.php#L215

That site doesn't seem to work that way anymore. Also, replacing `www.` was never fully proper because the URL can be something like `http://thewww.com`

effone:
Attempt to fix #2445 
host details fetching drop

### DIFF
--- a/inc/functions_serverstats.php
+++ b/inc/functions_serverstats.php
@@ -205,6 +205,10 @@ function build_server_stats($is_install=1, $prev_version='', $current_version=''
 	// Host URL & hostname
 	// Dropped fetching host details since v.1.8.16 as http://www.whoishostingthis.com API seems to be down and this info is not required by MyBB.
 	$info['hosturl'] = $info['hostname'] = "unknown/local";
+	if($_SERVER['HTTP_HOST'] == 'localhost')
+	{	
+		$info['hosturl'] = $info['hostname'] = "localhost";	
+	}
 
 	if(isset($_SERVER['HTTP_USER_AGENT']))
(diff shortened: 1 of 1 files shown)


inc/functions_serverstats.php
```diff
@@ -203,55 +203,8 @@ function build_server_stats($is_install=1, $prev_version='', $current_version=''
 	}
 
 	// Host URL & hostname
+	// Dropped fetching host details since v.1.8.16 as http://www.whoishostingthis.com API seems to be down and this info is not required by MyBB.
 	$info['hosturl'] = $info['hostname'] = "unknown/local";
-	if($_SERVER['HTTP_HOST'] == 'localhost')
```
Shade-: Shouldn't this check be kept? `$info['hosturl']` and `$info['hostname']` are sent, we might need `localhost` informations.

effone: Are we at all keeping the track? I thought no. Just left the default values to avoid any unwanted error.
Anyway, added back. No harm with it.

Shade-: No idea. However, as the unknown/local part was there, I assumed it was correct to detect localhost too. If unwanted, we can delete it later on. In the meantime, this PR fixes the hostcheck not working anymore. Good job!

PR #3253, opened 2018-06-11
Working on #2972 Template Conversion to Twig Format (footer)

Shade-:
- [x] footer
- [x] footer_contactus
- [x] footer_languageselect
- [x] footer_languageselect_option
- [x] footer_themeselect
- [x] footer_themeselector

Note that this also includes `task_image` in the footer template, but its conversion will be handled separately.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -4199,7 +4199,7 @@ function get_current_location($fields=false, $ignore=array(), $quick=false)
         if (!is_array($ignore)) {
             $ignore = array($ignore);
         }
-		$fields = [];
+        $fields = [];
         if (!empty($mybb->input)) {
             foreach ($mybb->input as $name => $value) {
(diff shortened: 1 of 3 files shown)


inc/functions.php
```diff
@@ -4199,19 +4199,18 @@ function get_current_location($fields=false, $ignore=array(), $quick=false)
         if (!is_array($ignore)) {
             $ignore = array($ignore);
         }
-
-        $form_html = '';
+		$fields = [];
```
euantorano: ANother case of tabs/spaces causing problems here 😉

inc/views/base/partials/footer.twig
```diff
@@ -1,3 +1,72 @@
+            <debugstuff>
```
euantorano: I'm surprised Twig doesn't complain about this honestly...

PR #3257, opened 2018-06-12
Adjustments for announcements (#2972)

Shade-:
This PR fixes some issues with announcements which weren't converted properly in my previous PR, such as alt_trow and announcement preview.

### DIFF
--- a/inc/views/base/modcp/announcements.twig
+++ b/inc/views/base/modcp/announcements.twig
@@ -1,7 +1,7 @@
 {% extends 'layouts/master.twig' %}
 
 {% block head %}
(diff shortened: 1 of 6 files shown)


inc/views/base/modcp/announcements.twig
```diff
@@ -22,8 +22,9 @@
                         </tr>
                         {% if globalAnnouncements %}
                             {% for aid, announcement in globalAnnouncements %}
+                        		{% set bgcolor = alt_trow() %}
```
euantorano: Indenting here looks a bit off.

Shade-:
Quick fix pushed :)

PR #3259, opened 2018-06-12
Working on #2972 Template Conversion to Twig Format (modcp finduser)

Shade-:
- [x] modcp_finduser
- [x] modcp_finduser_noresults
- [x] modcp_finduser_user

### DIFF
--- a/inc/views/base/modcp/finduser.twig
+++ b/inc/views/base/modcp/finduser.twig
@@ -20,32 +20,32 @@
                         <td class="tcat" align="center" width="15%"><span class="smalltext"><strong>{{ lang.lastvisit }}</strong></span></td>
                         <td class="tcat" align="center" width="10%"><span class="smalltext"><strong>{{ lang.postnum }}</strong></span></td>
                     </tr>
-                    {% if users %}
-                        {% for user in users %}
-                            {% set bgcolor = alt_trow() %}
-                            <tr>
-                                <td class="{{ bgcolor }}">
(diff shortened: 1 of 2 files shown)


inc/views/base/modcp/finduser.twig
```diff
@@ -0,0 +1,125 @@
...
+                        <td class="thead" colspan="5"><strong>{{ lang.users }}</strong></td>
+                    </tr>
+                    <tr>
+                        <td class="tcat" width="30%"><span class="smalltext"><strong>{{ lang.username }}</strong></span></td>
+                        <td class="tcat" align="center" width="15%"><span class="smalltext"><strong>{{ lang.usergroup }}</strong></span></td>
+                        <td class="tcat" align="center" width="15%"><span class="smalltext"><strong>{{ lang.regdate }}</strong></span></td>
+                        <td class="tcat" align="center" width="15%"><span class="smalltext"><strong>{{ lang.lastvisit }}</strong></span></td>
+                        <td class="tcat" align="center" width="10%"><span class="smalltext"><strong>{{ lang.postnum }}</strong></span></td>
+                    </tr>
+                    {% if users %}
```
euantorano: I believe this `if` is redundant. if we needed an else clause for an empty array, we can use:

```twig
{% for user in users %}
    {# user #}
{% else %}
    {# no users #}
{% endfor %}
```

Shade-: Fixed, and added no_user row back (somehow it didn't make it in the first commit).

PR #3261, opened 2018-06-12
Working on #2972 Template Conversion to Twig Format (usercp2)

Shade-:
- [x] usercp_addsubscription_thread

This PR also includes @effone's commit to remove usercp2 which has been pushed for 1.8.16.


inc/views/base/usercp/subscribe_thread.twig
```diff
@@ -0,0 +1,57 @@
+{% extends 'layouts/master.twig' %}
+
+{% block head %}
+    <title>{{ mybb.settings.bbname }} - {{ lang.forum_subscriptions }}</title>
+{% endblock head %}
+
+{% block body %}
+    <form action="usercp2.php" method="post" name="input">
```
euantorano: Shouldn't this action be `usercp.php` instead since we've removed `usercp2.php`?

inc/views/base/usercp/subscribe_thread.twig
```diff
@@ -0,0 +1,57 @@
...
+        <input type="hidden" name="my_post_key" value="{{ mybb.post_code }}" />
```
euantorano: I wonder if we should move this into a partial template since it's so commonly used...

install/resources/mybb_theme.xml
```diff
@@ -1814,7 +1814,7 @@ div.jqi .jqisubstate .jqibuttons button{
 
 /** Impromptu End */
 ]]></stylesheet>
-		<stylesheet name="usercp.css" attachedto="usercp.php|usercp2.php|private.php" version="1808" disporder="2"><![CDATA[.usercp_nav_item {
+		<stylesheet name="usercp.css" attachedto="usercp.php|private.php" version="1808" disporder="2"><![CDATA[.usercp_nav_item {
```
euantorano: The `version` here should probably be updated, but since we plan to look at altering the theme system a bit it should be ok.

Shade-: Yeah, I cherry-picked @effone's PR and applied it; he updated template versioning in a separate PR which I did not include since we're going to remove it anyway.

inc/views/base/usercp/subscribe_thread.twig
Shade-: I agree. However, I feel like it should be done when all templates are converted.

inc/views/base/usercp/subscribe_thread.twig
Shade-: Yup, I'll update it in a moment.

PR #3265, opened 2018-06-16
Working on implementing SwiftMailer

### LINKED ISSUE #2909: Add SwiftMailer library to handle the sending of emails
The current mailhandler setup works for most cases, but over the years we have encountered issues with several email systems where emails would fail to send.

As email is a fairly important component of the system, making use of a well tested and highly capable third party library used by many other software projects should help aleviate some of the issues we've sene in the past and take some pressure of maintaining the mail transport mechanisms off the team.

[SwiftMailer](https://swiftmailer.symfony.com/) is just such a library and is used in a wide variety of frameworks and applications. It should be fairly easy to integrate with the MyBB core since all emails are sent through the `my_mail` funciton, meaning we should eb able to modify a single function to get things working.

euantorano:
Fixes #2909

This removes the old mail handler class and substitutes SwiftMailer for sending emails.

This code is not yet tested, but I will be testing it during tomorrow.


Shade-:
Two issues found:
- errors are not handled properly, internal Exceptions are displayed;
- `send_mail_queue` is registered as a shutdown function, but it seems shutdown functions do not fire in 1.9. I don't know if it's an issue with my local install or if it's screwed for everyone, but a test might be worth. I don't know to debug shutdown functions unfortunately.

euantorano:
Yep, I just need to rebase the `develop/1.9` branch @zygzagZ, which I've been meaning to do for ages.

rajat315315:
I can't understand, there are too many commits for this SwiftMail feature.

euantorano:
That is due to the ongoing rebase - this PR will be recreated after the rebase is complete.

PR #3272, opened 2018-06-23
Remove unnecessary inserting of settings.

euantorano:
Also fixed update query for captcha.


dvz:
These also seem unnecessary: https://github.com/euantorano/mybb/blob/3c2b97aed8df0064056d2da38eb89c4c56de4435/install/resources/upgrade43.php#L31-L32

euantorano:
Removed.

PR #3273, opened 2018-06-23
Fixed partial rebase issue with forumdisplay

Shade-:
For #2972. This PR fixes a lot of issues due to a partial conversion likely originated from a corrupted/wrong rebase and conflicts fix.

Note that some things are marked as TODO in Twig comments (mainly `get_thread_link` lacking from Twig functions).

PR #3274, opened 2018-06-23
#2976 - add `my_date` filter and other useful twig functions

euantorano:
Part of #2976.

This PR adds the following:

- `my_date` filter:

```twig
{{ post.dateline|my_date('relative', 'offset', true) }}
```

This means that language strings are changed slightly, moving HTML into a partial template rather than a language string. Dates now also use the HTML 5 `<time>` element, with eh `datetime` attribute set and a `title` attribute that shows the date as a full string whether relative formatting is enabled or not.

I've tried to keep this functioning the same as the old `my_date`, but it was unclear to me what the different values of the `$ty` parameter to that function actually achieved...
- `my_number_format` filter:

```twig
{{ memprofile.threadnum|my_number_format }}
```

I plan to add other functions to this PR over this weekend.

### DIFF
--- a/inc/languages/english/global.lang.php
+++ b/inc/languages/english/global.lang.php
@@ -132,8 +132,6 @@ $l['no_subscribe_notification'] = "Subscribe without receiving any notification 
 $l['instant_email_subscribe'] = "Subscribe and receive email notification of new replies";
 $l['instant_pm_subscribe'] = "Subscribe and receive PM notification of new replies";
 
-$l['today_rel'] = "<span title=\"{1}\">Today</span>";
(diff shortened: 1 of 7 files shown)


Shade-:
`my_number_format` and `get_thread_link` are worth too IMHO.

euantorano:
@Shade- Yep, see the linked issue. I'm going to add some more functions to this PR, but don't want to do them all here.

Shade-:
Wooops sorry, didn't see it. Will post it there.

euantorano:
Added `my_number_format` filter.

euantorano:
Merging as `my_number_format` is required for #3273.

PR #3277, opened 2018-06-23
Fixes #3276 -PHP 7 multipage issue

### LINKED ISSUE #3276: PHP 7 multipage issue
Today I was debugging stuff on my forum and I found an issue with multipage. The issue happens in some pages that uses multipage. The issue happens if you specify a big number for page because in many pages there aren't conditions that check if specified page is greater then max page.
This is an example:
https://community.mybb.com/memberlist.php?page=999999999999999999
Many admin pages and some forum page are affected by this. It is a low bug but it should be fixed.

SvePu:


SvePu:
#3277

euantorano:
This fixes one occurrence of this issue, but unfortunately the same problem can be found all over the core :(

SvePu:
Ok I'll closing this PR .... #3278 is a complete fix

PR #3278, opened 2018-06-24
Fixes #3276 - PHP 7 multipage issue

### LINKED ISSUE #3276: PHP 7 multipage issue

chack1172:


euantorano:
This looks good. I'm leaving it open for now until 1.8.16 is released then we can merge.

SvePu:
I've closed my PR #3277 ... This is the better fix.

euantorano:
As 1.8.17 is a hot fix release, this will be part of 1.8.18. There isn't a milestone for 1.8.18 yet, but once there is I will assign it.

misc.php
```diff
@@ -293,9 +302,6 @@
 	}
 
 	// Do Multi Pages
-	$query = $db->simple_select("helpdocs", "COUNT(*) AS total", "hid IN(".$db->escape_string($search['querycache']).")");
-	$helpcount = $db->fetch_array($query);
-
 	if($upper > $helpcount)
```
Shade-: This doesn't make any sense; `$upper` is a number, `$helpcount` is an array.

private.php
```diff
@@ -303,9 +312,6 @@
 	}
 
 	// Do Multi Pages
-	$query = $db->simple_select("privatemessages", "COUNT(*) AS total", "pmid IN(".$db->escape_string($search['querycache']).")");
-	$pmscount = $db->fetch_array($query);
-
 	if($upper > $pmscount)
```
Shade-: Likewise.

PR #3281, opened 2018-06-25
Fixes #3241 .prop('checked') doesn't trigger change events

### LINKED ISSUE #3241: .prop('checked') doesn't trigger change events

Eric-Jackson:
#3241


dvz:
Same can be applied to `feature` (1.8) now

euantorano:
Yes, would be a good idea. Could you duplicate this PR for 1.8 @Eric-Jackson?

Eric-Jackson:
Yup sorry, meant to do that!

PR #3305, opened 2018-07-05
Fix #3304 - conditional sepator

### LINKED ISSUE #3304: Redundant `|` next to `Forum Team` on index for guests
screen shot 2018-07-05 at 19 16 53

As above, there is a redundant `|` next to the `Forum Team` link on the index when browsing as a guest.

euantorano:
Fixes #3304 The separator is moved into a language variable and the templates including the separator are updated.


Eldenroot:
Will be there a hotfix (1.8.16) or a new release with 10 issues fixed (according to the milestone?). Because because of that it would require bumt the template version to 1817.

About this PR, works fine.

euantorano:
It'll likely be 1.8.17 by the looks of it. We'll bump the template versions.

Issue #3311, opened 2018-07-05
Portal Page Login Issue


effone:
Portal page login form is missing `my_post_key` and lead to fail login attempt.

effone:
Fixed by this PR:
https://github.com/mybb/mybb/pull/3301

dvz:
https://github.com/mybb/mybb/issues/3323

Issue #3308, opened 2018-07-05
Missed out template versioning in 1.8.16 update


effone:
It's a bug. Version code still says 1800:
Code:&lt;template name="index" version="1800"&gt;

Original thread: [Missed out template versioning in 1.8.16 update](https://community.mybb.com/thread-218456.html)

effone:
Fixed by this PR:
https://github.com/mybb/mybb/pull/3302

PR #3314, opened 2018-07-05
Bump template version to 1817 for merged PRs

Eldenroot:
Bump template version to 1817 for already merged PRs - #3302 (issue #3308) and #3301 (issue #3311)


effone:
DO we need this? Its against the rectifications made in 1.8.16 so, the template should carry that version only.

euantorano:
For people who have already upgraded to 1.8.16, I think so yes. Not sure though.

Eldenroot:
Unfortunately there are a lot of users who already updated to MyBB v1.8.16 :(

effone:
Alright. Skipping this from assigning any milestone considering it an extension

dvz:
https://github.com/mybb/mybb/blob/mybb_1816/admin/inc/functions_themes.php#L147
https://github.com/mybb/mybb/blob/mybb_1816/admin/modules/style/templates.php#L1280

The upgrade script seems to be updating all master templates and the ACP to be checking for any version difference so it might be fine to go with `1816` (verification needed).

euantorano:
@dvz I think we're better safe than sorry and apply `1817` anyway, meaning we should merge this as is. Thoughts?

PR #3315, opened 2018-07-05
Fixes user stars not showing correctly

### LINKED ISSUE #3303: Userstars/groupimages are not displayed
Userstars (and groupimages as well) are not displayed for some (majority) of users... MyBB v1.8.16 (updated from 1.8.15)

Check this, Ben doesnt have any star, StefanT has :-O

https://imgur.com/J2rEN35

And this is the output in my devtools (Google Chrome)
https://imgur.com/bLknv0Z

That happens on MyBB official forum too :(

Original thread: https://community.mybb.com/thread-218440...pid1308085

Original thread: [Userstars/groupimages are not displayed](https://community.mybb.com/thread-218454.html)

euantorano:
Fixes #3303. This issue seems to have been introduced [here](https://github.com/mybb/mybb/commit/2c1e3448b14d4dfcf60d0c416e6cfa075efafc4c), so I've covered every instance of `array_merge` there.

`usergroup_displaygroup` fields can be null, so this filters out NULL fields but not empty fields so that if a users user group has stars or a group image but the display group doesn't, it will not show the stars.

Note that I haven't changed the calls to `usergroup_displaygroup` in `inc/class_session.php` as it's unclear if they need to be changed and since I'm testing on localhost I don't have any spiders online...

If anybody can confirm those calls need updating, I will update them.

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -4978,7 +4978,7 @@ if(use_xmlhttprequest == "1")
 </tr>]]></template>
 		<template name="member_profile_findposts" version="1807"><![CDATA[<br /><span class="smalltext">(<a href="search.php?action=finduser&amp;uid={$uid}">{$lang->find_posts}</a>)</span>]]></template>
 		<template name="member_profile_findthreads" version="1807"><![CDATA[<br /><span class="smalltext">(<a href="search.php?action=finduserthreads&amp;uid={$uid}">{$lang->find_threads}</a>)</span>]]></template>
-		<template name="member_profile_groupimage" version="120"><![CDATA[<img src="{$memperms['image']}" alt="{$usertitle}" title="{$usertitle}" /><br />]]></template>
+		<template name="member_profile_groupimage" version="1817"><![CDATA[<img src="{$memperms['image']}" alt="{$usertitle}" title="{$usertitle}" /><br />]]></template>
 		<template name="member_profile_modoptions" version="1812"><![CDATA[<table border="0" cellspacing="{$theme['borderwidth']}" cellpadding="{$theme['tablespace']}" width="100%" class="tborder">
 <tr>
 <td colspan="2" class="thead"><strong>{$lang->mod_options}</strong></td>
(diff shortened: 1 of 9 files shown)


euantorano:
Note: An alternative would be to check for NULL values inside `usergroup_displaygroup()`, but I haven't checked that approach to see if it breaks anything else...

euantorano:
This seems to work in my testing now, would be good if somebody else could test and confirm.

Eldenroot:
image

Works fine, thank you!

AndreRl:
Can also confirm this works fine for me.

install/resources/mybb_theme.xml
```diff
@@ -4978,7 +4978,7 @@ if(use_xmlhttprequest == "1")
 </tr>]]></template>
 		<template name="member_profile_findposts" version="1807"><![CDATA[<br /><span class="smalltext">(<a href="search.php?action=finduser&amp;uid={$uid}">{$lang->find_posts}</a>)</span>]]></template>
 		<template name="member_profile_findthreads" version="1807"><![CDATA[<br /><span class="smalltext">(<a href="search.php?action=finduserthreads&amp;uid={$uid}">{$lang->find_threads}</a>)</span>]]></template>
-		<template name="member_profile_groupimage" version="120"><![CDATA[<img src="{$memperms['image']}" alt="{$usertitle}" title="{$usertitle}" /><br />]]></template>
+		<template name="member_profile_groupimage" version="1817"><![CDATA[<img src="{$memperms['image']}" alt="{$usertitle}" title="{$usertitle}" /><br />]]></template>
```
JordanMussi: Did you forget to make the changes to the template here or is it just an accidental version bump?

euantorano:
The template was changed in 1.8.16 but the version wasn’t bumped there, this just fixes that missed bump.

euantorano:
Anybody else @mybb/developers @mybb/sqa?

AndreRl:
Can we begin to merge this @euantorano? The general response I've gotten from the community when applying the fix seems positive. Don't think we'll get much more testing for this unfortunately.

Shade-:
I can confirm this is working as it should.

PR #3325, opened 2018-07-06
Use bot name as array key

### LINKED ISSUE #3316: Spieders activity not updating on who's online
Hello!

Here i'm with a new bug on the latest version of MyBB.

I have noticed that the spiders activity isn't updating on who's online on the index page.

After refreshing the page many times,i don't see the update on the who's online.

Please check the screenshots to understand the problem:

https://community.mybb.com/attachment.php?aid=40562

https://community.mybb.com/attachment.php?aid=40563

https://community.mybb.com/attachment.php?aid=40564

Please look into it.

Thanks a lot.

Original thread: [Spieders activity not updating on who's online](https://community.mybb.com/thread-218477.html)

euantorano:
Fixes #3316 

From the array_merge PHP docs:

> Values in the input array with numeric keys will be renumbered with incrementing keys starting from zero in the result array.

Rather than using integer keys, use string keys to stop this from being a possibility.

Note that I haven't tested this at all, but believe it should work...


index.php
```diff
@@ -124,8 +124,17 @@
...
 		{
+			if($mybb->settings['wolorder'] == 'username')
+			{
+				$key = $spiders[$botkey]['name'];
+			}
+			else
+			{
+				$key = $user['time'];
+			}
+
```
effone: Why you need to sort here by key? The query result is already sorted from database. If its failing to sort there the issue should be fixed in query itself...

index.php
```diff
@@ -124,8 +124,17 @@
...
 			// The user is a search bot.
-			$onlinebots[] = format_name($spiders[$botkey]['name'], $spiders[$botkey]['usergroup']);
+			$onlinebots[$key] = format_name($spiders[$botkey]['name'], $spiders[$botkey]['usergroup']);
 			++$botcount;
```
effone: Also here why key even matter if I loop through already sorted result and assign (auto) numeric keys?

index.php
```diff
@@ -140,6 +149,16 @@
...
+	if($mybb->settings['wolorder'] == 'activity')
+	{
+		// activity ordering is DESC, username is ASC
+		krsort($onlinebots);
+	}
+	else
+	{
+		ksort($onlinebots);
+	}
+
```
effone: As stated earlier - sorting is already there in query results. So why sorting by key is again required?

euantorano:
The bot username isn’t taken from the query, but is taken from the cache instead so the query won’t use the same username.

effone:
But we are looping through $user only which is sorted. When we d'ont have uid we are using sid to confirm it as a bot.

euantorano:
Yes, but what is the username value from the query when it is a bot?

effone:
Why we are concerning about username from query? In the loop for every user we are creating $botkey and if the user id is not greater than 0 checking the user is bot or not anf getting botname using botkey from cache. What we are concerned about here are `uid` and `sid`, not `username`.

euantorano:
@effone The setting can be set to order the list by either username or by last activity. If it is set to sort by username, then it should be sorted by username, including bots... That's why we care about the bot username from the cache using the `$botkey`.

effone:
Ok. But this whole thing is not introduced by the commit mentioned elsewhere. It was there for a long while and not a mess up of 1.8.16 commits.

All I have to say if it was working before it should work fine now too.

euantorano:
Yes, it could well have been broken for a long time. It's not something I usually look at, so I have no idea if it's ever worked properly for bots.

AndreRl:
@euantorano Portal additions work fine when implemented with #3326. Do you want me to implement your changes into the fix?

effone:
Or just include @AndreRl 's changes in this PR itself rather than merging 2 separate PRs for single issue...

euantorano:
Yeah, I think I'll just merge the changes from #3326 into this PR if that's ok @AndreRl? Makes it easier to review one PR that way.

I'm busy most of the day today, but will merge it together this afternoon.

AndreRl:
That's fine Euan.

euantorano:
Merged the changes from your PR into this one @AndreRl, so I'm closing #3326 - thanks for submitting it though and providing the code needed 😄

Eldenroot:
Seems to working fine

euantorano:
Anybody else @mybb/developers @mybb/sqa?

PR #3326, opened 2018-07-07
Fix for Spiders not displaying on Portal

AndreRl:
Even though Spiders are called in the portal, they've never actually been displayed. Well at least dating back to 2015.


portal.php
```diff
@@ -293,6 +283,16 @@
 				}
 			}
 		}
+		elseif(my_strpos($user['sid'], 'bot=') !== false && $spiders[$botkey])
+		{
+			// The user is a search bot.
+			$onlinebots[] = format_name($spiders[$botkey]['name'], $spiders[$botkey]['usergroup']);
```
effone: Need to use explicit key while including bots in array, inline with @euantorano 's PR

AndreRl:
No code from this segment has been altered in this PR.

PR #3335, opened 2018-07-08
Trigger event on selectall change

### LINKED ISSUE #3241: .prop('checked') doesn't trigger change events

Eric-Jackson:
Fixes #3241

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -3773,7 +3773,7 @@ function build_users_view($view)
 	}
 
 	$built_view .= '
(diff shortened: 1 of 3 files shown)


euantorano:
I'm leaving this open until after 1.8.17 is released.

The `mybb_theme.xml` file will also need to be updated and the version for this script updated anywhere that it's referenced.

euantorano:
@Eric-Jackson Can you bump the version from `ver=1800` to `ver=1818` [here](https://github.com/mybb/mybb/blob/b0873d4d441630a5b6c91fda23d2730e18825827/admin/modules/user/users.php#L3776) please?

PR #3338, opened 2018-07-11
Fixes #3337 Unable to retrieve guest permissions while logged in

### LINKED ISSUE #3337: Unable to retrieve guest permissions while logged in
As the current `user_permissions()` function resets the uid to current user's uid if the uid is 0; it is impossible to get the user permission of a guest while logged in.

Problem instance:
https://community.mybb.com/thread-218439-post-1308791.html#pid1308791

effone:
Attempt to fix #3337 
Guest group permission fix

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1210,16 +1210,16 @@ function user_permissions($uid=null)
 {
 	global $mybb, $cache, $groupscache, $user_cache;
 
-	// Its a guest. Return the group permissions directly from cache
-	if($uid === 0)
-	{
-		return $groupscache[1];
-	}
-
 	// If no user id is specified, assume it is the current user
-	if(!$uid)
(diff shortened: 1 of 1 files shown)


inc/functions.php
```diff
@@ -1206,16 +1206,22 @@ function fetch_page_url($url, $page)
  * @param int $uid The user ID
```
euantorano: Could you update this comment to mention that passing `null` will use the current user's ID?

effone: Done.

euantorano:
As this isn’t a bug introduced by MyBB 1.8.16 I’m going to postpone merging until 1.8.18 so that we don’t delay 1.8.17 any further.

euantorano:
Appears to work fine for me.

PR #3342, opened 2018-07-15
Fixes #3341 Thread Subscription Bug

### LINKED ISSUE #3341: Thread subscription bug
When subscribing to a specific thread, after posting on it the user then automatically becomes unsubscribed.

Original thread: [Thread subscription bug](https://community.mybb.com/thread-218661.html)

effone:
Attempt to fix #3341 
Subscription bug while reply

- Fixed subscription assignment on quickreply action
- Created `get_subscription_method()` function, optimized code.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3456,7 +3456,7 @@ function get_subscription_method($tid = 0, $postoptions = array())
 {
 	global $mybb;
 
-	$subscription_methods = ['dont', 'none', 'email', 'pm']; // Define methods
+	$subscription_methods = array('dont', 'none', 'email', 'pm'); // Define methods
 	$subscription_method = (int)$mybb->user['subscriptionmethod']; // Set user default
 
 	// If no user default method available then reset method
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3481,7 +3481,7 @@ function get_subscription_method($tid = 0, $postoptions = array())
 	{
 		global $db;
 
-		$query = $db->simple_select("threadsubscriptions", "tid, notification", "tid='".$tid."' AND uid='".$mybb->user['uid']."'", array('limit' => 1));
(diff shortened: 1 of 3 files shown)


newreply.php
```diff
@@ -126,21 +126,10 @@
...
-	}
-	else if($mybb->user['subscriptionmethod'] == 2)
-	{
-		$mybb->input['postoptions']['subscriptionmethod'] = "email";
-	}
-	else if($mybb->user['subscriptionmethod'] == 3)
-	{
-		$mybb->input['postoptions']['subscriptionmethod'] = "pm";
-	}
+	$mybb->input['postoptions']['subscriptionmethod'] = get_subscription_method($mybb->get_input('pid', MyBB::INPUT_INT));
```
Shade-: Haven't set up a local test yet, but from bare reading the code this should be `$mybb->get_input('tid', MyBB::INPUT_INT)` as the new `get_subscription_method` definition requires `tid`, not `pid`.

effone: Right. Corrected.

inc/functions.php
```diff
@@ -3446,6 +3446,53 @@ function build_mycode_inserter($bind="message", $smilies = true)
...
+	if(isset($postoptions['subscriptionmethod']))
+	{
+		$method = trim($postoptions['subscriptionmethod']);
+		return (in_array($method, $subscription_methods)) ? $method : $subscription_methods[0];
+	}
+	else
+	{
+		global $db;
+
+		$query = $db->simple_select("threadsubscriptions", "tid, notification", "tid='".$tid."' AND uid='".$mybb->user['uid']."'", array('limit' => 1));
```
euantorano: Should probably cast `$tid` here just to be sure it is safe - function could be used elsewhere or by plugins in the future.

effone: Done!

PR #3343, opened 2018-07-15
Working on Fixes #3339 Counting of non-countable objects

### LINKED ISSUE #3339: Counting of non-countable objects in ACP's Banning results in Warning on PHP 7.2
The `$banned_groups` array is being assigned to within a loop that can be executed 0 times, which results in PHP Warnings being emitted when performing operations such as `count()` later on.

https://github.com/mybb/mybb/blob/mybb_1816/admin/modules/user/banning.php#L45

Reproducible when visiting the _Users & Groups → Banning_ page with no groups selected with _Yes, this is a banned group_ on PHP 7.2 or higher.

reported in https://community.mybb.com/thread-218637.html

effone:


showthread.php
```diff
@@ -897,7 +897,11 @@
             WHERE p.tid='$tid'
             $visible
             ORDER BY p.dateline
-        ");
+		");
```
Shade-: Indentation here seems a bit off.

inc/functions.php
```diff
@@ -4337,12 +4336,10 @@ function build_breadcrumb()
 				eval("\$nav .= \"".$templates->get("nav_bit")."\";");
 			}
 		}
+		$navsize = count($navbits);
+		$navbit = $navbits[$navsize-1];
 	}
 
-	$activesep = '';
```
Shade-: This should be retained as otherwise an undefined notice might be prompted due to `$activesep` being defined within a condition later on.

Shade-:
Apart for the mentioned little issues, this looks good to me.

inc/functions.php
effone: Isn't it a duplicate assignment as this has already been declared here?:
https://github.com/mybb/mybb/blob/feature/inc/functions.php#L4297

Shade-: Wooops, good spot. I used GitHub's diff and didn't see the whole function, my bad 😄

showthread.php
effone: Fixed now.

dvz:
Is `is_array()` the best approach, as opposed to making types consistent (creating undefined arrays, changing `''` to `array()`, etc.)? That's what we should arrive at after a rewrite.

Eldenroot:
Good question, others MyBB devs?

effone:
I know performance wise `is_array` is expensive but is it the alternative of it is set it and let it go? Is it even possible in every case to avoid the check? I am not sure...

euantorano:
I think `is_array()` is okay in the cases I've looked at here, we can always clean up and improve later if we notice any whilst working towards 1.9/1.10/etc.

PR #3348, opened 2018-07-18
prevent array error message popping up

partycodingtiger:
$collapse can be null and then in_array will throw an error message. to prevent this, ensure a proper array is used at all times.

I came upon this issue after my host (forcefully) changed from PHP 5.8.2 to 7.2.7

### DIFF
--- a/index.php
+++ b/index.php
@@ -360,7 +360,7 @@ if(($mybb->settings['showwol'] != 0 && $mybb->usergroup['canviewonline'] != 0) |
 		$stats = $cache->read('stats');
 	}
 	
-	$expaltext = (in_array("boardstats", $collapse)) ? "[+]" : "[-]";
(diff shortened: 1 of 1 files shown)


euantorano:
Collapse cannot be null: https://github.com/mybb/mybb/blob/feature/global.php#L1142

I'm assuming you've upgraded to MyBB 1.8.17 from MyBB 1.8.17 using the changed files package? if so, that won't work. if you are more than one version number behind, you have to use the full package to upgrade - changed files only contain files changed in one minor point release.

partycodingtiger:
@euantorano 
Actually, I upgraded from 1.8.16 to 1.8.17 but it appears my global.php file was outdated beforehand as several lines are missing. Gotta look into what happened there.


Sorry for the inconvenience.

PR #3352, opened 2018-07-21
Fixes #3340 Embedded videos have extraneous vertical space above and …

### LINKED ISSUE #3340: Embedded videos have extraneous vertical space above and below
Members of the MyBB-powered forum I help administer have complained that embedded videos have extra spaces inserted above and below them after the post is submitted. I had noticed this issue too. I fixed it on our forum by adding second (value: 1) and third (value: 0) parameters to the call &#36;templates-&gt;get("video_{&#36;video}_embed") in the eval() on the second-to-last line of postParser::mycode_parse_video() in inc/class_parser.php.

It's the third parameter that fixes the problem (the second is simply the  default), by omitting comments from the eval()d template. When comments are inserted, along with them are inserted a newline above and a newline below the embedded video, and these are ultimately converted into &lt;br&gt;s by the call to nl2br() in postParser::parse_message(), resulting in the complained-about extra spaces.

These extra parameters are present in all other calls to &#36;templates-&gt;get() in the parser, so perhaps their omission here was just a minor oversight.

Original thread: [Embedded videos have extraneous vertical space above and below](https://community.mybb.com/thread-218422.html)

lairdshaw:
…below

Fixes #3340.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1495,7 +1495,7 @@ class postParser
 
 		$id = htmlspecialchars_uni($id);
 
-		eval("\$video_code = \"".$templates->get("video_{$video}_embed")."\";");
+		eval("\$video_code = \"".$templates->get("video_{$video}_embed", 1, 0)."\";");
 		return $video_code;
 	}
 


kawaii:
@lairdshaw I've tested this and whilst the patch does indeed work I believe there was some debate over whether or not there should visually be some kind of padding between the video and any text in the post.

For reference, here's how the following post would be rendered on the default 1.8 theme currently, with the patch applied:
```bbcode
Hello world!
[video=youtube]https://www.youtube.com/watch?v=yPuI4l0jK7s[/video]
Goodnight world!
```
image
Perhaps the PR should be extended to wrap all video elements inside a common `<div>` or something?

@mybb/designers any input from you guys?

euantorano:
A common wrapper `<div>` for all videos would make it much easier to add adding if needed for definite. At the moment with this PR there is only a single `<br> before and after the embeds, meaning videos and text render right next to each other, which doesn't look great.

lairdshaw:
The problem I see with wrapping embedded videos in a div (especially with added padding) is twofold. Firstly, it prevents users from choosing inline embedding, which this PR supports:

inline-embedded-video-in-mybb

Whilst I can't think of a reason why a user might want to do this, I'm at even more of a loss as to why it should be forbidden.

Secondly, without updates to the visual editor, and especially if padding is added, it continues the problem of the final rendering being inconsistent with what the user sees in the visual editor.

lairdshaw:
There's also a third problem, inconsistency: images are currently rendered as embedded videos are in this PR (inline images are possible; no padding is added to images). Adding a div and/or padding to embedded videos would make their rendering inconsistent with that of images.

Eric-Jackson:
I agree with the points of @lairdshaw. Users can add extra vertical spacing where necessary in their posts. Looks good!

PR #3357, opened 2018-07-22
Fixes #3144 MyCode affects signature length Issue

### LINKED ISSUE #3144: MyCode affects signature length Issue

effone:
Attempt to fix #3144 
Signature length count considering only visible characters

After several discussions and [attempts](https://github.com/mybb/mybb/pull/3179) I have come up with this simplest solution to consider only visible characters to take in account while counting. Tested and the counts appearing as desired.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -765,7 +765,7 @@ class postParser
 		}
 		else
 		{
-			$replace = "\n{$lang->quote}\n--\n$1\n--\n";
+			$replace = empty($this->options['signature_parse']) ? "\n{$lang->quote}\n--\n$1\n--\n" : "$1";
 			$replace_callback = array($this, 'mycode_parse_post_quotes_callback2');
 		}
 
@@ -923,7 +923,7 @@ class postParser
 
 		if($text_only == true)
 		{
-			return "\n{$lang->code}\n--\n{$code}\n--\n";
+			return empty($this->options['signature_parse']) ? "\n{$lang->code}\n--\n$code\n--\n" : $code;
 		}
 
 		// Clean the string before parsing.
@@ -971,7 +971,7 @@ class postParser
 
 		if($text_only == true)
 		{
-			return "\n{$lang->php_code}\n--\n$str\n--\n";
+			return empty($this->options['signature_parse']) ? "\n{$lang->php_code}\n--\n$str\n--\n" : $str;
 		}
 
 		// Clean the string before parsing except tab spaces.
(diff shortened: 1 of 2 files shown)


Shade-:
Unfortunately, this is still not working perfectly. Tested using:

```[quote]Twenty characters 20[/quote]```

Settings: limit 20, MyCode count off

It's converted to:

```QuoteTwenty characters 20```

Which gives me a "signature too long, please remove 5 characters" error.

**Comment by effone** [comment]
Where it is converted? If you are talking about backend then all the spaces are also stripped out before counting. If it is about the starting quote then that is visible, hence covers the scope of countable.

I am sorry but what exactly the issue. I am not sure I get you perfectly.

**Comment by Shade-** [comment]
The parsed signature contains extra characters for certain MyCodes (such as quotes) which are not accounted in strip_tags. Hence why it still fails under certain circumstances like the above.

**Comment by euantorano** [comment]
The text “Quote” is being counted towards the limit. I don’t think it should, but it might be hard to remove in custom MyCode.

**Comment by effone** [comment]
The 'quote' word is visible in signature. Why not to count that? Its visible, it should be counted, no?

**Comment by Shade-** [comment]
Yeah well, it's visible but it should not be counted IMHO. It's not an user-generated text, it's part of the MyCode.

**Comment by effone** [comment]
Alright. Modified. Is it okay now?

File: inc/class_parser.php [hunk 1]
```diff
@@ -1746,6 +1746,13 @@ function text_parse_message($message, $options=array())
 		{
 			$this->options = $options;
 		}
+		else
+		{
+			foreach($options as $option_name => $option_value)
```
euantorano: Couldn't this just be `$this->options = array_merge($this->options, $options);`?

effone: I tried this first but it malfunctioned and didn't append the new key-value pair or override any old. That was odd but that happened repeatedly. So I placed it this way and it worked.

euantorano: Weird, this way works too anyway 😄

effone:
Foreach is slow but not too costly performance wise anyway, 
https://gist.github.com/Ocramius/8399625

considering the fact that the option update will not happen very frequent I will not go for something well obvious but failed during test for any unknown weird reason.

PR #3358, opened 2018-07-22
Fixes #3192 Avatar dimension separator consistency

effone:
Attempt to fix #3192 

Added flexibility in defining and usage of avatar dimensions:
- Added validation while setting dimensions through ACP
- Modified ACP settings description to define both separator usage is acceptable.
- Modified width - height separation process (used `preg_split()` in place of `explode()`) to generate values when any of the two separators is used.

PR #3360, opened 2018-07-23
Fixes #3359 Settings `gid` is blank in ACP settings search page

### LINKED ISSUE #3359: Settings `gid` is blank in ACP settings search page
While searching for settings through ACP the matched settings are grouped perfectly but generated form's `gid` field is blank.

`<input type="hidden" name="gid" value="">`

Reason behind is it is designed to have separate forms per `gid` but actually is combined into one.

This may lead to failing / skipping of operations in ACP settings module which are triggered based on `gid`.

effone:
Attempt to fix #3359 

- Individual form generation per settings group against search
- Enabling form-wise settings `gid` value.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -1158,10 +1158,6 @@ if($mybb->input['action'] == "change")
 			$page->add_breadcrumb_item($lang->settings_search);
 			$page->output_header($lang->board_settings." - {$lang->settings_search}");
 		}
-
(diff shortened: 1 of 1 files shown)


euantorano:
This looks good from looking at the code, I just need to test it and confirm, which I will do tonight.

Eldenroot:
Tested together with #3358 , seems working fine.

PR #3361, opened 2018-07-23
Fixes #2659 Double recipients with private messages

### LINKED ISSUE #2659: Double recipients with private messages
How to reproduce with MyBB 1.8.10 default theme and language.

Write a private message to three or more members, one member receives and read the message and do Reply All, now if the member click on "preview" recipients have doubled!

effone:
Attempt to fix #2659 
Duplicate PM Receipient Fix

### DIFF
--- a/private.php
+++ b/private.php
@@ -705,8 +705,8 @@ if($mybb->input['action'] == "send")
 		{
 			$optionschecked['readreceipt'] = 'checked="checked"';
 		}
-		$to = htmlspecialchars_uni(implode(', ',array_unique(array_map('trim', explode(',', $mybb->get_input('to'))))));
-		$bcc = htmlspecialchars_uni(implode(', ',array_unique(array_map('trim', explode(',', $mybb->get_input('bcc'))))));
+		$to = htmlspecialchars_uni(implode(', ', array_unique(array_map('trim', explode(',', $mybb->get_input('to'))))));
+		$bcc = htmlspecialchars_uni(implode(', ', array_unique(array_map('trim', explode(',', $mybb->get_input('bcc'))))));
 	}
 
 	$preview = '';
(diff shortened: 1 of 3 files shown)


effone:
I don't know if I did correctly but I have investigated every possible backend output and it appeared like backend is working perfectly. This repeated name is generated by js only while malprocessing a space character. May be I am wrong.

Shade-:
You're right – even myself proposed this fix back in the days. However, I feel like an array_unique on the backend side is still necessary just to filter out any mistakes on the front end side.

effone:
Is it okay now?

xmlhttp.php
```diff
@@ -281,6 +281,9 @@
 		}
 	}
 
+	// Filter out duplicates, if any, from recursive array data
+	$data = array_intersect_key($data, array_unique(array_map('serialize', $data)));
```
euantorano: This should use `my_serialize` rather than plain old `serialize`: https://github.com/mybb/mybb/blob/feature/inc/functions.php#L2289

That way things like multibyte strings and such are taken care of, which could be possible for user names.

effone: Done.

private.php
```diff
@@ -704,8 +705,8 @@
 		{
 			$optionschecked['readreceipt'] = 'checked="checked"';
 		}
-		$to = htmlspecialchars_uni($mybb->get_input('to'));
-		$bcc = htmlspecialchars_uni($mybb->get_input('bcc'));
+		$to = htmlspecialchars_uni(implode(', ',array_unique(array_map('trim', explode(',', $mybb->get_input('to'))))));
```
Shade-: I know I might seem pedant here, but there should be a space between `implode(', ',` and `array_unique` :)

effone: Done.

xmlhttp.php
```diff
@@ -281,6 +281,9 @@
 		}
 	}
 
+	// Filter out duplicates, if any, from recursive array data
+	$data = array_intersect_key($data, array_unique(array_map('my_serialize', $data)));
```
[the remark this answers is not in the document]
effone: For any impossible duplicate returned from database 👯‍♂️ 
I'll remove it ...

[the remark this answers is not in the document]
Shade-: 😁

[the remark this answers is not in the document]
effone: Removed.

PR #3371, opened 2018-07-30
Fixed send_mail_queue call

zygzagZ:
In inc/init.php:243 there is function call that registers send_mail_queue to be run at shutdown.
`add_shutdown('send_mail_queue');`
But just after include init.php in global.php, shutdown_functions is erased, not allowing send_mail_queue to  run.
`// Load main MyBB core file which begins all of the magic
require_once $working_dir.'/inc/init.php';

$shutdown_queries = $shutdown_functions = array();`


euantorano:
We’ve had a few reports of mails not sending... this could be the cause :P

zygzagZ:
Mails are still sent when running task.php and some other files. Nonetheless this was annoying enough for me to debug it ;)

euantorano:
Yeah, it’s definitely been an annoying one. Very glad to have a fix at last.

PR #3180, opened 2018-04-30
Fixes #3139 Attachment: a bug and a mod to allow multi-attachment selection

Sama34:
Fixes #3139 Attachment: a bug and a mod to allow multi-attachment selection

Please review.

PR #3373, opened 2018-08-01
Fixes #3139 Attachment: a bug and a mod to allow multi-attachment selection

### LINKED ISSUE #3139: Attachment: a bug and a mod to allow multi-attachment selection
First, the bug (1.8.14).  The situation: one has uploaded an attachment for a new/edit post and one has selected another file to be uploaded but "Add Attachment to upload the second attachment has not been clicked yet.  If at that point, one clicks on "Remove Attachment", the selected attachment will be uploaded. 

Now an improvement.  I was interested in allowing multi-attachments adds.  I found a plugin that added that capability, but it also added a number of other features.  I found the plugin unusable because of various bugs, many of which were related to features other than supporting multifile upload.  I implemented the following fairly straight-forward changes to allow the multi-file uploads.  If anyone wants to try it, and provide feedback, at some point I may turn it into a plugin.

Uploaded files come into the *.php code via the variables "&#36;_FILES['attachment'][&#36;field]", where &#36;field has various values like 'name' and 'size' for various aspects of the uploaded file. For multi-file uploads, those variables become arrays instead of singular values.  Where those variables are used, the following code segment must be inserted.

Code:&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;fields = array ('name', 'type', 'tmp_name', 'error', 'size');
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;if (is_array(&#36;_FILES['attachment']['name']))
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;{
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;// Already in multi-attachment array format
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;attachments = &#36;_FILES;
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;}
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;else
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;{
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;// Convert original-style non-array &#36;_FILES['attachment'][&#36;field] to array format in &#36;attachments
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;attachments = array('attachment' =&gt; array());
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;foreach (&#36;fields as &#36;field)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;attachments['attachment'][&#36;field] = array(&#36;_FILES['attachment'][&#36;field]);
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;}
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;foreach (&#36;attachments['attachment']['name'] as &#36;key =&gt; &#36;name)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;{
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;// Convert array &#36;attachments['attachment'][&#36;field][&#36;key] to non-array format in &#36;FILE
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;    &#36;FILE = array('attachment' =&gt; array());&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;    foreach (&#36;fields as &#36;field)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&#36;FILE['attachment'][&#36;field] = &#36;attachments['attachment'][&#36;field][&#36;key];
Step 1: The affected files are newthread.php, editpost.php and newreply.php, around line 225, give or take, in each.  In each file, insert the code before the first "if" statements that process the attachment (and add another closing curly brace for the added "foreach" statement). 

Step 2: In that "if" staement, chnage all references to "&#36;_FILES" to "&#36;FILE".

Example: in newthread.php, the new code segment would look like this:
[attachment=40094]

Step 4: Change the "input" tag in template "post_attachments_new" to have name "attachment[]" (braces added) and the keyword "multiple".
Code:&lt;input type="file" name="attachment[]" size="30" class="fileupload" multiple/&gt;
Older browsers may still be limited to single attachment uploads but it still should work for that.

Original thread: [Attachment: a bug and a mod to allow multi-attachment selection](https://community.mybb.com/thread-216548.html)

zygzagZ:
Polished version of PR #3180

Fixes #3139 

I added proper parsing of php limits and javascript to post.js


Eldenroot:
Great! Good job, thank you. I will test ASAP

Eldenroot:
Please bump template versions to 1818 for all templates which are affected.

jscripts/post.js
```diff
@@ -116,7 +117,44 @@ var Post = {
...
+		fileUpload.attr("multiple","multiple");
+		$('form').on('submit', Post.checkAttachments);
+	},
+
+	checkAttachments: function()
+	{
+		var files = $("input[type='file']");
+		var file = files.get(0);
+		if (!file) 
+			return true;
```
effone: Please put this line in curly braces (`{` `}`) just to maintain coding style consistency.

inc/functions_upload.php
```diff
@@ -679,6 +679,81 @@ function upload_attachment($attachment, $update_attachment=false)
...
 }
 
+/**
+ * Process adding attachment(s) when the "Add Attachment" button is pressed.
+ *
+ * @param int $pid The ID of the post.
+ * @param array $forumpermission The permissions for the forum.
+ * @param string $attachwhere Search string "pid='$pid'" or "posthash='".$db->escape_string($mybb->get_input('posthash'))."'"
+ * @param string $action Where called from: "newthread", "newreply", or "editpost"
+ */
```
effone: Please indicate `@return` type as well.

inc/functions_upload.php
```diff
@@ -679,6 +679,81 @@ function upload_attachment($attachment, $update_attachment=false)
...
+
+	$ret = array();
+
+	if($forumpermissions['canpostattachments'])
+	{
+		$attachments = array();
+		$fields = array ('name', 'type', 'tmp_name', 'error', 'size');
+
+		$total = count($_FILES['attachments']['name']);
+
```
effone: Isn't an empty check required here?

install/resources/mybb_theme.xml
```diff
@@ -3617,21 +3617,14 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
...
-</table>
-<br />]]></template>
 		<template name="error_inline" version="120"><![CDATA[<div class="error">
 	<p><em>{$title}</em></p>
 	<ul>
 		{$errorlist}
 	</ul>
 </div>
 <br />]]></template>
+		<template name="error_inline_item" version="1816"><![CDATA[<li>{$error}</li>]]></template>
```
effone: Bump template version to `1818`.

inc/functions_post.php
```diff
@@ -1021,3 +1021,28 @@ function get_post_attachments($id, &$post)
...
+/**
+ * Returns bytes count from human readable string
+ * Used to parse ini_get human-readable values to int
+ *
+ * @param string $val Human-readable value
+ */
+function return_bytes($val) {
+    $val = trim($val);
+    if ($val == "")
+    	return 0;
```
effone: Use curly braces....
Also this check is not required if a default is set in switch-case.

inc/functions_post.php
```diff
@@ -1021,3 +1021,28 @@ function get_post_attachments($id, &$post)
...
+
+    $last = strtolower($val[strlen($val)-1]);
+    switch($last) {
+        // The 'G' modifier is available since PHP 5.1.0
+        case 'g':
+            $val *= 1024;
+        case 'm':
+            $val *= 1024;
+        case 'k':
+            $val *= 1024;
```
effone:
```
        case 'g':
        case 'm':
        case 'k':
            $val *= 1024;
        default:
            $val = 0;
```
Better?

inc/functions_upload.php
zygzagZ: In PHP 7.2.7 everything's alright, it evaluates to NULL giving $total = 0
I do not know what happens in PHP 5 though and how it should be written to be crossplatform.

inc/functions_post.php
zygzagZ: Without this condition, given empty string it would attempt to get string offset -1, giving php notice.
The function is copy-pasted from here: http://php.net/manual/en/function.ini-get.php

inc/functions_post.php
zygzagZ: No, as 'g' is supposed to multiply by 2^24, 'm' by 2^16 and 'k; by 2^8

zygzagZ:
I hope everything is OK now. 
Btw: Is it better to allow no-script people to use multiple attachments without client-side size and count check, or to limit them to one attachment per request? In #3180 someone suggested second approach, but I like the first one more...

editpost.php
```diff
@@ -938,6 +943,17 @@
 		}
 	}
 
+	$php_max_upload_filesize = return_bytes(ini_get('max_upload_filesize'));
+	$php_post_max_size = return_bytes(ini_get('post_max_size'));
+
+	if ($php_max_upload_filesize != 0 && $php_post_max_size != 0)
```
euantorano: Our coding style enforces braces (`{`, `}`) for all statements, would be great if you could add them here 😄

inc/functions_post.php
```diff
@@ -1021,3 +1021,30 @@ function get_post_attachments($id, &$post)
...
+ * @param string $val Human-readable value
+ */
+function return_bytes($val) {
+	$val = trim($val);
+	if ($val == "")
+	{
+		return 0;
+	}
+
+	$last = strtolower($val[strlen($val)-1]);
```
euantorano: These should probably be `my_strtolower` and `my_strlen` to handle any possible multibyte strings.

inc/functions_upload.php
```diff
@@ -679,6 +679,83 @@ function upload_attachment($attachment, $update_attachment=false)
...
+			}
+		}
+
+		foreach($attachments as $FILE)
+		{
+			if(!empty($FILE['name']) && !empty($FILE['type']))
+			{
+				if($FILE['size'] > 0)
+				{
+					$query = $db->simple_select("attachments", "aid", "filename='".$db->escape_string($FILE['name'])."' AND {$attachwhere}");
```
euantorano: Running a query per file doesn't seem like the best approach to me - would be nice to have one query with `WHERE filename IN (filenames...)` and then you could build an array of `[filename => aid]`.

newreply.php
```diff
@@ -1470,6 +1446,17 @@
 		}
 	}
 
+	$php_max_upload_filesize = return_bytes(ini_get('max_upload_filesize'));
+	$php_post_max_size = return_bytes(ini_get('post_max_size'));
+
+	if ($php_max_upload_filesize != 0 && $php_post_max_size != 0)
```
euantorano: Again, would be nice to add braces to this condition.

newthread.php
```diff
@@ -1132,6 +1106,17 @@
 		}
 	}
 
+	$php_max_upload_filesize = return_bytes(ini_get('max_upload_filesize'));
+	$php_post_max_size = return_bytes(ini_get('post_max_size'));
+
+	if ($php_max_upload_filesize != 0 && $php_post_max_size != 0)
```
euantorano: Braces again.

euantorano:
Regarding users with JavaScript disabled, it would better to allow them to add multiple attachments IMO then error server side if there's a problem.

inc/functions_post.php
effone: Not if you declare default value which will pass 0 always other than a match.

inc/functions_post.php
zygzagZ: I do not think so... This is only useful to parse possible values of ini_get, and they are [0-9]+[kmg] - no use for user-specified input, thus no place for multibyte strings

editpost.php
zygzagZ: Sorry for all the lacking braces, I am used to other coding style :P

inc/functions_post.php
euantorano: Ah, fair enough. Should be fine as is then!

inc/functions_upload.php
zygzagZ: I fixed it, I doubt however if it is useful, as one query per attachment is little compared to 4-5 queries generated by `upload_attachment`

euantorano: Nice one, thanks. Anywhere where we can reduce the number of queries easily is worth doing IMO.

I'll give this a proper test tomorrow after work and hopefully merge. Thanks for all the hard work!

Eldenroot:
> Regarding users with JavaScript disabled, it would be better to allow them to add multiple attachments IMO then error server side if there's a problem.

Yes, it would be nice, @zygzagZ can you add this last missing thing? Fallback for this group of users, thank you.

I am testing this PR, seems to be working fine but I will need few more days (I will have more time during weekend).

zygzagZ:
It is already there in the last commit ;)

Eldenroot:
I am sorry, overlooked that

zygzagZ:
Anything else to change in this PR? All requested changes were fixed

Eldenroot:
I have been testing this PR for last 4 days, seems to be fine. Anyway @euantorano needs to approve and merge

euantorano:
I'm doing some testing of this PR now. If all goes to plan, I will merge it today - thanks for the contribution and sorry it's taken so long!

zygzagZ:
Out of curiosity: how do you test things? Do you put the code live on your forum? Maybe some better approach?

euantorano:
@zygzagZ I use [the awesome Dockerfile](https://github.com/mybb/docker) that @kawaii wrote! I test each PR and issue report in a new container to verify that everything seems to work and that the issue that's meant to be fixed is actually fixed.

If a patch is a small one, like #3379 was (a single template change) then I try to batch those together so I can test multiple changes at the same time.

euantorano:
Uh oh, looks like we have some conflicts here due to another PR - do you have time to quickly fix them, or would you like me to?

inc/functions_upload.php
```diff
@@ -679,6 +679,100 @@ function upload_attachment($attachment, $update_attachment=false)
...
+			$FILE = $attachments[$i];
+			if(!empty($FILE['name']) && !empty($FILE['type']) && $FILE['size'] > 0)
+			{
+				$filenames .= $delim . "'" . $db->escape_string($FILE['name']) . "'";
+				$delim = ",";
+			}
+
+		}
+
+		$query = $db->simple_select("attachments", "filename, aid", "{$attachwhere} AND `filename` IN (".$filenames.")");
```
euantorano: Backticks (\`) around `filename` break Postgres with the following error:

```
0 - ERROR: operator does not exist: ` character varying LINE 1: ...E posthash='30d9dd80a5391cdb895b43fd4d844326' AND `filename`... ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

The name `filename` isn't a reserved word in mySQL, Postgres or SQLite as far as I can tell so I'm just going to remove these backpacks and continue testing.

zygzagZ:
I can do this in 3h from now

euantorano:
This all seems to work for me otherwise, nice work. I'll try to fix the conflicts my self if GitHub will let me 😄

euantorano:
Looks like there may be a bug that can occur if `$filenames` is empty here: https://github.com/mybb/mybb/pull/3373/files#diff-860768c521d9895f2eb32600cb1bcf03R723

SHould be easy enough to fix by adding an `empty()` call before running the query and exiting the function early. Thoughts @zygzagZ?

Reference: https://community.mybb.com/thread-219241-post-1311615.html#pid1311615

Eldenroot:
Yes

euantorano:
Yeah another PR is the easiest way, thanks.

Eldenroot:
Now we need to add "select boxes" to remove multiple attachments at once :)

@zygzagZ

zygzagZ:
Wouldn't it be enough just for the "remove attachment" popup not to reload the page? User would have to click "remove" at each attachment and then confirm it in popup, but the page doesn't reload and user doesn't lose his focus, page doesn't scroll to the top and everything is nearly as quick as with checkboxes.

Eldenroot:
@zygzagZ - anything will be better than how it is now :)

zygzagZ:
Check it out #3439

Issue #3215, opened 2018-05-16
Add SameSite=lax flag to authentication cookies


dvz:
Attach the [SameSite](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7) flag set to `lax` for authentication cookies (`mybbuser` and `adminsid`).

The [`my_setcookie()`](https://github.com/mybb/mybb/blob/mybb_1815/inc/functions.php#L1863) function could be extended with additional parameter that would depend on a new setting in the ACP (for cases when cross-origin POST requests are expected).

https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/
https://caniuse.com/#search=samesite

PR #3374, opened 2018-08-02
change the my_post_key regularly

frostschutz:
MyBB uses "my_post_key" to prevent forged requests. Ideally this token should be unique for each request, or unique for each session ([see on wikipedia](https://en.wikipedia.org/wiki/Cross-site_request_forgery#Synchronizer_token_pattern)) but in MyBB, the token never changes, ever.

At the same time it's easy for the my_post_key to be leaked. It's not only used in page sourcecode, as a hidden form input field, but also openly in links (Mark Forums Read, Star Rating, etc.) so a simple Copy & Paste operation may inadvertently include it, it's in the browser history, etc.

So here is an attempt to change the my_post_key regularly.

It adds a time-based element to the hash, causing a new code to be generated every 3 hours. By generating older codes in the verification step (It's an older code, sir, but it checks out...) each code may be valid for up to 12 hours. After that, the code is useless.

In terms of usability, this means if you send a post request with a tab that had been open for half a day, it will no longer work since the my_post_key of the old tab had already expired. You have to reload to get a new key.


euantorano:
Yep, we've been meaning to improve this, I think @dvz had some ideas about this and things like the `loginkey`. Thanks for the patch!

frostschutz:
I've considered rotating the loginkey instead, but that would have to be a much larger timeframe, or it would make "persistent logins" pointless. So like instead of rotate every 3 hours, you'd have to rotate every 3 weeks, or even 3 months.

I also tried adding $session->sid and $adminsession['sid'] into the my_post_key hash, but that breaks posting as a guest who does not allow cookies. ;)

It would be good to have a key that is bound to a purpose. Actually MyBB uses this in one instance: the logoutkey. It's like the my_post_key, but it can only be used to logout, nothing else. The my_post_key on the other hand is global and used for everything, everywhere, and it's even the same for the frontend and for the ACP.

But generating unique keys for everything (a key to delete post 123, a different key to edit post 456) would require slightly extensive changes. Might actually be possible with the 1.9 template system though...

But for now, I'm mainly worried about leaking keys that are still valid weeks later. Even if the fix might annoy some tab hoarders...

euantorano:
We're definitely looking to overhaul and improve session handling and such in 1.9/1.10/future major releases. We want to switch password hashing schemes to, for fairly obvious reasons. We're planning an ACP rewrite too for 1.10, which should see a change to how admin sessions and admin post keys are handled which can then later be rolled into the front end too.

dvz:
In the end the anti-CSRF tokens may be implemented in the form of HMAC codes (so they can be pre-generated for every action separately without having to worry about short timeouts or tab overriding) or stored in the database, giving us (and, conversely, to administrators) more flexibility when it comes to enforcing expiration in case of security accidents.

Similarly, we should arrive at registered session keys (multiple per user with information on devices they're tied to) stored in the form of plaintext selector and hashed verifier; these could be terminated automatically basing on time, or on request (user's own or the administrator's).

The [SameSite](https://github.com/mybb/mybb/issues/3215) flag should alleviate some of the risk associated with XSS attacks taking advantage of leaked anti-CSRF tokens - while the ACP should be mostly covered by this, there may be places where write actions are performed via GET requests in the frontend, which should be addressed as the code is successively rewritten.

Azareal:
Beware of the BREACH attack exploit, I'm not sure if it's still exploitable, but it works by exploiting the fact that HTTP compression tends to shrink the size of a request when there are multiple identical strings.

frostschutz:
in case anyone is using this patch

for key rotation to work, all key validation has to go through verify_post_check

this is one instance I overlooked where key is checked directly instead, so older keys fail:

https://github.com/mybb/mybb/blob/4f6ff6f115e02b077b0e36edfcecbe52ab44e6de/admin/index.php#L809

change to verify_post_check (also used elsewhere in the acp and even in the same file)

euantorano:
@frostschutz I'd consider the usage of a manual chekcing in the ACP index a bug, would you mind opening a PR to patch it? If you don't get a chance, I can do so myself on Monday.

Eldenroot:
@dvz @euantorano @frostschutz - quite old PR, is possible to finish somehow and get it to merge?

Ben-MyBB:
Thanks for your contribution @frostschutz but this has been replaced by #4022 to fix merge conflicts and other improvements.

PR #3379, opened 2018-08-05
Fixes #3378 - Add a check box "remember_me"

### LINKED ISSUE #3378: Check box "remember_me" missing from No Permissions Login
In the template 'error_nopermission', the checkbox "remember_me" is missing since v1.40 !!!

I did not look for the older versions ...
Maybe it's missing since the beginning of mybb?

To fix this, insert the following code :
Code:
`snipped`

after this row :
Code:
`snipped`

to obtain :
Code:
`snipped`

For next version please update /install/resources /mybb_theme.xml.

Thanks

Original Thread: [Check box "remember_me" missing](https://community.mybb.com/thread-219117.html)

Eldenroot:
#3378

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3656,6 +3656,9 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 <td class="trow2"><input type="password" class="textbox" name="password" tabindex="2" /></td>
 </tr>
 <tr>
+<td class="trow1" colspan="2" align="center"><label title="{$lang->remember_me_desc}"><input type="checkbox" class="checkbox" name="remember" checked="checked" value="yes" /> {$lang->remember_me}</label></td>
+</tr>
+<tr>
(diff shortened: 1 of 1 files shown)


AndreRl:
Good job! I was wondering about the positioning of it. It looks a bit weird as is.

Eldenroot:
OK, what do you suggest, no problem to change it.

AndreRl:
Maybe aligned to the right? But when you look at the error no permission login and member login they’re actually really different. We should change them to be the same for consistency. @effone what do you think?

Eldenroot:
I will check and try to find the beat way. Feel free to share your ideas :)

euantorano:
In my opinion we can leave the layout as it is for now. The layout of these pages will change in 1.9 and if this looks the same as existing 1.8 pages it should be fine 😄

justinsoltesz:
Center aligned matches the full login page, so it's good enough for now.

As you said, layout will change in 1.9.

euantorano:
Awesome, thanks for the quick response Justin. I'll merge this now in that case!

PR #3386, opened 2018-08-11
Working on #2972 Template Conversion to Twig Format (nav)

Shade-:
- [x] nav
- [x] nav_bit
- [x] nav_bit_active
- [x] nav_dropdown
- [x] nav_sep
- [x] nav_sep_active

These were missing.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3531,59 +3531,28 @@ function fix_mktime($format, $year)
  */
 function build_breadcrumb()
 {
-    global $nav, $navbits, $templates, $theme, $lang, $mybb;
-
-    eval("\$navsep = \"".$templates->get("nav_sep")."\";");
-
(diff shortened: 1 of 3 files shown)


Shade-:
Yeah sure I can fix that asap.

euantorano:
Great, thanks!

Shade-:
Thing is, we still need to add the breadcrumb in the end of processing, given bits are dynamically added by each file. I don't know yet how to handle Twig extensions, and I don't know what'd be the best approach here

euantorano:
Hm, I’ll have a look at it and see if I can come up with something. If it’s done as an extension then the extension function will only be called when the layout file is rendered, which should be within `output_page` so we should be fine.

euantorano:
This was superseded by #3404, so I'm closing this.

PR #3389, opened 2018-08-11
Fix #2921 SQL error on install due to reserved word in MySQL 8

### LINKED ISSUE #2921: SQL error on install due to reserved word in MySQL 8.0.2

dvz:
#2921

skips non-MySQL engines used in the reverted #2955 and adds quoting for some DB methods; the `ADMIN` keyword, originally reserved in MySQL 8, became unreserved again in [8.0.12](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-12.html)


inc/db_mysql.php
```diff
@@ -1353,7 +1353,7 @@ function replace_query($table, $replacements=array(), $default_field="", $insert
 	 */
 	function drop_column($table, $column)
 	{
-		return $this->write_query("ALTER TABLE {$this->table_prefix}{$table} DROP {$column}");
+		return $this->write_query("ALTER TABLE {$this->table_prefix}{$table} DROP `{$column}`");
```
euantorano: I think it's worth checking if `$column` is already quoted for these methods to prevent duplicate quotes.

dvz: Not in the core (adding custom quoting for such methods would be hacky, but these plugins could break - not sure if worth the overhead)

dvz: Overhead of mitigating creative approaches, rather than performance; will add

Issue #3100, opened 2018-04-06
Improve `fetch_remote_file()` to support post requests with other types of request body


euantorano:
The `fetch_remote_file()` function will currently accept a second parameter which is an array of post data. This data is converted into the form request format (`key=val`), though there are several web services that expect POST requests with JSON bodies and such. 

It would be good to extend this function to support other types of request. There are a few enhancements that would be very nice to include:

- If the post data parameter is a string, pass that as is as the post body. 
- Support other HTTP request methods such as `PUT`, `PATCH`, `DELETE`, etc. 
- Support setting the request content type. 

If there are any other enhancements that could be made, please list them here and I’ll see what I can do.

effone:
Multiple request support?

yuliu:
Reference a [community post](https://community.mybb.com/thread-224406.html) here for a discussion to disregard the letter case in header's `Location: `.

euantorano:
HTTP headers are supposed to be case insensitive according to the specification if I remember correctly. If we’re doing case sensitive checks, that’s a bug for sure that should be fixed in feature.

yuliu:
[RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2) says it's case-insensitive:

> 3.2.  Header Fields
>
>   Each header field consists of a case-insensitive field name followed
>   by a colon (":"), optional leading whitespace, the field value, and
>   optional trailing whitespace.

Seems these fields are case-sensitive: request method, URI (but not hostname) and some fields in Protocol Versioning.

euantorano:
Yep, that’s what I thought. If we’re doing case sensitive checks, that’s 1000% a bug that needs fixed 😂

Ben-MyBB:
Fixed in https://github.com/mybb/mybb/issues/3850

PR #3391, opened 2018-08-14
Fixes #3058 File Verification can wrongly parse mybb.com responses

### LINKED ISSUE #3058: File Verification can wrongly parse mybb.com responses
The ACP's _File Verification_ currently loads expected file checksums by [splitting data with newline character](https://github.com/mybb/mybb/blob/mybb_1815/admin/modules/tools/file_verification.php#L39), which can result in incorrect interpretation, e.g. when the response contains an error page's HTML instead.

effone:
Attempt to fix #3058

Using `file_get_contents( )` in place of `fetch_remote_file( )` to avoid accumulating unwanted garbage error page codes resulting to trigger the condition check `empty($file[0])` properly.

_Test result with 1.8.18 pre-release (where checksum file is missing remotely)_

Using `fetch_remote_file( )`:

snap

Using `file_get_contents( )`:

snap

### DIFF
--- a/admin/modules/tools/file_verification.php
+++ b/admin/modules/tools/file_verification.php
@@ -36,7 +36,7 @@ if(!$mybb->input['action'])
 
 		$page->output_header($lang->file_verification." - ".$lang->checking);
 
-		$file = explode("\n", fetch_remote_file("https://mybb.com/checksums/release_mybb_{$mybb->version_code}.txt"));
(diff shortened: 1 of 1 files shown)


dvz:
This solution might increase reported connection errors; ideally [`fetch_remote_file()`](https://github.com/mybb/mybb/blob/mybb_1817/inc/functions.php#L6944) would be extended to handle HTTP error codes.

euantorano:
Yes, there is another open issue for fixes/improvements to `fetch_remote_file`: #3100 

This is a quick fix that will do for now, and can be extended in later versions.

PR #3392, opened 2018-08-15
Fixes #3402 - Memberlist-add numbers for quick search

### LINKED ISSUE #3402: Weird search oversight on all mybb forums
> I noticed that there is a search oversight on all mybb forums.

Here is the issue. When looking at member listing, there is a search bar that has the letters A-Z, you can quick search by clicking on a letter to see members with names that start with that letter,

But what about users whos name starts with a number. Example: 420man

The problem is, there is no way to quick search names that start with numbers because the search bar is only A-Z, what about 0-9?

I don't get the oversight on this.

link to MyBB original thread for discussion and more info
https://community.mybb.com/thread-219267-post-1311758.html#pid1311758

Eldenroot:
https://community.mybb.com/thread-219267-post-1311758.html#pid1311758

#3402


effone:
Where is the 'issue' against this PR? Every PR should be followed by an issue in place.
Have we come to a solution finally for how to handle the raised concern?

install/resources/mybb_theme.xml
```diff
@@ -5581,7 +5581,7 @@ if(use_xmlhttprequest == "1")
 </table>
 </div>
 </div>]]></template>
-		<template name="memberlist" version="1808"><![CDATA[<html>
+		<template name="memberlist" version="1818"><![CDATA[<html>
```
effone: Bump version suitably, this is not part of 1.8.18

Eldenroot:
This is a small improvement, a dropdown menu or whatever should be added in 1.9.

effone:
@Eldenroot whatever you are addressing should be driven through issues only. PR should not carry milestone. This is required to show up in the page linked from release note, like:
https://github.com/mybb/mybb/milestone/25?closed=1

Please make an issue and link the PR to that.

euantorano:
Any other thoughts on this - @Eric-Jackson and @justinsoltesz this kind of change is your kind of thing, since it's a cosmetic change.

If we do go ahead, the template versions will need bumping to `1820`.

Eldenroot:
Template version is bumped to 1820. Feel free to merge/reject :)

justinsoltesz:
I'm not sure this is necessary. I was actually considering removing the letter list in 1.9 as it really isn't all that useful.

Eldenroot:
So feel free to close this. Will be there any easy way how to filter users in 1.9 by first letter of username?

euantorano:
As we seem to be agreed, I'm going to close this.

PR #3401, opened 2018-08-18
Fixes #3395 - Regression - Force Custom profile fields on Admins & Mods

### LINKED ISSUE #3395: Regression on PR #2812 - Force Custom profile fields on Admins & Mods
I see no reason to keep #2812. Just because it was there before doesn't mean it should stay, it causes more trouble than not IMO. At this precise moment you can have a Custom profile field "Viewable to" setting set to "None" and that said custom profile field will still show for Admins and Moderators on the Profile. As it's an Administrator setting the "Viewable to" option to "None" (or not their own group), they clearly do not want to see the said field. Forcing them to see it is just silly, especially if it doesn't enhance security. 

If it stays as is, if admins & mods really do not want to see it they have to make a core edit. Without this change, if admins & mods really do not/or want to see it they only have to set the option to in the ACP.

Eldenroot:
#3395 

This is the easiest way how to solve this, we can create another issue with a complete solution which should be added in MyBB 1.9/1.10.


AndreRl:
Nicely done, however I was going to wait as @Shade- may of done this for other unknown reasons and it may not get accepted.

Eldenroot:
@AndreRl - it is fine, can be closed via one click :)

dvz:
The setting description lays out current logic, so it also would have to be changed:
https://github.com/mybb/mybb/blob/mybb_1818/inc/languages/english/admin/config_profile_fields.lang.php#L42
(https://github.com/mybb/mybb/blob/mybb_1818/admin/modules/config/profile_fields.php#L209)

Eldenroot:
Hmm ok, what do you suggest?

Eldenroot:
@dvz - ok, I changed the lang file

Eldenroot:
@euantorano and others MyBB devs - would be nice to get any feedback from you

Issue #3385, opened 2018-08-10
Unable to edit users in Admin CP


justinsoltesz:
You are not able to save users in the Admin CP in 1.9. The form's submit button does not appear, and the account settings tab has no content besides this error:

```
Fatal error: Uncaught Error: Call to undefined function MyBB\template() in /path-to-forum/inc/functions.php:4296 Stack trace: #0 /path-to-forum/admin/modules/user/users.php(1375): build_theme_select('style', '0', 1, '', true, false, true) #1 /path-to-forum/admin/index.php(821): require('/partial-path...') #2 {main} thrown in /path-to-forum/inc/functions.php on line 4296
```

rajat315315:
Is it resolved?

Ben-MyBB:
No, This is for 1.9 I believe

________________________________

rajat315315:
Yes, for 1.9 I was asking?

euantorano:
I can't remember off the top of my head if I've already fixed this or not. I don't believe I have.

rajat315315:
Hey, I was not able to replicate this scenario but there was a wierd error coming when I went to edit a `User's` profile in `Users & Groups` setting tab and tried to `Save` it.

![Screenshot_2020-07-19 Edit User](https://user-images.githubusercontent.com/15379278/87862468-1f7b9600-c96e-11ea-832d-54b5bf27b461.png)

cronlabspl:
> Hey, I was not able to replicate this scenario but there was a wierd error coming when I went to edit a `User's` profile in `Users & Groups` setting tab and tried to `Save` it.
> 
> ![Screenshot_2020-07-19 Edit User](https://user-images.githubusercontent.com/15379278/87862468-1f7b9600-c96e-11ea-832d-54b5bf27b461.png)

:) Minor wording error.

arcalys:
Could not reproduce.

PR #3404, opened 2018-08-18
Nav template conversion

euantorano:
For #2972, adaptation of #3386.

This uses a Twig function to render the breadcrumb, and adds a new `BreadcrumbManager` class to manage breadcrumbs.

This isn't yet complete, as I plan to add a `multipage` Twig function as part of this PR to use in the breadcrumb template. It also doesn't do anything with the nav on the archive pages, which will need some work.

Pinging @Shade- since he worked on the original nav template conversion.

### DIFF
--- a/archive/global.php
+++ b/archive/global.php
@@ -175,10 +175,6 @@ $lang->load("global");
 $lang->load("messages");
 $lang->load("archive");
 
-// Draw up the basic part of our naviagation
-$navbits[0]['name'] = $mybb->settings['bbname_orig'];
-$navbits[0]['url'] = $mybb->settings['bburl']."/archive/index.php";
-
 // Check banned ip addresses
 if(is_banned_ip($session->ipaddress))
 {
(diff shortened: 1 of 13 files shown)


euantorano:
This PR also fixes some other issues related to where `inc/src/bootstrap.php` was being included, which caused the archive page to error out for example. It may also fix #3385 too, but I haven't tested that yet.

euantorano:
Archive navigation is now implemented and this should be fixed.

Would be great if somebody could give it a quick look over and make sure I haven't missed anything, @PaulBender @Shade-.

euantorano:
Interesting, I’ll have a look at that tonight.

euantorano:
Fixed both of the above issues, only issue left is that the last entry is also showing as a link - it seems `loop.last` isn't working as it should...

euantorano:
This should now be complete, the above mentioned issues have both been fixed.

Unfortunately the breadcrumb template still uses `|raw` for the URL, as every URL passed to the old `add_breadcrumb` function runs `htmlspecialchars_uni()` before passing into that function. As time goes on we can replace those with calls to `BreadCrumbManager::addBreadcrumb()` and move the escaping logic.

PaulBender:
Looks good now.

PR #3405, opened 2018-08-19
Fixes #2020 Referrals Are Lost on Account Merge

### LINKED ISSUE #2020: Referrals Are Lost on Account Merge
As the title says.

If account A has 6 referrals and is merged into account B with only 1 referral then it will still only say they have 1 referral.

It should carry over IMHO, same with reputation actually.

Original thread: [Referrals Are Lost on Account Merge](http://community.mybb.com/thread-167259.html)

effone:
Attempt to fix #2020 
- Referrer / Referrals carry over
- Undefined / Same user error logic fine tuning
- Slight code optimization

Note: Addressing only the titled issue for now. If the extended part as mentioned in comments (related to reputation, not referrals) are also required that can be done too ...

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1918,16 +1918,13 @@ if($mybb->input['action'] == "merge")
 
 	if($mybb->request_method == "post")
 	{
-		$source_user = get_user_by_username($mybb->input['source_username'], array('fields' => '*'));
-		if(!$source_user['uid'])
+		foreach(array('source', 'destination') as $target)
 		{
-			$errors[] = $lang->error_invalid_user_source;
-		}
-
-		$destination_user = get_user_by_username($mybb->input['destination_username'], array('fields' => '*'));
-		if(!$destination_user['uid'])
-		{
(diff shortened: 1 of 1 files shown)


Eldenroot:
@effone - for others mentioned issues - will you add them into this PR later or a new PR in future? Thank you, I will test this soon.

effone:
I kept the decision open as to target the issues in an optimized way. As @euantorano directed big / complicated modifications are part of 1.9 as many of those may get resolved during coding itself. If it is required I have no issue adding the fix to this PR.

Eldenroot:
Maybe would be nice to get things sorted properly... if it is not a big change, I vote to add it to this PR (or the new one). @euantorano

euantorano:
I would class these as bugs rather than new features, and bug fixes can still be done in 1.8.x.

The stuff that I really want to leave until 1.9 is anything that has lots of template changes, as that makes merging 1.8 and 1.9 extremely difficult.

Eldenroot:
Can someone create a new ticket for that? I agree that this is a bug rather that a new feature.

PR #3408, opened 2018-08-22
fix #3383 - unable to add images to register questions

### LINKED ISSUE #3383: unable to add images to register questions
After mybb 1815 release, the security questions enable to add html, I use it to add images in the questions, if it was by security reasons, will be a good feature to add bbcode support instead of html, to support the image questions:

Before 1815:
screenshot_10

After 1815:
screenshot_8

The code added was:
screenshot_11

MinusPL:
Added parser for security questions at lines 1221 to 1237.
Still possible that 200 characters for security questions may be too short when mycode formatting is used.

### DIFF
--- a/admin/modules/config/questions.php
+++ b/admin/modules/config/questions.php
@@ -36,22 +36,25 @@ if($mybb->input['action'] == "add")
 
 		if(!$errors)
 		{
-			$answer = preg_replace("#(\r\n|\r|\n)#s", "\n", trim($mybb->input['answer']));
+			if(!isset($mybb->input['preview']))
+			{
+				$answer = preg_replace("#(\r\n|\r|\n)#s", "\n", trim($mybb->input['answer']));
 
-			$new_question = array(
-				"question" => $db->escape_string($mybb->input['question']),
-				"answer" => $db->escape_string($answer),
-				"active" => $mybb->get_input('active', MyBB::INPUT_INT)
-			);
-			$qid = $db->insert_query("questions", $new_question);
+				$new_question = array(
+					"question" => $db->escape_string($mybb->input['question']),
+					"answer" => $db->escape_string($answer),
+					"active" => $mybb->get_input('active', MyBB::INPUT_INT)
+				);
+				$qid = $db->insert_query("questions", $new_question);
 
-			$plugins->run_hooks("admin_config_questions_add_commit");
+				$plugins->run_hooks("admin_config_questions_add_commit");
 
-			// Log admin action
-			log_admin_action($qid, $mybb->input['question']);
+				// Log admin action
+				log_admin_action($qid, $mybb->input['question']);
 
-			flash_message($lang->success_question_created, 'success');
-			admin_redirect("index.php?module=config-questions");
+				flash_message($lang->success_question_created, 'success');
+				admin_redirect("index.php?module=config-questions");
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


Eldenroot:
Is possible to make limit higher?

MinusPL:
Most likely yeah, but I'd rahter leave that for someone else or something, because I don't want to mess with db related stuff. It needs to have changed one field's limit of length in mybb_questions table.

Eldenroot:
OK, anyway thank you for your PR, good job!

euantorano:
Ping @mybb/developers - thoughts on changing the security question field to type `TEXT`?

PaulBender:
Sounds like a good idea.

dvz:
Seems like it could also use parsing in the ACP (https://github.com/mybb/mybb/blob/feature/admin/modules/config/questions.php)

MinusPL:
I thought about ACP too, but knowing how images are parsed and all other bbcode stuff I choosed not to do so. It would make ACP really "wide" in that one column due to images beeing to big, or any other element. But if you want it is easy to add parser (better option would be adding a preview for added question to see how it would look).

euantorano:
I think it would be worthwhile adding to the ACP. Otherwise, this looks good.

Eldenroot:
@MinusPL - ACP part would be nice

@euantorano - what about to merge this and open a new issue with ACP setting?

euantorano:
I'd like to get the ACP part in before merging, but if @MinusPL doesn't have time I can always implement that.

MinusPL:
Feel free to implement that - currently I have too much other things to do, so I can't make any changes. Thanks 👍

Sama34:
What is the need to run the question through the parser? I couldn't find a discussion.

euantorano:
We used to allow HTML in them, but that caused issues if you had admins who might add a rogue script there or something, so we added `htmlspecialchas_uni()` to the questions. That then meant that you couldn’t use bold or italics or anything, so we then started using the parser instead.

Sama34:
Is there a real need for it? If we run the parser we should then add settings for the parser options for every single question.

Seems redundant to me to allow back the use of HTML (parser options) having reverted the use of it in the first place.

euantorano:
We probably should have just left HTML enabled and saved a loft of trouble, but we can’t really undo it now. The only people who can edit the questions are admins anyway and if you can’t trust your admins you’ve likely got bigger problems.

Sama34:
I see this was added in a607821ffb46e1d3c1accfa5f415f78bafef038c

I'm unsure about whether the change should be reverted but in my opinion it should stay as inserting HTML in many places is not a intended feature.

Regarding adding the parser, it simply looks unnecessary.

An win-win solution could be to add new a new variable, one being parsed with `htmlspecialchars_uni()` and the other none, using the parsed one in templates by default.

But this could lead to inconsistency across the script and a unnecessary waste of human resources.

Eldenroot:
OK, so make a conclusion, how to solve this, regression or add a setting for that?

euantorano:
I'm not sure, thoughts @dvz and other @mybb/developers? I'd be for just reverting the `htmlspecialchars_uni()` patch and just warning admins to be careful.

Eldenroot:
Or use the patch from @MinusPL with settings but with a limitation 200 characters length. If not, I agree with @euantorano to remove htmlspecialchars_uni() and add a note into lang file that this file is not parsed and html should be used with a caution.

dvz:
I'd go ahead with MyCode support ([also in ACP](https://github.com/mybb/mybb/pull/3408#issuecomment-419201266)).

Eldenroot:
Admins use images in security questions. However it was changed and now it is not possible.

MinusPL:
@Sama34 to make some things clear: I copied parser options from one MyBB script and changes few due to how registration form works. If admins use badwords in security questions then you have really bad admins ;)
Also I didn't had time to add ACP thing simply because I'm studying now and all my free time is beeing consumed by projects needed to pass some subjects >.>

MinusPL:
Okay I have more time now so I'd like to redo some things in my code. I'm going to fix some issues with my original fix and add few mentioned ideas to do it properly. Sorry guys that you had to wait several months, but my university just don't want me to have much free time.

Anyone - it is possible to commit to this pull request with fixed code, or I need to create new one?

euantorano:
You should be able to just commit to this PR, but really just do whatever is easiest for you 😄

MinusPL:
Okay, also I have noticed some other mistake - ajax version of refreshing question won't work with my fix and will display only text, but already fixed that in upcomming commit. Right now I'm writting some necessary changes to ACP.

MinusPL:
Okay I did some changes:
- added xmlhttp version of parsing questions
- changed parser options to filter badwords
- added preview function for questions in ACP:
Here I've created preview table with parsed content of question which will appear on top of editing form in question editing in ACP. Considering that parsing question on list would cause it to just expand and make it look weird I decided to take this approach. You can check preview by button in edit / add form or by option added to popup menu on list of questions.
Please give any feedback if this is good or I need to change it to make it look better with other things in MyBB. (I'm not really advanced in php or anything like that. I did this as I thought would be best to avoid expanding question list in ACP).

MinusPL:
@euantorano please check changes that I created and say if something needs further polish.

Eldenroot:
Good job @MinusPL ! This needs to be tested but looks promising!

PR #3416, opened 2018-08-25
Fixes #3415 - Quick reply auto-subscribes

### LINKED ISSUE #3415: Quick reply auto-subscribes users
Hi there, 

I upgraded my board to MyBB 1.8.18 yesterday. Most things have run smoothly; however, we did run into a bug resulting from the upgrade. One of our users reported that she was receiving alerts from replies to her subscribed threads - even though she hadn't set up any subscriptions, and her default thread subscription mode was set to "Do not subscribe". 

After a bit of investigation, I found that the code to handle thread subscriptions from Quick Reply (in newreply.php) has changed in MyBB 1.8.18, and this seems to be mishandling the "Do not subscribe" option. 

I've reverted this section of code to what was in MyBB 1.8.17 and earlier, which seems to have fixed the issue for now; however, I'm hoping that a more permanent fix can be found for future releases.

Thanks Smile !

Original thread: [Quick reply auto-subscribes users](https://community.mybb.com/thread-219406.html)

AndreRl:
Extended from #3342, fixes #3415

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3456,7 +3456,7 @@ function get_subscription_method($tid = 0, $postoptions = array())
 {
 	global $mybb;
 
-	$subscription_methods = array('dont', 'none', 'email', 'pm'); // Define methods
+	$subscription_methods = array('', 'none', 'email', 'pm'); // Define methods
 	$subscription_method = (int)$mybb->user['subscriptionmethod']; // Set user default
 
 	// If no user default method available then reset method


AndreRl:
@euantorano @Eldenroot @effone When you have the time please review if necessary.

effone:
This is not the desired fix. I placed the method value as `dont` for a reason. And the reason is this:
https://github.com/mybb/mybb/blob/feature/newreply.php#L835

If you blank out the first value the new reply page will appear with no radio checked instead of first one checked in case of not subscribed:
snap

The desired fix should be keeping the `dont` as it is and change this line:
https://github.com/mybb/mybb/blob/feature/newreply.php#L382

From
`if(!isset($postoptions['subscriptionmethod']))`
to
`if(!isset($postoptions['subscriptionmethod']) || ($postoptions['subscriptionmethod'] == "dont" && $mybb->get_input('method') == "quickreply"))`

dvz:
This above seems to work fine, but it would introduce internal inconsistency (some functions returning `dont` while the handler expects an empty string, requiring additional processing). Perhaps the first radio input could be checked when the `get_subscription_method()` output is empty?

effone:
In that case additional modifications in var naming and templates are required.
I will try to push some changes in the branch (if required).

effone:
Alright I have made some changes in the PR. Hope its okay now ...

dvz:
Looks like the changes of https://github.com/mybb/mybb/pull/3342/files and of this PR still need to be applied to `editpost.php`.
The template version should be bumped to `1819`.

PR #3418, opened 2018-08-26
Fixes #3417 - Last Active shown for hidden users

### LINKED ISSUE #3417: Last Active shown for hidden users
Users who have set "Hide me from the Who's Online list" to "Yes" will have their Last Visit information set to "(Hidden)". This is the correct behaviour. The same behaviour is no longer displayed when that said user is on someone's buddy list - their Last Visit information will be displayed.

Eldenroot:
#3417 

I am think that it is a intended behaviour - same as disable PMs - buddies are not affected.


Eldenroot:
@Ben-MyBB - changed :)

Eldenroot:
What do you think guys?

PR #3421, opened 2018-08-30
Fixes #3233 Cascading usergroup moderator permissions broken

### LINKED ISSUE #3233: Cascading usergroup moderator permissions broken
If you create a user moderator on a category, and enable, say, open/close threads and stick/unstick threads, the user will have these powers in all forums inside the category. If you then add them as a moderator again to one of the forums inside this category, and only give them the ability to stick/unstick, then they lose the ability to open/close threads in this specific forum. This is correct.

However, the same does not apply to usergroup moderators. Using the same example as above, the moderators would still be able to open/close threads in the forum were specific moderator permissions were set, where they should only be granted stick/unstick permissions.

The issue is that in get_moderator_permissions, when setting permissions for users, it checks if the value is 0:

PHP Code://&nbsp;Figure&nbsp;out&nbsp;the&nbsp;user&nbsp;permissionsif(&#36;value&nbsp;==&nbsp;0){&nbsp;&nbsp;&nbsp;&nbsp;//&nbsp;The&nbsp;user&nbsp;doesn't&nbsp;have&nbsp;permission&nbsp;to&nbsp;set&nbsp;this&nbsp;action&nbsp;&nbsp;&nbsp;&nbsp;&#36;perms[&#36;action]&nbsp;=&nbsp;0;}else{&nbsp;&nbsp;&nbsp;&nbsp;&#36;perms[&#36;action]&nbsp;=&nbsp;max(&#36;perm[&#36;action],&nbsp;&#36;perms[&#36;action]);}&nbsp;
This will override the 1 set at category level with the 0 set at forum level, thus removing the ability to open/close threads (again, correct).

However, a few lines down when setting permissions for usergroups, it doesn't do this:

PHP Code:&#36;perms[&#36;action]&nbsp;=&nbsp;max(&#36;perm[&#36;action],&nbsp;&#36;perms[&#36;action]);&nbsp;
Simply using the same code for usergroups as we do for users fixes the issue.

Usually in MyBB, we say a yes overrides a no - i.e. you can be in 10 groups that disallow a permission, but 1 that allows it, so you will be allowed. However, I think moderator permissions are a bit different.

Personally I think usergroups should behave the same as users, as it is reasonable to want to override permissions for a specific forum, maybe 3 levels deep, while setting everything else on a category level. This already works for users, just not usergroups.

Ideally, these permissions wouldn't be set up with a simple yes/no, they really need an 'inherit' option. This way, when you edit the moderators on specific forums, you could choose whether to specifically restrict a permission, or keep it in line with the permissions set on the category. However, this is best suited for 2.0 :D

Original thread: [Cascading usergroup moderator permissions broken](https://community.mybb.com/thread-206952.html)

Starpaul20:
Fixes #3233

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1709,7 +1709,16 @@ function get_moderator_permissions($fid, $uid=0, $parentslist="")
 					continue;
 				}
 
-				$perms[$action] = max($perm[$action], $perms[$action]);
+				// Figure out the usergroup permissions
+				if($value == 0)
+				{
+					// The usergroup doesn't have permission to set this action
+					$perms[$action] = 0;
+				}
+				else
+				{
+					$perms[$action] = max($perm[$action], $perms[$action]);
+				}
 			}
 		}
 	}


dvz:
`$perms` values can be inadvertently downgraded when a group with lower permissions is checked after a group with higher permissions in the same forum.

Some variables, like `$perms[$action]`, can be undefined when used with `max()` (same with existing code, but we can start fixing those given the code will be carried over to future branches).

Eldenroot:
@PaulBender - are you willing to complete this?

PaulBender:
I can't really, I just changed computers and I'm still settling it in.

Ben-MyBB:
@PaulBender Ping, if you get time.

Eldenroot:
@PaulBender - please can you finish this one? To review and merge :) thank you!

PaulBender:
Don't know if I can, I had to delete and recreate my repository awhile back.

Eldenroot:
This PR is a small one... maybe recreate it? :) thx

PaulBender:
Yeah, I'll try to fix this and recreate it this week if I can.

PaulBender:
After playing around with this issue last night, I've come to the conclusion that the original issue probably isn't worth fixing, at least in 1.8.

yuliu:
euantorano [has said in Discord](https://discordapp.com/channels/215876847634743296/215877079911104522/614857713112842241) on a relating issue #3776:
> Yes, permissions can be confusing sometimes especially for users with lots of groups
> That’s why 2.0 was going to use a yes/no/never system like XenForo’s
> I’d also like a way to visualise the exact permissions for any given user

Ben-MyBB:
@euantorano thoughts on this one?

euantorano:
Will have to review and reacquaint myself, but based on comments it sounds like we may be looking at a different approach in the future.

yuliu:
I have some idea about resolving #3996 (including #3233 that this PR tries to resolve), however the PR resolving #3996 and this PR will conflict with each over.

Anyway, the way getting mod permissions from multiple user groups of this PR seems fine to me, though primitive but does some work and should have no side effect.

lairdshaw:
With the intention of reviewing unmerged 1.8 PRs and issues to try to get them merged where possible, I've just revisited this PR, and I think the concerns raised by @dvz above, and @yuliu in #3996 mean that merging this PR would create as many problems as it solves  - e.g., it would mean that the comment `// User settings override usergroup settings` no longer applies unconditionally, given that a `$value` of zero in the usergroup code block would override an earlier non-zero permission set in the preceding user code block.

Recommendation: close this PR without merging it, and revisit permissions in 1.9, taking careful account of @yuliu's and @dvz's comments/analyses.

PR #3423, opened 2018-08-30
Fix #3241 .prop('checked') doesn't trigger change events

### LINKED ISSUE #3241: .prop('checked') doesn't trigger change events

dvz:
Fix #3241 

Changes to non-moderator functionality.

If "all" checkbox is used, item checkboxes are synchronized; after that it's unchecked when an item checkbox is unchecked (using `.trigger()` with extra parameter to distinguish the origin of this action).

### DIFF
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -108,7 +108,7 @@ class DefaultPage
 
 		echo "	<script type=\"text/javascript\" src=\"../jscripts/jquery.js?ver=1813\"></script>\n";
 		echo "	<script type=\"text/javascript\" src=\"../jscripts/jquery.plugins.min.js?ver=1813\"></script>\n";
-		echo "	<script type=\"text/javascript\" src=\"../jscripts/general.js?ver=1813\"></script>\n";
+		echo "	<script type=\"text/javascript\" src=\"../jscripts/general.js?ver=1820\"></script>\n";
 		echo "	<script type=\"text/javascript\" src=\"./jscripts/admincp.js\"></script>\n";
 		echo "	<script type=\"text/javascript\" src=\"./jscripts/tabs.js\"></script>\n";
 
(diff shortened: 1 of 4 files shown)


admin/inc/class_page.php
```diff
@@ -108,7 +108,7 @@ function output_header($title="")
 
 		echo "	<script type=\"text/javascript\" src=\"../jscripts/jquery.js?ver=1813\"></script>\n";
 		echo "	<script type=\"text/javascript\" src=\"../jscripts/jquery.plugins.min.js?ver=1813\"></script>\n";
-		echo "	<script type=\"text/javascript\" src=\"../jscripts/general.js?ver=1813\"></script>\n";
+		echo "	<script type=\"text/javascript\" src=\"../jscripts/general.js?ver=1819\"></script>\n";
```
euantorano: Needs bumping to `1820`.

install/resources/mybb_theme.xml
```diff
@@ -14039,13 +14039,13 @@ $(function() {
 </html>]]></template>
 		<template name="debug_summary" version="1800"><![CDATA[<div id="debug">{$generated_in} {$debug_weight} <br />{$sql_queries} / {$server_load} / {$memory_usage}<br />[<a href="{$debuglink}" target="_blank">{$lang->debug_advanced_details}</a>]<br /></div>]]></template>
 		<template name="gobutton" version="120"><![CDATA[<input type="submit" class="button" value="{$lang->go}" />]]></template>
-		<template name="headerinclude" version="1817"><![CDATA[<link rel="alternate" type="application/rss+xml" title="{$lang->latest_threads} (RSS 2.0)" href="{$mybb->settings['bburl']}/syndication.php" />
+		<template name="headerinclude" version="1819"><![CDATA[<link rel="alternate" type="application/rss+xml" title="{$lang->latest_threads} (RSS 2.0)" href="{$mybb->settings['bburl']}/syndication.php" />
```
euantorano: Again, `1820` for templates too.

PR #3426, opened 2018-08-31
Remove the archive/ directory and associated functions file for 1.9

kawaii:
Since the archive is being depreciated in 1.9, this PR removes it and the associated functions file from `inc/`. I belive @euantorano is handling the hard-coded navigation templates that reference it in a future PR.


justinsoltesz:
everything is wrong

euantorano:
I think you broke it @kawaii

PR #3438, opened 2018-09-03
fixes #3430 ACP settings search option hardcoded only for english language

### LINKED ISSUE #3430: ACP settings search option - hardcoded only for english language.
myBB version :
1.8.18

Issue :
search  in ACPs configuration tab is "hardcoded" to always use english language, independently from language chosed by administrator. Therefore search does not find anything when administrator use different language than english.

Steps to reproduce :
1. Goto ACP
2. Change ACP language to anything but english.
3. Click "configuration" tab at top of page or go to admin/index.php?module=config-settings
4. Insert non english word in search textbox and click search.
5. Obviously nothing will be found.

Details :
admin\modules\config\settings.php at line 1147 perform search by query on DB data which is in english.
It should rather create an array of translated data and search this instead. (or similar ideas)

Original thread: [ACP settings search option - hardcoded only for english language.](https://community.mybb.com/thread-219527.html)

avril-gh:
Fixes issue where ACP settings search option, has searched directly DB data (which is in english) without considering translation. Because of this, ACP options search feature didnt worked for any other language than english.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -1145,18 +1145,41 @@ if($mybb->input['action'] == "change")
 		// Search for settings
 		$search = $db->escape_string_like($mybb->input['search']);
 		$query = $db->query("
-			SELECT s.*
+			SELECT s.* , g.name as gname, g.title as gtitle, g.description as gdescription
 			FROM ".TABLE_PREFIX."settings s
 			LEFT JOIN ".TABLE_PREFIX."settinggroups g ON(s.gid=g.gid)
-			WHERE s.name LIKE '%{$search}%' OR s.title LIKE '%{$search}%' OR s.description LIKE '%{$search}%' OR g.name LIKE '%{$search}%' OR g.title LIKE '%{$search}%' OR g.description LIKE '%{$search}%'
 			ORDER BY s.disporder
 		");
 		while($setting = $db->fetch_array($query))
 		{
-			$cache_settings[$setting['gid']][$setting['sid']] = $setting;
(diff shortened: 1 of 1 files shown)


avril-gh:
fixes #3430

WildcardSearch:
@avril-gh I have a little time to test this week.

Make sure I have steps to reproduce correct:

- Load ACP using a translation
- ACP->Config->Settings [search box]
- current behavior (doesn't find setting because it is looking for English name)
- expected behavior (searches for translated language when appropriate)

avril-gh:
If admin use in ACP another language than english, he obviously will name options in his own language, but since search box directly query on DB tables which are in english. he will not find the option he is looking for.

The solution is to not search directly on DB tables (like now) but instead, fetch them and search them localy while applying translation in the fly.

like usual, tested closely at each stage before submitting PR
Thanks for attention @WildcardSearch

WildcardSearch:
Thank you, @avril-gh.

I was able to reproduce the error. Your PR solves the issue. Maybe @dvz or @euantorano will have time to merge it.

admin/modules/config/settings.php
```diff
@@ -1145,18 +1145,41 @@
 		// Search for settings
 		$search = $db->escape_string_like($mybb->input['search']);
 		$query = $db->query("
-			SELECT s.*
+			SELECT s.* , g.name as gname, g.title as gtitle, g.description as gdescription
```
avril-gh: we'll fetch all settings

WildcardSearch:
Fixes #3430

PR #3440, opened 2018-09-06
Fix #3428 Missing database inserts for engines other than MySQL

### LINKED ISSUE #3428: Missing database inserts for engines other than MySQL
MyBB has a [`mysql_db_inserts`](https://github.com/mybb/mybb/blob/feature/install/resources/mysql_db_inserts.php) file in the installer for inserting data into the database during install. It also has logic to choose other insert files for other database engines, but unfortunately that logic is flawed and there are no other insert files (`$dboptions` doesn't seem to be defined or in scope):

https://github.com/mybb/mybb/blob/feature/install/index.php#L1678

This causes a problem with PostgreSQL since #3389 was merged as the MySQL database inserts file now quotes the word `groups` with a backpack, which is invalid in Postgres.

dvz:
fixes #3428

### DIFF
--- a/install/index.php
+++ b/install/index.php
@@ -106,7 +106,7 @@ if(function_exists('pg_connect'))
 		'title' => 'PostgreSQL',
 		'short_title' => 'PostgreSQL',
 		'structure_file' => 'pgsql_db_tables.php',
-		'population_file' => 'mysql_db_inserts.php'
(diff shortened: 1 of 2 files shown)


kawaii:
Confirmed working, patched a fresh copy of 1.8.18 and the install went smoothly.

PR #3460, opened 2018-09-30
Fixes #3459 - Search feature improvements

### LINKED ISSUE #3459: Search feature improvements
Issue :
search.php - limited usability, due to required input of keyword or author fields.

Details :
Currently user is forced to input a keyword or author to perform search, or else search will not be executed. This limits search feature to only these 2 options and not allow to use additional search options that are present in form alone without providing keyword and author, that could allow to search for new posts, old posts, posts without reply and many other combinations where a keyword and author have no meaning.

For example, with current options, one could search for "posts from last week that have not less than 10 replies", but one cant, since keyword and author has not been provided. - Note that for this criteria, keyword and author have no meaning.

Solution :
allow empty keyword and author input, so the search options like show posts with X replies, or show posts not older than 1 week and other combinations of search options, become search criteria. 

+ add new options to sort by number of replies/views

link: https://community.mybb.com/thread-219812.html

Eldenroot:
#3459


avril-gh:
woah ! i overslept this one. :cat2: 
Great work @Eldenroot  @WildcardSearch  !!

PR #3469, opened 2018-10-14
Fix #3468 per page in manage group

### LINKED ISSUE #3468: Page Change in Managegroup causes "Division by zero" Error
Hello, in my forum I have certain usergroups that can be joined/approved etc. I noticed that in these group pages (managegroup.php) when I click on page 2 or 3 to see other users I get this error:

Quote:Warning [2] Division by zero - Line: 358 - File: managegroup.php PHP 7.0.29 (Linux)

This doesn't happen at page 1. Happens only at other pages.

My managegroup.php doesn't have any changes.

Can someone check this and provide a fix if it's possible?

Original thread: [Page Change in Managegroup causes "Division by zero" Error](https://community.mybb.com/thread-220212.html)

euantorano:
Fixes #3468 as described [here](https://community.mybb.com/thread-220212-post-1316567.html#pid1316567).

### DIFF
--- a/managegroup.php
+++ b/managegroup.php
@@ -355,7 +355,7 @@ else
 	if($page && $page > 0)
 	{
 		$start = ($page-1) * $perpage;
-		$pages = ceil($numusers / $per_page);
+		$pages = ceil($numusers / $perpage);
 		if($page > $pages)
 		{
 			$start = 0;


Eldenroot:
I can confirm that this fix is working fine :) thx!

PR #3472, opened 2018-10-16
Fix #3467 - load all css.php stylesheets at once

### LINKED ISSUE #3467: css.php - multiple queries
Apologies if this is in the wrong place. This is a request to enhance css.php

Many (like myself) choose to not pre-cache stylesheets as a file (for various reasons), but unfortunately this can require a dozen extra MySQL queries. When it only requires 1.

Can this:
```
<link type="text/css" rel="stylesheet" href="https://domain.com/css.php?stylesheet=32" />
<link type="text/css" rel="stylesheet" href="https://domain.com/css.php?stylesheet=38" />
<link type="text/css" rel="stylesheet" href="https://domain.com/css.php?stylesheet=17" />
```
Be changed to this:
```
<link type="text/css" rel="stylesheet" 
href="https://domain.com/css.php?stylesheet=32,38,17&lastmod=1539446075" />
```
Please.

euantorano:
Fixes #3467 by loading all stylesheets loaded through `css.php` at once.

This does cause some issues with the stylesheet display order options, unfortunately. It could be possible to have the display order set in the following style:

- Sheet from `css.php`
- Sheet from cache
- Another sheet from cache
- ...

With this patch, the above would become:

- Sheet from cache
- Another sheet from cache
- All sheets from `css.php`, in order according to display order

We may also want to add some extra hooks into `css.php`, but I just wanted to whip up a first draft implementation to get more thoughts on the proposal.

### DIFF
--- a/css.php
+++ b/css.php
@@ -60,14 +60,14 @@ SQL;
 	{
 		$stylesheet = $row['stylesheet'];
 
-		$plugins->run_hooks('css_start');
+		$plugins->run_hooks('css_start', $stylesheet);
 
 		if(!empty($mybb->settings['minifycss']))
 		{
 			$stylesheet = minify_stylesheet($stylesheet);
 		}
 
-		$plugins->run_hooks('css_end');
+		$plugins->run_hooks('css_end', $stylesheet);
 
 		$content .= $stylesheet;
 	}
--- a/css.php
+++ b/css.php
@@ -22,13 +22,14 @@ if(!empty($stylesheets))
 	$stylesheet_list = implode(', ', array_map('intval', $stylesheets));
 
 	$content = '';
+	$prefix = TABLE_PREFIX;
 
 	switch($db->type)
 	{
 		case 'pgsql':
 		case 'sqlite':
 			$sql = <<<SQL
-SELECT stylesheet FROM themestylesheets
+SELECT stylesheet FROM {$prefix}themestylesheets
   WHERE sid IN ({$stylesheet_list})
   ORDER BY CASE sid
 SQL;
@@ -46,7 +47,7 @@ SQL;
 			break;
 		default:
 			$sql = <<<SQL
-SELECT stylesheet FROM themestylesheets
+SELECT stylesheet FROM {$prefix}themestylesheets
   WHERE sid IN ({$stylesheet_list})
   ORDER BY FIELD(sid, {$stylesheet_list});
 SQL;
--- a/css.php
+++ b/css.php
@@ -19,21 +19,41 @@ $stylesheets = $mybb->get_input('stylesheet', MyBB::INPUT_ARRAY);
 
 if(!empty($stylesheets))
 {
-	$stylesheet_list = '';
-	$sep = '';
-
-	foreach($stylesheets as $stylesheet)
-	{
-		$stylesheet_list .= $sep . (int) $stylesheet;
-		$sep = ', ';
-	}
+	$stylesheet_list = implode(', ', array_map('intval', $stylesheets));
 
 	$content = '';
 
-	$query = $db->simple_select('themestylesheets', 'stylesheet', "sid IN ({$stylesheet_list})", array(
-		'order_by' => 'sid',
-		'order_dir' => 'ASC',
-	));
+	switch($db->type)
+	{
+		case 'pgsql':
+		case 'sqlite':
+			$sql = <<<SQL
+SELECT stylesheet FROM themestylesheets
+  WHERE sid IN ({$stylesheet_list})
+  ORDER BY CASE sid
+SQL;
+
+			$i = 0;
+			foreach($stylesheets as $sid)
+			{
+				$sid = (int) $sid;
+
+				$sql .= "WHEN {$sid} THEN {$i}\n";
+				$i++;
+			}
+
+			$sql .= 'END;';
+			break;
+		default:
+			$sql = <<<SQL
+SELECT stylesheet FROM themestylesheets
+  WHERE sid IN ({$stylesheet_list})
+  ORDER BY FIELD(sid, {$stylesheet_list});
+SQL;
+			break;
+	}
+
+	$query = $db->query($sql);
 
 	while($row = $db->fetch_array($query))
 	{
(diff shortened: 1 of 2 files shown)


woostarUK:
This is great. Thank you!

Can I recommend an addition...

In 'themestylesheets' there is a column that lists 'lastmodified'. Can you append the url with highest number (eg. domain.com/css.php?1,2,3&lastmod=1537993210).

This will allow for browser and CDN caching :)

euantorano:
That would make sense, but should probably be applied to all style sheets (including cached) so it may make sense to create a new issue for it.

euantorano:
Ping @mybb/developers - any thoughts before merging? Would be good to get at least another member of the team's opinion.

woostarUK:
Is &lastmod= going to make it into the next release? Would be really really handy for serving stylesheets via CDN :)

Sama34:
Is this going to be forced? I haven't tested yet but reading the changes lets me confused about whether administrators will be forced to use this new method, as loading all CSS style sheets in a single file only might not always be the best approach.

woostarUK:
I've tested it. It doesn't make any difference other than save queries if using css.php

Static stylesheets are still served individually.

euantorano:
It is forced at the minute, as using css.php should be a last resort used by a minority of sites. There’s no downside to combining the files at all.

Sama34:
Is more testing required before merging? If it simply combines files but still respects the `Attached to` field then I see no issue.

euantorano:
I need to check the ordering of the css.php stylesheets, but once I’ve done that (tomorrow) it can be merged.

euantorano:
Ordering of stylesheets changed as per the comment from @woostarUK here: https://github.com/mybb/mybb/issues/3482#issuecomment-450576156

This should be good to merge, but would appreciate testing @kawaii - especially against postgreSQL/SQLite since they have a slightly different SQL syntax for the ordering.

woostarUK:
The query is missing TABLE_PREFIX

euantorano:
Good catch, serves me right for not testing. Patched.

woostarUK:
🙈 So sorry. Is it:

SELECT {$prefix}stylesheet FROM themestylesheets
or 
SELECT stylesheet FROM {$prefix}themestylesheets

euantorano:
👀 you saw nothing... 😜

woostarUK:
😂😂

Eldenroot:
Good job! Will try to test

Eldenroot:
I applied all code changes yesterday, testing. @woostarUK - can you please confirm that all is fine with this PR?

woostarUK:
Works great for me on mysql 👍

Eldenroot:
Yes, I have mysql too

woostarUK:
Can I make a suggestion before this gets merged.

Really really minor, but very helpful. 

Can $args be added to the hooks? This will allow the css to be manipulated.

	while($row = $db->fetch_array($query))
	{
		$stylesheet = $row['stylesheet'];
		
		$args = array(
			'stylesheet' => &$stylesheet
		);
		
		$plugins->run_hooks('css_start', $args);
		if(!empty($mybb->settings['minifycss']))
		{
			$stylesheet = minify_stylesheet($stylesheet);
		}
		$plugins->run_hooks('css_end', $args);
		$content .= $stylesheet;
	}

woostarUK:
It's been bugging me why redundant queries are still running. Dug through the code:

This loop doesn't need run the multiple queries any more. The ID's can simply be put in an Array (we already have id's - no need to pull them from the database).
```
				if(strpos($page_stylesheet, 'css.php') !== false)
				{
					// We need some modification to get it working with the displayorder
					$query_string = parse_url($stylesheet_url, PHP_URL_QUERY);
					$id = (int) my_substr($query_string, 11);

					// **** REMOVE THE FOLLOWING LINES ****
					//$query = $db->simple_select("themestylesheets", "name", "sid={$id}");
					//$real_name = $db->fetch_field($query, "name");
					//$theme_stylesheets[$real_name] = $id;

					// **** ADD THIS INSTEAD ****
					$css_ids[]=$id;
				}
```

Use $css_ids instead of $css_php_script_stylesheets
```
if(!empty($css_ids))
{
	$sheet = $mybb->settings['bburl'] . '/css.php?';
	$sep = '';
	foreach($css_ids as $sid)
	{
		$sheet .= $sep . 'stylesheet[]=' . (int) $sid;
		$sep = '&amp;';
	}
	$stylesheets .= "<link type=\"text/css\" rel=\"stylesheet\" href=\"{$sheet}\" />\n";
}
```

euantorano:
@woostarUK Added a parameter to the hooks.

I saw your earlier comment regarding the query for the stylesheet name. Unfortunately we can't remove that at the moment without making a bigger change to how CSS is loaded/managed. This will be done in 1.9.

woostarUK:
Brilliant! Thank you.

Yeh, I deleted the post because I realised it would have been too complicated to make changes on the 1.8 branch (not to mention what I posted didn't work). I did come up with a version that worked, but since 'css.php' is hardcoded in serialized data within the 'theme' db table, it just ended up being to messy str_ replacing.

css.php
```diff
@@ -15,26 +15,64 @@
 require_once "./inc/init.php";
 require_once MYBB_ROOT . $config['admin_dir'] . '/inc/functions_themes.php';
 
-$stylesheet = $mybb->get_input('stylesheet', MyBB::INPUT_INT);
+$stylesheets = $mybb->get_input('stylesheet', MyBB::INPUT_ARRAY);
```
dvz: It looks like the `stylesheet=$sid` format might still exist in the database - it might be a good idea to convert these to `[$sid]` arrays.

global.php
```diff
@@ -364,17 +364,33 @@
...
+			{
+				$stylesheets .= $theme_stylesheets[$style_name];
+			}
 		}
 	}
 }
 
+if(!empty($css_php_script_stylesheets))
+{
+	$sheet = $mybb->settings['bburl'] . '/css.php?' . http_build_query(['stylesheet' => $css_php_script_stylesheets]);
```
dvz: old `array()` syntax needed

euantorano: Fixed.

css.php
euantorano: I'll have to have a look at the database for an install using `css.php` and see. I think that will be handled by the `global.php` changes though.

euantorano:
Are we good to merge this then? If so, I'll go ahead so we can get to releasing 1.8.20.

woostarUK:
> Are we good to merge this then? If so, I'll go ahead so we can get to releasing 1.8.20.

Works as expected here :)

PR #3478, opened 2018-10-20
Fix #3477 - Include back trace in error logs and error emails

### LINKED ISSUE #3477: Improve error log details
The current error logging mechanism logs where errors occurs, but no stack trace.

It's likely best explained with an example (taken from `whiten` on Discord):

```
<error>
	<dateline>1539607234</dateline>
	<script>inc/db_mysqli.php</script>
	<line>469</line>
	<type>2</type>
	<friendly_type>Warning</friendly_type>
	<message>mysqli_fetch_array() expects parameter 1 to be mysqli_result, boolean given</message>
</error>
```

A stack trace would help track down where `mysqli_fetch_array` is actually being called from (eg: where the failing query was generated).

euantorano:
Fixes #3477

### DIFF
--- a/inc/class_error.php
+++ b/inc/class_error.php
@@ -331,6 +331,20 @@ class errorHandler {
 		// Do not log something that might be executable
 		$message = str_replace('<?', '< ?', $message);
 
+		if(function_exists('debug_backtrace'))
+		{
+			ob_start();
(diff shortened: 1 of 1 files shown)


euantorano:
That would be great, thanks!

PR #3480, opened 2018-10-22
Last poster avatars

### LINKED ISSUE #3425: Avatars in forum display and forum index
The 1.9 theme shows the avatars of the last posters in the forum display and the forum index. In the demo forum and the current code the avatars are hardcoded to the default avatar, so we need to add support to `build_forumlist()` and associated functions.

euantorano:
Fixes #3425.

This is still a work in progress, I need to add handling for default avatars for the `render_avatar()` function, and make sure the avatars work in announcements, subforms, etc.


Azareal:
I'm curious what happened to the work with the query builder? It would be nice to eliminate some of these raw queries, although not the end of the world.

euantorano:
Tracking areas still left to do, any I've missed @justinsoltesz?

- [X] User CP index - subscribed threads
- [X] User CP index - recent threads
- [ ] User CP - thread subscriptions
- [ ] User CP - forum subscriptions
- [X] Forum Display - sub-forum listing
- [X] Forum Display - thread listing
- [X] Forum bits - `build_forumbits()` function
- [X] Index - forum list

euantorano:
This PR also fixes an issue with the `CoreExtension` that I discovered while testing, and updates some composer dependencies in the lock file.

justinsoltesz:
Need search results and private messages as well

euantorano:
I knew there was something I'd missed! Thanks, added to the list.

euantorano:
Ok, this should now be complete. Last poster avatars and first poster avatars are now available everywhere I can think of. They're also available for the "users browsing this forum" list, not sure if we want them there but it may make sense?

PR #3488, opened 2018-11-06
Fixes #3486 Missing Moderation Notices

### LINKED ISSUE #3486: Missing Moderation Notices
I remember previously stumbling across this issue myself but I do not know why I never actually pushed to get a solution for it. After reading [How To?] Notifications to moderators about unapproved posts I was once again reminded that MyBB does not give notice of all moderation notices.

As it stands all posts, threads and attachments that need moderator approval are not being notified with a moderation notice like you would get for reported content.

This solution from 2011 does work (I just tested it) however I am not sure how up to par the coding is and haven't actually looked at it hard enough. I'm also TERRIBLE at trying to maneuver or work GitHub so I figured I would post this here for someone else that may be able to actually do pull requests and changes and what not that are necessary.

I do believe this should have always been standard/core MyBB, after all if it is in the moderation queue it should be notified to all moderators or those with Mod CP access.

Original thread: [Missing Moderation Notices](https://community.mybb.com/thread-220422.html)

effone:
Attempt to fix #3486

### DIFF
--- a/global.php
+++ b/global.php
@@ -637,6 +637,8 @@ if($mybb->user['uid'] != 0 && is_array($groupleaders) && array_key_exists($mybb-
 }
 
 $modnotice = '';
+$moderation_queue = array();
+
 // This user is a moderator, super moderator or administrator
 if($mybb->settings['reportmethod'] == "db" && ($mybb->usergroup['cancp'] == 1 || ($mybb->user['ismoderator'] && $mybb->usergroup['canmodcp'] == 1 && $mybb->usergroup['canmanagereportedcontent'] == 1)))
 {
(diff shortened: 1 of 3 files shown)


Eldenroot:
Great, testing since yesterday, seems to be fine. I had been using plugin for that. Thank you!

effone:
Thanks for testing @Eldenroot, There is a very neat-pick issue I have observed with this PR but failed to find the root cause, is an additional space after each link.
snap
I have cross-checked templates, used `trim()` but no luck. That may be my cluttered experiment-exhausted code base. Can you confirm and / or try to find the root cause?

Eldenroot:
I will try to reproduce in 5 hours, back in my office right now. Maybe sometjing similar with board stats (online members) in footer bug which was solved in 1.8.17 IIRC?

effone:
Okay I have found the issue. Its the new line inserted by `template` class.
```

		if($htmlcomments)
		{
			if($mybb->settings['tplhtmlcomments'] == 1)
			{
				$template = "<!-- start: ".htmlspecialchars_uni($title)." -->\n{$template}\n<!-- end: ".htmlspecialchars_uni($title)." -->";
			}
			else
			{
				$template = "\n{$template}\n";
			}
		}
```
If I remove all `\n` from it issue gets resolved. I don't know how to handle this.
@euantorano ?

effone:
Alright I have fixed it. There was a parameter `$htmlcomments` for `get()` method that required to be set to `false`. Love MyBB.

euantorano:
Nice catch @effone!

PR #3490, opened 2018-11-07
Fixes #3483 Add 'exact' option to memberlist username search

### LINKED ISSUE #3483: Add 'exact' option to memberlist username search
Currently the memberlist has advanced search options to search for a username 'beginning with' or 'containing' a string. I'd like to suggest we also add an 'exact' option. On large forums with thousands of members it can be very difficult to find certain users when it's searching in a regular expression style.

effone:
Attempt to fix #3483


euantorano:
This looks good, but it's made me notice another possible issue with username searches: #3492

Issue #1201, opened 2014-08-14
Problem clicking links within AdminCP Template-Sets


sthag:
While working with a new Template I noticed that I've got problems clicking the names of the templates within a fold. I noticed that there is a padding on the span tag around the title of the template. This padding is overlapping with the next and previous title so only half of the text is clickable.

It is coded in the file `admin / modules / style / templates.php` line 894
I changed this `padding: 20px` to `padding-left: 20px` in my code.
But I'm not shure if the surrounding span is even necessary.

I use Windows 7 & 8 and Chrome 34+

I know I could have made a fork and pull request but it's the first time looking into this code for me.

PR #3491, opened 2018-11-07
Fix unclickable bottom 50% of sub-menu items

haydster7:
As reported in:
https://github.com/mybb/mybb/issues/1201
https://community.mybb.com/thread-157489-page-2.html


Regular-Merc:
Such a simple change. Means I don't need to concentrate when clicking on the sub menu links in my templates. Cheers.

effone:
We need an issue created on prior for every PR, don't we?

euantorano:
Yeah, I'm not entirely sure why that issue was closed without a fix, but I'm happy to just re-use it.

PR #3493, opened 2018-11-10
Fixes #3419 View Own Unapproved Threads / Posts

### LINKED ISSUE #3419: Thread and posts should not be hidden from the OP when approval is required!
When creating a new thread or post that is not approved, no one but mods/admins can view it.

That is wrong and not intuitive at all. It should allow for the person who created that post / thread to also view it. It just doesn't make any sense, why does it have to be hidden for even the person who made it.

That causes the person who created the thread to:

1. Think it was not created/published thus trying to post it multiple times
2. There is a yellow warning while writing the thread and another after submitting it which apparently no one reads. The warnings are fine in my opinion but they do not help with the fact the post is not displayed.
3. And last but not least after its posted there is no trace of it, see GIF at the bottom.

The proper behavior should be to still allow the creator of the post / thread be taken back to it once posted, just like a regular post and allow the creator of it, to still view/read it, with a message at the footer of the post saying the message is still waiting for approval and won't be viewed by other users until then.

example

Thanks for the amazing forum!

effone:
Attempt to fix #3419 

- [x] Allow viewing unapproved threads - forumdisplay
- [x] Allow viewing unapproved threads - showthread
- [x] Allow viewing unapproved posts - showthread
- [x] Allow operating postbit buttons
- [x] Add ACP setting

### DIFF
--- a/editpost.php
+++ b/editpost.php
@@ -71,7 +71,10 @@ $forum = get_forum($fid);
 
 if($thread['visible'] == 0 && !is_moderator($fid, "canviewunapprove") || $thread['visible'] == -1 && !is_moderator($fid, "canviewdeleted") || ($thread['visible'] < -1 && $thread['uid'] != $mybb->user['uid']))
 {
-	error($lang->error_invalidthread);
+	if($thread['visible'] == 0 && !($mybb->settings['showownunapproved'] && $thread['uid'] == $mybb->user['uid']))
+	{
+		error($lang->error_invalidthread);
+	}
 }
 if(!$forum || $forum['type'] != "f")
 {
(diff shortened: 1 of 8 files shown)


Eldenroot:
One of the most annoying thing in MyBB, looking forward! Welcome back effone!

Eldenroot:
Sure, I will in the evening. I am just curious - how users will know when their post/thread/attachment was approved by mods? I think that there is no way how to let them know. This should be considered (at least for MyBB 1.9) - right now you have to use 3rd party plugin to send a PM when the post/thread is approved by board staff.

WildcardSearch:
@effone and @Eldenroot is there no visual representation that the user's post/thread is unapproved?

To clarify, when the user themselves navigate to the unapproved post, is there anything to indicate that it has been unapproved? If not, I think there should be. Maybe just like soft delete where you can show the post to view it with a button?

@mybb/developers

Eldenroot:
@WildcardSearch - I agree - at least the background color of post should have different color or display a note.

effone:
As soft delete it also gets hidden with a button to show.
For visual representation it is light red background, same as for unapproved shown to mods.

WildcardSearch:
Great @effone I'll try to find time to test sometime today.

PR #3439, opened 2018-09-04
Enable attachments to be removed quickly using ajax

zygzagZ:
Implemented ajax handling for removing attachments, enabling users to quickly remove multiple attachments.

PR #3515, opened 2018-12-01
Fixes #3514 Show only unread private messages

### LINKED ISSUE #3514: Show only unread private messages
Implement a link and process to show only unread private messages.

effone:
Attempt to fix #3514

### DIFF
--- a/inc/functions_user.php
+++ b/inc/functions_user.php
@@ -725,8 +725,11 @@ function get_pm_folder_name($fid, $name="")
 
 	switch($fid)
 	{
-		case 1:
+		case 0:
 			return $lang->folder_inbox;
+			break;
+		case 1:
+			return $lang->folder_unread;
 			break;
 		case 2:
 			return $lang->folder_sent_items;
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -9862,7 +9862,7 @@ img { max-width: 100%; height: auto; }
 <br />
 <br />
 <hr size="1" />]]></template>
-		<template name="private" version="1813"><![CDATA[<html>
+		<template name="private" version="1820"><![CDATA[<html>
 <head>
 <title>{$mybb->settings['bbname']} - {$lang->private_messaging}</title>
 {$headerinclude}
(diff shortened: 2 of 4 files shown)


Eldenroot:
Works fine, thank you

euantorano:
This seems to work for me, but as it's a bigger change it could maybe do with some more testing. I'm going to assign it to 1.8.21 and then we'll get it merge pretty soon after 1.8.20 is released so that we have a full merge window to test it.

install/resources/mybb_theme.xml
```diff
@@ -9862,7 +9862,7 @@ img { max-width: 100%; height: auto; }
 <br />
 <br />
 <hr size="1" />]]></template>
-		<template name="private" version="1813"><![CDATA[<html>
+		<template name="private" version="1820"><![CDATA[<html>
```
euantorano: As we're planning on targeting 1.8.21 this will need bumped to `1821`.

install/resources/mybb_theme.xml
```diff
@@ -10293,7 +10293,7 @@ if(use_xmlhttprequest == "1")
 		<template name="private_nomessages" version="120"><![CDATA[<tr>
 <td colspan="6" class="trow1">{$lang->nomessages}</td>
 </tr>]]></template>
-		<template name="private_orderarrow" version="1800"><![CDATA[<span class="smalltext">[<a href="private.php?fid={$folder}&amp;sortby={$sortby}&amp;order={$oppsortnext}">{$oppsort}</a>]</span>]]></template>
+		<template name="private_orderarrow" version="1820"><![CDATA[<span class="smalltext">[<a href="private.php?fid={$fid}&amp;sortby={$sortby}&amp;order={$oppsortnext}">{$oppsort}</a>]</span>]]></template>
```
euantorano: This will also need to be `1821`.

inc/functions_user.php
```diff
@@ -725,9 +725,12 @@ function get_pm_folder_name($fid, $name="")
 
 	switch($fid)
 	{
-		case 1:
```
euantorano: As these numbers are used in multiple places, I would be tempted to `define()` some constants for them to prevent possible future bugs.

effone:
I sincerely didn't mess with the folder structure in database for old plugin compatibility. The new folder IDs are modified on the fly.
Tested in live forums for more than 2 weeks.

euantorano:
Ok, if you can add a constant for the magic numbers then I'd be happy to go for 1.8.20 and delay it for a couple of days so that we can install it on the MyBB Community and test it there to make 100% sure.

Eldenroot:
Yes, same here, testing this and others PRs (#3516 #3439 and #3525 ) from effone and others for more than 2 weeks, works fine. Quite useful addition.

inc/functions_user.php
effone: Are you sure? `$fid` has been used as is in multiple locations of templates.
A constant in template?

euantorano:
You won’t be able to use it in the templates, but it could be used elsewhere. It’s not a deal breaker, but I don’t like having to remember special numbers like 0, 1, 2.

Eldenroot:
@effone do you have any chance to commit requested changes to get this into 1820?

euantorano:
We can always change to using constants or something in a future release, as I said it's not a deal breaker and this seems to work otherwise.

PR #3516, opened 2018-12-02
Upgrade jQuery to 3.0.0

### LINKED ISSUE #3545: Upgrade jQuery to jQuery 3.0.0
We are currently using an old version of jQuery, we should update to a more modern version.

effone:
Fixes #3545.

Ran Migrate 1.4.1 Console Logger.
Resolved deprecated event binding method.

Known deprecation usage by third party plugins:
- `$.fn.load()` used by sceditor
- Synchronous XMLHttpRequest (`async: false`) by jeditable


euantorano:
Note that this will need to be documented in the release notes and release blog post, preferably in big bold letters as it may break some custom JS used by themes/plugins. CC @dvz

PR #3528, opened 2018-12-27
Syndication Fix + JSON Format Support

effone:
Outdated PR. Refer: #3682

### DIFF
--- a/inc/class_feedgeneration.php
+++ b/inc/class_feedgeneration.php
@@ -18,11 +18,11 @@ class FeedGenerator
 	public $feed_format = 'rss2.0';
 
 	/**
-	 * The XML to output.
+	 * The feed to output.
 	 *
 	 * @var string
 	 */
-	public $xml = "";
+	public $feed = "";
 
 	/**
 	 * Array of all of the items
@@ -45,7 +45,11 @@ class FeedGenerator
 	 */
 	function set_feed_format($feed_format)
 	{
-		if($feed_format == 'atom1.0')
+		if($feed_format == 'json')
+		{
+			$this->feed_format = 'json';
+		}
+		elseif($feed_format == 'atom1.0')
 		{
 			$this->feed_format = 'atom1.0';
 		}
@@ -75,8 +79,8 @@ class FeedGenerator
 		$this->items[] = $item;
 	}
 
-	/**
-	 * Generate and echo XML for the feed.
+	/** 
+	 * Generate the feed.
 	 *
 	 */
 	function generate_feed()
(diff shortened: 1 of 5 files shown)


Eldenroot:
I would like to see at least point 2 in the list above for MyBB 1.8.x branch. What do you think?

effone:
I have tested further, since the restricted forums are no way accessible by readers I think we can live with this. The concern what bothered me is if you are logged in the feed shows as per your permission, its good actually - coz it becomes dynamic to produce feed to you respecting your permission to the site the feed is being fetched from.

I am differing the point.

Eldenroot:
I have been testing this since yesterday. Personally I am not a fan of RSS and I do not use JSON format. Some users on my forum use it, I contacted them to test this feature and let me know. So far so good.

install/resources/mybb_theme.xml
```diff
@@ -6192,7 +6192,7 @@ if(use_xmlhttprequest == "1")
 <td class="{$class}">{$smilie['name']}</td>
 <td class="{$class}">{$smilie['find']}</td>
 </tr>]]></template>
-		<template name="misc_syndication" version="120"><![CDATA[<html>
+		<template name="misc_syndication" version="1820"><![CDATA[<html>
```
euantorano: This should be `1821`.

inc/class_feedgeneration.php
```diff
@@ -121,17 +135,42 @@ function generate_feed()
 			}
 			switch($this->feed_format)
 			{
-				// Output Atom 1.0 format feed.
+				// Output JSON formatted feed.
+				case "json":
+					++$serial;
+					$end = $serial < count($this->items) ? "," : "";
+					$item_id = explode('tid=', $item['link']);
```
euantorano: Wouldn't it be easier to just build an army for the entry, then `json_encode` the whole array? eg:

```php
$entry = array(
	'url' => $item['link'],
	'title' => $item['title'],
	...
);

if(!empty($item['author']))
{
	$entry['author'] = array(
		...
	);
}

$this->feed .= json_encode($entry);
$this->feed .= $end . "\n";
```

The same would apply to the feed header. If nothing else, it's probably easier to read with an array in my opinion.

effone: I understand what you say. The reason doing it that way is to maintain the indentation in the similar manner and inline with how it has been done already for xml feeds.
If you don't bother about manual beautifying and insist I can do the suggested change. Please confirm.

euantorano:
I’m not really bothered either way to be honest - both ways work after all.

effone:
Superseded by #3682

PR #3532, opened 2018-12-29
Update member.php

labrocca:
Reflected XSS exploit in the $code because it's not sanitized.

### DIFF
--- a/member.php
+++ b/member.php
@@ -1849,7 +1849,7 @@ if($mybb->input['action'] == "resetpassword")
 				break;
 		}
 
-		$code = $mybb->get_input('code');
+		$code = htmlspecialchars_uni($mybb->get_input('code'));
 
 		if(!isset($user['username']))
 		{


euantorano:
For future reference, security issues should be reported following the guide available here: https://mybb.com/get-involved/security/

I’m on mobile at the minute so I’ve only looked at the minimal diff that GitHub’s UI shows, but am I right in thinking this patch is sanitising input before inserting into the database rather than before displaying? If so, we prefer to sanitise before display so that raw values are always available in the database for inspection. Sanitising on display also fixes any existing data.

labrocca:
Sorry but I'm not sure if you are agreeing this is an exploit or not.   It is, I assure you.  A malformed link sent to a user and clicked is then executed.  btw, this was reported to me at HF, I confirmed it working and that it's branch MyBB.  If you need a working link PoC please LMK and I'll PM you at MyBB.

Look at the code, you can see that a few lines down the $template is called. The user input is not sanitized.  That's an XSS reflection attack. It's minor exploit but still should be fixed.

As for keeping this private.   I didn't discover the exploit.  I don't know the intent of the person who did report it to me.  They could be very malicious or they may believe in selling the exploit in private blackhat forums.  So I don't believe that keeping this private is good policy and the fix is easy.  Also it appears that policy for reporting security exploits from MyBB is a suggestion not a requirement.  

If I found this bug myself I would contact MyBB privately but this is not the case. I'm not taking credit for this.

euantorano:
I’m not disagreeing that it is an exploit at all, just saying that we usually prefer that people follow our responsible disclosure guidelines when disclosing issues or fixes rather than posting patches publicly with next to no description or explanation.

labrocca:
Oh I see what you mean about what you thought about the code.  

I will post at MyBB to discuss my thoughts on MyBB's "responsible disclosures guidelines".

PR #3536, opened 2019-01-02
Display username before bbname on profiles

kawaii:
This is something that has irked me since I started using MyBB, and is a change I always made in all of my custom templates.

It simply reverses the order of the `{{ trans('profile', memprofile.username) }}` and `{{ mybb.settings.bbname }}` variables so that the browser tab displays the text as;
```
Profile of andrewjs18 - MyBB 1.9 Demo
```
Instead of;
```
MyBB 1.9 Demo - Profile of andrewjs18
```
Why exactly? Because when you have more than one tab open the title text is truncated which leaves you with;
```
MyBB 1.9 Demo -  Profile o...
```
Which is entirely useless. As far as I'm aware this issue exists across MyBB in many places - so consider this a "concept PR" or a very small quality of life improvement at the slightest. Input from @justinsoltesz and @euantorano would be appreciated. :)

### DIFF
--- a/inc/views/base/member/profile.twig
+++ b/inc/views/base/member/profile.twig
@@ -1,7 +1,7 @@
 {% extends 'layouts/master.twig' %}
 
 {% block head %}
-    <title>{{ mybb.settings.bbname }} - {{ trans('profile', memprofile.username) }}</title>
+    <title>{{ trans('profile', memprofile.username) }} - {{ mybb.settings.bbname }}</title>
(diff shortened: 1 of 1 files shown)


euantorano:
Yes please. This has also annoyed me for way too long now.

Ben-MyBB:
Totally agree, very annoying.


justinsoltesz:
@kawaii Yes, I plan on going through and consistently applying `Page title - bbname`

PR #3541, opened 2019-01-04
Fixes #3540 Omit unwanted links / buttons

### LINKED ISSUE #3540: Omit unwanted links / buttons
There are few places where the unwanted links still appear, I believe those should be removed / modified.

- [x] "Subscribe to this thread" - appears for logged in users, but not for guests. Fine. But "Subscribe to this forum" appears for all; however clicking on it as a guest will show 'No permission' page.
- [x] "Ban this user in MCP / ACP" appears in every member's profile even if they are already banned. That can be removed or changed for already banned members to "Modify Ban / Unban".
- [x] "Open Buddy List" appear for those who don't have buddies.
- ~~"Mark All Forums Read" appears even if there is not a single unread post.~~ (deferred. lightbulbs buggy).
- [x] "Show Forum Team" link appears in Memberlist even if the Forum Team listing page is disabled.
- [x] Find all posts / Find all threads in profile even when there is no post / thread made by user. ~~Same for reputation details link. Same for Reputation report of Reputation-less user.~~ (Skipped reputation link as reputation can be 0 in case of equal positive and negative, thus; reputation 0 can also have report).
- [ ] Empty forum shows thread filter and all good thread controls along with postbit frame.
- ~~Showthread-> Moderator Tools : "Open/Close Thread" and "Stick/Unstick Thread" should be specific as per state like "Approve/Unapprove".~~ ( covered and to be addressed by #3654 )
- [x] PM / email links in own post / profile.
- [x] Report buttons for unreportable posts.

effone:
Attempt to fix #3540

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -5632,7 +5632,7 @@ if(use_xmlhttprequest == "1")
 </table>
 </div>
 </div>]]></template>
-		<template name="memberlist" version="1808"><![CDATA[<html>
+		<template name="memberlist" version="1820"><![CDATA[<html>
 <head>
 <title>{$mybb->settings['bbname']} - {$lang->member_list}</title>
 {$headerinclude}
@@ -5684,9 +5684,6 @@ if(use_xmlhttprequest == "1")
 </tr>
 {$users}
 </table>
-<div class="float_right" style="padding-top: 4px;">
-	<a href="showteam.php"><strong>{$lang->forumteam}</strong></a>
-</div>
 {$multipage}
 <br class="clear" />
 <br />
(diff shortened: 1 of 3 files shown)


install/resources/mybb_theme.xml
```diff
@@ -5684,9 +5684,6 @@ if(use_xmlhttprequest == "1")
 </tr>
 {$users}
 </table>
-<div class="float_right" style="padding-top: 4px;">
```
effone: Showteam link moved to footer and available globally.
This is redundant.

effone:
Will continue in common PR:
https://github.com/mybb/mybb/pull/3542

Issue #616, opened 2014-05-29
Move hard-coded words to language files


ATofighi:
- [x] modcp_modqueue_attachments_attachment Template -> Post: (fixed in #768 )
- [x] report_reasons Template -> If 'Other' please provide a reason for reporting this content. (fixed in #768 )
- [x] admin/forum/attachments.php: line 417 & 576 -> Scanning... (fixed in #768 )
- [x] managegroup_adduser template -> Search for a user (fixed in #768 )
- [x] member_register_referrer template -> Search for a user (fixed in #768 )
- [x] memberlist template -> Search for a user (fixed in #768 )
- [x] memberlist_search -> Search for a user (fixed in #768 )
- [x] modcp_banuser -> Search for a user (fixed in #768 )
- [x] modcp_warninglogs -> Search for a user (fixed in #768 )
- [x] private_send_autocomplete -> to select2 -> Search for a user (fixed in #768 )
- [x] private_send_autocomplete -> bcc select2 -> Search for a user (fixed in #768 )
- [x] search -> Search for a user (fixed in #768 )
- [x] usercp_editlists -> Search for a movie -> :o ( must change to user! ) (fixed in #768 )
- [x] usercp_editlists -> Search for a user (fixed in #768 )
- [x] install/upgrade.php lines 255/259

PR #3548, opened 2019-01-06
Language: use horizontal ellipsis instead of three dots or two dots

BKPepe:
I'm not a native English speaker, but I think two dots are wrong, what I have been able to find.
Also, I have replaced three dots (...) with the [horizontal ellipsis](https://en.wikipedia.org/wiki/Ellipsis#Computer_representations) (…), which is unicode character.

Sources:
https://english.stackexchange.com/questions/60452/usage-of-double-dots-is-it-formal
https://english.stackexchange.com/questions/443527/can-a-sentence-be-ended-with-two-dots


Eldenroot:
I agree, looks better

inc/languages/english/admin/forum_attachments.lang.php
```diff
@@ -72,11 +72,11 @@
 $l['step1of2_line1'] = "Please wait, the file system is currently being scanned for orphaned attachments.";
 $l['step2of2_line1'] = "Please wait, the database is currently being scanned for orphaned attachments.";
 $l['step_line2'] = "You'll automatically be redirected to the next step once this process is complete.";
-$l['scanning'] = 'Scanning..';
+$l['scanning'] = 'Scanning…';
```
effone: Why its 2 here?

BKPepe: Hard to tell. Maybe its some mistake, who left it there?

BKPepe: I looked at it and it was added by this commit to language files:
https://github.com/mybb/mybb/commit/9dc1d0a73fae4e2ca0775925e245c500bbd01dfa#diff-bd2eb23b594576096e0256c0f4d8d62d ( https://github.com/mybb/mybb/issues/616) from this commit https://github.com/MyBBIr/mybb/commit/44de55595b5b1fc4014d9b80e6aa9b0017ae93bb

This was there for a long time.

euantorano:
I believe this should use the HTML entity for compatibility, same as we use `&amp;` usually. The entity for this character seems to be `&hellip;`.

euantorano:
Superseded by #3650 which uses the HTML entity.

PR #3557, opened 2019-01-26
Add check for ajax in newthread and newreply

### LINKED ISSUE #3556: AJAX removal of attachment is not working
Related to #3547.

> 1. In a forum - Press 'Post Thread' button (which will take you to the post editor).
> 2. Under 'New attachment', press the 'Choose Files' button.
> 3. Select a file
> 4. Press 'Add Attachment' 
> *The attachment is now shown under 'Attachments'*
> 5. Press 'Remove' next to the file you've just selected (assume you selected the wrong file and want to pick a different file).
>
> Even though 'Remove' is pressed, the attachment is still shown under 'Attachments'.
> The attachment doesn't look like its been removed (although when you submit the post, the file isn't attached).

Reported [here](https://community.mybb.com/thread-221648-post-1322912.html#pid1322912).

euantorano:
Fixes #3556

### DIFF
--- a/editpost.php
+++ b/editpost.php
@@ -249,7 +249,7 @@ if($mybb->settings['enableattachments'] == 1 && $mybb->get_input('attachmentaid'
 		update_thread_counters($post['tid'], array('attachmentcount' => "-1"));
 	}
 
-	if($mybb->input['ajax'] == 1)
+	if($mybb->get_input('ajax', MyBB::INPUT_INT) == 1)
 	{
 		header("Content-type: application/json; charset={$lang->settings['charset']}");
 		echo json_encode(array("success" => true));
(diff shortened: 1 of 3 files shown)


woostarUK:
Patched. Fixes the original problem 👍, but:
Not sure if it's a bug, or just annoys me...

Shouldn't the 'Update Attachment' (button) refresh work automatically?

If you:

1. Add two attachments
2. Remove one of them
3. Add another attachment

it doesn't work as I would expect it too, without pressing 'Update Attachment' (which triggers a complete page refresh).

---------

Side note: Is the 'Add Attachment' button really necessary. You already confirm the file selected to upload with the dialog box. It feels like it should automatically ajax refresh once the confirm/submit/choose button is pressed.

PR #3571, opened 2019-02-05
Resolve #3530 Modern password hashing

### LINKED ISSUE #3530: Modern password hashing
Use [`\Illuminate\Hashing`](https://github.com/laravel/framework/tree/5.7/src/Illuminate/Hashing) and extensions for modern password hashing.

Codebase:
- modify `create_password()`, `verify_user_password()` wrapper functions to use MyBB's extension of `\Illuminate\Hashing\HashManager`,
- implement `bcrypt` password hashing (`Illuminate\Hashing\BcryptHasher`) as default,
- create `MybbHasher` hashing driver to handle old hashes,
- modify `verify_user_password()` to allow plugins overwrite parameters (e.g. to decrypt values before checking),
- modify `LoginDataHandler::complete_login()` to upgrade algorithm and parameters to default on the fly.

Database:
- extend `mybb_users.password` column length,
- add `mybb_users.password_algorithm` column for algorithm identification.

Defer to 1.9-compatible version of [DVZ Hash](https://community.mybb.com/mods.php?action=view&pid=976) for algorithm wrapping and encryption with intent to implement in subsequent branches; set `argon2id` as default algorithm once PHP 7.3 is required.

dvz:
Resolve #3530

### DIFF
--- a/inc/functions_user.php
+++ b/inc/functions_user.php
@@ -207,6 +207,7 @@ function create_password($password, $salt = false, $user = false)
 			$salt = generate_salt();
 		}
 
+		/** @var Illuminate\Hashing\HashManager $hashManager */
 		$hashManager = \MyBB\app('hash');
 
 		$driverName = $hashManager->getDefaultDriver();
@@ -251,6 +252,7 @@ function verify_user_password($user, $password)
 	}
 	else
 	{
+		/** @var Illuminate\Contracts\Hashing\Hasher $hashDriver */
 		$hashDriver = \MyBB\app('hash')->driver($parameters['user']['password_algorithm']);
 
 		return $hashDriver->check($parameters['password'], $parameters['user']['password'], [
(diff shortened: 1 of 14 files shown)


inc/functions_user.php
```diff
@@ -201,16 +201,24 @@ function create_password($password, $salt = false, $user = false)
...
 	else
 	{
+		// salt may be needed in mechanisms other than passwords
 		if(!$salt)
 		{
 			$salt = generate_salt();
 		}
 
-		$hash = md5(md5($salt).md5($password));
+		$hashManager = \MyBB\app('hash');
```
euantorano: Mind adding a `/** @var ...` comment here to help IDEs and such?

inc/init.php
```diff
@@ -229,6 +229,12 @@
 
 require_once __DIR__.'/src/bootstrap.php';
 
+MyBB\app()->instance('config', new \Illuminate\Config\Repository());
```
euantorano: This may make more sense inside `inc/src/bootstrap.php`, or a service provider.

dvz: I was thinking about introducing a proper service provider in 1.10 when config values would be moved & stored in the standardized `return []` format - in 1.9 additional config can be added to `$config` (which is why it's eventually unprefixed).

euantorano: It may make sense to introduce the service provider in 1.9 anyway - that way we can simply extend it in 1.10 and have the service provider available for plugin developers and such to start using sooner.

PR #3576, opened 2019-02-19
README: fix URL for MyBB logo

BKPepe:
Currently, the MyBB logo in the readme is not displayed, because it seems like the URL was changed from raw.github.com to raw.githubusercontent.com


euantorano:
Good catch. Thanks for the PR!

PR #3582, opened 2019-02-25
Moderators can edit soft deleted posts.

hlstriker:
Moderators should be able to edit soft deleted posts again as long as they also have permission to view deleted posts. It's a bit silly a mod can view the a soft deleted post but they cannot edit it. There are many instances in which modifying a soft deleted post will come in handy.

### DIFF
--- a/editpost.php
+++ b/editpost.php
@@ -49,7 +49,7 @@ else
 	$post = get_post($pid);
 }
 
-if(!$post || ($post['visible'] == -1 && $mybb->input['action'] != "restorepost"))
+if(!$post)
 {
 	error($lang->error_invalidpost);
 }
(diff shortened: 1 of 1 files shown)


dvz:
Community discussion: https://community.mybb.com/thread-219509.html

euantorano:
This doesn't seem to restore the `edit` button for soft deleted posts for moderators. You can edit soft deleted posts, but only by manually going to `editpost.php?pid=X`...

effone:
Its not as simple as the PR shows it to be.
The modification was to make it that way. A deleted thing can't be edited, by whomsoever, need a restore if edit is required. Why a deleted thing needs to be edited at the first place? If such a thing is at all required - Unapproving a thread should be the right way for further "secret operation".

hlstriker:
> 
> 
> This doesn't seem to restore the `edit` button for soft deleted posts for moderators. You can edit soft deleted posts, but only by manually going to `editpost.php?pid=X`...

This was only intended to allow the edit functionality to exist again. Our custom theme still has the edit button, the default theme will need it added back as well.

> 
> 
> Its not as simple as the PR shows it to be.
> The modification was to make it that way. A deleted thing can't be edited, by whomsoever, need a restore if edit is required. Why a deleted thing needs to be edited at the first place? If such a thing is at all required - Unapproving a thread should be the right way for further "secret operation".

If there is a way to unapprove a single post within a thread let me know since I'm unaware if that exists. However, if that feature doesn't exist you shouldn't need to unapprove an entire thread just to modify a deleted post within said thread.

euantorano:
You can unapprove a single post by checking the checkbox for the post, then scroll to the bottom of the page to the moderation tools and select “unapprove post”

hlstriker:
> 
> You can unapprove a single post by checking the checkbox for the post, then scroll to the bottom of the page to the moderation tools and select “unapprove post”

Thank you I had no idea. Everything effone said makes sense now after realizing you can do that.

PR #3586, opened 2019-02-28
New hook on recount_rebuild.php

demtor:
May be useful add a do_recount_rebuild hook in order to have directly included the post request check and the input page check, avoiding to rewrite them in the development of the plugin.


euantorano:
Any thoughts @mybb/developers?

PaulBender:
Sounds like a good idea.

PR #3595, opened 2019-03-05
Variadic hooks for plugin system

### LINKED ISSUE #3594: Plugin system should allow hooks with multiple arguments
The current plugin system allows you to pass a single argument to `run_hooks`. This leads to workarounds such as the following pseudo code:

```php
$args = [
'file' => &$file,
'line' => &$line,
];

$plugins->run_hooks('foo', $args);
```

I propose that we adapt `run_hooks` to be a variadic function that takes multiple arguments. We can then dispatch hooks using `call_user_func_array()` when there is more than 1 argument.

euantorano:
Fixes #3594 

This adds a new `MyBB\PluginSystem` class as there are some subtle differences:

- Methods are now `camelCase` rather than `snake_case`
- `run_hooks` now returns `void` rather than the passed argument - returning `$arguments` would return an array, which isn't what would be expected. It should instead return any value returned by the called function in the future...
- `hooks` and `currentHook` are now private, with public getter methods. These variables should not be changed outside the plugin system class

I've started writing some unit tests for the new plugin system class, and will be expanding upon these to increase coverage.

This PR touches some other files to:

- lock `Illuminate` dependencies to 5.7.x as 5.8 causes breakage (particularly for `MyBB\Application`) - a future PR will likely upgrade to 5.8
- Add `declare(strict_types = 1)` to all new code


maciejfikus:
If you wish I can send a PR with 5.8 dependencies as I solved it during implementing HttpKernel & middlewares about week ago. It is a bit messy tho, mostly empty methods to satisfy interface.
Also, please consider existing attaching main classes (such as $mybb) to IoC container in global.php. It would allow people to extend their copies using MVC without need to modifying core files.

Adding one new line between php tag and declare statement would be nice:
https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-import-statements

euantorano:
Sure, that would be very useful @maciejfikus. Note that it may take me a while to merge though, as my current priority is on rebasing the `develop/1.9` branch on top of `feature` - something I'm trying to do in chunks that is going to result in a lot of PRs over the coming days/weeks.

Regarding binding core classes, I've been considering this myself. My eventual goal is to move all of the core classes into the `\MyBB\` namespace (as this PR starts to do with the plugin system). We already register some of the main core classes which are needed by things like twig extensions here: https://github.com/mybb/mybb/blob/develop/1.9/inc/src/ServiceProvider.php#L16 - I'm not sure how many are missing at the moment though.

I'm more than happy to add a new blank line before `declare()`. I actually wasn't aware of that proposed standard. I'll open another PR for that today.

maciejfikus:
Okay, I will send a PR as soon as I will get a moment to clear the mess there. Propably during the week.

Yeah, I saw this provider, however it binds only fresh instance of i.e. MyBB class while another instance is already initialized in the global.php. Binding $mybb variable to container in the end of global.php, requiring it in the new page entrypoint, like new admin.php allows us to use it (authenticated user, whole session, everything) throughout whole "new" app. It is not the clearest solution, but it does the work for now as a bridge between old mybb and new one. Current way gives as just an empty \MyBB object etc.

I guess I will send just split that into two PR's, one for just 5.8 dependencies and one including concept of HttpKernel implementation (even though it is basically just adapted Laravel's one). 

Cheers :-)

euantorano:
@maciejfikus 

> Yeah, I saw this provider, however it binds only fresh instance of i.e. MyBB class while another instance is already initialized in the global.php. Binding $mybb variable to container in the end of global.php, requiring it in the new page entrypoint, like new admin.php allows us to use it (authenticated user, whole session, everything) throughout whole "new" app. It is not the clearest solution, but it does the work for now as a bridge between old mybb and new one. Current way gives as just an empty \MyBB object etc.

It binds `$GLOBALS['mybb']`, so as long as it's resolved after `global.php` then it should return the value created there I believe as it's created as a global. I haven't tested it though to be fair. We need to bind it early enough for it to be passed in to the Twig extensions where it's used, so getting it right is tough. I'll need to have a bit of a play and test it.

euantorano:
Added a blank line before `declare()`.

euantorano:
This is ready for review.

euantorano:
I am going to rebase this this weekend, then would like to hear some thoughts @mybb/developers @mybb/sqa so that I can merge it without everything breaking again.

PR #3605, opened 2019-03-11
Fixes #3520 Invisible Recaptcha - PW Reset Error

### LINKED ISSUE #3520: Invisible Recaptcha - PW Reset Error
A member has tried to reset his PW, but couldn't get past the email stage as it flags up "Please complete Recaptcha", but the Recaptcha doesn't appear

I was able to replicate and to solve I simply switched from Invisible Recaptcha to the normal Recaptcha

So I think there is an issue with Invisible Recaptcha on the reset PW page

Original thread: [Invisible Recaptcha - PW Reset Error](https://community.mybb.com/thread-220033.html)

effone:
Attempt to fix #3520

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -4787,7 +4787,7 @@ if(use_xmlhttprequest == "1")
 {$footer}
 </body>
 </html>]]></template>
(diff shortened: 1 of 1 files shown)


skmedix:
Tested and works fine 👍

PR #3622, opened 2019-03-15
Fixes #3610 - Disabling extensions - better description

### LINKED ISSUE #3610: BugDoc - Please fix documentation for Attachment Type
Hello there,

This is something you need to consider changing in the next 1.8.x maintenance release (in the event of 1.8.12 comes out).

If if admin user goes to ACP &gt;&gt; Home » Attachment Types and disables some extensions. The instructions on the page "mislead" the admin to think that disabling that extension only disables the ability to attach files, but it doesn't specify that exiting attachments wont' be available.

See: https://community.mybb.com/thread-210502.html

Original thread: [BugDoc - Please fix documentation for Attachment Type](https://community.mybb.com/thread-210522.html)

Eldenroot:
Fixes #3610 

If you have a better description in your mind, let me know :)

### DIFF
--- a/inc/languages/english/admin/config_attachment_types.lang.php
+++ b/inc/languages/english/admin/config_attachment_types.lang.php
@@ -6,7 +6,7 @@
  */
 
 $l['attachment_types'] = "Attachment Types";
-$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts.";
+$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts. Note: Disabling an attachment type will lead to removing all attachment of this type from all posts!";
 $l['add_new_attachment_type'] = "Add New Attachment Type";
 $l['add_attachment_type'] = "Add Attachment Type";
 $l['add_attachment_type_desc'] = "Adding a new attachment type will allow members to attach files of this type to their posts. You have the ability to control the extension, MIME type, maximum size and show a small icon for each attachment type.";
--- a/inc/languages/english/admin/config_attachment_types.lang.php
+++ b/inc/languages/english/admin/config_attachment_types.lang.php
@@ -6,7 +6,7 @@
  */
 
 $l['attachment_types'] = "Attachment Types";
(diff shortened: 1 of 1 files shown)


inc/languages/english/admin/config_attachment_types.lang.php
```diff
@@ -6,7 +6,7 @@
  */
 
 $l['attachment_types'] = "Attachment Types";
-$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts.";
+$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts. Note: Disabling an attachment type will lead to removing all attachment of this type from all posts!";
```
WildcardSearch: That verbiage is a accurate, but could be a little more concise. I'd suggest:

```suggestion
$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts. Note: Disabling an attachment type will remove all existing attachments of this type from the forum!";
```

Eldenroot:
@WildcardSearch thank you for your feedback, corrected :)

inc/languages/english/admin/config_attachment_types.lang.php
```diff
@@ -6,7 +6,7 @@
  */
 
 $l['attachment_types'] = "Attachment Types";
-$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts.";
+$l['attachment_types_desc'] = "Here you can create and manage attachment types which define which types of files users can attach to posts. Note: Disabling an attachment type will remove all existing attachments of this type from the forum!";
```
dvz: Looks like such attachments are made inaccessible, rather than deleted, and can be made available again when matching type is restored.

Eldenroot: So what is your suggestion?

dvz: How about:
```diff
- will remove all existing attachments of this type from the forum!
+ will make existing attachments of this type inaccessible.
```

Eldenroot: OK, better. Fixed

PR #3623, opened 2019-03-15
Fixes #3609 - Fulltext search and stopwords

### LINKED ISSUE #3609: Fulltext Search and Stopwords
I hope you consider this a bug but if not it's an improvement.

Currently MySQL defaults with fulltext stopwords.  These are words that do not get part of the index.

By default MyBB has a search add a + to each word you search so a search for "my setup" is actually "+my +setup".  

So any MyBB forum with fulltext search enabled, including this one, will show zero results anytime search has a word that's part of stopwords. 

And to add to that if you have a ft_min_word_len set in MySQL (my.cnf) it will also do 0 results.  

Here is default list of stopwords (36 words):
https://dev.mysql.com/doc/refman/5.7/en/...ch-indexes

I propose that MyBB does 2 things.
1. Hard code into the cleanwords functions inside functions_search.php an array to simply remove any default stopwords. Could also be put into a setting or table with the &#36;cache for adding, subtracting.

2. Remove any words from search under the setting of "Minimum Search Word Length" which is part of MyBB already.  Just remove the word.  A search for "my setup" should then just do a search for "setup".  

Returning zero results is bad.  Especially if you have 40 million posts.  

This should be relatively easy to fix and will improve search for MyBB.

Thank you.

EDIT: Btw, this is an issue with default MySQL and default MyBB.  No special setup or alteration is required for this behavior.

Original thread: [Fulltext Search and Stopwords](https://community.mybb.com/thread-221156.html)

Eldenroot:
Fixes #3609

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -272,6 +272,8 @@ function clean_keywords_ft($keywords)
 	// Separate braces for further processing
 	$keywords = preg_replace("#((\+|-|<|>|~)?\(|\))#s", " $1 ", $keywords);
 	$keywords = preg_replace("#\s+#s", " ", $keywords);
+	//Replaces less than 3 characters
+	$keywords = preg_replace('/(\b.{1,2}\s)/','',$keywords);
 
 	$words = array(array());
 
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -283,7 +283,10 @@ function clean_keywords_ft($keywords)
 	$min_word_length -= 1;
 	
 	// Replaces less than 3 characters
-	$keywords = preg_replace("/(\b.{1,{$min_word_length}}\b)/",'',$keywords);
+	$keywords = preg_replace("/(\b.{1,{$min_word_length}})(\s)|(\b.{1,{$min_word_length}}$)/", '$2', $keywords);
+	// Collapse multiple spaces
+	$keywords = preg_replace('/(\s)+/', '$1', $keywords);
+	$keywords = trim($keywords);
 
 	$words = array(array());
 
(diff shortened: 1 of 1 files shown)


euantorano:
This isn't a full fix for #3609 - it only implements suggestion 2. It does help though, so I'm going to merge it.

inc/functions_search.php
```diff
@@ -272,6 +272,8 @@ function clean_keywords_ft($keywords)
 	// Separate braces for further processing
 	$keywords = preg_replace("#((\+|-|<|>|~)?\(|\))#s", " $1 ", $keywords);
 	$keywords = preg_replace("#\s+#s", " ", $keywords);
+	//Replaces less than 3 characters
+	$keywords = preg_replace('/(\b.{1,2}\s)/','',$keywords);
```
euantorano: This should use the setting for "Minimum Search Word Length" (`$mybb->settings['minsearchword']}` and fall back to sensible defaults if it is set to `0` (or a value less than mySQL's stop words default length (4 for full text).

euantorano:
I've pushed a minor change to this PR to use the `minsearchword` setting rather than forcing `2`. I also changed the regex slightly based on some testing.

euantorano:
@mybb/developers Any chance somebody else can review and test given the changes I've made? This seems to work on my test install with mYSQL - perhaps @kawaii can double check Postgres?

Eldenroot:
@euantorano - thank you for your time and improvements :) yes, this fixes only one issue but still better than nothing.

kawaii:
@euantorano this issue doesn't apply to PostgreSQL, but I can confirm applying the patch does not break searching or cause any unexpected behaviour.

kawaii:
One minor note I wasn't sure about however:
```
&highlight=my+old+friend
```
Results in a highlight like this:
image
Should stopwords also be highlighted?

euantorano:
I think that behaviour is fine since it shows what’s actually matched.

PR #3648, opened 2019-05-16
Hooks position improvement

demtor:
The two main pages of user CP (edit options and edit profile) do not give the chance to integrate new parameters. Every plugin that want to add fields in this two pages needs to run a separate query from the "global" one that MyBB already do. There may be a workaround (doing a nesteed add_hook, first one to the _start hook and from here call a datahandler hook) but imho it is not that good to see since in fact is a workaround and it is detached from the logic that MyBB adopts (also you are going to write and execute extra code that can be automatically done if the hook is correctly positioned).

Some _start hooks are literally positioned at the start of the action, when they could be positioned after important declarations or checks (ex. permission checks) that may be useful to have already included.


effone:
This is good.
I remember `xmlhttp` hook starts before the forum close check. I felt it should be after.
Can this be addressed in this PR?

Eldenroot:
#3872

Eldenroot:
@demtor - please can you fix conflicts? Thank you, if not let me know.

Eldenroot:
It seems to be OK! Thx!

@euantorano

usercp.php
```diff
@@ -226,8 +228,8 @@
 		"birthdayprivacy" => $mybb->get_input('birthdayprivacy'),
 		"away" => $away,
 		"profile_fields" => $mybb->get_input('profile_fields', MyBB::INPUT_ARRAY)
-	);
-	foreach(array('icq', 'skype', 'google') as $cfield)
+	));
+	foreach(array('icq', 'yahoo', 'skype', 'google') as $cfield)
```
euantorano: Yahoo has been removed from contact feeds recently, this needs removed here.

PR #3649, opened 2019-05-16
Use `my_serialize` for cache and update PHPDoc

euantorano:
The PHPDoc for `datacache::update()` wrongly states the content can only be a string when actually arrays and other types are supported. This PR corrects the documentation comment.

It also replaces `serialize` and `unserialize` with our `my_serialize()` and `my_unserialize()` functions for consistency with the rest of the code base. I don't foresee this causing any problems unless anybody is purposefully storing objects in the cache, which isn't something I've personally seen done before...


kawaii:
Seems to work. Tested using `db` and `memcached` cache backends - everything still functioning as normal.

PR #3650, opened 2019-05-18
Language: use horizontal ellipsis instead of three dots or two dots

euantorano:
This updates #3548 to use `&hellip;` rather than the actual character.


BKPepe:
Wow. I don't know how did you merge it, but you did it completely wrong. There are so many ways how to do it better.

Looking at the commit: 
https://github.com/mybb/mybb/commit/9001292c4186a498d973c369cccdf18420f7f2eb

Now it seems that you authored it. No co-authored, not even included in the commit message. For the next time, I would appreciate if you can ping me.

However, honestly, I'm not sure if the HTML entity is the right thing to do in all files. For example, looking at:
https://stackoverflow.com/questions/6536182/how-to-handle-horizontal-ellipsis-three-dots-character-in-xml-output-through-p

It's not so easy afterall. ;-)

euantorano:
The link you included is talking about XML rather than HTML. We use entities in all other language strings which is why I made the change. I did mention it on the original PR a while back. 

Sorry for not altering the existing PR - I did a mass find/replace which was easier to do by checking out your branch, altering it and pushing it to my fork. Somewhere in that process it seems to have lost your original commit.

PR #3653, opened 2019-05-18
Fix 3647 - add missing parameters to modify_column() and rename_column() in DB classes

### LINKED ISSUE #3647: Incompatible arguments of DB_Base::modify_column() and DB_Base::rename_column() implementations
Accepted arguments for [PostgreSQL implementations](https://github.com/mybb/mybb/blob/mybb_1820/inc/db_pgsql.php) of [`DB_Base::modify_column()`](https://github.com/mybb/mybb/blob/mybb_1820/inc/db_base.php#L380) and [`DB_Base::rename_column()`](https://github.com/mybb/mybb/blob/mybb_1820/inc/db_base.php#L390) `ALTER COLUMN` methods deviate from the other ones, requiring additional database engine checks to call the method with additional parameters for `NULL` and `DEFAULT` attributes in the case of PostgreSQL.

These methods can be extended in other implementations to similarly accept such attributes in separate arguments for compatibility.

euantorano:
Fixes #3647 - Incompatible arguments of DB_Base::modify_column() and DB_Base::rename_column() implementations

### DIFF
--- a/inc/db_base.php
+++ b/inc/db_base.php
@@ -375,19 +375,25 @@ interface DB_Base
 	 *
 	 * @param string $table The table
 	 * @param string $column The column name
-	 * @param string $new_definition The new column definition
+	 * @param string $new_definition the new column definition
+	 * @param boolean $new_not_null Whether to drop or set a column
+	 * @param boolean $new_default_value The new default value (if one is to be set)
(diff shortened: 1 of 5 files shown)


dvz:
Added implementations for the added arguments; will need testing on all RDBMSs.

kawaii:
Patch applied, nothing exploded so looks good to me.

PR #3655, opened 2019-05-20
Fixes #3540 Omit unwanted links / buttons (v.2)

### LINKED ISSUE #3540: Omit unwanted links / buttons

effone:
Second phase attempt to fix #3540
_(finger crossed)_

### DIFF
--- a/global.php
+++ b/global.php
@@ -531,10 +531,10 @@ if($mybb->user['uid'] != 0)
 
 	$buddylink = $searchlink = $pmslink = '';
 
-    if(!empty($mybb->user['buddylist']))
-    {
-        eval('$buddylink = "' . $templates->get('header_welcomeblock_member_buddy') . '";');
-    }
+	if(!empty($mybb->user['buddylist']))
+	{
+		eval('$buddylink = "' . $templates->get('header_welcomeblock_member_buddy') . '";');
+	}
     
 	if($mybb->usergroup['cansearch'] == 1)
 	{
--- a/global.php
+++ b/global.php
@@ -529,14 +529,19 @@ if($mybb->user['uid'] != 0)
 	// Format the welcome back message
 	$lang->welcome_back = $lang->sprintf($lang->welcome_back, build_profile_link(htmlspecialchars_uni($mybb->user['username']), $mybb->user['uid']), $lastvisit);
 
-	$searchlink = '';
+	$buddylink = $searchlink = $pmslink = '';
+
+    if(!empty($mybb->user['buddylist']))
+    {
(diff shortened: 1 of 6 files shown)


global.php
```diff
@@ -529,14 +529,19 @@
 	// Format the welcome back message
 	$lang->welcome_back = $lang->sprintf($lang->welcome_back, build_profile_link(htmlspecialchars_uni($mybb->user['username']), $mybb->user['uid']), $lastvisit);
 
-	$searchlink = '';
+	$buddylink = $searchlink = $pmslink = '';
+
+    if(!empty($mybb->user['buddylist']))
```
euantorano: Indentation here doesn't seem right.

member.php
```diff
@@ -2335,8 +2335,14 @@
 	$findposts = $findthreads = '';
 	if($mybb->usergroup['cansearch'] == 1)
 	{
-		eval("\$findposts = \"".$templates->get("member_profile_findposts")."\";");
-		eval("\$findthreads = \"".$templates->get("member_profile_findthreads")."\";");
+		if((int)$memprofile['postnum'])
```
euantorano: This cast seems like it's redundant?

effone: Done

global.php
effone: Done

PR #3664, opened 2019-05-23
Fix #3663 - Moved thread link name not changing properly

### LINKED ISSUE #3663: Moved thread link name not changing properly
I was doing some testing with my friend and we noticed that when you move thread to other forum and change name of "Moved" link that was created in place of that thread you will update original thread name, but not the Moved link name.

I looked up in few files and found that there's no change of "Moved link" thread. Only original thread has been written in these files.
I will check what would happen when you move it few times and have few "Moved" links lying around.

For me it is a bug because you can use jeditable on moved threads and it will allow you to edit it's name, but when you accept and refresh page it's name going to be the same as before change.

I might've an idea how to change this to include all "Moved" links, but firstly I'd like to hear from others if they too consider this to be a bug.

Original thread: [Moved thread link name not changing properly](https://community.mybb.com/thread-223132.html)

MinusPL:
resolves #3663

There's new code for updating thread subject in moved thread links.
From my testing it works both in editor and in inline edit of thread name (in both ways - if you change on moved thread link it will update original thread's name and vice versa).
Please check if everything is working as it should, also please tell me if there's something wrong here (wanna get better at programing in php to make better contributions to MyBB in future).


Ben-MyBB:
The code solves the issue correctly, just need a developer to check the code standards before merging.

@euantorano @mybb/developers

inc/datahandlers/post.php
```diff
@@ -1879,6 +1879,20 @@ function update_post()
...
+
+			// Update any moved thread links to have corresponding new subject.
+			if(isset($post['subject']))
+			{
+				$query = $db->simple_select("threads", "tid, closed", "closed='moved|".$this->tid."'");
+				if($db->num_rows($query) > 0)
+				{
+					while($result = $db->fetch_array($query))
+					{
+						$update_data['subject'] = $db->escape_string($post['subject']);
```
dvz: This definition can be moved outside the `while` loop, since values will be the same

MinusPL: Changed that, thanks for your suggestion.

PR #3667, opened 2019-05-24
Fix #3666 - PM search - incorrect folder

### LINKED ISSUE #3666: PM search - incorrect folder
When you search PMs, incorrect folders are displayed. It seems it reports "Unread" for all messages in "Inbox".

Original thread: [PM search - incorrect folder](https://community.mybb.com/thread-222668.html)

MinusPL:
List of changes:
- Added moving read messages to inbox folder
- Changed how inbox selects messages to show
- Changed calculation of total amount of messages depending on which folder we're looking into

Now some thoughts:
- This system is not properly designed. I've had to make some changes to work around few weird things.
- To now this system of PMs wasn't using inbox folder for anything - it was just there to be there.
- It was using only message status to determine if message was read or not. From now it also uses folders to determine that.

To be honest this system is in need of changes. Most likely complete redesign and rework of about  half to almost all code. For now and for keeping it as it should work these changes I've created should make it possible to fix this issue without changing messaging system.
Remember that this fix is only a way to make it work properly.
Also there's one important note.

To properly show people their messages it is needed to update whole privatemessages table.
It need to change folder from 1 to 0 of all messages that are currently in folder 1 and have status 1 (So it means that we move read messages from unread folder back to inbox folder).

I've checked it and it seems to work, but please test it and look through code to avoid unnecessarry bugs.


effone:
Refer my comment for better understanding:
https://github.com/mybb/mybb/issues/3666#issuecomment-495842409

PR #3674, opened 2019-05-29
Fix #3656 - SQL Error on Post Update With Removed Attachment

### LINKED ISSUE #3656: SQL Error on Post Update With Removed Attachment
```

SQL Error:
0 - ERROR: invalid input syntax for integer: "" LINE 3: WHERE aid='' ^
Query:
DELETE FROM mybb_attachments WHERE aid=''
```
Software:
- MyBB 1.8.20
- PostgreSQL 11.2

Steps to reproduce issue:
- Add an attachment to a post and publish the post
- Edit the post, remove the attachment and save the updated post

euantorano:
Fixes #3656 

There are some other issues with editing posts, especially when JS is disabled. I'll open another issue regarding these.

### DIFF
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -21,7 +21,7 @@ function remove_attachment($pid, $posthash, $aid)
 	global $db, $mybb, $plugins;
 	$aid = (int)$aid;
 	$posthash = $db->escape_string($posthash);
-	if($posthash != "")
+	if(!empty($posthash))
 	{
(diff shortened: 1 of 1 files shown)


kawaii:
This patch resolves the issues I experienced in #3656, however as a side effect it seems that once the attachment is removed, clicking the `Update Post` button sends me to the top of the page, where I then need to scroll down and click it again to commit my edit.

euantorano:
Yeah, see the other issue I opened about some other issues I found with editing posts. This issue has existed since we added the AJAX management of attachments I believe.

kawaii:
@euantorano alright let's go ahead and merge this one then since it fixes the root issue.

PR #3708, opened 2019-06-10
Fix #3707 - "&hellip;" is displayed in redirect button

### LINKED ISSUE #3707: "&hellip;" is displayed in "Redirecting" button
There is a regression in 1.8.21 - "&hellip;" is displayed in "Redirecting" button in ACP.

image

euantorano:
This should fix #3707, but I haven't tested it yet.

### DIFF
--- a/admin/modules/tools/recount_rebuild.php
+++ b/admin/modules/tools/recount_rebuild.php
@@ -431,7 +431,6 @@ function check_proceed($current, $finish, $next_page, $per_page, $name, $name2, 
 					var textElement = $('<div/>').html('{$lang->automatically_redirecting}');
 				
 					button.val(textElement.text());
-					button.html(\"{$lang->automatically_redirecting}\");
 					button.attr(\"disabled\", true);
 					button.css(\"color\", \"#aaa\");
 					button.css(\"borderColor\", \"#aaa\");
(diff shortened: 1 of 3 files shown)


euantorano:
Good catch, that might be right.

JordanMussi:
Yeah just tested and that works 👍

euantorano:
Change applied, and form change from 1.8.21 applied to `admin/modules/forum/management.php` as it was previously missed there.

admin/modules/tools/recount_rebuild.php
```diff
@@ -426,8 +426,12 @@ function check_proceed($current, $finish, $next_page, $per_page, $name, $name2,
...
 			$(function() { 
 				var button = $(\"#proceed_button\"); 
-				if(button.length > 0) {
-					button.val(\"{$lang->automatically_redirecting}\");
+				if (button.length > 0) {
+					// create a temporary div element to render the text within, unescaping HTML entities
+					var textElement = $('<div/>').html('{$lang->automatically_redirecting}');
+				
+					button.val(textElement.text());
+					button.html(\"{$lang->automatically_redirecting}\");
```
JordanMussi:
```suggestion

```

euantorano:
It might make sense to create a function actually, yeah. I’ll do that tonight.

PR #3715, opened 2019-06-12
Fix user postnum calculation in delete_user_posts

demtor:
Exclude soft deleted posts and drafts on the postnum calculation in the delete_user_posts function.
It can bring to a SQL Error:

> SQL Error: 1690 - BIGINT UNSIGNED value is out of range in 'mybb_users.postnum- x'


euantorano:
This is conflicted unfortunately :(

Eldenroot:
#3871

Eldenroot:
It seems to be OK. 

@euantorano

PR #3716, opened 2019-06-12
Fix member_referrals tables

demtor:


effone:
While I appreciate your contribution; please make an issue and then place a relevant PR to resolve the same. If you are unsure about the issue and need discussions / opinions then the community forums is the right place to place the issue. Upon agreement the same will be pushed to Github.

PR #3722, opened 2019-06-14
Poll Voting Efficiency Fix

labrocca:
Removes a query against the non-indexed `poll` column in the `threads` table in favor of get_thread() using the poll tid.

### DIFF
--- a/global.php
+++ b/global.php
@@ -206,7 +206,7 @@ if(in_array($current_page, $valid))
 	// If we're accessing poll results, fetch the forum theme for it and if we're overriding it
 	else if(isset($mybb->input['pid']) && THIS_SCRIPT == "polls.php")
 	{
-		$query = $db->simple_select('threads', 'fid', "poll = '{$mybb->input['pid']}'", array('limit' => 1));
+		$query = $db->query("SELECT t.fid FROM ".TABLE_PREFIX."polls p LEFT JOIN ".TABLE_PREFIX."threads t ON (t.tid=p.tid) WHERE p.pid = '{$mybb->input['pid']}' LIMIT 1");
 		$fid = $db->fetch_field($query, 'fid');
 
 		if($fid)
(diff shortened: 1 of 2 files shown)


labrocca:
This was added to HF because we found polls loaded very slowly when both voting and showing results.  This was because the poll column in the mybb_threads table is not indexed so a large forum with a lot of threads could have slow query issues.  This fixes that with simple code.

global.php
```diff
@@ -206,7 +206,7 @@
 	// If we're accessing poll results, fetch the forum theme for it and if we're overriding it
 	else if(isset($mybb->input['pid']) && THIS_SCRIPT == "polls.php")
 	{
-		$query = $db->simple_select('threads', 'fid', "poll = '{$mybb->input['pid']}'", array('limit' => 1));
+		$query = $db->query("SELECT t.fid FROM ".TABLE_PREFIX."polls p LEFT JOIN ".TABLE_PREFIX."threads t ON (t.tid=p.tid) WHERE p.pid = '{$mybb->input['pid']}' LIMIT 1");
```
euantorano: Shouldn't this be an `INNER JOIN` rather than a `LEFT JOIN`?

labrocca: So make it INNER JOIN if you think that's better.

euantorano: Ok. Not everybody likes to have their PRs changed without prior discussion or warning.

euantorano: Also it looks like this PR isn't commits from maintainers. If you enable this, I can edit this PR directly so that it shows up in the log correctly: https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

labrocca: PR is from Xerotic so he will have to do the the configuration change.

Eldenroot: Please do it )

euantorano:
I can always merge this then edit it afterwards, which may work out easier.

PR #3727, opened 2019-06-16
Fix #3723 PostgreSQL `LIMIT` correction code

### LINKED ISSUE #3723: PostgreSQL `LIMIT` correction code doesn't apply to queries ending in `;`
The PostgreSQL database driver has a special check in the `query()` method to replace `LIMIT offset, limit` syntax with the `LIMIT limit OFFSET offset` syntax:

```
$string = preg_replace("#LIMIT (\s*)([0-9]+),(\s*)([0-9]+)$#im", "LIMIT $4 OFFSET $2", trim($string));
```

However, this block of code will only apply for queries that are not terminated by a semicolon `;`.

Matslom:
Fixing #3723

### DIFF
--- a/inc/db_pgsql.php
+++ b/inc/db_pgsql.php
@@ -288,7 +288,7 @@ class DB_PgSQL implements DB_Base
 	{
 		global $mybb;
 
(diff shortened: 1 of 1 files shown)


kawaii:
LGTM.

PR #3730, opened 2019-06-17
Sceditor set http:// in placeholder

### LINKED ISSUE #3721: Change video default link value "http://" to placeholder
I think that default value "http://" for links in video dropdown in editor is making some problems, because when we copy link with http already in it will add another one in front of original link and break it.

It would be better to change in dropdown value to placeholder with same value. This way we still have information that this is field for links, but we don't get duplicated http protocol in front of our link.

chack1172:
Fixed #3721


jscripts/bbcodes_sceditor.js
```diff
@@ -474,7 +474,7 @@ $(function ($) {
 				'</div>' +
 				'<div>' +
 				'<label for="link">' + editor._('Video URL:') + '</label> ' +
-				'<input type="text" id="videourl" value="http://" />' +
+				'<input type="text" id="videourl" placeholder="http://" />' +
```
WildcardSearch, line 561: As mentioned [here](https://github.com/mybb/mybb/issues/3721#issuecomment-502755183) the `placeholder` attribute is for instructions, and `http://` doesn't fit the bill. I would suggest creating a new language entry with something like `paste URL here` instead of using the placeholder attribute for an unintended purpose.

chack1172: Well sceditor already use it for normal links and images, if we change it here we must change it also for the other fields.

WildcardSearch: The description of the `placeholder` attribute:

>Specifies a short hint that describes the expected value of the input field

No matter what is going on elsewhere in the code, we do not want to introduce bad design into our software.

WildcardSearch (changes requested):
Improper use of `placeholder` attribute. Change needed.

Ben-MyBB:
@chack1172

Eldenroot:
So what is the result? Just add a lang var mentioned above?

jscripts/bbcodes_sceditor.js
Sama34: I think this falls into the definition of _hint_. If no one decides to push a different approach I think this could be merged.

Sama34 (approved):

jscripts/bbcodes_sceditor.js
Eldenroot: One issue is with https, it is missing in the "hint"

Sama34: But `http` is a _hint_ that `https` could be used. I mean, if somebody wants to create a new PR, go ahead, otherwise I see no reason to reject this contribution.

euantorano:
Yeah, I think it’s best to just go with this.

Eldenroot:
What about to change it to Http(s):// ?

euantorano:
> What about to change it to Http(s):// ?

Other inputs already just use `http://` I believe, and I'd rather keep them the same.

Ben-MyBB:
@chack1172 Could you fix the conflict please?

chack1172:
Sorry, i didn't understand how we should change it...

Eldenroot:
@chack1172 - leave it as it is, just resolve merge conflict :)

chack1172:
Ok, I'll fix it after dinner.

PR #3735, opened 2019-06-27
Fixed error in thread.twig

billthefish:
Found a small error while setting up a mybb 1.9 test platform


euantorano:
Thanks for the contribution, that was an easy one to miss :smile:

PR #3749, opened 2019-07-17
resolves #3745, resolves #3632

### LINKED ISSUE #3632: UI inconsistency: moving threads from forum view vs topic view
Reporting as requested here.

Inline Moderation ("forum view") allows to move topic(s), only.

Moderation Options ("topic view") allows to move/copy/move with redirects

As far as I understand, all 3 options should really be available in either case.

Original thread: [UI inconsistency: moving threads from forum view vs topic view](https://community.mybb.com/thread-222449.html)

### LINKED ISSUE #3745: New hook in newreply.php page for lastposts view
There is no hook in Mybb docs for management content in lastposts section in newreply.php

screenshot-www mageguildwars pl-2019 07 01-00-25-24

In my plugin I need to modify content of posts in this section, but I don't see any way to do it. All **newreply_** hooks don't work.

It will be grateful if someone could confirm if I have right and devteam change it in future.

callumthomson:
resolves #3745
resolves #3632


callumthomson:
RE #3632 :
- I also created a new hook `moderation_do_multimovethreads`, as there wasn't one already.
- Instead of modifying `Moderation::move_threads`, I've iterated over `Moderation::move_thread`, and unless there is a good reason to have both methods, it seems smart to remove move_threads, to avoid repeating code and, the method is weaker anyway as it does not support redirects / copying.

install/resources/mybb_theme.xml
```diff
@@ -8455,6 +8455,14 @@ if(use_xmlhttprequest == "1")
 <td class="trow1"><strong>{$lang->new_forum}</strong></td>
 <td class="trow2">{$forumselect}</td>
 </tr>
+<tr>
```
euantorano: The parent `<template>` tag (`moderation_inline_movethreads`) should have its `version` attribute updated to `1822`.

callumthomson: Requested changes have been made

euantorano: Brilliant, thanks!

PR #3750, opened 2019-07-18
Configure PHPStan as part of MyBB 1.9's CI #3551 PR

### LINKED ISSUE #3551: Configure PHPStan as part of MyBB 1.9's CI
[PHPStan](https://github.com/phpstan/phpstan) is a static analysis tool for PHP that can help catch common errors in code.

We should add it as part of our CI pipeline in MyBB 1.9 to catch obvious errors.

We should likely only target the `inc/src` and `tests` folders, but it's worth trying it on the full codebase first to see how bad it actually is...

ks129:
I hope this is it. Please review it. Close #3551


composer.json
```diff
@@ -57,3 +86,16 @@
         "sort-packages": true
     }
 }
+
+{
```
euantorano: This is causing the CI to fail due to the JSON being invalid (missing `,` after the above closing `}`).

composer.json
```diff
@@ -27,13 +27,42 @@
         "illuminate/events": "5.7.*",
         "illuminate/hashing": "5.7.*",
         "illuminate/routing": "5.7.*",
-        "twig/twig": "^2.4"
+        "twig/twig": "^2.4",
```
euantorano: This seems like a lot of extra dependencies just to add PHPStan. I would have expected there to only be changes to `require-dev` and that should just be:

```json
"phpstan/phpstan": "^0.11.12"
```

inc/src/phpstan.neon
```diff
@@ -0,0 +1 @@
+
```
euantorano: This config file should at least specify the paths to analyse (`inc/src`)

ks129:
@euantorano It's now done

euantorano:
Thanks, I will have another look after dinner.

composer.json
```diff
@@ -55,5 +56,15 @@
         "bin-dir": "inc/vendor/bin",
         "optimize-autoloader": true,
         "sort-packages": true
+    },
+    "scripts": {
+        "phpstan": "phpstan analyse src/ -c inc/src/phpstan.neon --level=0 --no-progress -vvv",
```
euantorano: It looks like using the same name for these scripts causes some issues:

```
Key phpstan is a duplicate in ./composer.json at line 62
```

Might be best to make them `phpstan-0`, `phpstan-1`, etc.

.travis.yml
```diff
@@ -44,3 +44,13 @@ script:
   - inc/vendor/bin/phpunit
   - inc/vendor/bin/phpcs
   - composer validate --strict
+
+test:
```
euantorano: This doesn't actually seem to be getting ran by Travis as far as I can tell: https://travis-ci.org/mybb/mybb/jobs/561740143

It might be easiest to just add it as a script step:

```
- vendor/bin/phpstan analyse -l max -c inc/src/phpstan.neon src --no-interaction --no-progress
```

inc/src/phpstan.neon
```diff
@@ -0,0 +1,5 @@
+parameters:
```
euantorano: I'm not sure if this is the best place for this config file. The idea behind `inc/src/` is to hold source code. All of our other config (`phpcs`, `phpunit`, etc.) is in the root `/` so it may make more sense too live there.

ks129:
@euantorano everything is now OK, only 2 allowed failures due PHP version conflict.

PR #3758, opened 2019-07-25
Fix #3757 attachment paging links missing forum info in AdminCP

### LINKED ISSUE #3757: Lost 'forum' query field during multipaging in AdminCP>Forum>Attachments>Find Attachments
Sorry for mistakenly submit the initial issue. My bad.

Here comes the problem. In the AdminCP > Forums & Posts > Attachments > Find Attachments, if a query is issued with forums selected, and the returned records are multipaged, the link of any page doesn't include a `&forum[]=` string in any subsequent GET request. Thus, all the following paging results are attachments queried in the board scope.

A quick fix is applicable. Should I post a diff first or pull a request after it's been confirmed?

yuliu:
Resolve #3757  .

### DIFF
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -870,7 +870,7 @@ if(!$mybb->input['action'])
 				{
 					foreach($mybb->input['forum'] as $fid)
 					{
-						$pagination_url .= "&forum[]=".urlencode($fid);
+						$pagination_url .= "&forum[]=".(int)$fid;
 					}
 				}
 				$pagination = draw_admin_pagination($mybb->input['page'], $mybb->input['perpage'], $num_results, $pagination_url);
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -866,7 +866,7 @@ if(!$mybb->input['action'])
 						$pagination_url .= "&{$var}=".urlencode($mybb->input[$var]);
 					}
 				}
-				if($mybb->input['forum'])
+				if(is_array($mybb->input['forum']) && !empty($mybb->input['forum']))
 				{
 					foreach($mybb->input['forum'] as $fid)
 					{
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -858,12 +858,19 @@ if(!$mybb->input['action'])
 			if($num_results > $mybb->input['perpage'])
 			{
 				$pagination_url = "index.php?module=forum-attachments&amp;results=1";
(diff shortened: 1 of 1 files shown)


WildcardSearch:
Looks good. I'll try to pull and test this if I get time before I go out of town.

admin/modules/forum/attachments.php
```diff
@@ -858,14 +858,21 @@ function() {
...
-				$pagination_vars = array('perpage', 'sortby', 'order', 'filename', 'mimetype', 'username', 'fid', 'downloads', 'downloads_dir', 'dateuploaded', 'dateuploaded_dir', 'filesize', 'filesize_dir');
+				$pagination_vars = array('perpage', 'sortby', 'order', 'filename', 'mimetype', 'username', 'downloads', 'downloads_dir', 'dateuploaded', 'dateuploaded_dir', 'filesize', 'filesize_dir');
 				foreach($pagination_vars as $var)
 				{
 					if($mybb->input[$var])
 					{
 						$pagination_url .= "&{$var}=".urlencode($mybb->input[$var]);
 					}
 				}
+				if($mybb->input['forum'])
```
WildcardSearch: Might be good to change to this:

```suggestion
				if(is_array($mybb->input['forum']) && !empty($mybb->input['forum'])
```

admin/modules/forum/attachments.php
```diff
@@ -858,14 +858,21 @@ function() {
...
+				{
+					foreach($mybb->input['forum'] as $fid)
+					{
+						$pagination_url .= "&forum[]=".urlencode($fid);
```
WildcardSearch: Since `fid`s are integers, we really just need to clean them to add them to the URL, nothing should need to be encoded.

```suggestion
						$pagination_url .= "&forum[]=".(int)$fid;
```

yuliu: I think it needs neither urlencode nor a type cast, since if any error exists, we can't get to this point. But it's OK we ensure this variable is an integer.

admin/modules/forum/attachments.php
yuliu: <del>
If we use `is_array()` on `$mybb->input['forum']`, we must ensure `$mybb->input['forum']` exists, thus an `isset()` should be called before `is_array()`. We will get an `Undefined variable` notice with your modification, if no forum is selected in the query form, I suppose.

Don't know how to make a change to your change, so I leave a comment. Hope you have time to check it, @WildcardSearch.
</del>

Oh, I've just seen that MyBB suppresses NOTICE error types. This code works!

yuliu:
It's the first time for me to use the commit option and resolve feature in a PR. Please check if everything's going right.

Eldenroot:
Applied to my board, testing :)

PR #3759, opened 2019-07-26
Fix #3757 attachment paging links missing forum info in AdminCP

### LINKED ISSUE #3757: Lost 'forum' query field during multipaging in AdminCP>Forum>Attachments>Find Attachments

yuliu:
Resolve #3757.

The URL looks better now, but I don't know if it's a MyBB-style URL 🤣, by using a separator `|`.

### DIFF
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -703,7 +703,7 @@ if(!$mybb->input['action'])
 		{
 			if(!is_array($mybb->input['forum']))
 			{
-				$mybb->input['forum'] = array($mybb->input['forum']);
+				$mybb->input['forum'] = array_unique(array_filter(explode("|", $mybb->input['forum'])));
 			}
 
 			$fid_in = array();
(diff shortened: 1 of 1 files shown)


euantorano:
The URL is a bit different, but I don't think that's a problem. Thoughts @mybb/developers @mybb/sqa?

admin/modules/forum/attachments.php
```diff
@@ -858,14 +858,18 @@ function() {
...
 				foreach($pagination_vars as $var)
 				{
 					if($mybb->input[$var])
 					{
 						$pagination_url .= "&{$var}=".urlencode($mybb->input[$var]);
 					}
 				}
+				if($mybb->input['forum'])
+				{
+					$pagination_url .= "&forum=".implode("|", $mybb->input['forum']);
```
WildcardSearch: This might not make sense out of context, but if you change the other line to expect `$mybb->input['forum']` to be an array of `fid`s, then you can just set it up here like this:

```suggestion
					$pagination_url .= "&forum[]=".implode("&forum[]=", $mybb->input['forum']);
```

The intended, end result for passing arrays over `GET` is `&var[]=1&var[]=2&var[]=3` and that yields `[1,2,3]`

yuliu:
I don't dig a lot in the old codebase, and I guess the original query form should be only supporting finding attachments in one forum, as the hint text `Forum is` and the field `fid` that was built in multi-paged URLs. Later on, a search on multiple forums is added as a new feature, so the #3757 issue comes in, again, I guess.

A quick fix is to add multiple `&forum[]=` as what I did in PR #3758, just like what you've suggested, @WildcardSearch. `$mybb->input['forum']` is formatted as an array, before we found the issue, so I just put an array in the GET request. It's very intuitive, right? As I said in #3757, the URL can be much longer than before, then I pull this request to mitigate this problem (?).

Whatever way we use, the URL will either be longer than before or look unfriendly. But to maintain MyBB's style, the other PR is better than this one, as the URL using a separator is weird across the whole MyBB forum.

Oh, by the way, I found another query similar to the attachments one is user query in AdminCP, allowing search in multiple usergroups, but the user query uses a search_id in adminsession to hide all the secret jobs.

@euantorano, @WildcardSearch, would you like take a look at #3758?

WildcardSearch:
I'll take a look. As to the URL not looking great, that's no big deal to me. Lets get it functioning properly, I say.

yuliu:
I peek the grouping of templates in AdminCP is using a `|` separator in it's URL, eg `./admin/index.php?module=style-templates&sid=1&expand=43|9|22|41#group_41`, when expanding more than one template groups.

Eldenroot:
Looks fine, I am testing on my board right now.

Eldenroot:
It seems to be fine on my end.

Eldenroot:
Still fine :)

PR #3769, opened 2019-08-12
Resolve #3762 Purging guest sessions by IP address may invalidate anti-CSRF tokens between requests

### LINKED ISSUE #3762: Purging guest sessions by IP address may invalidate anti-CSRF tokens between requests

dvz:
Resolves #3762.

To continue using the `mybb_sessions` table / `session` class for CSRF prevention tokens for guest users:
- a `mybb_sessions.isunique` column is created to indicate whether the session should be representative of a unique user and included in online counts/lists,
- the session system logic is modified to set old, conflicting sessions as not unique instead of having them deleted (resulting in additional `UPDATE` query for guest users in `session::update_session()`, which is called instead of `session::create_session()` as sessions continue with no disruption),
- online statistics sections are limited to unique sessions by extending general `SELECT` queries of `mybb_sessions` with a `isunique = 1` condition.

Alternatively to `isunique`-like column, `SELECT` queries could be extended to return most recent rows by IP; this approach would need to be compatible with database engine versions [supported by MyBB 1.8](https://docs.mybb.com/1.8/install/requirements/).

These changes are not intended to modify the original behavior of the session system in relation to numbers and listings of active users.

### DIFF
--- a/install/resources/mysql_db_tables.php
+++ b/install/resources/mysql_db_tables.php
@@ -724,9 +724,11 @@ $tables[] = "CREATE TABLE mybb_sessions (
   nopermission tinyint(1) NOT NULL default '0',
   location1 int(10) unsigned NOT NULL default '0',
   location2 int(10) unsigned NOT NULL default '0',
+  isunique tinyint(1) NOT NULL default '0',
   PRIMARY KEY(sid),
   KEY location (location1, location2),
   KEY time (time),
+  KEY isunique (isunique),
   KEY uid (uid),
   KEY ip (ip)
 ) ENGINE=MyISAM;";
--- a/install/resources/sqlite_db_tables.php
+++ b/install/resources/sqlite_db_tables.php
@@ -645,7 +645,8 @@ $tables[] = "CREATE TABLE mybb_sessions (
 	anonymous tinyint(1) NOT NULL default '0',
 	nopermission tinyint(1) NOT NULL default '0',
 	location1 int(10) NOT NULL default '0',
-	location2 int(10) NOT NULL default '0'
+	location2 int(10) NOT NULL default '0',
+    isunique tinyint(1) NOT NULL DEFAULT '0'
 );";
 
 $tables[] = "CREATE TABLE mybb_settinggroups (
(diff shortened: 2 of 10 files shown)


install/resources/sqlite_db_tables.php
```diff
@@ -645,7 +645,8 @@
 	anonymous tinyint(1) NOT NULL default '0',
 	nopermission tinyint(1) NOT NULL default '0',
 	location1 int(10) NOT NULL default '0',
-	location2 int(10) NOT NULL default '0'
+	location2 int(10) NOT NULL default '0',
+    isunique tinyint(1) NOT NULL default '0'
```
euantorano: Indentation is incorrect here.

install/resources/upgrade49.php
```diff
@@ -31,6 +31,20 @@ function upgrade49_dbchanges()
...
+	if (!$db->field_exists("isunique", "sessions"))
+	{
+		switch ($db->type)
+		{
+			case "pgsql":
+			case "sqlite":
+				$db->add_column("sessions", "isunique", "smallint NOT NULL default '0'");
+				break;
+			default:
+				$db->add_column("sessions", "isunique", "tinyint(1) NOT NULL default '0'");
```
euantorano: Installer DB structure specifies a key for `isunique` for MySQL, but the upgrade script doesn't.

install/resources/mysql_db_tables.php
```diff
@@ -724,9 +724,11 @@
   nopermission tinyint(1) NOT NULL default '0',
   location1 int(10) unsigned NOT NULL default '0',
   location2 int(10) unsigned NOT NULL default '0',
+  isunique tinyint(1) NOT NULL default '0',
   PRIMARY KEY(sid),
   KEY location (location1, location2),
   KEY time (time),
+  KEY isunique (isunique),
```
euantorano: It looks like only the MySQL structure adds this key - is this on purpose?

dvz: Yes, it looks like MyBB doesn't create performance-related indexes for SQLite and PostgreSQL.

euantorano:
We should probably fix that then, that seems like a major omission. 

On Thu, 15 Aug 2019, at 16:59, Tomasz Mlynski wrote:
> *@dvz* commented on this pull request.

> 
> In install/resources/mysql_db_tables.php <https://github.com/mybb/mybb/pull/3769#discussion_r314380233>:

> >    PRIMARY KEY(sid),
   KEY location (location1, location2),
   KEY time (time),
+  KEY isunique (isunique),

dvz:
https://community.mybb.com/thread-225595.html

Issue #3037, opened 2018-02-19
maxwarningsday Bug


euantorano:
https://community.mybb.com/thread-214294.html

This bug is similar to that one.

```php
// Check we haven't exceeded the maximum number of warnings per day
    if($mybb->usergroup['maxwarningsday'] != 0)
    {
        $timecut = TIME_NOW-606024;
        $query = $db->simple_select("warnings", "COUNT(wid) AS given_today", "issuedby='{$mybb->user['uid']}' AND dateline>'$timecut'");
        $given_today = $db->fetch_field($query, "given_today");
        if($given_today >= $mybb->usergroup['maxwarningsday'])
        {
            error($lang->sprintf($lang->warnings_reached_max_warnings_day, $mybb->usergroup['maxwarningsday']));
        }
    }
```

Example Bug:

Main Usergroup :  Cannot warn users / maxwarningsperday 0
Secondary group:  Can warn users / maxwarningsperday 3

User can give unlimited warnings.

Please confirm and fix.

EDIT: Btw I'm going to make a simple suggestion that eliminates a setting and could make things like this less buggy.

Remove the canwarnusers setting and use maxwarningsday with -1 as unlimited, 0 as cannot warn, and anything over 0 as the actual limit. In this way you simplify the code and remove a setting but achieve the same results.  MyBB continues to use 0 as unlimited and I think that's a poor choice.  Consider a better implementation.

Original thread: [maxwarningsday Bug](https://community.mybb.com/thread-215994.html)

Issue #2913, opened 2017-11-22
Signature Permissions Bug


euantorano:
Example bug.

Member in 2 groups:
1. Registered
2. Custom (type 4)

Registered is set with canusesig=1 and canusesigxposts=250
Custom group is set canusesig=0 and canusesigxposts=0

Member is added to the custom group.  They are able to add a signature even if their post count is under 250.  This is because &#36;groupzerogreater (class_core.php) has the canusesigxposts but not the canusesig permissions.

Logic grabs the canusesig as 1 and the canusesigxposts as 0.  It should grab the canusesigxposts permission where it get the canusesig. Or at least the lowest number if greater than 0.  

Best of luck fixing.

Original thread: [Signature Permissions Bug](https://community.mybb.com/thread-214294.html)

kawaii:
I have confirmed this issue on MyBB 1.8.13.

PR #3778, opened 2019-08-24
Resolve #3776, #2913, #3037. Fix numerical permissions logic for user with multiple user groups.

### LINKED ISSUE #3776: Generating numerical permissions with on/off switches from multiple user groups
Edited title and with new content.

Many numerical permissions that controlled by on/off switches may have been wrongly assigned to a user in multiple user groups. See #2913 and #3037 for example. The problem in a simple word is that each numerical permission is set up to with the simple `0 or greater` (or `0 or lesser, not implemented correctly in MyBB right now) method regardless of its controlling switch permission in its corresponding user group.

The permissions are not limited to warning and signature, but seems most of the numerical permissions are not well computed. Here's a list (might not be the complete one):

|              switch permission               |         numerical permission      |   permission type   |
|:------------------------------------:|--------------------------------|:-------------------:|
|           **Forums and Posts**           |                                               |                              |
| `canpostthreads` & `canpostreplys` | `maxposts`                              |      0 or greater     |
| `canpostattachments`                     | `attachquota`                          |      0 or greater     |
| `caneditposts`                                 | `edittimelimit`                          |      0 or greater     |
|                          --                          |--------------------------------|            --             |
|       **Users and Permissions**        |                                                |                              |
| `canusesig`                                     | `canusesigxposts`                    |       0 or lesser      |
| `cangivereputations`                       | `reputationpower`                   |  **1 or greater** or 0   |
|                                                       | `maxreputationsperthread`      |      0 or greater     |
|                                                       | `maxreputationsperuser`         |      0 or greater     |
|                                                       | `maxreputationsday`               |      0 or greater     |
| `canwarnusers`                               | `maxwarningsday`                   |      0 or greater     |
| `canusepms`                                   | `pmquota`                               |      0 or greater     |
|                                                       | `maxpmrecipients`                   |      0 or greater     |
|                          --                          |--------------------------------|            --             |
|           **Miscellaneous**                 |                                               |                              |
| `cansendemail`                               | `maxemails`                            |     0 or greater     |
|                                                       | `emailfloodtime`                    |     0 or lesser     |

`reputationpower` is a numerical permission different from others. If it's value is `0`, it means numerical 0 but not unlimited. So I label it with `1 or greater or 0`.

_**Original investigation (Some statements might be wrong):**_

First, this is an old issue reported in #2913 and #3037 with a PR #3191 closed them. Then, here comes the story. This time it gets a TL;DR at the bottom ✌️ .

Various MyBB's permissions for a user having multiple user groups assigned are generated by [`usergroup_permissions()`](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/inc/functions.php#L1262) with some particular permissions defined as [`$grouppermignore`](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/inc/class_core.php#L601) and [`$groupzerogreater`](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/inc/class_core.php#L602).

For `$groupzerogreater` permissions, MyBB use the most largest value as the privileged one for this user, i.e., the greatest value among its user groups or unlimited for this user. For example, MyBB sets private message quota with the biggest `pmquota` or unlimited if any group has set this value to `0`. This should be the policy how MyBB interprets user permissions with multiple groups, I suppose. If it's not, there will be inconsistency need to solve. 

I peek that the `canusesigxposts` and `maxwarningsday` were removed from `$groupzerogreater` in #3191, which in my opinion should not be because it violets MyBB's policy on user permissions I mentioned above.

A simple removal of those two from `$groupzerogreater` doesn't resolve those issues (#2913 and #3037), even breaks the permission design. Imagine an administrator with unlimited warning has been denied with a specific number of warning per day, or a user with a lowered number of post to use signature is prevented using it by a higher threshold.

I comes with a solution (TL;DR):

1. Revert changes to have `maxwarningsday` in `$groupzerogreater`.

1. Add another `$groupzerolesser` to hold `canusesigxposts`, since the smaller this value is the most privilege the user gets.

1. Change permission validation to verify whether a user can do first, with `canusesig` and ~~`canwarnusers`~~, then verify their thresholds. (warning.php already verifies `canwarnusers` first.)

1. **Edited to add**: `maxwarningsday` should be going with `canwarnusers` together. **So do reputation, attachment, private messages related permissions.**

It turns out more troublesome than I've expected. I would submit a PR as best as I can.

yuliu:
Try to resolve #3776 and extend #3191 (w.r.t #2913, #3037) with extra code to compute 
**"0 or greater"** numerical permissions: 
`maxposts`, `attachquota`, `edittimelimit`, `maxreputationsperthread`, `maxreputationsperuser`, `maxreputationsday`, `maxwarningsday`, `pmquota`, `maxpmrecipients`, `maxemails`; 

and **"0 or lesser"** numerical permissions: 
`canusesigxposts`, `emailfloodtime`;

and general **"greater"** numerical permissions (value 0 for permission does not mean unlimited): 
`reputationpower`

**Note:** `maxposts` is compared and set to the bigger one if either of those switches is on.
Code and performance optimization.

All numeric permissions should be _well_ handled now.

~~`maxposts` and `reputationpower` are still handled in MyBB's old way.~~

~~Codes are a little bit ugly, sort of unnoticeablly inefficient.~~ Updated with optimization. Performance should be fine.


yuliu:
Numerical permissions other than `maxposts` can now be well handled when a user is assigned with multiple user groups.

`maxposts` plays its role when [either of another **two** permissions](https://github.com/mybb/mybb/issues/3776#issue-484452174) is set on. It should be manually defined or determined by MyBB Group the way how this permission is interpreted, under current MyBB settings scenario.

yuliu:
`maxposts` or other permissions with multiple switches is handled in the latest commit, in such a way for multiple user groups:

- In traversing a user's groups, if either current group or previously traversed group enables either `canpostthreads` and `canpostreplys` (or other switches), this permission will be checked.

- If this permission's value is greater (or lesser for other permissions) than previously updated one, then update it.

Currently there's only `maxposts` is such a permission requiring multiple switches.

I suppose this PR has resolved #3776.

yuliu:
I've just read that the upgrade script has its own declared `$grouppermignore` and `$groupzerogreater` as it said "Include the necessary contants for installation". Seems it doesn't use them but I'm not sure.

Yes, there's a typo :laughing: , and the install script has the correct one. Still, I don't know how they're being used.

Eldenroot:
@yuliu - this one should be finished including the upgrade script. This issue is quite annoying in ACP where some options cannot be used properly because of numerical permissions...

yuliu:
> @yuliu - this one should be finished including the upgrade script. This issue is quite annoying in ACP where some options cannot be used properly because of numerical permissions...

There's no database upgrade needed. Only the logic how user's permission is calculated is corrected.

I had manually tested several combinations of permissions various permission groups, i.e., `$groupzerogreater`, `$groupzerolesser`, `$groupxgreater` and `$grouppermbyswitch`, seems working as desired as I've depicted in #3776.

Seems @euantorano is self-assigned as a reviewer and looks like a unit test would be good for testing this PR, which is applicable in 1.9.

Eldenroot:
This is a continuation of what had been done in past @effone

inc/class_core.php
```diff
@@ -599,7 +599,41 @@ function __destruct()
  */
 
 $grouppermignore = array("gid", "type", "title", "description", "namestyle", "usertitle", "stars", "starimage", "image");
-$groupzerogreater = array("pmquota", "maxpmrecipients", "maxreputationsday", "attachquota", "maxemails", "maxposts", "edittimelimit", "maxreputationsperuser", "maxreputationsperthread", "emailfloodtime");
```
euantorano, line 602: Not a problem with this PR, but I really don't like how 1.8 mixes classes and global variables in a single file 😆

Eldenroot: Thats why we need to make a list of changes which must be done in next major version of MyBB :)

euantorano (commented):
This looks good, but I want to do some testing of it over this weekend. I'm going to assign myself.

Ben-MyBB:
@Sama34 / @JordanMussi any chance you could get time to review this?

inc/functions.php
```diff
@@ -1299,10 +1330,116 @@ function usergroup_permissions($gid=0)
...
+					$perm_numerical_lowerbound = $groupxgreater[$perm];
+				}
+
+				if($perm_is_numerical != 0)
 				{
-					$usergroup[$perm] = 0;
+					$update_current_perm = true;
+
+					// Ensure it's an integer.
+					$access = (int)$access;
```
Sama34, line 1363: Where is `$access` defined?

Sama34 (approved):
Could we get rid of the `yes` for compatibility? It is just annoying at this point, 1.8 has been released over 6 years ago by now.

Also, won't it be easier to add constants for these magic number? It is really confusing to read the code.

inc/functions.php
yuliu: > Where is `$access` defined?

https://github.com/mybb/mybb/pull/3778/files#diff-f3a2689b39e6a1a9532d63ed0bc25e99cafc51ae5d7486c07b1e68e5a9681a3fR1320

Sama34:
Hi, I would like to see the changes I mentioned above, but if anyone thinks it is better to left for later then I'm fine with merging this as is (I did test and it seems to work for me of course, but I think at least one extra tester would be wonderful).

PR #3782, opened 2019-08-30
Resolve #3777 and #3779, issues of some front-end outputs

### LINKED ISSUE #3777: Some outputs to the front-end have issues: empty post subject, etc.
Edited:
_**I'll stop tracking similar problems in this issue. Open a new issue, if another related problem is found.
The last two problems should be addressed in new issues, too.**_

~~Still collecting more related issues, please don't close it. 
If you have any related problem stated here, please confirm it or comment on this issue.
Any pull request that fix problems listed here, please reference this issue. Thank you!~~

It should be a low priority issue requiring improvement.

(Edited title and included similar issues.)

### Empty post subject not well handled in warning & modcp pages
Empty post subject usually comes from data imported from other forums (or old MyBB? not sure). Latest MyBB **does allow** users to leave the subject blank when posting replies (but not new thread), and it will set the post's subject by prefixing the thread's subject of this post with `RE: `, and this is how make quick reply works.

Currently found the `warning` and `modcp` pages are not well handling an empty post subject, only output HTML like `<a href="{$post_link}"></a>`. There are also several occurrences in the modcp page. **Not sure if it's a complete list:**

- `./warnings.php?uid={$uid}`: warnings for post
- `./warnings.php?action=warn&uid={$uid}&pid={$pid}`: warnings for post
- `./warnings.php?action=view&wid={$wid}`: warnings for post
- `./modcp.php?action=modqueue&type=posts`: posts in moderation queue
- `./modcp.php?action=modqueue&type=attachments`: attachments in moderation queue
- **`./modcp.php?action=modlogs`:** `pid` value in table `moderatorlog` is not correctly stored for some mod actions on posts, can't really depend on that.
- `./usercp.php`: warnings for post
- `./usercp.php?action=attachments`: attachments attached to posts
- `./usercp.php?action=drafts`: compatible for converted posts with empty subject

The post subject is going to be removed in 1.9, as issued in #3499, and [the community post](https://community.mybb.com/thread-220552.html). I think MyBB should handle some empty subject cases for the the most compatibility.

### Empty template variables
On user profile page, `$timeremaining` is not generated for displaying template `member_profile_banned_remaining` in [this block](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/member.php#L2782), but done [here](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/member.php#L2811).

### File size unit in Alert box text for attachment size over PHP limit not inconsistent with language file
[Alert box text](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/jscripts/post.js#L184) uses a value of `php_max_upload_size` divided by 1e6 (1,000,000), but
[Message](https://github.com/mybb/mybb/blob/f8066f4a4c92bfd0bf4b79567a7938206769c293/inc/languages/english/global.lang.php#L126) in the default English language pack uses `bytes`.
Source: [community post](https://community.mybb.com/thread-224427-post-1335046.html#pid1335046).
Original PR: #3373.

### HTML tag's attributes starting with unwanted space in HTML table output
Such as first/last `<tr>` or first `<th>` row: `<th class=" first">Forum</th>`. Mostly used in AdminCP area via function `output_row_header`.

### Inconsistent language usages of text for navigation breadcrumb on UserCP page.
Some use `$lang->nav_*`  language variables and others use `$lang->ucp_nav_*` ones.

### LINKED ISSUE #3779: hardcoded SID causes missing CSS when exporting PMs in HTML, after upgrading
Hello,

I open this issue following noyle's advice in MyBB Community Forums : https://community.mybb.com/thread-224378.html 

In _private.php_ , the following code is supposed to retrieve the global.css stylesheet :
`$query = $db->simple_select("themestylesheets", "stylesheet", "sid = '1'", array('limit' => 1)); `

However, after upgrading myBB (from 1.8.14 to 1.8.20 or 1.8.21), this SID is not in the _themestylesheets_ table anymore because of stylesheets updates (delete+insert, with an auto_increment PK)

As a result, the "Export Private Messages" feature is not working as expected (when selecting HTML format) : CSS is not applied.

More info / screenshots [in the original thread](https://community.mybb.com/thread-224378.html)

Thanks,

JY.

yuliu:
Resolve #3777, partially (see the note below):

Fix **empty post subject not well handled in warning & modcp pages** in (mostly for compatible issues):
- `./warnings.php?uid={$uid}`
- `./warnings.php?action=warn&uid={$uid}&pid={$pid}`
- `./warnings.php?action=view&wid={$wid}`
- `./modcp.php?action=modqueue&type=posts`
- `./modcp.php?action=modqueue&type=attachments`
- **`./modcp.php?action=modlogs`:** `pid` value in table `moderatorlog` is not correctly stored for some mod actions on posts, can't really depend on that.
- `./modcp.php?action=ipsearch&ipaddress={$ip}&search_posts=1` _(added in a recent commit)_
- `./usercp.php`
- `./usercp.php?action=attachments`
- `./usercp.php?action=drafts`
- `./showthread.php?mode=threaded&tid={$tid}&pid={$pid}` _(added in a recent commit)_

Fix **empty template variables**: `$timeremaining` for displaying banned time on member profile.

Fix **file size unit in Alert box text for attachment size over PHP limit not inconsistent with language file**. It's a quick fix and it's somehow unfriendly to end-user. Hope anyone experienced with JavaScript will code a function similar to `get_friendly_size()` to display the value.

And fix #3779 hard-coded CSS stylesheet id when exporting PMs as HTML.

**Note:**

1. As for the inconsistent language usages of text for navigation breadcrumb on UserCP page, changes upon that might lead to more trouble for current users. So I leave it untouched. If we have to deal with this minor problem, better do it in feature major version upgrade.
1. I'd like to suggest changes for non-well-formatted table's `class` attribute output in AdminCP in another PR, since this one focuses on the front-end.
1. Since a literal string `RE: ` is hardcoded in current MyBB deeply, it's better to address its problem in a separated issue.


Eldenroot:
Good job keep it

modcp.php
```diff
@@ -2452,6 +2460,10 @@
 			$attachment['filename'] = htmlspecialchars_uni($attachment['filename']);
 			$attachment['threadsubject'] = htmlspecialchars_uni($parser->parse_badwords($attachment['threadsubject']));
 			$attachment['filesize'] = get_friendly_size($attachment['filesize']);
+			if(empty($attachment['postsubject']))
+			{
+				$attachment['postsubject'] = "RE: ".$attachment['threadsubject'];
```
euantorano: This should probably be translatable via a language variable.

usercp.php
```diff
@@ -3855,6 +3859,10 @@
 			$attachment['postlink'] = get_post_link($attachment['pid'], $attachment['tid']);
 			$attachment['threadlink'] = get_thread_link($attachment['tid']);
 			$attachment['threadsubject'] = htmlspecialchars_uni($parser->parse_badwords($attachment['threadsubject']));
+			if(empty($attachment['subject']))
+			{
+				$attachment['subject'] = "RE: ".$attachment['threadsubject'];
```
euantorano: Again, this should probably be translatable.

usercp.php
```diff
@@ -3317,6 +3317,10 @@
 				$type = "thread";
 			}
 
+			if(empty($draft['subject']))
+			{
+				$draft['subject'] = "[".$lang->post_no_subject."]";
```
euantorano: This might be nice to be translatable, since we're trying to minimise the amount of hardcoded HTML and other text.

usercp.php
```diff
@@ -4057,8 +4065,12 @@
 			while($warning = $db->fetch_array($query))
 			{
 				$post_link = "";
-				if($warning['post_subject'])
+				if(!empty($warning['pid']))
 				{
+					if(empty($warning['post_subject']))
+					{
+						$warning['post_subject'] = "[".$lang->post_no_subject."]";
```
euantorano: See above regarding translations.

jscripts/post.js
```diff
@@ -181,7 +181,7 @@ var Post = {
 
 		if (totalSize > php_max_upload_size && php_max_upload_size > 0)
 		{
-			var php_max_upload_size_pretty = Math.round(php_max_upload_size / 1e4) / 1e2;
+			var php_max_upload_size_pretty = php_max_upload_size;
```
euantorano: We definitely need a friendly size for this, I can write one if you open this PR to allow modifications from team members - see here for instructions: https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

euantorano (changes requested):
Looks great, just some minor comments that are mostly nitpicks.

jscripts/post.js
yuliu: The `Allow edits from maintainers. ` option is already ticked. Should I change any other settings?

Now working on the `no subject` language entries.

euantorano: Great, I'll work on a function to format file sizes for JS as part of this PR later on today.

modcp.php
yuliu: The `RE:` prefix for thread replies is indeed hardcoded deeply in current MyBB.

I'll only try to update those variables in this PR with a translatable entry.

euantorano: I feared that may be the case. In that case, just leave the ones in this PR for now, and we'll open another issue to try and patch them all at once.

usercp.php
yuliu: I followed the way of handling empty post subject in `showthread.php`. Then I find that I should also make `[]` as translatable, you can see I haven't enclosed them with brackets in `modcp.php`.

I'll try to update them and `showthread.php` as well.

euantorano: Ok. It seems we still have a lot of hardcoded stuff scattered about in that case 😢

yuliu: Seems `no_subject` is only hardcoded in `showthread.php` before this PR. But for `RE: `... pretty a lot.

euantorano: We'll definitely do a separate PR for `RE:` then, otherwise it would make following PRs nearly impossible.

yuliu (commented):
Forgot to mention the empty subject when searching IP in posts is also replaced with the language variable in this commit. Will add this information in the main PR description.

Eldenroot:
Ok, I didnt know that... Good job!

Eldenroot:
I have applied to my board :) more testers needed 👍

euantorano:
See my above comment about friendly size - I will write the code for this this weekend and add it to this PR.

jscripts/post.js
yuliu: Borrowed a function from https://stackoverflow.com/a/20732091/6681141

yuliu:
> See my above comment about friendly size - I will write the code for this this weekend and add it to this PR.

Have borrowed one from stackoverflow which seems good, and size units are backed by language entries.

Ben-MyBB:
@yuliu

PR #3786, opened 2019-09-02
Fix enforced case sensitivity check on HEADER's location field

yuliu:
It's a quick fix that should be working.

BTW, seems it's conventional in MyBB to use `#` as the delimiter for PCRE functions, but some outliers use `/`. I'm not sure so I didn't change it.

Source: [community post](https://community.mybb.com/thread-224406.html)

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -7112,7 +7112,7 @@ function fetch_remote_file($url, $post_data=array(), $max_redirects=20)
 
 			if(in_array(curl_getinfo($ch, CURLINFO_HTTP_CODE), array(301, 302)))
 			{
-				preg_match('/Location:(.*?)(?:\n|$)/i', $header, $matches);
+				preg_match('/^Location:(.*?)(?:\n|$)/im', $header, $matches);
 
 				if($matches)
 				{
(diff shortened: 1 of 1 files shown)


euantorano:
IMO the regex should also be changed to `^Location:(.*?)(?:\n|$)` to only match headers that are only `Location`, not other headers that end with `Location:`

yuliu:
I think you're right, but can't get it work with `^` right now. I'll try to figure it out. Regex is a trouble to me.

euantorano:
It's likely because the headers string is multi-line, and `^` matches on the start of text, not the start of line by default. Try adding the `m` (multi-line) flag.

yuliu:
Yep, I just figure it out while you're commenting. :tada: 

Oh, seems it should be added with an `m` modifier to enable multi-line mode.

yuliu:
The ending part `(?:\n|$)` seems can be optimized, but I've no idea. Will leave it for MyBB members. :sheep:

euantorano:
The whole thing can probably be optimised a bit, but network requests are already slow anyway.

yuliu:
Fixes #3850.

PR #3788, opened 2019-09-05
Replace "sex" with "gender"

gonzalezjo:
The main utility behind making one's sex public on a forum is that people are better informed as to what pronouns to use in reference to a given person. That said, one's assigned biological sex might not align with their gender. Changing "sex" to "gender" gives a more fitting denotation.


euantorano:
Thanks for the contribution.

Note that this won't update existing installs, it'll only affect new installs. In order to update existing installs, [the upgrade script](https://github.com/mybb/mybb/blob/feature/install/resources/upgrade49.php) will need updated to add a new SQL update query, something like:

```php
$db->update_query('profilefields', array(
    'description' => "REPLACE(description, 'sex', 'gender')",
), 'fid = 3');
```

This should be added before line 34. If you don't get a chance to add this, I can always do it myself later.

dvz:
I believe we were settled to remove the field altogether (https://community.mybb.com/thread-208445.html), possibly in 1.9.

A discussion platform may benefit from gender pronoun information (so asking for _sex_ or _gender_ may be unnecessary - similar to some hardcoded IM fields), but that's still language-dependent; one way or another, a more flexible Profile Fields system with a _select/search list + custom_ could make this easier for administrators.

gonzalezjo:
I updated the upgrade script. Not sure if this should be closed or not because of the above comment, though.

Azareal:
I'd lean towards not having it at all, for privacy reasons.

yuliu:
Really would like to see changes in future versions. v1.8 is in its 5 years which should finish its era destiny now, would mostly only accept security & bug fixes. Besides, for existing installations, forum owners may have already remove/modified this field.

apple-fan-pl:
> Really would like to see changes in future versions

Hehe optimist.... new version? of mybb? good joke. Made my day :)

Eldenroot:
@gonzalezjo - please can you solve merge conflicts? Thank you

gonzalezjo:
Seem good now?

Sama34:
Won't this force incorrect changes? Like changing `sexo` to `gendero`?

MyBB is language friendly and I would rather go on only applying the change for new installs. I would rather vote for its removal altogether.

Sama34:
#3869

euantorano:
Yep, this would affect those fields. We could always remove the update script to only affect new installs.

dvz:
From Discord:
> it may need different values than male/female (and meaning of saved values would change, unintended, if it's just renamed)

— _Man_ / _Woman_ may be more appropriate for _gender_.

Sama34:
@dvz If I have to take a side I would agree because male or female are general terms whereas Man or Woman are specific to humans.

PaulBender:
Also, the field name should be changed to Gender as well. It seems weird to change the description but not the name.

install/resources/upgrade49.php
```diff
@@ -35,6 +35,9 @@ function upgrade49_dbchanges()
 
 	$db->delete_query("settings", "name='allowyahoofield'");
 
+	$db->update_query('profilefields', array(
+		'description' => "REPLACE(description, 'sex', 'gender')",
+	), 'fid = 3');
```
yuliu, line 40: This database upgrade script `upgrade49.php` is for 1.8.21 to 1.8.22. If this PR is going to be merged before 1.8.23 release, changes should be made accordingly in `upgrade50.php`.

euantorano: See the above comment by @Sama34 - we want to remove this upgrade script completely.

yuliu: > See the above comment by @Sama34 - we want to remove this upgrade script completely.

Makes sense to only affect new installs. I vote for this. Maybe existing forums have already altered those fields.

gonzalezjo: I think this has been addressed? Hoping you can double check.

install/resources/mysql_db_inserts.php
```diff
@@ -69,7 +69,7 @@
...
 $inserts[] = "INSERT INTO mybb_profilefields (fid, name, description, disporder, type, regex, length, maxlength, required, registration, profile, postbit, viewableby, editableby, postnum) VALUES (2, 'Bio', 'Enter a few short details about yourself, your life story etc.', 2, 'textarea', '', 0, 0, 0, 0, 1, 0, -1, -1, 0);";
-$inserts[] = "INSERT INTO mybb_profilefields (fid, name, description, disporder, type, regex, length, maxlength, required, registration, profile, postbit, viewableby, editableby, postnum) VALUES (3, 'Sex', 'Please select your sex from the list below.', 3, 'select\nUndisclosed\nMale\nFemale\nOther', '', 0, 0, 0, 0, 1, 0, -1, -1, 0);";
+$inserts[] = "INSERT INTO mybb_profilefields (fid, name, description, disporder, type, regex, length, maxlength, required, registration, profile, postbit, viewableby, editableby, postnum) VALUES (3, 'Sex', 'Please select your gender from the list below.', 3, 'select\nUndisclosed\nMale\nFemale\nOther', '', 0, 0, 0, 0, 1, 0, -1, -1, 0);";
```
Sama34, line 72: Hi @gonzalezjo, sorry to bother you so much. The field `name` value should be updated for both db engines of course as well, as @PaulBender mentioned it makes no sense to keep it as `Sex`.

install/resources/upgrade49.php
```diff
@@ -35,8 +35,8 @@ function upgrade49_dbchanges()
 
 	$db->delete_query("settings", "name='allowyahoofield'");
 
-	$db->modify_column('attachments', 'filename', 'varchar(255)', true, "''");
 	$db->modify_column('attachments', 'attachname', 'varchar(255)', true, "''");
+	$db->modify_column('attachments', 'filename', 'varchar(255)', true, "''");
```
JordanMussi, line 39:
```suggestion
	$db->modify_column('attachments', 'filename', 'varchar(255)', true, "''");
	$db->modify_column('attachments', 'attachname', 'varchar(255)', true, "''");
```

Looks like this got accidentally reordered in the previous commit. You should be able to apply this suggestion to make the change. :-)

Issue #3785, opened 2019-09-02
showteam page doesn't work with group leaders


yuliu:
This is a v1.9 problem.

The `showteam` page outputs an `Uncaught TypeError: Argument 1 passed to MyBB\Twig\Extensions\UrlExtension::getProfileLink() must be of the type integer, null given, called in...` error when the board has group leaders to show.

Probably caused by neglect during theme migration to Twig. I think I can fix it.

BTW, for the 1.9 development branch, should I wait for the code being rebased before I can pull a request? New to git rebase and current MyBB 1.9 code rebase.

euantorano:
Feel free to open PRs - the rebase is a long and arduous task that I’m working on as and when possible. I prioritise PRs otherwise they end up conflicted.

yuliu:
I'm not good at CSS scripting, and thus I really can't fix most of the style issues posted here. Better mention, maybe justinsoltesz and effone and others, to check on those problems.

yuliu:
Looks like the doubled `&&amp;` in some URLs are caused by double escaping, one by MyBB's function `htmlspecialchars_uni()`, and the other by Twig'a auto-escaping that defaults to `html`.

I didn't find any MyBB's environment settings for Twig for now, so variables fetched in `{{ }}` will be escaped by Twig by default. Or we should remove MyBB's `htmlspecialchars_uni()` on those `get_*_link()`.

See https://twig.symfony.com/doc/2.x/templates.html#html-escaping and https://twig.symfony.com/doc/2.x/api.html#environment-options .

euantorano:
We configure the environment here: https://github.com/mybb/mybb/blob/develop/1.9/inc/src/Twig/ServiceProvider.php#L67

The `get_*_link()` functions should be changed IMO though - one of the biggest advantages of the move to Twig is the auto escaping.

yuliu:
Looks there are a lot of code to be examined for `get_*_link()` functions. Besides this. I have found other possible/questionable issues on (have been listed in the main issue):

- Avatar for guest/non-user should be outside a `<a href>` link.

- Some controls that are originally permissions are calculated in the template. Shouldn't we have them in PHP code? Eg, the `newthread` button in `forumdisplay`. There are a few questionable controls like this.

euantorano:
It might be nice to have the `get_*_link()` functions available within Twig, but I'm not sure.

- Agreed. Not sure what would be best for that, perhaps another argument to the function like `bool $renderLink`?
[...]

Issue #3498, opened 2018-11-17
Deprecate Thread Ratings in 1.9


Ben-MyBB:
Just following up on the vote we took on this at: https://community.mybb.com/thread-220554.html

The vote was to remove these in 1.9 so this issue is to track that.

effone:
(lol @justinsoltesz `Depreciate` is also a correct word and a synonym of `deprecate`).

justinsoltesz:
@effone Depreciate and deprecate have different meanings.

https://stackoverflow.com/questions/9208091/the-difference-between-deprecated-depreciated-and-obsolete/9208164

justinsoltesz:
Also, on topic, I believe the plan that has been discussed for thread ratings in 1.9 is to remove them from the templates (which is essentially already the case in the new theme) but not to remove the ratings that have already been stored in the database. This will allow a plugin to use that existing data if desired.

kawaii:
^ and keeping that data will also allow for someone (me) to create the plugin which reinstates thread ratings to 1.9. :wink:

euantorano:
> Also, on topic, I believe the plan that has been discussed for thread ratings in 1.9 is to remove them from the templates (which is essentially already the case in the new theme) but not to remove the ratings that have already been stored in the database. This will allow a plugin to use that existing data if desired.

Yes, that is the plan exactly.

yuliu:
Current template shows this field and there's no style for correctly displaying it, just raw list items.

justinsoltesz:
> Current template shows this field and there's no style for correctly displaying it, just raw list items.

Correct, there is no style because thread ratings are being deprecated but the template is displaying it because the work to deprecate thread ratings hasn't been finished yet.

PR #3789, opened 2019-09-05
Add missing PM menu, fix fetch_remote_file(), remove thread rating, minor template fix, etc.

yuliu:
- Add Empty Folders & Export Messages back on PM menu. I'm not sure they're missing for an intentional reason. Icons might need replacement, ping @justinsoltesz here.
- Fix wrong `,` (comma) position in displaying moderators in forum listing.
- Fix post thread button showing on `forumdisplay` page for guest uses. Let Twig do less PHP job.
- Fix #3786 enforced case sensitivity check in `fetch_remote_file()`.
- Fix moderators not showing on forumdisplay page and minor tweak on using `,`(comma).
- #3498 code clean up, remove thread rating by commenting them in templates `showthread` and `forumdisplay`. (PHP scripts including `forumdisplay.php`, `showthread` and ratethread.php` and JavaScript `./jscripts/ratings.js` are left as what they were, where codes of thread rating are not touched.)

These problems have been tracked in #3785. I'll try to fix other issues in separated PRs.

### DIFF
--- a/inc/views/base/forumdisplay/forumdisplay.twig
+++ b/inc/views/base/forumdisplay/forumdisplay.twig
@@ -39,11 +39,10 @@
             <p>{{ lang.moderated_by }}
                 <strong>
                     {% for moderator in moderators.users %}
-                        <a href="{{ moderator.profilelink }}">{{ moderator.username }}</a>{% if loop.last == false %}{{ lang.comma }} {% endif %}
+                        <a href="{{ moderator.profilelink }}">{{ moderator.username|raw }}</a>{% if loop.last == false or moderators.groups is not empty %}{{ lang.comma }} {% endif %}
                     {% endfor %}
-                    {% if moderators.groups %}{{ lang.comma }} {% endif %}
                     {% for moderator in moderators.groups %}
-                        {{ moderator.title }}{% if loop.last == false %}{{ lang.comma }} {% endif %}
+                        {{ moderator.title|raw }}{% if loop.last == false %}{{ lang.comma }} {% endif %}
                     {% endfor %}
                 </strong>
             </p>
--- a/inc/views/base/forumdisplay/forumdisplay.twig
+++ b/inc/views/base/forumdisplay/forumdisplay.twig
@@ -71,7 +71,7 @@
             </section>
         {% endif %}
 
-        {% if foruminfo.type == 'f' and foruminfo.open and fpermissions.canpostthreads and mybb.user.suspendposting is not empty %}
+        {% if foruminfo.type == 'f' and foruminfo.open and fpermissions.canpostthreads %}
             {% set newthread %}
                 <a href="newthread.php?fid={{ foruminfo.fid }}" class="button button--medium button--full-width">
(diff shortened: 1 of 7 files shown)


yuliu:
@euantorano please let this PR be open. I'd like to push more commits in this, and looks like @justinsoltesz will check some problems in what has already been pushed for now (some may be intentionally missed but fixed by me). But I'll ping you when commits in this PR can be merged regardless of his response. Any check on this is welcome anytime.

yuliu:
Hi @euantorano, I think that's enough for this PR to me. It should be ready to be reviewed and then merged unless there's something that I'm doing backwards.

A lot more to say, I was thinking I can fix the missing mod queue notice in this PR, but I recall you've said that we might do more flexible work on the URL, and I've got some thoughts on that but not so easy to implement. I'd like to address that problem in a separated PR if I can figure it out, otherwise will just fix it as how v1.8 did (not equivalently good because v1.9 uses Twig).

inc/views/base/showthread/showthread.twig
```diff
@@ -205,7 +205,7 @@
         </div>
 
         {% if mybb.settings.allowthreadratings and forum.allowtratings %}
```
skmedix: Isn't this a useless check right now? The whole if could be commented

inc/views/base/forumdisplay/forumdisplay.twig
```diff
@@ -242,7 +241,7 @@
                                 <option value="{{ field }}"{% if mybb.input[field] %} selected{% endif %}>{{ attribute(lang, 'sort_by_' ~ field) }}</option>
                             {% endfor %}
                             {% if displayRating %}
```
skmedix: Same here

inc/views/base/showthread/showthread.twig
yuliu: The script is still doing the same as what we do in v1.8, fetch/calculate/output thread rating. For the theme engine, it can filter what will/won't show. Seems the team's idea is to preserve exiting data and just redact the thread rating view from front-end, but allow plugins to reuse this data field.

- The most important for contributors would be the Twig templates conversion.
- And we can make sure that the thread rating is deprecated in 1.9, and some modifications to user profile editing section.
- If anything in doubt, preserve what 1.8 has, and accommodate it with 1.9.

Besides, the new templates should have been written in maybe nearly 10 or 11 months ago. The original idea is fully transform templates to use Twig, without considering too much. @euantorano for clarification.

euantorano:
Yes, the original approach was just straight port from 1.8 to 1.9. Thread ratings are still in templates as they’re still in the DB at the moment. Once it’s fully removed, we’ll update the templates too.

Ben-MyBB:
What shall we do with this?

justinsoltesz:
Regarding the Empty Folders and Download Messages links, I don't think those have ever been on the side menu. In 1.8 they are above the message listing. In 1.9, they are below it. The links only appear if you have more than 1 message.

image

PR #3801, opened 2019-09-09
Fix #3793 use `unserialize` instead of `my_unserialize` when read…

### LINKED ISSUE #3793: Caching fallback breaks when external cache system has problems
Community [post](https://community.mybb.com/thread-224537.html). I've discussed much with @dvz on this issue, and I'll post it here. It's a little bit long, but I think it's worth reading, especially for those who are not familiar with how MyBB's caching works. Sorry, no TL;DR this time. 😆 

I think it's an issue that really breaks whole MyBB's running, eg. forums/forumpermissions/usergroups/etc.

Well, I did some research on MyBB's caching mechanism which supports using database and external cache system as cache handler, and will fallback to using database if the cache handler fails. Let me assume some failures that can really happen when using an external cache system, taking Memcache (I've tested with it for sure, probably Memcached too) for example:

1. If MyBB's cache setting is correct:
- The cache system, eg. Memcache, is gone/malfunctioning/etc. MyBB can't fetch cache from it.
- The cache system is restarted, but currently no cache exists. MyBB can fetch cache but will get none, then MyBB tries to create cache on the cache system.

2. If MyBB's cache setting is wrong with Memcache settings, eg `host` and `port`. MyBB can't fetch cache from it.

It's worth noting that when an external cache system is invalid or MyBB gets invalid cache from it (probably a `bool(false)`), MyBB tries to fetch the cache content stored in the database, which is a fallback mechanism, and then recreate the cache on external cache system (if the cache system's still working).
For the second situation, MyBB will not test the cache server's connectivity before trying to fetch cache from it, yielding MyBB gets invalid cache from it, [a `bool(false)` as documented by PHP](https://www.php.net/manual/en/memcache.get.php#refsect1-memcache.get-returnvalues).

The whole caching work should be working just fine if we have a valid data-cache configuration (in `./inc/config.php`). But for large forum, whatever with a large number of categories/subforums or a large number of user groups or something else, the cache fallback mechanism will be broken.

Internally (and externally in a very few modules), MyBB uses a custom (un)serialize function to store and recover some internal data, for this issue, it's cache content. The corresponding functions of PHP's built-in ones are [`serialize()`](https://www.php.net/manual/en/function.serialize.php) and [`unserialize()`](https://www.php.net/manual/en/function.unserialize.php). For security concerns, MyBB changes to use custom (un)serialize functions instead, i.e., `my_(un)serialize()`, see #1733.

However, the `_safe_unserialize()` function used by `my_unserialize()` has a hard limit for the length of input string to be unserialized. When the length exceeds this limit, the function just refuses to work and returns a `bool(false)`. This is indeed a problem/limitation, see also #3775, for related issues.

Let me get back to the cache problem. When MyBB gets a `bool(false)` from the external cache system, it'll fall back to query on database (table `datacache`) for that value. When MyBB encounters a long input string, which is read from a record in table `datacache` to be unserialized, the custom `my_unserialize()` will return a `bool(false)` to the fallback query. Yes, it's the final answer, any failed unserialization will lead to a cache to be invalid with a `bool(false)` value that is used everywhere where the cache is read.

For the community post, OP observed this problem when his Memcached server is restart, meaning it's working, but MyBB only gets `false` upon cache reading. Actually this `false` means the key doesn't exist. Then MyBB falls back on database query to fetch the original cache data, and fails, probably a very long field.

Well, the story thing can be stopped now. Here's some thoughts and resolutions to the problem. Not really well-designed for all but only a mitigation and something to be expected in future version:

- A quick fix for OP's problem is replace `my_unserialize` with `unserialize` in the [datacache class](https://github.com/mybb/mybb/blob/fda816dfab9d93c24f95ba76a925ac6bde72b3c4/inc/class_datacache.php#L166). See also what the cache reading is done using database as the cache store in [this line](https://github.com/mybb/mybb/blob/fda816dfab9d93c24f95ba76a925ac6bde72b3c4/inc/class_datacache.php#L195). This should solve most of the problems, including following concerns.

- Using an external cache system (sorry, I just realized really should use `cache store` instead `cache system` 🐞), to specific, Memcache/Memcached, MyBB doesn't check the connectivity to the cache server. This could be because the need of using clustering. However, the [Memcache::get](https://www.php.net/manual/en/memcache.get.php) during [`$cache->fetch()`](https://github.com/mybb/mybb/blob/fda816dfab9d93c24f95ba76a925ac6bde72b3c4/inc/cachehandlers/interface.php#L29) for Memcache/Memcached does throw a **notice** (not sure if it's E_NOTICE or E_USER_NOTICE ,it's even not a warning) if the connection from MyBB to the cache server is timeout (probably by a misconfiguration or a real network issue). MyBB suppresses notice by default, so the forum users (really should be owners/admins) are not aware of what really happens. Don't know if this kind of notice can be caught and alert forum admins.

- This one is a little off current topic. @dvz mentioned this when we were discussing about this issue. The `datacache` database table is volatile, meaning as its table name indicates, which is full of caches. As the circumstance for MyBB for now, almost all scripts of it rely on a global `$cache` variable which is built by reading from an external cache store or the `datacache` table. And also taking the cache fallback mechanism into account, this table should be really really really reliable. However currently, there lacks of a checking policy and rebuilding mechanism on existence/correctness of records in this table. There's still more, missing records if any will not be recreate by "Rebuild & Reload All" in AdminCP's Cache Manager.

- This one is a bit more off current topic. As I've mentioned, seems MyBB needs a **self** health checking mechanism, which is not just considered as a tool that requires user to click to run, on database and files integrity. The [DVZ Integrity Tools](https://community.mybb.com/mods.php?action=view&pid=980) is a great reference, but we might need some of its feature to run automatically like a task. Current design permits admins to delete/disable a task, so we need to add some fields to prevent MyBB itself to delete/disable health check tasks.

Oh.. seems pretty long is this post, but I think I've got the problem revealed clearly.

yuliu:
The PR is supposed to solve the most problems specified in #3793, with the cache check work that should be done through a scheduled task is processed upon each visiting and checked in a hard coded time of 10 mins.

- Use `unserialize` instead of `my_unserialize` when reading cache from database.
- Add an availability test of cache handler when cache reading fails in cache store level, issue an `E_USER_WARNING` warning when failure continues.
- Add a persistent check on cache. Upon pivot cache inexistence or every 10 minutes, cache records will be examined and missing records will be recreated then recached. Currently support only MyBB's core (internal) caches, most of. The code also hints a future use for cache checking for plugins and/or MyBB packages (if core MyBB is package-ized).

I was about to upload a patch/diff file but looks like GitHub doesn't support that directly. So I draft it. The cache check is only tested with database cache store, but should also work with an external cache handler. **There's an output when specific cache is checked and/or rebuilt upon its missing. I forgot to remove the `echo` part.**


yuliu:
Looks like mybb has own method of raising error, would replace with that later.

yuliu:
Seems the PR is ready for review now.

Tested with environments set up as follows:

- **`db` as datastore, database table `datacache` is damaged:** essential records in this database table will be rebuilt.

- **`memcache` as datastore, but memcache is not running, database table `datacache` is fine:** a warning is issued, forum continues to run without problem.

- **`memcache` as datastore, but memcache is not running, database table `datacache` is damaged:** a warning is issued,  essential records in this database table will be rebuilt and forum continues to run without problem.

Currently, the warning of `cache_runtime_error` is issued to front-end to all users. It's a bit difficult to only display this warning to some particular users such as admins since session is created and checked really far behind the cache can be checked. The alternative is to trigger an error rather than a warning. However, an error will halt the forum's running.

Ben-MyBB:
@yuliu Can you check the conflicts please?

yuliu:
> @yuliu Can you check the conflicts please?

Sure I will.

But please do know that this PR was intended to be a demonstration for ideally resolving issues specified in #3793 and the way it fixes things is quite brutal:
- Converting to use PHP built-in `unserialize()` instead of `my_unserialize()` in cache reading may not be safe, though it looks safe to me.
- Testing cache handler's availability whenever cache reading may fail may have performance concerns.
- Throwing runtime errors when cache fails does hit performance, though mostly because of cache availability test that may have to wait the cache reading timed out.

euantorano:
It seems most cache handlers use `unserialize()`:

- [`apc`](https://github.com/mybb/mybb/blob/feature/inc/cachehandlers/apc.php#L63)
- [`eaccelerator`](https://github.com/mybb/mybb/blob/feature/inc/cachehandlers/eaccelerator.php#L66)
- [`redis`](https://github.com/mybb/mybb/blob/feature/inc/cachehandlers/redis.php#L97)

However, the main `datacache` class uses `my_unserialize()` since #3649 (which is my fault 😉). I find it interesting that some drivers do serialisation themselves, which doesn't quite seem right...

Regarding the future approach to caching, I really like how [Laravel's cache interface allows passing a callback as a second parameter to provide a default value](https://laravel.com/docs/7.x/cache#retrieving-items-from-the-cache). This means you an do this:

```php
$forums = $cache->get('forums', function () {
    // forums cache does not exist, read forums from DB and return them - this will add them to the cache and return the rsult
});
```

We've already discussed adopting the Laravel caching library in the future, I was looking at it for 1.9 in #2961 but ended up closing it - perhaps I'll have another look at it soon.

euantorano:
Another note is that the PHPDoc for a lot of the cache handlers seems wrong, such as the APC cache handler's `fetch()` function saying:

> @return boolean True if successful, false on failure

However, it returns false if not in the cache, otherwise it returns mixed.

yuliu:
> It seems most cache handlers use `unserialize()`:
> 
>     * [`apc`](https://github.com/mybb/mybb/blob/feature/inc/cachehandlers/apc.php#L63)
> 
>     * [`eaccelerator`](https://github.com/mybb/mybb/blob/feature/inc/cachehandlers/eaccelerator.php#L66)
> 
>     * [`redis`](https://github.com/mybb/mybb/blob/feature/inc/cachehandlers/redis.php#L97)
> 
> 
> However, the main `datacache` class uses `my_unserialize()` since #3649 (which is my fault 😉). I find it interesting that some drivers do serialisation themselves, which doesn't quite seem right...

And the `cache()` function uses PHP built-in `unserialize()` https://github.com/mybb/mybb/blob/mybb_1822/inc/class_datacache.php#L112 where as `read()` uses the custom one.

From the `datacache` class's design, we can see that raw content is fed to cache handlers, so each driver will do serialization on putting or unserialization on getting, in accordance with the internal `db` handler.

euantorano:
Yeah, we need to try be a little more consistent with it.

inc/class_core.php
```diff
@@ -581,6 +581,9 @@ function trigger_generic_error($code)
 				$message = "Xcache needs to be configured with PHP to use the Xcache cache support.";
 				$error_code = MYBB_CACHEHANDLER_LOAD_ERROR;
 				break;
+			case "cache_runtime_error":
+				$message = "The cache store is either misconfigured or not working properly.";
+				$error_code = E_USER_WARNING;	// TODO: or MYBB_CACHEHANDLER_LOAD_ERROR for triggering an error?
```
Sama34, line 586:
```suggestion
				$error_code = E_USER_WARNING;	// TODO: or MYBB_CACHEHANDLER_LOAD_ERROR for triggering an error?
				break;
```

inc/class_datacache.php
```diff
@@ -450,6 +462,113 @@ function size_of($name='')
...
+	 * @param array $name The cache components to check.
+	 * @return bool True on successfully rebuilt cache, false otherwise.
+	 */
+	function check_cache($package = 'core', $cache_names = array())
+	{
+		// Ensure the value is fetched from the original datacache table.
+		$this->read('cache_checked', true);
+
+		// Only continue check if forced and cache is out-dated or non-existent.
+		if(isset($this->cache['cache_checked']) && TIME_NOW - $this->cache['cache_checked'] < 60 * 10)	//TODO: hard coded time?
```
euantorano, line 499: I think a hard-coded time is fine, though perhaps define it as a constant rather than magic numbers?

inc/class_datacache.php
```diff
@@ -450,6 +462,113 @@ function size_of($name='')
...
+		{
+			return false;
+		}
+
+		// The order is concerned, since some cache is relying on others, i.e., rebuilding 'forumpermissions' relies on 'forums' and 'usergroups' cache.
+		$cache_core = array('version', 'attachtypes', 'smilies', 'badwords', 'forums', 'usergroups', 'moderators', 'forumpermissions', 'stats', 'statistics', 'usertitles', 'reportedcontent', 'awaitingactivation', 'mycode', 'profilefields', 'posticons', 'spiders', 'bannedips', 'banned', 'bannedemails', 'birthdays', 'groupleaders', 'threadprefixes', 'forumsdisplay', /*'plugins', 'internal_settings', */'default_theme', 'reportreasons');
```
euantorano, line 505: Can we split this up into lines? This single lone line is quite difficult to read.

```suggestion
		$cache_core = array(
			'version',
			'attachtypes',
			'smilies',
			'badwords',
			'forums',
			'usergroups',
			'moderators',
			'forumpermissions',
			'stats',
			'statistics',
			'usertitles',
			'reportedcontent',
			'awaitingactivation',
			'mycode',
			'profilefields',
			'posticons',
			'spiders',
			'bannedips',
			'banned',
			'bannedemails',
			'birthdays',
			'groupleaders',
			'threadprefixes',
			'forumsdisplay',
			/*'plugins', 'internal_settings', */'default_theme',
			'reportreasons'
		);
```

inc/class_core.php
```diff
@@ -581,6 +581,10 @@ function trigger_generic_error($code)
...
```
euantorano, line 586: I think `E_USER_WARNING` is fine here, `TODO` can be removed.

inc/class_datacache.php
```diff
@@ -450,6 +462,113 @@ function size_of($name='')
...
+		foreach($cache_to_check['core'] as $core_cache)
+		{
+			// If forced to read from database, the cache key should exist with value of false.
+			if($this->read($core_cache, true) === false)
+			{
+				$update = 'update_'.$core_cache;
+				if(method_exists($this, $update))
+				{
+					call_user_func(array($this, $update));
+					continue;
```
euantorano, line 563: This `continue` seems unnecessary?

euantorano (changes requested):
This looks good, just some minor comments.

Sama34:
All related/referenced PRs/issues seem merged/fixed by now.

Has this PR become superseded at this point ? So maybe we should close this.

Issue #3724, opened 2019-06-16
Plugins hooks in forum display


chack1172:
I think I new hook should be added in forumdisplay.php after the $threadcache and $tids are defined.
This will help making only one new query to the database if extra data is needed for a plugin.
Otherwise they should run one extra query for each thread displayed.

Eldenroot:
@chack1172 - could you please open a PR for that?

veryard:
I'm not 100% sure if this is correct, a bit more detail would help, if someone can check see if this is okay.

Thanks

PaulBender:
Since this issue has come up, there is one hook I want to add to forumdisplay.php, just above the eval for $threadslist. I'm making a PR for that right now.

Eldenroot:
@veryard where is your PR? If you have any issues, let me know and I will help you

Eldenroot:
@PaulBender and @chack1172 - are you willing to create a PR for the first part of this issue? Thank you

veryard:
@Eldenroot Looks like I made a commit to my fork rather than a PR.

https://github.com/veryard/mybb/commit/149d4329cdc5631ee319a12a8efb09692aad788e

PR #3805, opened 2019-09-19
New plugin hook for forumdisplay.php (as mentioned in #3724)

PaulBender:
A new hook I need in forumdisplay.php.


euantorano:
Related to #3724.

PR #3808, opened 2019-09-27
Resolves #2440 Image paste from clipboard support

### LINKED ISSUE #2440: Image paste from clipboard broken
Hi,

just installed myBB on our server. First thing I tested was image paste from the clipboard ... but this seems to be broken. Is this a known issue? I couldn't find it in the forum or on github.
Short description:
Paste image works in the editor, but displays as source code in the preview and in the final posted version

If its not a known issue, I'm happy to provide more details.

Jan

Original thread: [Image paste from clipboard broken](http://community.mybb.com/thread-194877.html)

effone:
Resolves #2440

Image paste from clipboard support added.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -451,20 +451,18 @@ class postParser
 		$message = $this->mycode_parse_quotes($message);
 
 		// Convert images when allowed.
-		if(!empty($this->options['allow_imgcode']))
+		$allow_imgcode = "";
+
+		if(empty($this->options['allow_imgcode']))
 		{
-			$message = preg_replace_callback("#\[img\](\r\n?|\n?)(https?://([^<>\"']+?))\[/img\]#is", array($this, 'mycode_parse_img_callback1'), $message);
-			$message = preg_replace_callback("#\[img=([1-9][0-9]*)x([1-9][0-9]*)\](\r\n?|\n?)(https?://([^<>\"']+?))\[/img\]#is", array($this, 'mycode_parse_img_callback2'), $message);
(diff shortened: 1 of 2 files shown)


inc/class_parser.php
```diff
@@ -1332,6 +1343,17 @@ function mycode_parse_img_disabled_callback4($matches)
...
 
+	/**
+	 * Parses IMG MyCode disabled.
+	 *
+	 * @return string Raw image code notice.
+	 */
+	function mycode_parse_img_disabled_callback5()
+	{
+		global $lang;
+		return $lang->posted_image_uri;
```
euantorano: Shouldn't this behave the same as the other disabled images by calling `mycode_parse_img_disabled`?

dvz:
Do we want inline image support in the long run (rather than handling them as attachments client-side when pasted/dropped)? Best to go over this before adding support for formats that may be dropped later @mybb/developers

euantorano:
IMO this is what should happen:

1. Image pasted/dropped in to editor
2. JS detects this, and starts an AJAX request to create an attachment
3. If attachment added, atatchment is inserted in post/thread/PM (note: PMs don't support attachments in 1.8...)

In the long run, base 64 content for what could be large (several MB) image files inside the posts table would only cause headaches.

However, implementing the above wouldn't be finished by 1.8.22 unless we want to delay the release by several weeks.

I wouldn't mind stripping out the base 64 support for now and creating a new issue for 1.8.23 to implement the above flow. Or we can leave it as is for 1.8.22 and implement the above for 1.8.23 - only problem will be any posts created between 1.8.22 and 1.8.23 could have inline base 64 image data.

dvz:
Yes, we could revert this one to avoid potential backtracking on MyCode support (individual forums could still install a plugin for this, which should be simple to create).

euantorano:
I think that would be for the best for now.

PR #3812, opened 2019-09-29
Fix #3123 - StopForumSpam confidence levels

### LINKED ISSUE #3123: StopForumSpam confidence levels
This is related to #2910 and the PR that resolved it - #3121.

Currently, the Stop Forum Spam spam prevention system has a single `confidence` level that encapsulates emaila ddress, username and IP address confidence from the Stop Forum Spam. In the future, it would be a good idea to instead have 4 confidence settings:

- A confidence level for usernames - if the username exceeds this confidence level, block the registration/post/whatever.
- A confidence level for email addresses - if the email address exceeds this confidence level, block the registration/post/whatever.
- A confidence level for IP addresses - if the IP address exceeds this confidence level, block the registration/post/whatever.
- An overall confidence level which is `username + email + ip address / 3` (or by however many things are actually checked) - this will cath cases where the username isn't 100% confident, nor is the email but when both are together, the system is triggered.

We can leave this to 1.9 or a future release, as it's an enahcnement rather than a bug fix.

euantorano:
Resolves #3123 

This is still a work in progress.

**TODO**

- [X] Add a new interface for classes to check against anti-spam interfaces
- [X] Better handling for errors when checking anti-spam services
- [X] Ability to check against individual confidence levels (username, email, IP address)
- [ ] New settings for the above
- [ ] Testing the new implementation
- [ ] Updating usages of the old implementation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "illuminate/events": "5.7.*",
         "illuminate/hashing": "5.7.*",
         "illuminate/routing": "5.7.*",
-        "twig/twig": "^2.11"
+        "twig/twig": "^2.11",
+        "ext-json": "*"
     },
     "require-dev": {
         "mockery/mockery": "^1.2",
--- a/inc/src/AntiSpam/AntiSpamServiceInterface.php
+++ b/inc/src/AntiSpam/AntiSpamServiceInterface.php
@@ -18,5 +18,5 @@ interface AntiSpamServiceInterface
      *
      * @return \MyBB\AntiSpam\CheckResult The result of the check.
      */
-    function checkUser(string $username, string $emailAddress, string $ipAddress): CheckResult;
+    public function checkUser(string $username, string $emailAddress, string $ipAddress): CheckResult;
 }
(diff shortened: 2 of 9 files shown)


euantorano:
This is called a [fluid interface](https://en.wikipedia.org/wiki/Fluent_interface) and it allows method chaining on an object:

```php
$checkResult = (new \MyBB\AntiSpam\CheckResult())
    ->setResult(\MyBB\AntiSpam\CheckResult::ERROR)
    ->setError('error checking...')
    ->setTotalConfidence(0.0)
    ->setNumberOfChecks(3);
```

It is a very well known design pattern.

apple-fan-pl:
> This is called a [fluid interface](https://en.wikipedia.org/wiki/Fluent_interface) and it allows method chaining on an object:
> 
> ```
> $checkResult = (new \MyBB\AntiSpam\CheckResult())
>     ->setResult(\MyBB\AntiSpam\CheckResult::ERROR)
>     ->setError('error checking...')
>     ->setTotalConfidence(0.0)
>     ->setNumberOfChecks(3);
> ```
> 
> It is a very well known design pattern.

Not so well known if I havent had an idea about this. Indeed seems very useful. Thanks @euantorano :)

Ben-MyBB:
@euantorano just a reminder about this one whenever you get time

euantorano:
Yep, planning to finish and merge as soon as I can.

PR #3817, opened 2019-10-10
Update helpdocs.lang.php

travelkris:
Location of User CP link is at upper left, not right in default theme.

Resolves #ISSUE_ID


yuliu:
I suggest these position-depended directions be replaced with actual links, better to do so in future versions.

euantorano:
Agreed, definitely.

travelkris:
I agree that a link would be more expedient. However, the purpose of this help page is to teach users how to use the forum without having to return to the help page to click a link each time they want to access the User CP.

PR #3821, opened 2019-10-29
Resolves #3813 Counter correction

### LINKED ISSUE #3813: mybb doesn't parse IP address urls in threads
as seen here: https://community.mybb.com/thread-224823.html

Original thread: [mybb doesn't parse IP address urls in threads](https://community.mybb.com/thread-224825.html)

effone:
Resolves #3813


dvz:
Regressions found (e.g. `https://example.net` is not parsed correctly)

PR #3822, opened 2019-11-02
Fixes #3702 - Case sensitivity in English

### LINKED ISSUE #3702: Case sensitivity in English (Mark forum read)
There is an irregular case sensitivity in language "English (American)".

forumdisplay.lang.php &gt; markforum_read:
"Mark this forum read"

global.lang.php &gt; bottomlinks_markread:
"Mark All Forums Read"

Better keep pervasive lower case to harmonise.
Maybe there are more case sensitivity issues on other lang-records worth to seek out.

[ExiTuS]

Original thread: [Case sensitivity in English (Mark forum read)](https://community.mybb.com/thread-223334.html)

Eldenroot:
#3702

Resolves #ISSUE_ID


yuliu:
But all the bottom links are capitalized in each word's first character.

euantorano:
Yes, IMO there’s no need for those links to be title case either. I guess that comes down to UI/UX though which Justin would be better placed to comment on.

PR #3824, opened 2019-11-04
Fix #3820 - extend attachment filename length

### LINKED ISSUE #3820: error: Attachment file name is too long
When uploading an attachment, if the filename is longer than the field length, I get an error. This can happen if the filename details photo content.
Quote:SQL Error: 1406 - Data too long for column 'filename' at row 1
Query: 
                        INSERT
                        INTO mybb_attachments (pid,posthash,uid,filename,filetype,filesize,attachname,downloads,dateuploaded,thumbnail,visible)
                        VALUES (0,'82308f13b0ceb97b6b3d8988ef3af7a2','262','Marlene Dietrich, Jean Harlow, Joan Blondell (looking bored unknown ( Elizabeth Allan), Edward G. Robinson, and Dolores del Rio..jpg','image/jpeg',80341,'201910/post_262_1571487304_daacdaa987c4bdbfdf717ea2d58b462c.attach',0,1571487304,'201910/post_262_1571487304_daacdaa987c4bdbfdf717ea2d58b462c_thumb.jpg',1)

Error is backtraced to /inc/functions_upload.php line 764.

I have increased field length in table mybb_attachments / filename from varchar(120) to varchar(200). I hope this does not have unintended consequences. This option is better than allowing the field to be truncated by dropping STRICT_TRANS_TABLES from sql-mode.

However, a better solution would be validation of the filename length and generating an error message similar to what I would get if filesize is too great.

Would this work? Please suggest corrections to syntax.

438      // Check the filename length
439      if(strlen($attachment['filename']) &gt; 120 != &quot;&quot;)
440      {
441          $ret['error'] = $lang-&gt;sprintf($lang-&gt;error_filenamelength, htmlspecialchars_uni($attachment['name']), &quot;120&quot;);
442          return $ret;
443      }

It should be obvious I'm attempting to modify an existing portion of code, and it should *really* be obvious I am not experienced. But I am always willing to learn. :)

Where would I insert the new error message?

Original thread: [error: Attachment file name is too long](https://community.mybb.com/thread-225080.html)

euantorano:
Resolves #3820 

The attachment filename length limit is now 260 characters, based upon the following references:

- https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations
- https://docs.microsoft.com/en-gb/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#maximum-path-length-limitation

I haven't tested this yet.


yuliu:
So here `260=256+4`?
256 characters for Windows system should be just fine. The other 4 characters that are a driver letter, a colon, a backslash and a terminating null character, together with a 256-character filename, will form a 260-character **path**.

BTW, I don't know why the wiki pages say the maximum length is 255 characters.

euantorano:
Looks like 255 might cover it, yes. I just read up to the part:

> the maximum length for a path is MAX_PATH, which is defined as 260 characters

Update incoming.

PR #3830, opened 2019-11-19
Resolve #3829 "Access Own Unapproved Threads / Posts" breaks thread filters & pagination

### LINKED ISSUE #3829: "Access Own Unapproved Threads / Posts" breaks thread filters & pagination
The `showownunapproved` setting-dependent thread visibility condition (#3493) breaks the `AND`/`OR` logic in https://github.com/mybb/mybb/blob/mybb_1821/forumdisplay.php#L859, causing additional thread filtering (prefixes, dates) to be ignored when the setting is enabled.
Additionally, the visibility condition for own threads is absent in https://github.com/mybb/mybb/blob/mybb_1821/forumdisplay.php#L615, which may result in incorrect page (thread) count when the setting is enabled.

reported by @Eldenroot, @yuliu

dvz:
Resolves #3829

Simplified construction of `t.visible` conditions. `$visibleonly` left out for (presumably) `posts` context.

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -387,15 +387,11 @@ if($foruminfo['rulestype'] != 0 && $foruminfo['rules'])
 $bgcolor = "trow1";
 
 // Set here to fetch only approved/deleted topics (and then below for a moderator we change this).
+$visible_states = array("1");
+
 if($fpermissions['canviewdeletionnotice'] != 0)
 {
-	$visibleonly = "AND visible IN (-1,1)";
-	$tvisibleonly = "AND t.visible IN (-1,1)";
-}
-else
-{
-	$visibleonly = "AND visible='1'";
-	$tvisibleonly = "AND t.visible='1'";
+	$visible_states[] = "-1";
 }
 
 // Check if the active user is a moderator and get the inline moderation tools.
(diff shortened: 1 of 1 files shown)


yuliu:
Using only `" OR t.uid=".(int)$mybb->user['uid']` for fetching user's own unapproved threads is not a smart way, or maybe not a right way. By doing so, the queries will just return all the users threads, including drafted threads (`visible = -2`) & soft-deleted threads (`visible = -1`), according to other criteria.

I think using a `$visible_states` array is great. And based on that idea, you should also make queries to use an extra variable for current logged in user who's viewing a forum, like `$visible_states_user`.

Suppose current user has no moderation permissions, for `showownunapproved`:
- if it is set to `off`, don't fetch threads of `visible = 0` by not appending `0` to `$visible_states_user` for them.
- if it is set to `on`, append `0` to `$visible_states_user`.

Suppose current user has no moderation permissions, for `canviewdeletionnotice`:
- if it is set to `off`, don't fetch threads of `visible = -1` by not appending `-1` to `$visible_states_user` for them.
- if it is set to `on`, append `-1` to `$visible_states_user`.

Actually the `$visible_states_user` variable can be constructed regardless of the user's having moderation permissions or not, as long as the user is logged in. Then that final SQL condition should become as follows, if `$visible_states_user` is empty:
```
visible IN (".implode(',', array_unique($visible_states)).")
```
if `$visible_states_user` is not empty:
```
visible IN (".implode(',', array_unique($visible_states)).")
OR
t.uid=".(int)$mybb->user['uid'] AND visible IN (".implode(',', array_unique($visible_states_user)).")
```

In addition, some control codes for correctly displaying unapproved & soft-deleted threads might need review, too.

dvz:
Looks like `canviewdeletionnotice` applies to all content, not just the user's, so I've extended the user condition with the `0` status instead of an array.

yuliu:
Seems working for me now.

Looks like `canviewdeletionnotice` is not limited to user's own threads as I thought before.

BTW, MyBB really should have these permissions/variables/definitions noted or documented somewhere publicly, as I've requested months ago, not only for convenient development for developers but also forum management for forum owners.

Eldenroot:
I applied this fix on my board... will testing for few more days, seems to be fine now :) thx!

Eldenroot:
Why it is not applied to 1.8.22 release? This regression is quite annoying and breaks one of the core feature :(

euantorano:
It needs some more testing, and will be applied to the community forums. 1.8.22 isn’t fully released yet, just a pre-release.

PR #3831, opened 2019-11-24
Fix #3828 - Alter the [font] MyCode

### LINKED ISSUE #3828: [font] tag parsing allows setting of some HTML attributes
https://github.com/mybb/mybb/blob/465dd9af27a6870e0481bac3bc8de6d9bcb63dfc/inc/class_parser.php#L371-L372

Examples:
```
[font=Arial" contenteditable ]This span will be editable.[/font]
[font=Arial" hidden ]This span will be hidden.[/font]
```

euantorano:
Fixes #3828 by altering the `[font]` MyCode to only allow quotes at the start and end and to ensure the used quotes match each other.

Tested locally using the provided broken tag examples.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -368,7 +368,7 @@ class postParser
 
 		if($mybb->settings['allowfontmycode'] == 1)
 		{
-			$nestable_mycode['font']['regex'] = "#\[font=(\\\"?)([a-z0-9 ,\-_']+)\\1\](.*?)\[/font\]#si";
+			$nestable_mycode['font']['regex'] = "#\[font=(\"?)([a-z0-9 ,\-_']+)\\1\](.*?)\[/font\]#si";
 			$nestable_mycode['font']['replacement'] = "<span style=\"font-family: $2;\" class=\"mycode_font\">$3</span>";
 
 			++$nestable_count;
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -368,7 +368,7 @@ class postParser
 
 		if($mybb->settings['allowfontmycode'] == 1)
 		{
-			$nestable_mycode['font']['regex'] = "#\[font=([\\\"']?)([a-z0-9 ,\-_]+)\\1\](.*?)\[\/font\]#si";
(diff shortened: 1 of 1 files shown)


dvz:
Note that dropping support for quotes may break functionality, where the font family name needs to be quoted: https://www.w3.org/TR/css-fonts-3/#family-name-value

Leaving single quotes may be a good compromise. We could also have a function to convert double quotes `"` to single `'`, but that may further interfere with intended/original behavior.

inc/class_parser.php
```diff
@@ -368,8 +368,8 @@ function cache_mycode()
 
 		if($mybb->settings['allowfontmycode'] == 1)
 		{
-			$nestable_mycode['font']['regex'] = "#\[font=([a-z0-9 ,\-_'\"]+)\](.*?)\[/font\]#si";
-			$nestable_mycode['font']['replacement'] = "<span style=\"font-family: $1;\" class=\"mycode_font\">$2</span>";
+			$nestable_mycode['font']['regex'] = "#\[font=([\\\"']?)([a-z0-9 ,\-_]+)\\1\](.*?)\[\/font\]#si";
```
dvz: Doesn't look like escaping the forward slash is needed in `[\/font\]`

euantorano: True. Will remove shortly.

inc/class_parser.php
```diff
@@ -368,8 +368,8 @@ function cache_mycode()
 
 		if($mybb->settings['allowfontmycode'] == 1)
 		{
-			$nestable_mycode['font']['regex'] = "#\[font=([a-z0-9 ,\-_'\"]+)\](.*?)\[/font\]#si";
-			$nestable_mycode['font']['replacement'] = "<span style=\"font-family: $1;\" class=\"mycode_font\">$2</span>";
+			$nestable_mycode['font']['regex'] = "#\[font=(\\\"?)([a-z0-9 ,\-_']+)\\1\](.*?)\[/font\]#si";
```
dvz: The `"` should be fine unescaped for regex, so `\\\"` can be changed to `\"`.

PR #3837, opened 2019-12-05
Resolve #3762 Purging guest sessions by IP address may invalidate anti-CSRF tokens between requests

### LINKED ISSUE #3762: Purging guest sessions by IP address may invalidate anti-CSRF tokens between requests

dvz:
Resolves #3762 
Replaces #3769, #3833

This solution filters data from the `mybb_sessions` table in `SELECT` queries instead of relying on deleting sessions from same IP addresses on creation (MyBB 1.8.21) or updating rows (referenced PRs).

The `INNER JOIN` subquery groups records by `{uid, ip}` and provides the most recent timestamp (`MAX(time)`), which may result in returning more than 1 guest session with the same IP address if the `time` value is also the same. User sessions, however, should be unique, as duplicates are still deleted at https://github.com/mybb/mybb/blob/mybb_1821/inc/class_session.php#L498.

### DIFF
--- a/online.php
+++ b/online.php
@@ -159,7 +159,9 @@ else
 	{
 		case "sqlite":
 			$sessions = array();
-			$query = $db->simple_select("sessions", "sid", "time > {$timesearch}");
+			$query = $db->simple_select("sessions", "MAX(sid)", "time > {$timesearch}", array(
+				'group_by' => 'uid, ip',
+			));
 			while($sid = $db->fetch_field($query, "sid"))
 			{
 				$sessions[$sid] = 1;
@@ -169,7 +171,9 @@ else
 			break;
 		case "pgsql":
 		default:
-			$query = $db->simple_select("sessions", "COUNT(sid) as online", "time > {$timesearch}");
+			$query = $db->simple_select("sessions", "COUNT(sid) as online", "time > {$timesearch}", array(
+				'group_by' => 'uid, ip',
+			));
 			$online_count = $db->fetch_field($query, "online");
 			break;
 	}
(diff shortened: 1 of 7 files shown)


online.php
```diff
@@ -159,7 +159,9 @@
 	{
 		case "sqlite":
 			$sessions = array();
-			$query = $db->simple_select("sessions", "sid", "time > {$timesearch}");
+			$query = $db->simple_select("sessions", "MAX(sid)", "time > {$timesearch}", array(
+			    'group_by' => 'uid, ip',
+            ));
```
euantorano: Indentation here is off - it seems to use spaces for the closing `));` rather than tabs.

```suggestion
			));
```

online.php
```diff
@@ -169,7 +171,9 @@
 			break;
 		case "pgsql":
 		default:
-			$query = $db->simple_select("sessions", "COUNT(sid) as online", "time > {$timesearch}");
+			$query = $db->simple_select("sessions", "COUNT(sid) as online", "time > {$timesearch}", array(
+			    'group_by' => 'uid, ip',
+            ));
```
euantorano: Indentation here is off - it seems to use spaces for the closing `));` rather than tabs.

```suggestion
			));
```

euantorano:
Conflicts according to the GH interface @dvz

dvz:
Reverted #3833 via #3839, and #3769 via 8f9b22d1566b3e905b620cd6848bd550682a7533, should be good now

euantorano:
Yep, that seems to have resolved the conflicts.

PR #3843, opened 2019-12-10
Fixes #3818

### LINKED ISSUE #3818: Header no longer displays red alert join requests to groupleaders
Hello there,

I've noticed that earlier this year after an update, the header no longer showed pending group requests for groupleaders. It used to be a red alert message. Here's a picture of the header on the forums:

Here's the header template:

A few plugins are actively used -- but none I recall being modified/added when the functionality went away. Like System, Username Changes limit and history, edit history logs, shoutbox, automedia, admin user assumption.

I found this prior service request, which lead me to the post the header pictures above. 

Any ideas would certainly be helpful and appreciated to get this feature working again.

Thank you!

Original thread: [Header no longer displays red alert join requests to groupleaders](https://community.mybb.com/thread-224958.html)

Eldenroot:
Resolves #3818

### DIFF
--- a/global.php
+++ b/global.php
@@ -630,7 +630,7 @@ if($mybb->user['uid'] != 0 && is_array($groupleaders) && array_key_exists($mybb-
 
 		$user['gid'] = (int)$user['gid'];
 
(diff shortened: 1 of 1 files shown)


euantorano:
This seems like the easiest fix and works for me locally, thanks!

PR #3844, opened 2019-12-16
Fix & optimize Who's Online queries

### LINKED ISSUE #3762: Purging guest sessions by IP address may invalidate anti-CSRF tokens between requests

dvz:
Resolves #3762
Extends #3837

- DB interaction modified to count guests in a separate query with `COUNT(DISTINCT ip)`,
- `index.php` guest count query extended to return counts grouped by `location1` when `$mybb->settings['showforumviewing'] != 0` for the _N users browsing_ forum counts,
- `online.php`:
  - more complex queries to filter guests by IP database engine version-dependent window functions, with a fallback for older versions (`MIN(sid)` only used to pick unique `sid` for each `{uid,ip}` pair when same timestamp is stored); no duplicate `{uid,ip}` pairs will be returned anymore (side effect in #3837), but spider and guest sessions associated with same IP may exclude each other in statistics
  - fixed sorting for SQLite to show users first

Queries in this state rely on https://github.com/mybb/mybb/blob/mybb_1821/inc/class_session.php#L498 to not show duplicate user entries when the forum is visited using the same account from different IP addresses.

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -283,49 +283,41 @@ if($mybb->settings['browsingthisforum'] != 0)
 	$onlinemembers = '';
 	$doneusers = array();
 
+	$query = $db->simple_select("sessions", "COUNT(DISTINCT ip) AS guestcount", "uid = 0 AND time > $timecut AND location1 = $fid AND nopermission != 1");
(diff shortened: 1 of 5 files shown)


PaulBender:
Looks good to me.

PR #3868, opened 2020-01-05
Fixes #3781 - Long avatar URL causes MySQL error

### LINKED ISSUE #3781: Long avatar URL causes MySQL error
There should probably be better validation for this.

SQL Error:
1406 - Data too long for column 'avatar' at row 1
Query:
UPDATE mybb_users SET `avatar`='http:...', `avatardimensions`='62|62', `avatartype`='remote' WHERE uid='1'

Eldenroot:
Resolves #3781

### DIFF
--- a/install/resources/mysql_db_tables.php
+++ b/install/resources/mysql_db_tables.php
@@ -1057,7 +1057,7 @@ $tables[] = "CREATE TABLE mybb_users (
   email varchar(220) NOT NULL default '',
   postnum int(10) unsigned NOT NULL default '0',
   threadnum int(10) unsigned NOT NULL default '0',
-  avatar varchar(200) NOT NULL default '',
+  avatar varchar(255) NOT NULL default '',
   avatardimensions varchar(10) NOT NULL default '',
   avatartype varchar(10) NOT NULL default '0',
   usergroup smallint unsigned NOT NULL default '0',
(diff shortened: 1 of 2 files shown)


Eldenroot:
Guys please review this :)

Ben-MyBB:
@dvz can you give this a quick glance please?

JordanMussi:
What about sqlite?
https://github.com/mybb/mybb/blob/b43cf6f7f6ec6aca745fc8ee469ea7ed0a6affa0/install/resources/sqlite_db_tables.php#L944

dvz:
Best to check input length and show a message (or quietly set to empty), otherwise it can still result in errors.

JordanMussi:
Also I can't see anything relating to ensuring the maximum length is 255 when updating in User CP:

https://github.com/mybb/mybb/blob/9feef312dc22abef36dcc1ddb1a703e66d9c2fe1/usercp.php#L2487-L2492

JordanMussi:
Beaten to it by @dvz! It might be a good idea to refactor to use the user datahandler and put the validation in there 😃

euantorano:
Yeah, if the data handler is used to update the handler, the validation should be done there. Most plugins affecting avatars would hopefully use the data handler so having the validation there makes sense.

dvz:
We'll need upgrade code as well.

Sama34:
I think is necessary to increase the field size, instead the user datahandler should validate user input.

PR #3875, opened 2020-01-06
Fix #3851 User CP -> pagination in Attachments is broken

### LINKED ISSUE #3851: User CP -> pagination in Attachments is broken
I have 21 pages of attachments in my UCP -> Attachments

However last 3 pages (19, 20, 21) are empty, no attachment is displayed... Also statistics are empty for these pages (this is fine, but the statistci should be counted for all attachments together, not just for one page...)

image

demtor:
Resolves #3851 + optimization that may most of the time save a query

### DIFF
--- a/usercp.php
+++ b/usercp.php
@@ -3892,7 +3892,7 @@ if($mybb->input['action'] == "attachments")
 		++$processedattachments;
 	}
 
-	if ($processedattachments >= $perpage) {
+	if ($processedattachments >= $perpage || $page > 1) {
 		$query = $db->query("
 			SELECT SUM(a.filesize) AS ausage, COUNT(a.aid) AS acount
 			FROM ".TABLE_PREFIX."attachments a
--- a/usercp.php
+++ b/usercp.php
@@ -3892,7 +3892,8 @@ if($mybb->input['action'] == "attachments")
 		++$processedattachments;
 	}
 
-	if ($processedattachments >= $perpage || $page > 1) {
+	if($processedattachments >= $perpage || $page > 1)
+	{
 		$query = $db->query("
 			SELECT SUM(a.filesize) AS ausage, COUNT(a.aid) AS acount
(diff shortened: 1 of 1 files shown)


Eldenroot:
No, it is not working properly. When you click to the latest page in attachment -> pagination will be hidden.

Moreover the statistic window below the pagination is still broken -> it should display values for all attachments, not just for the viewed page.

Eldenroot:
More info:
- only last page is broken
- total downloads and bandwith in statistics are broken (counted only for page before the page you are on, for example when you are on page 15, only 1-15 are counted into statistic; all after page 15 is not).

demtor:
My bad lol, should be okay now. Anyway I think that total downloads and bandwidth usage needs a separate issue since I guess is not really a bug but something done on purpose?

Eldenroot:
I dont think so... others values are calculated correctly. Could you please fix it, or at least @euantorano your opinion?

euantorano:
I'd be happy with a separate issue for the bandwidth/total downloads problem.

Eldenroot:
This fix works fine as it is... for the mentioned issue above we should open a new ticket as euan wrote.

Sama34:
Do you have reproduction steps to review this? I couldn't reproduce in my local envioroment.

Eldenroot:
@Sama34 this PR fixes one bug. Anyway I have found the second one, so lets merge this PR and I will open a new ticket, right? @euantorano

Sama34:
I checked this and everything seems to be as before, nothing changed for me and I didn't find the query being saved, so I won't review this but I suppose it is fine by me to merge if someone else reviews is able to test properly.

Eldenroot:
@Sama34 @euantorano @yuliu 

This PR fixes pagination, so one issue fixed. However we need to open a second ticket for wrong statistics calculation

PR #3879, opened 2020-01-14
Resolve #3878 Remove "banned" datacache

### LINKED ISSUE #3878: Remove "banned" datacache
Remove the `banned` data cache entry and rely on the user group's `isbannedgroup` value instead to determine banned status, fetching additional data from the `mybb_banned` table on positive result.

The status is used for instant ban lift checks for the visiting user (https://github.com/mybb/mybb/blob/mybb_1822/inc/class_session.php#L251).
MyBB installations that don't use the `isbannedgroup` type (_Yes, this is a banned group_) will expire account bans through the [_Half-hourly User Cleanup_ task](https://github.com/mybb/mybb/blob/mybb_1822/inc/tasks/usercleanup.php#L53).

The [query used for the _banned_ notice](https://github.com/mybb/mybb/blob/mybb_1822/global.php#L813) may no longer be necessary.

Related: #3876, #3685, https://community.mybb.com/thread-214400.html

dvz:
Resolves #3878

### DIFF
--- a/install/resources/upgrade50.php
+++ b/install/resources/upgrade50.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * MyBB 1.8
+ * Copyright 2014 MyBB Group, All Rights Reserved
+ *
+ * Website: http://www.mybb.com
+ * License: http://www.mybb.com/about/license
+ *
+ */
+
+/**
+ * Upgrade Script: 1.8.22
+ */
+
+$upgrade_detail = array(
+	"revert_all_templates" => 0,
+	"revert_all_themes" => 0,
+	"revert_all_settings" => 0
+);
+
+@set_time_limit(0);
+
+function upgrade50_dbchanges()
+{
+	global $output, $cache;
+
+	$output->print_header("Updating Cache");
+
+	echo "<p>Updating cache...</p>";
+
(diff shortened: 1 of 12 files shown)


yuliu:
Seems appropriate. The `banned` row in table `data_cache` is just a duplicate of the table `banned`.

However, directly removing the `banned` cache row and its read/update functions might break plugins. Deprecating it now and then removing it after 1 another release seems a nicer way.

dvz:
> Deprecating it now and then removing it after 1 another release seems a nicer way.

We can make it dependent on a temporary `$config` value, if we go that route - override to continue rebuilding the cache (or a more careful, 2-step policy to first let administrators force-disable rebuilding in 1.8.23 and then force-enable it 1.8.24, and in 1.8.25 it would get removed).

install/resources/upgrade50.php
```diff
@@ -0,0 +1,35 @@
...
+
+function upgrade50_dbchanges()
+{
+	global $output, $cache;
+
+	$output->print_header("Updating Cache");
+
+	echo "<p>Updating cache...</p>";
+
+    $cache->delete("banned");
```
euantorano: Nit: indentation.

PR #3881, opened 2020-01-18
#3874: SQL Error when sorting by username in online.php

### LINKED ISSUE #3874: SQL Error when sorting by username in online.php
```
SQL Error: 1054 - Unknown column 'u.username' in 'order clause'
Query: 
	SELECT * FROM (
		SELECT
			s.sid, s.ip, s.uid, s.time, s.location, u.username, s.nopermission, u.invisible, u.usergroup, u.displaygroup,
			row_number() OVER (PARTITION BY s.uid, s.ip ORDER BY time DESC) AS row_num
			FROM
				mybb_sessions s
				LEFT JOIN mybb_users u ON (s.uid = u.uid)
			WHERE s.time > 1578249748
		) s
		WHERE row_num = 1
		ORDER BY u.username ASC, s.time DESC
		LIMIT 0, 30
```

euantorano:
Resolves #3874

### DIFF
--- a/online.php
+++ b/online.php
@@ -48,7 +48,7 @@ if($mybb->get_input('action') == "today")
 	{
 		$mybb->settings['threadsperpage'] = 20;
 	}
-	
+
 	// Add pagination
 	$perpage = $mybb->settings['threadsperpage'];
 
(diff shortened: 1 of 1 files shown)


demtor:
This would give the same error when the other query occurs (the one for older versions). Probably a `$sql` replace on `u.` or `u.username` with `s.` or `s.username` before the query call should be the cleanest way to fix the issue.

euantorano:
I don't have an old version of MySQL/PostgreSQL handy to test against unfortunately, but I've just tested on SQLFiddle and you seem to be right. Incoming patch.

PR #3882, opened 2020-01-18
#2962 error page display

### LINKED ISSUE #2962: Only show sensitive errors to admins by default
At the moment MyBB will show almost all errors including SQL errors to the end user regardless of status. On forums I've run in the past I've always modified the error handler to only show the sensitive DB errors to the administrator group.

euantorano:
Resolves #2962

This makes the following changes relating to how errors are displayed:

- Default the `Error Type Medium` setting to `Hide Errors and Warning`, meaning that by default full error details are not shown, which could leak potentially sensitive information such as database table prefixes.
- Prevent the `Error Type Medium` setting being hidden when the `Error Logging Medium` is `Neither`.
- When displaying an error, if the user is not an administrator change the text in the footer. For administrators, the text is:

    > Please contact the MyBB Group for technical support.

    For users, the text is:

    > Please contact the site owner for technical support.

    This should cut back on the number of emails we receive to `contact@` relating to an error on a site that we can do nothing about.

## Other possible future enhancements (for a future PR)

- For SQL errors, show the file where the error happened
- During install, if the URL is configured to localhost, set the `Error Type Medium` to `Warnings and Errors`
- Additional option for `Error Type Medium` to show errors to logged in admins, and hide them from normal users

### DIFF
--- a/inc/class_error.php
+++ b/inc/class_error.php
@@ -539,6 +539,28 @@ class errorHandler {
 			$charset = 'UTF-8';
 		}
 
+		if(isset($mybb->user) && is_array($mybb->usergroup) && $mybb->usergroup['cancp'] == 1)
+		{
+			// admins directed to contact MyBB
+			$contact = 'Please contact the <a href="https://mybb.com">MyBB Group</a> for technical support.';
+		}
+		else
+		{
+			// users directed to contact the site owner
+			$contact_us = 'site owner';
+			if(($mybb->settings['contactlink'] == "contact.php" && $mybb->settings['contact'] == 1 && ($mybb->settings['contact_guests'] != 1 && $mybb->user['uid'] == 0 || $mybb->user['uid'] > 0)) || $mybb->settings['contactlink'] != "contact.php")
+			{
+				if(!my_validate_url($mybb->settings['contactlink'], true, true) && my_substr($mybb->settings['contactlink'], 0, 7) != 'mailto:')
+				{
+					$mybb->settings['contactlink'] = $mybb->settings['bburl'].'/'.$mybb->settings['contactlink'];
+				}
+
+				$contact_us = '<a href="' . $mybb->settings['contactlink'] . '">' . $contact_us . '</a>';
+			}
+
+			$contact = 'Please contact the ' . $contact_us . ' for technical support.';
+		}
+
 		if(!headers_sent() && !defined("IN_INSTALL") && !defined("IN_UPGRADE"))
 		{
 			@header('HTTP/1.1 503 Service Temporarily Unavailable');
@@ -579,7 +601,7 @@ class errorHandler {
 
 			<div id="error">
 				{$error_message}
-				<p id="footer">Please contact the <a href="https://mybb.com">MyBB Group</a> for technical support.</p>
+				<p id="footer">{$contact}</p>
 			</div>
 		</div>
 	</div>
@@ -604,7 +626,7 @@ EOF;
 		<h2>{$title}</h2>
 		<div id="mybb_error_error">
 		{$error_message}
-			<p id="mybb_error_footer">Please contact the <a href="https://mybb.com">MyBB Group</a> for technical support.</p>
+			<p id="mybb_error_footer">{$contact}</p>
 		</div>
 	</div>
 EOF;
--- a/inc/class_error.php
+++ b/inc/class_error.php
@@ -555,11 +555,11 @@ HTML;
 
 			$contact = <<<HTML
 <p>
-	<strong>If you're a visitor of this website</strong>, please wait a few minutes.{$contact_site_owner}
+	<strong>If you're a visitor of this website</strong>, please wait a few minutes and try again.{$contact_site_owner}
 </p>
 
 <p>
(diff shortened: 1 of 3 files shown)


dvz:
> if the URL is configured to localhost

We can rely on [`MyBB::$dev_mode`](https://github.com/mybb/mybb/blob/mybb_1822/inc/class_core.php#L108) instead.

euantorano:
> We can rely on MyBB::$dev_mode instead.

I always forget that's there... Would certainly make sense for the future. We'd need to make a couple of installer changes to make it work though I guess. It's been a while since I looked at the install code.

inc/class_error.php
```diff
@@ -539,6 +539,28 @@ function output_error($type, $message, $file, $line)
...
+		if(isset($mybb->user) && is_array($mybb->usergroup) && $mybb->usergroup['cancp'] == 1)
+		{
+			// admins directed to contact MyBB
+			$contact = 'Please contact the <a href="https://mybb.com">MyBB Group</a> for technical support.';
+		}
+		else
+		{
+			// users directed to contact the site owner
+			$contact_us = 'site owner';
+			if(($mybb->settings['contactlink'] == "contact.php" && $mybb->settings['contact'] == 1 && ($mybb->settings['contact_guests'] != 1 && $mybb->user['uid'] == 0 || $mybb->user['uid'] > 0)) || $mybb->settings['contactlink'] != "contact.php")
```
Sama34, line 551: ` && $mybb->user['uid'] == 0` here seems redundant.

euantorano: This is copied from `global.php`: https://github.com/mybb/mybb/blob/128434e0b038a0b28f9cfc9e0d42bdd5cc28972b/global.php#L1038

Sama34: Still redundant but the main reason to ask for changes was actually what I mentioned above.

> Also, maybe you can check THIS_SCRIPT against contact.php to display the admin email instead of the contact link.

I'm fine with everything else. If users aren't able to use the core contact form maybe we should display something else like the contact email or simply no link.

euantorano: Yes, makes sense. Will change it, but I don’t think we should expose the contact email address itself - will probably just not have a contact link.

dvz:
Note that administrators may never see the support link if they're logged out and cannot log back in, or the error occurs early on and the status cannot be determined.

JordanMussi:
What about doing something similar to Cloudflare error pages where they give info for visitors & owners?

image

For example:
> **If you are browsing this website**
> Please try again in a few minutes or contact the site owner.
> 
> **If you're the owner of this website**
> Please check the [MyBB Documentation](https://docs.mybb.com) for help resolving [common issues](https://docs.mybb.com/1.8/faq/), or get technical help on the [MyBB Community Forums](https://community.mybb.com).

euantorano:
That’s probably the better option IMO.

euantorano:
Updated with new text for site owners and guests. It now looks like this:

Viewing index (board contact link shown):

Screenshot 2020-05-10 at 15 49 41

Viewing contact page (board contact link not shown):

Screenshot 2020-05-10 at 15 49 49

inc/class_error.php
```diff
@@ -539,6 +539,30 @@ function output_error($type, $message, $file, $line)
...
+HTML;
+		}
+
+			$contact = <<<HTML
+<p>
+	<strong>If you're a visitor of this website</strong>, please wait a few minutes.{$contact_site_owner}
+</p>
+
+<p>
+	<strong>If you are the site owner</strong>, lease check the <a href="https://docs.mybb.com">MyBB Documentation</a> for help resolving <a href="https://docs.mybb.com/1.8/faq/">common issues</a>, or get technical help on the <a href="https://community.mybb.com/">MyBB Community Forums</a>.
```
JordanMussi:
```suggestion
	<strong>If you are the site owner</strong>, please check the <a href="https://docs.mybb.com">MyBB Documentation</a> for help resolving <a href="https://docs.mybb.com/1.8/faq/">common issues</a>, or get technical help on the <a href="https://community.mybb.com/">MyBB Community Forums</a>.
```

euantorano: Just fixed that a moment ago 😉

euantorano:
Any other feedback? Would like to merge this for 1.8.23 if possible.

Ben-MyBB:
@Eldenroot Can you test this one please? I will apply this to my board and confirm results.

Ben-MyBB:
This works without issue locally. I suggest we give this special attention in the blog post to make forum administrators aware of the changes.

Eldenroot:
I agree with you @Ben-MyBB . This is a nice addition and improvement - in a matter of security.

yuliu:
An irrelevant question, isn't here for warning?
https://github.com/mybb/mybb/blob/634e2234978497d1d46bc40a4fe35b59e84afda9/inc/class_error.php#L210

euantorano:
Yeah, it's always been like that as far as I know. In the future, I'd like to improve it a little further to handle more cases. I also want to look into the comment at https://github.com/mybb/mybb/blob/634e2234978497d1d46bc40a4fe35b59e84afda9/inc/class_error.php#L164

PR #3891, opened 2020-02-09
Forumdisplay_thread - add start date

Eldenroot:
These two small improvements could be really useful. I think both of them are missing in core for a no reason.

I would like to open a discussion.

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3966,14 +3966,14 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 {$forums}
 </table>
 <br />]]></template>
-		<template name="forumdisplay_thread" version="1813"><![CDATA[<tr class="inline_row">
+		<template name="forumdisplay_thread" version="1823"><![CDATA[<tr class="inline_row">
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%"><span class="thread_status {$folder}" title="{$folder_label}">&nbsp;</span></td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%">{$icon}</td>
 	<td class="{$bgcolor}{$thread_type_class}">
 		{$attachment_count}
 		<div>
-          <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']}</div>
+          <span>{$prefix} {$gotounread}<a href="{$mybb->settings['bburl']}/forumdisplay.php?fid={$mybb->input['fid']}&prefix={$thread['prefix']}">{$thread['threadprefix']}</a><span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
+			<div class="author smalltext">{$thread['profilelink']} <span class="smalltext" title="{$lang->thread_started_time_date} {$threaddate}, {$threadtime}">({$threaddate}, {$threadtime})</span></div>
 		</div>
 	</td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}"><a href="{$mybb->settings['bburl']}/misc.php?action=whoposted&tid={$thread['tid']}" onclick="MyBB.whoPosted({$thread['tid']}); return false;">{$thread['replies']}</a>{$unapproved_posts}</td>
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3973,7 +3973,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 		{$attachment_count}
 		<div>
           <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']} <span class="smalltext" title="{$lang->thread_started_time_date} {$threaddate}, {$threadtime}">({$threaddate}, {$threadtime})</span></div>
+			<div class="author smalltext">{$thread['profilelink']} <span class="smalltext" title="{$thread_started_time_date_title}">{$thread_started_time_date}</span></div>
 		</div>
 	</td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}"><a href="{$mybb->settings['bburl']}/misc.php?action=whoposted&tid={$thread['tid']}" onclick="MyBB.whoPosted({$thread['tid']}); return false;">{$thread['replies']}</a>{$unapproved_posts}</td>
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3972,7 +3972,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 	<td class="{$bgcolor}{$thread_type_class}">
 		{$attachment_count}
 		<div>
-          <span>{$prefix} {$gotounread}<a href="{$mybb->settings['bburl']}/forumdisplay.php?fid={$mybb->input['fid']}&prefix={$thread['prefix']}">{$thread['threadprefix']}</a><span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
+          <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
 			<div class="author smalltext">{$thread['profilelink']} <span class="smalltext" title="{$lang->thread_started_time_date} {$threaddate}, {$threadtime}">({$threaddate}, {$threadtime})</span></div>
 		</div>
 	</td>
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1335,6 +1335,8 @@ if(!empty($threadcache) && is_array($threadcache))
 
 		if($fpermissions['canviewdeletionnotice'] != 0 && $thread['visible'] == -1 && !is_moderator($fid, "canviewdeleted"))
 		{
+			$threaddate = my_date($mybb->settings['dateformat'], $thread['dateline']);
+			$threadtime = my_date($mybb->settings['timeformat'], $thread['dateline']);
 			eval("\$threads .= \"".$templates->get("forumdisplay_thread_deleted")."\";");
 		}
 		else
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1337,6 +1337,8 @@ if(!empty($threadcache) && is_array($threadcache))
 		{
 			$threaddate = my_date($mybb->settings['dateformat'], $thread['dateline']);
 			$threadtime = my_date($mybb->settings['timeformat'], $thread['dateline']);
+			$thread_started_time_date_title = $lang->sprintf($lang->thread_started_time_date_title, $thread['date'], $thread['time']);
+			$thread_started_time_date = $lang->sprintf($lang->thread_started_time_date, $thread['date'], $thread['time']);
 			eval("\$threads .= \"".$templates->get("forumdisplay_thread_deleted")."\";");
 		}
 		else
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1335,8 +1335,8 @@ if(!empty($threadcache) && is_array($threadcache))
 
 		if($fpermissions['canviewdeletionnotice'] != 0 && $thread['visible'] == -1 && !is_moderator($fid, "canviewdeleted"))
 		{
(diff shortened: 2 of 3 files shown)


Sama34:
I agree with prefix filtering, I actually gave it for granted tbh.

Can you post before and after captures of this change?

Eldenroot:
Prefixes are clickable... so no change in visual.

About date/time for thread creation - I can share my customized output, give me few days and I will set a testing board again ;)

Eldenroot:
image
 from my board (CSS a little bit different)

Sama34:
Oh, I thought you meant adding start date filtering.

> I would like to open a discussion.

I think, from my small UIX experience, that adding the start date might be unnecessary and add no overall value to the end user as the last post date is enough as a activity indicator.

However, as the MyBB default theme is meant to showcase "possible _ness_" and not necessarily to be used as the live theme for all boards. So we can assume that designers who see it as unnecessary for their concepts can simply strip it out.

Therefore, if nobody disagrees I have no issue with it being implemented.

But, I think it should be properly implement, so I would like to ask you to update the code for it to use the relative time feature of the script.

Finally but important to highlight, thank you for your constant contribution to the project Eldenroot.

yuliu:
> Oh, I thought you meant adding start date filtering.

Oh, I was thinking in the same way and finally get understood..

My thoughts here are:

- Thread prefix is not necessarily clickable here. Theme authors could always make it clickable in their template since there's no extra variables needed to be added in PHP scripts. Another reason is similar to what @Sama34 has said.

- The thread time is useful which in MyBB is missing IMO in comparison with other forum software.

What kind of time forms should a time be displayed is something really bothering me. I really like the way how GitHub shows time, for example the HTML page outputs (maybe) ISO formatted time like `2020-03-01 12:56` and then the actual text is rendered through Javascript to be a relative time like `2 Hours ago`. In this way, the forum could be controlled to display each time field in different forms on the fly, which I'd like to see in v1.9 😆

Eldenroot:
I am not sure of relative time is useful in this case... anyway it uses your time/date settng from ACP

Eldenroot:
Should I split this PR into two PRs? One for thread filtering and the second one for displaying time and date?

euantorano:
I think it’s fine as is.

Sama34:
Let me test this today.

install/resources/mybb_theme.xml
```diff
@@ -3966,14 +3966,14 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
...
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%"><span class="thread_status {$folder}" title="{$folder_label}">&nbsp;</span></td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%">{$icon}</td>
 	<td class="{$bgcolor}{$thread_type_class}">
 		{$attachment_count}
 		<div>
-          <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']}</div>
+          <span>{$prefix} {$gotounread}<a href="{$mybb->settings['bburl']}/forumdisplay.php?fid={$mybb->input['fid']}&prefix={$thread['prefix']}">{$thread['threadprefix']}</a><span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
+			<div class="author smalltext">{$thread['profilelink']} <span class="smalltext" title="{$lang->thread_started_time_date} {$threaddate}, {$threadtime}">({$threaddate}, {$threadtime})</span></div>
```
Sama34, line 3976: This link should be built using core functions, i.e. Google SEO won't be able to replace this hard coded link.

Sama34: Actually, I think this should probably be a new template, becase the link code will be present even if the thread has no prefix.

Sama34: Actually, I think this should probably be a new template, becase the link code will be present even if the thread has no prefix.

Eldenroot: Ok, but that would be quite tricky... we do not even have a separate template for a thread prefix in forumdisplay, right?

lairdshaw: > This link should be built using core functions, i.e. Google SEO won't be able to replace this hard coded link.

But do core functions even currently support getting a forum link with a query parameter like "prefix"? It doesn't look like it to me. It looks as though that support would have to be added, which probably is beyond the scope of this PR?

yuliu: > But do core functions even currently support getting a forum link with a query parameter like "prefix"? It doesn't look like it to me. It looks as though that support would have to be added, which probably is beyond the scope of this PR?

It has a parameter for filtering prefixes in this way `/forumdisplay.php?fid={fid}&prefix={prefix_id}`.

yuliu:
I've got some concern about having a link for `$thread['threadprefix']`. Because it's a style of a real thread prefix which can be customized to have a link, though rarely to be seen. doing so in this PR would result unexpected results.

install/resources/mybb_theme.xml
lairdshaw: > It has a parameter for filtering prefixes in this way `/forumdisplay.php?fid={fid}&prefix={prefix_id}`.

Perhaps we're miscommunicating. The core function to which I'm referring is `get_forum_link()` in inc/functions.php. The two parameters to that function are `$fid` and `$page`. I see no scope for a prefix in those parameters.

yuliu: > > It has a parameter for filtering prefixes in this way `/forumdisplay.php?fid={fid}&prefix={prefix_id}`.
> 
> Perhaps we're miscommunicating. The core function to which I'm referring is `get_forum_link()` in inc/functions.php. The two parameters to that function are `$fid` and `$page`. I see no scope for a prefix in those parameters.

Oh I see. But `&prefix=` can be always appended to a URL formed by `get_forum_link()` or what ever if it's a link of current viewing forum.

yuliu: > > Perhaps we're miscommunicating. The core function to which I'm referring is `get_forum_link()` in inc/functions.php. The two parameters to that function are `$fid` and `$page`. I see no scope for a prefix in those parameters.
> 
> Oh I see. But `&prefix=` can be always appended to a URL formed by `get_forum_link()` or what ever if it's a link of current viewing forum.

Sorry, inaccurate comment was made.

The forum URL has to be formed as a variable maybe `$forum_url` just like how `$page_url` was formed, except that there's no other parameter, so that in the template we could use `{$forum_url}?prefix={$prefixid}` as the link href. 

Note `{$prefixid}` should be changed to the var accordingly. Maybe `bburl` should be added as well?

lairdshaw: > in the template we could use `{$forum_url}?prefix={$prefixid}`

Yes. But only sometimes. Other times we would need that `?` to be an `&amp;`, especially when basic URLs rather than core/SEO-plugin SEO-friendly URLs were operative.

lairdshaw: So, I guess that's really a "No, it's more complicated than that".

lairdshaw:
> I've got some concern about having a link for `$thread['threadprefix']`. Because it's a style of a real thread prefix which can be customized to have a link, though rarely to be seen. doing so in this PR would result unexpected results.

Right. I just want to say first-up that I appreciate how well you pick up problems that are not always obvious.

Re a solution to this particular problem that you've picked up, perhaps we could add in forumdisplay.php a key into the $thread array whose value uses the plain-text prefix - something like `$thread['plainthreadprefix']` - and that we could use in place of the styled `$thread['threadprefix']` in this template?

lairdshaw:
> Re a solution to this particular problem that you've picked up, perhaps we could add in forumdisplay.php a key into the $thread array whose value uses the plain-text prefix - something like `$thread['plainthreadprefix']` - and that we could use in place of the styled `$thread['threadprefix']` in this template?

Meh. Nevermind. That's obviously not a useful solution, since forumdisplay.php is one of the main places where the styled version of the prefix is _intended_ to be used.

lairdshaw:
A more workable potential alternative solution: we add, for threads with a prefix, a small "filter" icon just prior to (or after) the (styled) prefix, which links to the URL in question.

JordanMussi:
There is already a prefix filter option at the bottom of the forum table with the other filtering options so I think just drop adding a link because the functionality is already there.

yuliu:
Thread prefix style, although is different from user group name style often used on a user name with a profile link attached, should be too rare to be seen to have a link href in it. 
---
Following comment may be too much off-topic and biased.

> > I've got some concern about having a link for `$thread['threadprefix']`. Because it's a style of a real thread prefix which can be customized to have a link, though rarely to be seen. doing so in this PR would result unexpected results.
> 
> Right. I just want to say first-up that I appreciate how well you pick up problems that are not always obvious.

I've been nitpicking for too much time if you say so 🤣 . It's mainly for the this stable 1.8 branch that has a life span more than half a decade. During the time when 1.8 branch is being supported, developers and forum owners may have done massive work of extending MyBB for what is missing for being an up-to-date forum software for years.

For the .23 and later patch version which will probably be released after its 6th year, we should take extra care of most of features/fixes that may break something maybe already established. This is an example that would probably break many things: https://github.com/mybb/mybb/pull/3891#discussion_r411256697 , in a patch release but not a major one. That will, probably, stop forums from being upgraded.

Another reason of my reluctance of _fixing the old time (out-dated) issues_ is that I think these issues really should get addressed in a future major release. For the exiting 1.8 branch, yes, we can always have _enhancement and new features_ for it as long as it's supported, but this will never end if we don't devote into 1.9's development.

However, for a future release we shouldn't stop getting further. If some one tries to upgrade an existing MyBB 1.8 installation with very a few customizations to 1.9, they'll notice a certain amount of breaking things. By breaking things, new features are provided.

I will cut off-topic talking here.

Sama34:
I also agree there is no need to add a link since users can already filter by prefixes.

@mybb/developers

Eldenroot:
Actually I made this PR primary because of easier filtering option... now you have to scroll to the filtering threads section... with this solution it is much friendlier :)

install/resources/mybb_theme.xml
euantorano: I think that the best solution would be introduce a new function to build links with a set of parameters. Any other manual way of trying to do it is sure to cause problems.

euantorano:
I think adding. link to the prefix is going to cause a lot more hassle than its worth, between hard-coded URL structures and the (very small) potential of breaking custom prefix HTML.

Eldenroot:
@euantorano - OK, so I will create a smaller new PR with just thread creation time and date support? I will use same approach, is it OK?

euantorano:
Yes, that would be perfect, or you can just update this PR to remove the links.

Eldenroot:
I have removed the thread filtering :)

Eldenroot:
Ok, URLs should be reworked in 1.9/1.10. Thats why I have removed the filtering. So it is not related to this PR anymore.

install/resources/mybb_theme.xml
```diff
@@ -3966,14 +3966,14 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
...
-		<template name="forumdisplay_thread" version="1813"><![CDATA[<tr class="inline_row">
+		<template name="forumdisplay_thread" version="1823"><![CDATA[<tr class="inline_row">
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%"><span class="thread_status {$folder}" title="{$folder_label}">&nbsp;</span></td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%">{$icon}</td>
 	<td class="{$bgcolor}{$thread_type_class}">
 		{$attachment_count}
 		<div>
           <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']}</div>
+			<div class="author smalltext">{$thread['profilelink']} <span class="smalltext" title="{$lang->thread_started_time_date} {$threaddate}, {$threadtime}">({$threaddate}, {$threadtime})</span></div>
```
euantorano, line 3976: I wonder if it would be best to have language a variable for the string, and to use `$thread['date']` and `$thread['time']` for the variables, then do `$thread_date_and_time_title = $lang->sprintf($lang->'thread_started_time_date, $thread['date'], $thread['time'])` and `$thread_date_and_time = $lang->sprintf($lang->'thread_started_time_date_value, $thread['date'], $thread['time'])`?

This way, the `,` isn't hard-coded as I don't know if that punctuation makes sense for all languages?

Eldenroot: @euantorano - please check now

dvz:
It could simply use the `relative` format with `my_date()` (that takes advantage of [`datetimesep`](https://github.com/mybb/mybb/blob/mybb_1822/install/resources/settings.xml#L472)).

Sama34:
I suggested the same before, I think this should be using relative date.

Eldenroot:
OK, in my opinion bets to use non-relative time output for this purpose but if you really wanna relative lets change it :) 

Moreover this will need small adjustments, I will fix all in one commit.

JordanMussi:
@Eldenroot To me it looks like you can take advantage of `my_date` using `$format='normal'` and `$ty=2` which gives a non-relative date & time using the separator:

https://github.com/mybb/mybb/blob/b81c692c1a7095f131a607b499a8a7f13f256435/inc/functions.php#L495-L516

lairdshaw:
@Eldenroot asked me to take care of this PR for him by replacing it with one of my own, which takes account of the suggestions in the above conversation, in particular, the suggestion that the date-time to be relative. I've honoured @Eldenroot's request in #3977.

PR #3892, opened 2020-02-09
Fixes #3890 - Forum order - make it wider

### LINKED ISSUE #3890: forum order needs to be wider in admin cp
the forum order is too narrow in the admin cp.  if you have forum orders with 3 numbers, it cuts them off.  

clean install of mybb 1.8.22:

Original thread: [forum order needs to be wider in admin cp](https://community.mybb.com/thread-226390.html)

Eldenroot:
This should be fine, but it depends on your resolution...

Resolves #3890 

image

### DIFF
--- a/admin/modules/forum/management.php
+++ b/admin/modules/forum/management.php
@@ -2286,7 +2286,7 @@ if(!$mybb->input['action'])
 		$form_container = new FormContainer($lang->manage_forums);
 	}
 	$form_container->output_row_header($lang->forum);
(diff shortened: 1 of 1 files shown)


Sama34:
Why not a little wider? I now 3 digits seems fair, but we shouldn't be updating this later on.

Eldenroot:
I think 3 digits are more than enough. 

Anyway in 1.9/2.0 there should be a total ACP rewrite

Sama34:
You might be right.

I have no issues leaving it with 3 digits.

Eldenroot:
Please review :)

PR #3893, opened 2020-02-16
managegroup do_manageusers: Fix check for whether any users are selected

dsyalex:
no_users_selected was only being shown if `removeuser` was not an array.
If it was an empty array the error would not be shown and MyBB would continue as if nothing were wrong.

### DIFF
--- a/managegroup.php
+++ b/managegroup.php
@@ -241,7 +241,7 @@ elseif($mybb->input['action'] == "do_manageusers" && $mybb->request_method == "p
 
 	$plugins->run_hooks("managegroup_do_manageusers_start");
 
-	if(is_array($users) && count($users))
+	if(!empty($users))
 	{
 		foreach($users as $uid)
 		{
--- a/managegroup.php
+++ b/managegroup.php
@@ -239,7 +239,6 @@ elseif($mybb->input['action'] == "do_manageusers" && $mybb->request_method == "p
 
 	$plugins->run_hooks("managegroup_do_manageusers_start");
 
-	$users = $mybb->get_input('removeuser', MyBB::INPUT_ARRAY);
 	if(is_array($users) && count($users))
 	{
 		foreach($users as $uid)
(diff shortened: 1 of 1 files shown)


managegroup.php
```diff
@@ -239,9 +239,10 @@
 
 	$plugins->run_hooks("managegroup_do_manageusers_start");
 
-	if(is_array($mybb->get_input('removeuser', MyBB::INPUT_ARRAY)))
+	$users = $mybb->get_input('removeuser', MyBB::INPUT_ARRAY);
+	if(is_array($users) && count($users))
```
Sama34: The line above warranties it will be always an array, doesn't it? It might be better and simpler for it to be `!empty($users)`.

managegroup.php
```diff
@@ -239,9 +239,10 @@
 
 	$plugins->run_hooks("managegroup_do_manageusers_start");
 
-	if(is_array($mybb->get_input('removeuser', MyBB::INPUT_ARRAY)))
+	$users = $mybb->get_input('removeuser', MyBB::INPUT_ARRAY);
```
Sama34: This should probably be added before running the `managegroup_do_manageusers_start` hook.

Eldenroot:
@dsyalex - hi, looks fine. Just please fix the issues mentioned above. Thank you for your contribution!

Ben-MyBB:
@dsyalex Hi, is there any chance you can update this PR with the changes requested?

Eldenroot:
@Ben-MyBB - can you recreate it?

managegroup.php
Sama34:
```suggestion
```

managegroup.php
```diff
@@ -239,9 +239,10 @@
 
 	$plugins->run_hooks("managegroup_do_manageusers_start");
```
Sama34, line 240:
```suggestion
	$users = $mybb->get_input('removeuser', MyBB::INPUT_ARRAY);

	$plugins->run_hooks("managegroup_do_manageusers_start");
```

managegroup.php
Sama34:
```suggestion
	if(!empty($users))
```

JordanMussi:
Does this need an issue ticket?

euantorano:
Yes, it probably should.

euantorano:
Fixed #4069

PR #3900, opened 2020-03-02
Fix #3852 incorrect counts for guests and x viewing forum for bots on Index page

### LINKED ISSUE #3852: Incorrect guests number counting
### Total number of online users
Seems happening to 1.8.22. Mentioned in [Discord chat](https://discordapp.com/channels/215876847634743296/215876989733568512/661567597287964701).

The guest count here is counting spiders/bots numbers and guest numbers, with the WHERE condition `uid = 0 AND time > $timesearch`:
https://github.com/mybb/mybb/blob/mybb_1822/index.php#L64-L89

And the total count here is the summation of member count & guest count & bot count:
https://github.com/mybb/mybb/blob/mybb_1822/index.php#L187

### Number of users viewing a specific forum
Another issue related here is that if `showforumviewing` is on, the number of users viewing a specific forum is doubled for spiders/bots:
https://github.com/mybb/mybb/blob/mybb_1822/index.php#L161

## The following issue will be tracked in a separated one.

~The number of guests that are printed on online page is not equal to the count of guests shown on the Index page's statistics~

~E.g., the MyBB community forum's online page shows `Pages (11): 1 2 3 4 5 … 11 Next » Jump to page  `. The total number of pages is 11 and per page count is 20, thus the online count of all members + bots + guests should be less than or equal to 11*20=220.~

~But at this time, the index page says `401 users active in the past 15 minutes (7 members, 1 of whom is invisible, and 391 guests).`.*~

yuliu:
Resolves #3852.

BTW, there are two questions I still doubt:
1. It's quite rare that a pure guest and a recognized bot/spider have a same IP, though possible. Current code won't distinguish such visitings from each other, when counting the number of total guests and guests of x viewing forum.

2. Looks like in the Index page and the Forumdisplay page, there's a variable called `$doneusers` that may be for legacy code which I'm not sure what's it really doing for now.

### DIFF
--- a/index.php
+++ b/index.php
@@ -78,7 +78,7 @@ if($mybb->settings['showwol'] != 0 && $mybb->usergroup['canviewonline'] != 0)
 
 			if($location['location1'])
 			{
-				$forum_viewers[$location['location1']] += $location['guestcount'];
+				$forum_viewers[$location['location1']] = isset($forum_viewers[$location['location1']]) ? $forum_viewers[$location['location1']] + $location['guestcount'] : $location['guestcount'];
 			}
 		}
 	}
@@ -101,7 +101,7 @@ if($mybb->settings['showwol'] != 0 && $mybb->usergroup['canviewonline'] != 0)
 	// Fetch spiders
 	$spiders = $cache->read('spiders');
 
-	// Loop through all users.
+	// Loop through all users and spiders/bots.
 	while($user = $db->fetch_array($query))
 	{
 		// Create a key to test if this user is a search bot.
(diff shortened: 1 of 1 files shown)


Eldenroot:
@yuliu - is it suppose to fix huge numbers of visitors in the stats page which I can see since last update of MyBB?

yuliu:
> @yuliu - is it suppose to fix huge numbers of visitors in the stats page which I can see since last update of MyBB? 

@Eldenroot no it's not. It's supposed to fix incorrect numbers of all users/guests and x viewing in the index page.

@dvz had his thoughts expressed in https://github.com/mybb/mybb/issues/3852#issuecomment-593427601which I haven't yet got the code done and thus this PR is not ready ATM.

yuliu:
Replaced by #3969.

PR #3904, opened 2020-03-21
Resolve #3422 Loose comparison used for forum passwords

### LINKED ISSUE #3422: Loose comparison used for forum passwords
Forum password checks use weak type comparison (e.g. `== ''`, `empty()​`), which may result in the check being disabled for values like `0`.

dvz:
Resolves #3422

- added function to check forum password validation status:
```php
function forum_password_validated($forum, $ignore_empty=false, $check_parents=false)
```
- weak comparisons (`==`, `empty()`) changed to identical (`===`) for no password check
- password verification comparison changed to use `my_hash_equals()`
- local optimizations

### DIFF
--- a/archive/index.php
+++ b/archive/index.php
@@ -26,7 +26,7 @@ switch($action)
 		if($announcement['fid'] != -1)
 		{
 			$forum = get_forum($announcement['fid']);
-			if(!$forum['fid'] || $forum['password'] != '')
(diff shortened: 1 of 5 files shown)


euantorano:
This looks good upon inspection, I just was to test it before merging. Will hopefully get round to that today or tomorrow.

Eldenroot:
Looks fine for me (applied to my board).

PR #3911, opened 2020-03-26
Develop/1.9 Add send attachments in private messages

R4z3l:
Resolves #ISSUE_ID

Send attachments in private messages


euantorano:
Did you open this intentionally? It seems to be a PR to merge develop/1.9 into feature, which we aren't quite ready to do just yet...

PR #3918, opened 2020-03-31
Fix #3894 managegroup.php can only be used to remove users for whom i…

### LINKED ISSUE #3894: managegroup.php can only be used to remove users for whom it is not the primary group
It will list users for whom the group is primary, but it calls the leave_usergroup function which only removes additional groups.
When this happens it pretends everything is fine but the user does not actually get removed. It should either:
* error (or block in the UI) if this operation is not possible for the particular user
* allow removing the user from their primary group (presumably selecting a new primary group from one of their other additional groups and erroring out if they have none?)

The latter sounds preferable but I don't know the full implications.

Sama34:
…t is not the primary group

Resolves #3894

image

When all checkboxes are disabled the button won't be displayed.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -5028,6 +5028,11 @@ function leave_usergroup($uid, $leavegroup)
 
 	$user = get_user($uid);
 
+	if($user['usergroup'] == $leavegroup)
+	{
+		return false;
+	}
+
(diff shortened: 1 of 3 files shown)


Eldenroot:
Seems to be fine, I applied to my board.

PR #3923, opened 2020-04-05
Fix a mismatched HTML tag in the footer template.

euantorano:
There was a mismatched closing HTML tag in the footer template - the opening tag was `<footer>` but the closing tag was `</div>`. Spotted by @yuliu and reported on Discord.


yuliu:
Tracked in #3785 >> Templates and templates to be worked with scripts & style >> Minor issues.

PR #3887, opened 2020-02-01
Fixes #3886 - $_FILES['attachments'] check

Eldenroot:
Resolves #3886

PR #3926, opened 2020-04-08
Fixes several bugs with attachments at once

### LINKED ISSUE #3675: Editing posts with JS disabled to manage attachments
There are some issues when editing posts with JS disabled and trying to manage attachments. First, the code refers to an input called `submit`, which never appears on the page as far as I can tell (it's called `submitbutton` instead):

```php
	if(!isset($mybb->input['submit']))
	{
		$mybb->input['action'] = "editpost";
	}
```

This means that if you disable JS and try to remove an attachment, you simply get redirected to the thread without the attachment being removed.

### LINKED ISSUE #3886: Thread Modes $_FILES['attachment'] instead of $_FILES['attachments']
https://community.mybb.com/thread-226313-post-1342675.html#pid1342675

lairdshaw:
Resolves #3886
Resolves #3675

1. When the "Post Thread" or "Post Reply" button was clicked after an
   attachment had been selected with the "Choose Files" button, the
   attachment was not saved.

2. When "Add Attachment" was clicked after a previous attachment had
   been removed via AJAX and a new attachment had been selected with
   the "Choose Files" button, the new attachment was not saved. I
   reported this issue here:

   https://community.mybb.com/thread-227178.html

3. When Javascript was disabled and "Remove", "Unapprove", or "Approve"
   was clicked, the corresponding action was not performed. This bug
   was reported in issue #3675.

These problems were caused by:

1. Various typos. Correcting the typos resolves the problem:

1.1. A typo - already noted in issue #3886 - in the conditionals in
     editpost.php, newreply.php, and newthread.php (singular should
     be plural):

     `$_FILES['attachment']` should be `$_FILES['attachments']`

1.2. A typo in two conditionals in editpost.php (as noted in issue
     #3675):

     `$mybb->input['submit']` should be `$mybb->input['submitbutton']`

2. The setting in jscripts/post.js of the `attachmentaid` input when
   removing an attachment, which, when the new attachment was
   submitted, caused those same conditionals as in 1.1 to fail due to
   the test:

   `!$mybb->get_input('attachmentaid', MyBB::INPUT_INT)`

   Removing that test resolves the problem without any side-effects.
   An alternative was to unset `attachmentaid` in post.js after
   the removal operation completed, however, it seems better to
   delete code than to add it.

3. The lack of a value for the `attachmentaid` and `attachmentact`
   inputs. This is fixed in this commit by:

   3.1. Adjusting the `name` property of the remove, approve, and
        unapprove submit buttons such that the attachment ID is
	appended after an underscore.

   3.2. Adding a function `detect_attachmentact()` which detects
        when an input with such a name is present, extracts the
	attachment ID from it, and then sets the missing values
	for `attachmentaid` and `attachmentact`.


inc/functions_post.php
```diff
@@ -1067,3 +1067,28 @@ function return_bytes($val) {
 
 	return intval($val);
 }
+
+/**
+ * Detects whether an attachment removal/approval/unapproval
+ * submit button was pressed (without triggering an AJAX request)
+ * and sets inputs accordingly (as for an AJAX request).
+ */
+function detect_attachmentact() {
```
euantorano, line 1076: Nitpick: `{` should be on a new line on its own.

```suggestion
function detect_attachmentact()
{
```

inc/functions_post.php
```diff
@@ -1067,3 +1067,28 @@ function return_bytes($val) {
...
+	global $mybb;
+
+	foreach ($mybb->input as $key => $val) {
```
euantorano, line 1079: Nitpick: this should follow our code style:

```suggestion
	foreach($mybb->input as $key => $val)
	{
```

inc/functions_post.php
```diff
@@ -1067,3 +1067,28 @@ function return_bytes($val) {
...
+		if (strpos($key, 'rem_') === 0) {
```
euantorano, line 1080:
```suggestion
		if(strpos($key, 'rem_') === 0)
		{
```

inc/functions_post.php
```diff
@@ -1067,3 +1067,28 @@ function return_bytes($val) {
...
+			$mybb->input['attachmentaid'] = (int)substr($key, 4);
+			$mybb->input['attachmentact'] = 'remove';
+			break;
+		} else if (strpos($key, 'approveattach_') === 0) {
```
euantorano, line 1084:
```suggestion
		}
		elseif(strpos($key, 'approveattach_') === 0)
		{
```

inc/functions_post.php
```diff
@@ -1067,3 +1067,28 @@ function return_bytes($val) {
...
+			$mybb->input['attachmentaid'] = (int)substr($key, 14);
+			$mybb->input['attachmentact'] = 'approve';
+			break;
+		} else if (strpos($key, 'unapproveattach_') === 0) {
```
euantorano, line 1088:
```suggestion
		}
		elseif(strpos($key, 'unapproveattach_') === 0)
		{
```

install/resources/mybb_theme.xml
```diff
@@ -9420,10 +9420,10 @@ if(use_xmlhttprequest == "1")
 <td class="trow2" width="60%" style="white-space: nowrap">{$attachment['filename']} ({$attachment['size']})</td>
 <td class="trow2" style="white-space: nowrap; text-align: center;">{$attach_mod_options} {$attach_rem_options} {$postinsert}</td>
 </tr>]]></template>
-		<template name="post_attachments_attachment_mod_approve" version="120"><![CDATA[<input type="submit" class="button" name="approveattach" value="{$lang->approve_attachment}" onclick="return Post.attachmentAction({$attachment['aid']},'approve');" />]]></template>
-		<template name="post_attachments_attachment_mod_unapprove" version="120"><![CDATA[<input type="submit" class="button" name="unapproveattach" value="{$lang->unapprove_attachment}" onclick="return Post.attachmentAction({$attachment['aid']},'unapprove');" />]]></template>
+		<template name="post_attachments_attachment_mod_approve" version="120"><![CDATA[<input type="submit" class="button" name="approveattach_{$attachment['aid']}" value="{$lang->approve_attachment}" onclick="return Post.attachmentAction({$attachment['aid']},'approve');" />]]></template>
```
euantorano, line 9423: Template version needs changing:

```suggestion
		<template name="post_attachments_attachment_mod_approve" version="1823"><![CDATA[<input type="submit" class="button" name="approveattach_{$attachment['aid']}" value="{$lang->approve_attachment}" onclick="return Post.attachmentAction({$attachment['aid']},'approve');" />]]></template>
```

install/resources/mybb_theme.xml
```diff
@@ -9420,10 +9420,10 @@ if(use_xmlhttprequest == "1")
...
+		<template name="post_attachments_attachment_mod_unapprove" version="120"><![CDATA[<input type="submit" class="button" name="unapproveattach_{$attachment['aid']}" value="{$lang->unapprove_attachment}" onclick="return Post.attachmentAction({$attachment['aid']},'unapprove');" />]]></template>
```
euantorano, line 9424: Template version needs changing:

```suggestion
		<template name="post_attachments_attachment_mod_unapprove" version="1823"><![CDATA[<input type="submit" class="button" name="unapproveattach_{$attachment['aid']}" value="{$lang->unapprove_attachment}" onclick="return Post.attachmentAction({$attachment['aid']},'unapprove');" />]]></template>
```

install/resources/mybb_theme.xml
```diff
@@ -9420,10 +9420,10 @@ if(use_xmlhttprequest == "1")
...
 		<template name="post_attachments_attachment_postinsert" version="1809"><![CDATA[<input type="button" class="button" name="insert" value="{$lang->insert_attachment_post}" onclick="$('#message').sceditor('instance').insertText('[attachment={$attachment['aid']}]'); return false;" />]]></template>
-		<template name="post_attachments_attachment_remove" version="1604"><![CDATA[<input type="submit" class="button" name="rem" value="{$lang->remove_attachment}" onclick="return Post.removeAttachment({$attachment['aid']});" />]]></template>
+		<template name="post_attachments_attachment_remove" version="1604"><![CDATA[<input type="submit" class="button" name="rem_{$attachment['aid']}" value="{$lang->remove_attachment}" onclick="return Post.removeAttachment({$attachment['aid']});" />]]></template>
```
euantorano, line 9426: Template version needs changing:

```suggestion
		<template name="post_attachments_attachment_remove" version="1823"><![CDATA[<input type="submit" class="button" name="rem_{$attachment['aid']}" value="{$lang->remove_attachment}" onclick="return Post.removeAttachment({$attachment['aid']});" />]]></template>
```

lairdshaw:
Thanks for fixing up my oversights!

euantorano:
That was quick 😄 I'll try and test this over this Easter weekend.

yuliu:
I had a comment in https://github.com/mybb/mybb/pull/3887#issuecomment-609875287 that is still relevant to this issue. What I mean is that maybe the `if`-controls need overhauls.

`detect_attachmentact()` seems a good way for working with AJAX. But template changes are pretty rushing, IMHO, which would result in incompatibility with a number of existing 3rd party themes. Will there be any other means to achieve the same goal without changing the `name=` attribute?

lairdshaw:
> I had a comment in [#3887 (comment)](https://github.com/mybb/mybb/pull/3887#issuecomment-609875287) that is still relevant to this issue. What I mean is that maybe the `if`-controls need overhauls.

I've just responded there: the conditionals seem fine to me as updated by this PR.

> `detect_attachmentact()` seems a good way for working with AJAX. But template changes are pretty rushing, IMHO, which would result in incompatibility with a number of existing 3rd party themes.

I don't think that this is too big of an issue: worst case (and expected) scenario seems to me to be that 3rd party themes _continue_ to experience this bug as they _already_ were, until their templates are updated. Yes, it would be nice if we could come up with a fix that applied no matter which theme was in use, but...

> Will there be any other means to achieve the same goal without changing the `name=` attribute?

...I can't think of one.

We need for a single button to communicate two pieces of information: (1) the action to perform (remove/approve/unapprove) and (2) the ID of the attachment on which to perform that action, _without_ the use of Javascript.

We can't use hidden inputs because - in the absence of Javascript - they don't change depending on which button is pressed, and we need the conveyed information to be conditional on the button pressed.

So... I don't know, hopefully I'm missing something and somebody _can_ come up with an alternative fix that doesn't require changes to templates - otherwise, I don't think they're avoidable.

yuliu:
OK, I'll reply to the comment https://github.com/mybb/mybb/pull/3887#issuecomment-611593117 here since it's response to this PR.

> > But now this "update post" action only confusingly results in uploading an attachment
> 
> It works correctly for me. Clicking "Update Post" uploads the file(s) and returns to the post in thread view mode. Are you sure you have applied the full PR of #3926?

Ya, it's working as intended now using this PR.


> > default theme doesn't allow multiple attachments uploading at once
> 
> Uploading multiple attachments at once works fine for me (using the default theme on a vanilla MyBB installation with PR #3926 applied). I do it by clicking "Choose Files" and then selecting multiple files in the dialogue by holding down ctrl when clicking them.

Well, it does work with _Ctrl_ pressed or by selecting multiple files in the opened file box, though it may not be the intuitive way for uploading multiple attachments :-/

> I don't think that this is too big of an issue: worst case (and expected) scenario seems to me to be that 3rd party themes _continue_ to experience this bug as they _already_ were, until their templates are updated. Yes, it would be nice if we could come up with a fix that applied no matter which theme was in use, but...

I've looked at the code and tried again. Looks like the only thing would get broken is the "Remove" button for uploaded attachment.

..

> > Will there be any other means to achieve the same goal without changing the `name=` attribute?
> 
> ...I can't think of one.
> 
> We need for a single button to communicate two pieces of information: (1) the action to perform (remove/approve/unapprove) and (2) the ID of the attachment on which to perform that action, _without_ the use of Javascript.

.. Since then, I have an idea but don't know whether they'll be OK, I'm not so good at working with POST form data and AJAX really 🤣 .

Here at https://github.com/mybb/mybb/pull/3926/files#diff-821efc69bc1cf4f87862693332f222deR108 , just using `$('input[name^=\'rem\']')` without the trailing underscore should provide compatibility for existing themes. Yep I've tried.


Oh, by the way, I just find another issue you may have missed. The "Remove" button doesn't work well with unapproved attachments since the attachment line doesn't have an `id=` for the table row to be erased. Hope you could fix it as well in this PR. Just add ` id="attachment_{$attachment['aid']}"` in `<tr>` on [mybb_theme.xml#L9426](https://github.com/mybb/mybb/blob/mybb_1822/install/resources/mybb_theme.xml#L9426)


Former thoughts that are invalid now:
~~- Just using input names like `name=rem{$attachment['aid']}`, `unapproveattach{$attachment['aid']}` and `approveattach{$attachment['aid']}` would provide compatibility for existing themes. Yep I've tried. Or if this kind of method~~

~~- Is it possible to use array format for POST variables? Like using `name=rem[{$attachment['aid']}]`, `unapproveattach[{$attachment['aid']}]` and `approveattach[{$attachment['aid']}]` will let them still be can be captured by using original `$mybb->input['rem']`/$mybb->unapproveattach['rem']`/$mybb->attachment['rem']`/`$('button[name=rem]')` while the script may also get the variable's index.~~

yuliu:
I also have some concerns about the "redo_mod_queue_header" idea. Actually, it's pretty ideal of sending a correct page indicating the umber of unapproved attachments, doing this "redo_mod_queue_header" is very good.

However, it has to modify the `$header` variable which will invalid any modifications to it by plugins that hook to `global_end`. Modifying a former processed variable whose scope is not in the present context seems plugin-unfriendly.

I do have an idea of re-post all the form's input fields to respective pages after attachments being approved/unapproved/removed, but I think it's so over-kill..

lairdshaw:
> Well, it does work with _Ctrl_ pressed or by selecting multiple files in the opened file box, though it may not be the intuitive way for uploading multiple attachments :-/

Yes, I guess it's not all that intuitive. I wonder what a better approach would be. Even if we could find a non-intrusive but "obvious" way of letting the user know that that's how to select multiple files, we'd be in a better position.

> Here at https://github.com/mybb/mybb/pull/3926/files#diff-821efc69bc1cf4f87862693332f222deR108 , just using `$('input[name^=\'rem\']')` without the trailing underscore should provide compatibility for existing themes. Yep I've tried.

Good one. I guess I was wrong about what the worst case scenario was! Anyhow, the commit I've just added now provides this fallback. Tested here too.

> Oh, by the way, I just find another issue you may have missed. The "Remove" button doesn't work well with unapproved attachments since the attachment line doesn't have an `id=` for the table row to be erased. Hope you could fix it as well in this PR. Just add ` id="attachment_{$attachment['aid']}"` in `<tr>` on [mybb_theme.xml#L9426](https://github.com/mybb/mybb/blob/mybb_1822/install/resources/mybb_theme.xml#L9426)

Nice catch. Yes, I had missed that. Also incorporated into the commit I've just added.

> Former thoughts that are invalid now

An interesting idea though. It was worth exploring.

> However, it has to modify the `$header` variable which will invalid any modifications to it by plugins that hook to `global_end`. Modifying a former processed variable whose scope is not in the present context seems plugin-unfriendly.

That's a valid concern.

> I do have an idea of re-post all the form's input fields to respective pages after attachments being approved/unapproved/removed, but I think it's so over-kill..

In the long term, maybe we could provide a function dedicated especially to generating the header, and then plugins could hook in to that function rather than into `global_end`. Then, the header could be regenerated at any time, and each plugin that needed to would be hooked in on the regeneration too.

In the short term... I don't know. If I come up with anything, I'll share it.

lairdshaw:
> Anyhow, the commit I've just added now provides this fallback. Tested here too.

Oops, I misread your suggestion, which was much simpler than the approach I took. Fixed in the latest commit.

yuliu:
> > Anyhow, the commit I've just added now provides this fallback. Tested here too.
> 
> Oops, I misread your suggestion, which was much simpler than the approach I took. Fixed in the latest commit.

Yeah that's what I mean.

> In the long term, maybe we could provide a function dedicated especially to generating the header, and then plugins could hook in to that function rather than into `global_end`. Then, the header could be regenerated at any time, and each plugin that needed to would be hooked in on the regeneration too.

I've suggested adding hooks to the template's `get()`/`render()` functions, which I think should be the most plugin-friendly way. Anyway, seems that it won't happen to the 1.8 branch.

> In the short term... I don't know. I agree that the approach you suggest, ~while it would work,~ is overkill. If I come up with anything, I'll share it. [Actually, I don't see how it would work when Javascript is disabled. I think it would take Javascript to automatically resubmit the form with its inputs as-is (unless I'm misunderstanding what you're suggesting).]

The method I was thinking about is to re-post data from server side, i.e., in PHP, not happening in user's browser. But after learning that although it's possible to do so, the actual POST request, using either curl (not ideal) or `header()`, is issued from the sever, I consider it not a good way to achieve the goal.

A second method is to use PHP's Session. Yep this seems the best applicable solution I can come up with at this moment: POST data fields can be all preserved though the `$_SESSION` variable between sessions, and we can redirect to `editpost.php?processed=1` with cleaned `$_POST` stored in `$_SESSION`, and then capture `$_SESSION` at some point for following processing (highly possibly needs modifications to core files in `./inc`).

However, since MyBB doesn't use PHP's built-in Session at all and Session has its own limitation, it's not a good idea, either.

Over-kill, right?

As for using JavaScript for auto-submit, the actions of approve/unapprove don't have a fully functioning AJAX operation that remove has, so such actions still go in PHP processing where JavaScript couldn't intervene. Any other possible method I could imagine also involve using JavaScript.

Sorry for telling the long story and my though to the "redo_mod_queue_header" idea is that _unless the `$hreader`/etc. modification issue could get resolved, it's better not to implement "redo_mod_queue_header" at this moment_.

lairdshaw:
> I've suggested adding hooks to the template's `get()`/`render()` functions, which I think should be the most plugin-friendly way.

That seems like a good idea.

> Over-kill, right?

Yep. But it was worth exploring those ideas.

> Sorry for telling the long story and my though to the "redo_mod_queue_header" idea is that _unless the `$hreader`/etc. modification issue could get resolved, it's better not to implement "redo_mod_queue_header" at this moment_.

OK. I don't have a strong opinion on it either way.

Eldenroot:
Guys, this PR fixes #3886.

This PR should be closed #3887 because it has been replaced by #3926

Sama34:
This is what I reviewed:
> - When the "Post Thread" or "Post Reply" button was clicked after an attachment had been selected with the "Choose Files" button, the attachment was not saved.
> - Clicking "Update Post" uploads the file(s) and returns to the post in thread view mode.
> - Uploading multiple attachments at once works by clicking "Choose Files" and then selecting multiple files in the dialogue by holding down ctrl when clicking them.
> - When "Add Attachment" was clicked after a previous attachment had
been removed via AJAX and a new attachment had been selected with
the "Choose Files" button, the new attachment was not saved. I
reported this issue here:
> - When Javascript was disabled and "Remove", "Unapprove", or "Approve"
was clicked, the corresponding action was not performed. This bug
was reported in issue #3675.

All seemed correct to me.

> Will there be any other means to achieve the same goal without changing the name= attribute?

There are already various template updates for this PR so I don't see a need for a work around. Theme authors or administrators will need to update their themes regardless.

> Well, it does work with Ctrl pressed or by selecting multiple files in the opened file box, though it may not be the intuitive way for uploading multiple attachments :-/

Maybe a better UI can be worked on the future. I suppose PHP & Javascript intuitive file upload libraries exists.

yuliu:
> > Will there be any other means to achieve the same goal without changing the name= attribute?
> 
> There are already various template updates for this PR so I don't see a need for a work around. Theme authors or administrators will need to update their themes regardless.

Looks like you've missed something that me and @lairdshaw have talked about.

---

@lairdshaw, I've got another question about the "redo_mod_queue_header" trick. Here at https://github.com/mybb/mybb/pull/3926/files#diff-7b6854708fdbd11fab65c1bcd9e280c7R239-R242 , there's one more situation that should also trigger this _redo_ thing, removing an attachment that is already unapproved. However, it may not work as intended if JavaScript/AJAX is enabled where the shown page needs self-refresh/self-post.

Since there are obvious obstacles hindering implementing a well-designed "redo_mod_queue_header" procedure, I suggest postpone it by leaving out the two `eval` lines (https://github.com/mybb/mybb/pull/3926/files#diff-7b6854708fdbd11fab65c1bcd9e280c7R329 and https://github.com/mybb/mybb/pull/3926/files#diff-7b6854708fdbd11fab65c1bcd9e280c7R331). This means it actually affects nothing in the front-end but the idea preserves.

lairdshaw:
> @lairdshaw, I've got another question about the "redo_mod_queue_header" trick. Here at https://github.com/mybb/mybb/pull/3926/files#diff-7b6854708fdbd11fab65c1bcd9e280c7R239-R242 , there's one more situation that should also trigger this _redo_ thing, removing an attachment that is already unapproved.

Another good catch. Basically, we need to trigger the redo unconditionally once we enter the main conditional for remove/approve/unapprove. My latest commit does that.

In the process of testing this change, I also picked up another change that needed to be made: the template `post_attachments_attachment_unapproved` had hard-coded HTML for the remove button rather than using the {$attach_rem_options} variable as its counter-part `post_attachments_attachment` does, so our changes to the remove button via its template `post_attachments_attachment_remove` weren't being reflected in the output of unapproved attachments.

I've replaced the hard-coded HTML with that variable.

> However, it may not work as intended if JavaScript/AJAX is enabled where the shown page needs self-refresh/self-post.

Tested and seems to work as intended - unless I'm misunderstanding your concern.

> Since there are obvious obstacles hindering implementing a well-designed "redo_mod_queue_header" procedure, I suggest postpone it by leaving out the two `eval` lines (https://github.com/mybb/mybb/pull/3926/files#diff-7b6854708fdbd11fab65c1bcd9e280c7R329 and https://github.com/mybb/mybb/pull/3926/files#diff-7b6854708fdbd11fab65c1bcd9e280c7R331). This means it actually affects nothing in the front-end but the idea preserves.

Earlier, I wrote that I don't have a strong opinion either way, but I do have an opinion, which I'll express here:

While it is possible that in regenerating the header here we will overwrite changes to it that had been made earlier in a hook such as `global_end` by some plugin(s), we do not know this for certain. We _do_ know for certain, though[1], that the header will be misrepresentative in terms of its moderation queue if we do not regenerate it here. We should act on what we know for sure, rather than what is merely possible. Therefore, we ought to regenerate the header here.

A reasonable argument?

[1] Except in the case that the attachment is approved and being removed, in which case our regeneration of the header is unnecessary.

yuliu:
> In the process of testing this change, I also picked up another change that needed to be made: the template `post_attachments_attachment_unapproved` had hard-coded HTML for the remove button rather than using the {$attach_rem_options} variable as its counter-part `post_attachments_attachment` does, so our changes to the remove button via its template `post_attachments_attachment_remove` weren't being reflected in the output of unapproved attachments.
> 
> I've replaced the hard-coded HTML with that variable.

I think replacing `name="rem"` with `name="rem_{$attachment['aid']}"` will just do fine on this line https://github.com/mybb/mybb/pull/3926/files#diff-05b404fc75b0e5bd6d57c79b667b44b8R9430. Anyway, your way seems good as well.

> > However, it may not work as intended if JavaScript/AJAX is enabled where the shown page needs self-refresh/self-post.
> 
> Tested and seems to work as intended - unless I'm misunderstanding your concern.

Ya, I mean when JavaScript and AJAX enabled, an unapproved attachment getting removed by clicking the "Remove" button will only make that row disappear, but the **unapproved attachments count** needs a page refresh to show the new correct one.

> Earlier, I wrote that I don't have a strong opinion either way, but I do have an opinion, which I'll express here:
> 
> While it is possible that in regenerating the header here we will overwrite changes to it that had been made earlier in a hook such as `global_end` by some plugin(s), we do not know this for certain. We _do_ know for certain, though[1], that the header will be misrepresentative in terms of its moderation queue if we do not regenerate it here. We should act on what we know for sure, rather than what is merely possible. Therefore, we ought to regenerate the header here.
> 
> A reasonable argument?
> 
> [1] Except in the case that the attachment is approved and being removed, in which case our regeneration of the header is unnecessary.

It's ideally and logically reasonable to properly send the latest and correct page to the end users. Either your thoughts on "redo header" or my concerns about `$header` re-generating falls into this point.

**"it is possible that in regenerating the header here we will overwrite changes to it that had been made earlier in a hook such as `global_end` by some plugin(s), we do not know this for certain"**

Developers could do anything with any hook provided by MyBB.

Ya, just take my example here. I have written a plugin accompanied with a theme which uses the `global_end` hook to modify `$header`, `$footer` and `$headerinclude` variables. Without attaching to the `global_end` hook, the program could also survive, but only by attaching to maybe every possible hook for each MyBB page just before the [`output_page()`](https://github.com/mybb/mybb/blob/mybb_1822/inc/functions.php#L16) is called. Further, this workaround will definitely fail on pages coming from other plugins which don't have hooks. Even they do have hooks, my theme or other developers can't know and further implement every hook provided by plugins.

Both of showing an re-generated page which invalidates plugins' work and hooking to numerous possible hooks are obscured, not reasonable and thus bad for plugin development.

**"We _do_ know for certain, though[1], that the header will be misrepresentative in terms of its moderation queue if we do not regenerate it here."**

Your concern here is also valid, since we have the same reason as I pointed out in this comment. I'm going to say something about the pros and cons of the "redo header" procedure in my opinion.

Pros:
- The page should show a correct number of unapproved attachments. _But_ I think it's not a major defect of not showing a correct number in the _edit post_ screen in where the user/admin/mod is focusing on editing a post.

Cons:
- Since currently there's no good way to re-generate the `$header` variable just like how `global.php` does, current implementation of the "redo header" procedure is not plugin friendly.
- The AJAX operation of removing an attachment (possibly approving/unapproving ones will also support AJAX in future) will have to trigger a full page refresh to show the correct number of unapproved attachments in the header, see the former part of this comment. This makes the AJAX action in vain.

BTW, for _[1] Except in the case that the attachment is approved and being removed, in which case our regeneration of the header is unnecessary."_ I could just say that this move is contradicting your original intention of showing correct numbers.

Sama34:
> Looks like you've missed something that me and @lairdshaw have talked about.

If I understand the conclusion was the same, wasn't it? I didn't _missed_ per se, but simply ignored commenting because the conclusion was ultimately to change the name attribute value.

> Ya, I mean when JavaScript and AJAX enabled, an unapproved attachment getting removed by clicking the "Remove" button will only make that row disappear, but the unapproved attachments count needs a page refresh to show the new correct one.

I would say to update its value via JavaScript if the page isn't being refreshed. Header "regenerating" doesn't seem like the best option for this (though building the header in its own function as mentioned could be discussed for future versions).

Also, @lairdshaw it seems you are still adding commits to this and I couldn't find where `{$attach_rem_options}` gets eval'd in your code?

I'm trying to follow this to get it merged as soon as possible, but if you are still discussing additions it might be better to clarify once it gets ready for us to review. I have no comments in your decisions because at least up to my last review and test I didn't disagree with anything.

lairdshaw:
> Ya, I mean when JavaScript and AJAX enabled, an unapproved attachment getting removed by clicking the "Remove" button will only make that row disappear, but the **unapproved attachments count** needs a page refresh to show the new correct one.

OK. I see. It doesn't seem to be a major issue to me - I think we can expect the user to understand that the header simply reflects the state at the time at which the page was loaded.

> I'm going to say something about the pros and cons of the "redo header" procedure in my opinion.

OK. Thanks for sharing your views. I don't have anything more to say on this, so I will leave it to the dev team to decide whether they want to revert or leave as-is that part of this PR's code which regenerates the header.

lairdshaw:
> Also, @lairdshaw it seems you are still adding commits to this and I couldn't find where `{$attach_rem_options}` gets eval'd in your code?

It was already being eval'd in the editpost.php, newthread.php, and newreply.php files for the `post_attachments_attachment` template. When there are unapproved attachments, the `post_attachments_attachment_unapproved` template is used instead, so the variable is already available to it.

Sama34:
Oh, I just saw your last commit for the header issue. As I mentioned early adding a function for the header generation sounds like a good idea for future versions (1.9, 1.10, etc), but not for a minor release.

Having said that, I don't understand why you are rebuilding the whole header, you only need to rebuild notification bars. To not break plugins compatibilities (if any) you can even wrap the numeric (or string) value in the notification with a ID and get only that refreshed.

Your change seems too aggressive for this bug fix only.

lairdshaw:
> Unless something new comes up, it is now ready for review.

Something new came up - I realised that attachments also weren't being saved when the "Preview Post" and "Save as Draft" buttons were clicked. The first new commit above ('Save attachments on "Preview Post" and "Save as Draft".') fixes that.

> Having said that, I don't understand why you are rebuilding the whole header, you only need to rebuild notification bars. To not break plugins compatibilities (if any) you can even wrap the numeric (or string) value in the notification with a ID and get only that refreshed.

Great idea. The second new commit above ('Replace mod queue in $header rather than regen $header') takes this approach. @yuliu, perhaps this satisfies you too?

Sama34:
Once more I apologize if I'm missing something but I'm having a hard time understanding your code.

There is one moderator notice located in the header which you have found requires update after performing some actions (attachment related). Now, what I'm confused about is that I thought you were performing the update via Ajax (no page refresh), in which case I don't get the reasoning for editing the `$header` variable where you do. Simply eval `$modnotice`, print it somehow and get Javascript to change the old div with the new one by using the div's ID attribute value.

Also, what happens if there was never a moderation notice to begin with, will this display the notice once moderators unapprove one attachment?

I think you are over doing it, even assuming I fully understand I think you are overdoing it.

I would even suggest not display the moderation notice at all in certain pages were user can interact with moderated-capable content via Ajax (new thread, new reply, edit post).

@mybb/developers

lairdshaw:
> Now, what I'm confused about is that I thought you were performing the update via Ajax (no page refresh)

Ah, no, the opposite is the case: this code runs in the event that Javascript is _disabled_, in which case clicking the remove/(un)approve buttons results in a full page query to the server, and the loading of a new page.

At the start of that page load, global.php is included, and it sets `$header` with the normal `eval()`. But then the content page - editpost.php - runs, and processes attachments. At that point, the header may need to be updated, because, for example, the attachment which was unapproved when global.php built the header is now approved given the (approve) button that the user clicked, and so the moderation queue notice (that there is one unapproved attachment) needs to be removed from the header. (A variation is when an approved attachment is unapproved by the user, and the header needs to be updated to indicate that there is now - assuming there were no others - one unapproved attachment).

> Also, what happens if there was never a moderation notice to begin with, will this display the notice once moderators unapprove one attachment?

Yes. That's the point of the new `global_modqueue_notice_placeholder` template, which ordinarily doesn't show up (it is empty and unstyled) but which is replaced with a full moderation queue notice in the scenario that you raise.

yuliu:
> > Looks like you've missed something that me and @lairdshaw have talked about.
> 
> If I understand the conclusion was the same, wasn't it? I didn't _missed_ per se, but simply ignored commenting because the conclusion was ultimately to change the name attribute value.

Yes you're right here but changing the input name attributes should not affect 3rd party themes. What would affect 3rd party themes is modifications to `post.js` and it has been enhanced via commit https://github.com/mybb/mybb/pull/3926/commits/5550766b994d720dc3941c8a8e9104ace7a3c8ea . I think there's no more theme compatibility issue here except for what the "redo header" or "redo mod_queue" is doing.

> I think you are over doing it, even assuming I fully understand I think you are overdoing it.

I'm feeling the same way.

> There is one moderator notice located in the header which you have found requires update after performing some actions (attachment related). Now, what I'm confused about is that I thought you were performing the update via Ajax (no page refresh), in which case I don't get the reasoning for editing the `$header` variable where you do. Simply eval `$modnotice`, print it somehow and get Javascript to change the old div with the new one by using the div's ID attribute value.
> Also, what happens if there was never a moderation notice to begin with, will this display the notice once moderators unapprove one attachment?
> I would even suggest not display the moderation notice at all in certain pages were user can interact with moderated-capable content via Ajax (new thread, new reply, edit post) if you want to "fix" that issue.

I agree on this point. Real-time updating on MyBB is not applicable currently, and maybe also not ideal unless someone wish it turns to be a Gmail-like site.

There could be many circumstances that the page won't show a modnotice or even a full header at all in 3rd party theme or plugin pages. The latest commit that has added an extra template and hard-coded HTML in scripts is somehow overdoing and probably theme/plugin unfriendly ~~incompatible~~.

---

> Something new came up - I realised that attachments also weren't being saved when the "Preview Post" and "Save as Draft" buttons were clicked. The first new commit above ('Save attachments on "Preview Post" and "Save as Draft".') fixes that.

I have some other concerns about this and the posting/editing screen's UX, but not so relevant to this PR and will cut this part of discussion.

> Great idea. The second new commit above ('Replace mod queue in $header rather than regen $header') takes this approach. @yuliu, perhaps this satisfies you too?

Sorry, I still hold my opinion in https://github.com/mybb/mybb/pull/3926#issuecomment-613229340 , the header (of course including anything in it) should not be modified in places other than `global.php`, unless MyBB provides other ways for evaluating it.

Eldenroot:
Just my 2 cents - 3rd party theme should follow changes between versions of MyBB. Yeah, it may lead to incompatibility but MyBB cannot stuck in development because of 3rd party themes... there is a list of updated/modified templates in every update blog spot, so theme devs should follow these and implement changes.

euantorano:
Note that 3rd party themes would only need updated if they’ve changed a template that has been updated. I doubt too many have altered the templates that are changed here.

yuliu:
> Just my 2 cents - 3rd party theme should follow changes between versions of MyBB. Yeah, it may lead to incompatibility but MyBB cannot stuck in development because of 3rd party themes... there is a list of updated/modified templates in every update blog spot, so theme devs should follow these and implement changes.

The problem here is that templates changes in this PR except **1)** modifications to input name attributes [https://github.com/mybb/mybb/pull/3926/commits/95381fa2ce2450179c63d25fc03deb2a9f67f564#diff-05b404fc75b0e5bd6d57c79b667b44b8] and **2)** adding missing id attribute [https://github.com/mybb/mybb/pull/3926/commits/c6b533019dc50171b83b4fcf26e842121bf7cbda#diff-05b404fc75b0e5bd6d57c79b667b44b8] are unnecessary, because they're for the "redo header" work which I've got concerns about.

The marked template changes (**1** and **2** in this comment) don't require existing theme to get updated for not breaking existing functions before this PR. Unless in order to fix issues that this PR fixes, one won't need to update their theme/template.

This PR has resolved issues I also care about, including 1) attachment operations when JS is disabled, 2) attachment uploading issues (after removing one) and 3) attachment removing issues. And since I've already made my thoughts clear in previous comments and the above-mentioned issues were resolved here, I probably couldn't give any further opinions.

Sama34:
> At the start of that page load, global.php is included, and it sets $header with the normal eval(). But then the content page - editpost.php - runs, and processes attachments.

I think your change should be done in `global.php`, if you can't then I would go with just hiding the notification in these areas as I mentioned in my last post.

Or you could trigger an JS action once the page is loaded to refresh the notification bar value.

Your approach looks unfriendly to me, maybe you can drop that change and open a separate issue for discussion.

lairdshaw:
> I think your change should be done in `global.php`, if you can't then I would go with just hiding the notification in these areas as I mentioned in my last post.

OK. I've gone with that.

> Or you could trigger an JS action once the page is loaded to refresh the notification bar value.

Oh, I think you're forgetting that this code deals with the scenario in which JS is disabled.

> Your approach looks unfriendly to me

To me too! It's messy and extravagant, but it seemed to be the only way (short of creating something like a `regenerate_header()` function) to maintain a consistent user-interface. I don't mind it being excised.

> maybe you can drop that change

Done. See above commit.

> and open a separate issue for discussion.

Perhaps we could have a discussion about a function to (re)generate the header? I have encountered a need for it separately in a plugin which I'm working on. It is probably a relatively common need.

global.php
```diff
@@ -745,8 +745,11 @@
 	}
 }
 
-// Get awaiting moderation queue stats
-if($can_access_moderationqueue || ($mybb->user['ismoderator'] && $mybb->usergroup['canmodcp'] == 1 && $mybb->usergroup['canmanagemodqueue'] == 1))
+// Get awaiting moderation queue stats, except if the page is editpost.php,
+// because that page can make changes - (un)approving attachments, or deleting
+// unapproved attachments - that would invalidate anything generated here.
+// Just leave this queue notification blank for editpost.php.
+if(!(defined('THIS_SCRIPT') && THIS_SCRIPT == 'editpost.php') && ($can_access_moderationqueue || ($mybb->user['ismoderator'] && $mybb->usergroup['canmodcp'] == 1 && $mybb->usergroup['canmanagemodqueue'] == 1)))
```
Sama34, line 752: This will simply hide the notification in the editpost page, is that correct?

Sama34:
> Perhaps we could have a discussion about a function to (re)generate the header? I have encountered a need for it separately in a plugin which I'm working on. It is probably a relatively common need.

Sounds good to me. It can be applied in future major releases.

global.php
lairdshaw: Correct. It hides _all_ moderation queue notices though, so on top of unapproved attachment notices, it also hides unapproved thread and unapproved post notices. If we wanted to be more fine-grained, and only hide unapproved _attachment_ notices for the editpost page, then we could instead conditionally alter the $modqueue_types array a few lines below (i.e., remove 'attachments' from it when THIS_SCRIPT == 'editpage.php').

Sama34: I'm fine with this, but lets hear @mybb/developers

yuliu: I don't think it's a wise move of making the header visually inconsistent across forum modules.

yuliu:
> > and open a separate issue for discussion.
> 
> Perhaps we could have a discussion about a function to (re)generate the header? I have encountered a need for it separately in a plugin which I'm working on. It is probably a relatively common need.

Yep, I think you could just open an issue for discussion. Altering a template content which should be a common need through plugin is pretty difficult, not only in current 1.8 but also future 1.9 using Twig.

global.php
lairdshaw: OK, @yuliu, but if we do nothing then the header will also be visually inconsistent across forum modules - it will at times for the editpage.php module alone show an unapproved attachment in the queue when there are none, and fail to show an unapproved attachment in the queue when there is one - so, we're kind of "damned if we do; damned if we don't".

Question for you: I've tried three different approaches to this issue (1. regenerate the entire header; 2. selectively replace only the mod queue in the header; 3. selectively disable the mod queue in the header for editpage.php), and you've rejected all three of them, so: what do you think _should_ be done; what is it that you _would_ accept?

Sama34: I myself would be fine with not fixing this right now, if this can be fixed by moving the header and footer build process to functions then this could as well be fixed then.

lairdshaw:
> Sounds good to me. It can be applied in future major releases.

> Yep, I think you could just open an issue for discussion

Thanks for the encouragement, @Sama34 and @yuliu. I've started that discussion in #3939.

global.php
lairdshaw: OK. Thank you, @Sama34. I will wait for a little while for others to provide their feedback, but, unless/until a contrary view is expressed, I will assume that we are going to go with the following approach: ignore this potential UI inconsistency in the header for now; don't try to fix it at all or in any way - we will leave any such fix until 1.9.*. @lairdshaw  should revert the (by now minor) changes that he has made in this PR (to global.php) to try to minimise the effects of the UI inconsistency.

Yes? No? Anybody want to express a contrary view?

yuliu: > but if we do nothing then the header will also be visually inconsistent across forum modules

Yes and such inconsistency may also appear across browser windows / sessions and that's something we're probably gonna to ignore.

> so, we're kind of "damned if we do; damned if we don't".

Exactly.

> Question for you: I've tried three different approaches to this issue (1. regenerate the entire header; 2. selectively replace only the mod queue in the header; 3. selectively disable the mod queue in the header for editpage.php), and you've rejected all three of them, so: what do you think _should_ be done; what is it that you _would_ accept?

Leave current problem (the volatile count number of unapproved attachments) as is. The reasons are: 
- editpost page is a screen focused on editing a post, users probably won't navigate to other pages without saving the post (means exiting editpost page) or will dismiss the changes made (means exiting editpost page).
- no more compatible way to redo the work in _global.php_, proposals by @lairdshaw may break more compared with fixing this one issue, especially on heavily customized forums (by custom codes and/or plugins).

So I suggested to open a discussion about re-design how a page's header/footer/etc. are evaluated which will probably resolve the problem in this PR and the common need of editing relevant templates by plugins.

yuliu: This one seems OK to me now after reading how the pm_notice is taken care of on https://github.com/mybb/mybb/blob/mybb_1822/global.php#L846 , looks like it's not a real solution but only a workaround for now.

Furthermore, the unread pm count has the same issue as yours, see https://github.com/mybb/mybb/blob/mybb_1822/global.php#L547 .

lairdshaw: Interesting, @yuliu. So, there's a precedent for this approach in the code already. Thanks for digging it up.

lairdshaw: Above, I wrote:

> unless/until a contrary view is expressed, I will assume that [...] @lairdshaw should revert the (by now minor) changes that he has made in this PR (to global.php) to try to minimise the effects of the UI inconsistency.

I took @yuliu's later approving comments to be the expression of a contrary view, and so I have not reverted those changes after all. At this point, a couple of days later, I think that unless anybody wants to argue otherwise, this PR is ready as-is.

Eldenroot:
I agree with @Sama34 - but it should be covered by another ticket. Applied to my board, seems to be fine.

euantorano:
Merge conflict resolved, this looks good to me. Thanks, and sorry for the long wait time!

PR #3930, opened 2020-04-14
Add Thread Prefix forum list & filtering

### LINKED ISSUE #3508: Thread prefixes - use style in ACP
Same as for usergroups we should use styles for thread prefixes in ACP too.

https://github.com/mybb/mybb/issues/3291#issuecomment-442525541

### LINKED ISSUE #3929: Add Thread Prefix forum list & filtering
List assigned forums for each prefix and allow to filter prefixes by forum in the ACP's _Thread Prefixes_ management.

dvz:
Resolves #3929
Resolves #3508

Prefixes sorted by:
1. _All Forums_
2. multiple forums
3. forum name ([natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order))
3. prefix (natural sort order)

### DIFF
--- a/admin/modules/config/thread_prefixes.php
+++ b/admin/modules/config/thread_prefixes.php
@@ -541,14 +541,6 @@ if(!$mybb->input['action'])
 					}
 				}
 			}
-			else
-			{
-				$title = $lang->thread_prefixes;
-			}
-		}
-		else
-		{
-			$title = $lang->thread_prefixes;
 		}
 
 		usort($prefixes, 'thread_prefix_sort');
@@ -588,6 +580,12 @@ if(!$mybb->input['action'])
 	{
 		$table->construct_cell($lang->no_thread_prefixes, array('colspan' => 3));
 		$table->construct_row();
+	}
+
+
+	if(!isset($title))
+	{
+		$title = $lang->thread_prefixes;
 	}
 
 	$table->output($title);
--- a/admin/modules/config/thread_prefixes.php
+++ b/admin/modules/config/thread_prefixes.php
@@ -568,7 +568,7 @@ if(!$mybb->input['action'])
 				$forum_names = implode($lang->comma, $forum_names);
 			}
 
-			$table->construct_cell("<a href=\"index.php?module=config-thread_prefixes&amp;action=edit_prefix&amp;pid={$prefix['pid']}\"><strong>".htmlspecialchars_uni($prefix['prefix'])."</strong></a>");
+			$table->construct_cell("<a href=\"index.php?module=config-thread_prefixes&amp;action=edit_prefix&amp;pid={$prefix['pid']}\" style=\"color: inherit;\" title=\"".htmlspecialchars_uni($prefix['prefix'])."\">".$prefix['displaystyle']."</a>");
 			$table->construct_cell($forum_names);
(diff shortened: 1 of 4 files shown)


admin/modules/config/thread_prefixes.php
```diff
@@ -510,14 +510,65 @@ function checkAction(id)
 
 	$table = new Table;
 	$table->construct_header($lang->prefix);
+	$table->construct_header($lang->forums);
 	$table->construct_header($lang->controls, array('class' => 'align_center', 'colspan' => 2));
 
 	$prefixes = build_prefixes();
+
+	foreach($prefixes as &$prefix)
```
yuliu, line 518: `build_prefixes()` could return `false` [ref](https://github.com/mybb/mybb/blob/mybb_1822/inc/functions.php#L3648), so it's better to check if it's an array.

admin/modules/config/thread_prefixes.php
```diff
@@ -510,14 +510,65 @@ function checkAction(id)
...
+	{
+		$prefix['forum_fids'] = explode(',', $prefix['forums']);
+	}
+	unset($prefix);
+
+	$forums = cache_forums();
```
yuliu, line 524: The [`get_forum()`](https://github.com/mybb/mybb/blob/mybb_1822/inc/functions.php#L6267) that wraps details of fetching forums data, IMO, is better than retrieving all forums data. Anyway, using `cache_forums()` is valid.

admin/modules/config/thread_prefixes.php
```diff
@@ -530,7 +581,33 @@ function checkAction(id)
...
+	{
+		return 1;
+	}
+
+	if(count($a['forum_fids']) !== count($b['forum_fids'])) {
+		return count($b['forum_fids']) - count($a['forum_fids']);
+	}
+	elseif(count($a['forum_fids']) === 1)
+	{
+		$forums = cache_forums();
```
yuliu, line 607: Same here. It's worth noting that forum IDs might not exist due to data inconsistency, though rarely happens.

yuliu (commented):
Some comments may be helpful.

Sama34:
Seems mostly fine:
image
image

Only issue is when there are no prefixes:
image

euantorano:
This seems like a big usability improvement to me, which is especially useful on forums with lots of prefixes.

Sama34 (approved):

Sama34:
Is this how the review process works? Haha, anyways seems fine to me now.

Eldenroot:
Can we add something like this alongside with this improvement?
https://github.com/mybb/mybb/issues/3508

It would be reallly coool!

Sama34 (approved):

Sama34:
image

Eldenroot:
@dvz @Sama34 - there is somwthing wrong, style is not loaded if you use classes in global.css :( is possible to ifx that?

dvz:
There's currently no dedicated CSS source that could be included in the ACP and frontend, but you could take advantage of this to avoid modifying core files: https://github.com/mybb/mybb/blob/mybb_1824/admin/inc/class_page.php#L103-L107

Eldenroot:
Please could you elaborate? Thank you!

I use css classes in global.css right now, so it works for frontend

dvz:
Never mind - there's already a `config.css`. In that case it might be best to create a separate file and add it to ACP's `class_page.php` - e.g.:
```php
echo "	<link rel=\"stylesheet\" href=\"../cache/themes/themeID/thread-prefixes.min.css\" type=\"text/css\" />\n";
```

Sama34:
Basically, you need to add your CSS to the AdminCP as well, not just the front end through the theme system.

PR #3932, opened 2020-04-15
Adds hCaptcha support

### LINKED ISSUE #3936: Add hCaptcha support
https://www.hcaptcha.com/

via #3932

n-thumann:
Resolves #3936

This PR adds hCaptcha as an alternative to Google reCAPTCHA.
It also resolves issues like [blocked reCAPTCHA in China](https://community.mybb.com/thread-220816.html) and privacy concerns.
More information on hCaptcha: [hCaptcha.com](https://www.hcaptcha.com/) & [article](https://blog.cloudflare.com/moving-from-recaptcha-to-hcaptcha/) by Cloudflare on moving from reCAPTCHA to hCaptcha.
I´ll be happy about any feedback 😊


inc/class_captcha.php
```diff
@@ -125,6 +135,16 @@ function __construct($build = false, $template = "")
 			{
 				$this->build_recaptcha();
 			}
+		}else if(in_array($this->type, array(6, 7)) && $mybb->settings['captchapublickey'] && $mybb->settings['captchaprivatekey'])
```
euantorano, line 138: Nitpick: code style:

```suggestion
		}
		elseif(in_array($this->type, array(6, 7)) && $mybb->settings['captchapublickey'] && $mybb->settings['captchaprivatekey'])
```

n-thumann: Thanks, done :)

yuliu:
I've got some suggestion, separate pub/private keys' storage of hCaptcha's from reCaptcha's.

euantorano:
I considered separating them, but you can obviously only be using one of them at a time so don’t really see a problem with sharing the setting.

yuliu:
Seems so. But in this way, switching CAPTCHA providers will require forum owners to re-enter the credentials that might be hard to retrieve. They're two sets of valid settings, using one provider doesn't and shouldn't invalidate/alter the other provider's set of settings. Same situation for the mail settings.

euantorano:
I guess so, we have to weigh up between having too few and too many settings and determine which option is clearer. The mail settings can already be confusing for some people, especially the SMTP settings.

Sama34:
I agree settings should be separated and not overwritten. But won't this be an issue if JavaScript disabled? As seeing fields that aren't valid can be confusing.

mhussaincov:
hi there,
i'm blind, this capture solution does it work for people lime me who are blind?
google capture is greatee in the fact that you tick a box and wate?
does this have an audio option where in you listen to the audio and
type the numbers in to the edit box?
Majid

On 16/04/2020, Omar Gonzalez <notifications@github.com> wrote:
> I agree settings should be separated and not overwritten. But won't this be
> an issue if JavaScript disabled? As seeing fields that aren't valid can be
> confusing.
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/mybb/mybb/pull/3932#issuecomment-614509734


-- 
kind regards,
Majid Hussain

n-thumann:
> I agree settings should be separated and not overwritten. But won't this be an issue if JavaScript disabled? As seeing fields that aren't valid can be confusing.

This problem exists with all settings that are shown or hidden base on the selection of another, but I agree on separating reCAPTCHA and hCaptcha Keys 😊 

> hi there, i'm blind, this capture solution does it work for people lime me who are blind? google capture is greatee in the fact that you tick a box and wate? does this have an audio option where in you listen to the audio and type the numbers in to the edit box? Majid

hCaptcha doesn´t offer and audio Captcha, but it has its own [solution for accessibility](https://www.hcaptcha.com/accessibility): You can sign up [here](https://dashboard.hcaptcha.com/signup?type=accessibility) and set an accessiblity cookie that allows you to pass the Captcha without solving it ✌️

euantorano:
I believe that when JS is disabled, MyBB just shows all settings. I could be wrong - and if I am, we need to fix that!

JordanMussi:
Not sure it’s something you can fix that well as if you change a setting inline without saving, the settings dependant on it won’t appear without the js. Is it more confusing and less usable to have to save the setting (ie changing captcha type without yet adding the keys) before the dependant ones are shown?

inc/class_captcha.php
```diff
@@ -126,6 +136,17 @@ function __construct($build = false, $template = "")
 				$this->build_recaptcha();
 			}
 		}
+		elseif(in_array($this->type, array(6, 7)) && $mybb->settings['captchapublickey'] && $mybb->settings['captchaprivatekey'])
+		{
+			// We want to use hCAPTCHA or hCAPTCHA invisible, set the server options
+			$this->server = "//www.hCaptcha.com/1/api.js";
```
JordanMussi, line 142:
```suggestion
			$this->server = "//www.hcaptcha.com/1/api.js";
```

euantorano:
I was meaning that in the case of JS being disabled, all settings should show all the time.

JordanMussi:
Oh right! Whoopsie, that’s ideal then :)

n-thumann:
There is now a separation between reCAPTCHA and hCaptcha Keys ✌️

install/resources/upgrade50.php
```diff
@@ -22,14 +22,19 @@
...
 
 	echo "<p>Updating cache...</p>";
 
 	$cache->delete("banned");
 
+	// Add hCaptcha support
+	$db->update_query("settings", array('optionscode' => 'select\r\n0=No CAPTCHA\r\n1=MyBB Default CAPTCHA\r\n4=NoCAPTCHA reCAPTCHA\r\n5=reCAPTCHA invisible\r\n6=hCaptcha\r\n7=hCaptcha invisible'), "name='captchaimage'");
+	$db->update_query("settings", array('name' => 'recaptchapublickey'), "name='captchapublickey'");
+	$db->update_query("settings", array('name' => 'recaptchaprivatekey'), "name='captchaprivatekey'");
+
```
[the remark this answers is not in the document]
dvz: [`sync_settings()`](https://github.com/mybb/mybb/blob/mybb_1822/install/upgrade.php#L796) updates settings by `name`, so updating `optionscode` here may be unnecessary, but modifying `name` looks correct to keep existing values.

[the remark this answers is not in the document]
yuliu: > [`sync_settings()`](https://github.com/mybb/mybb/blob/mybb_1822/install/upgrade.php#L796) updates settings by `name`, so updating `optionscode` here may be unnecessary, but modifying `name` looks correct to keep existing values.

OK, I get it. That means those changes suggested by me are not essentially necessary. And probably [this line](https://github.com/mybb/mybb/pull/3932/files/0041b46d7e46157b48eb72788874a3d751a53877#diff-1da9a318f936a17cf4b748d0c8043296R34) can be safely removed as well?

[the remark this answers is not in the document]
n-thumann: Yes, I just tested an upgrade an everything works as expected 👍

Sama34:
I think it is a little confusing for the wording to ask for a "public key" and a "private key" when such words aren't used by the service, they call them " Sitekey" and "Account Secret". I think the same applies to reCaptcha but that would a different issue.

Sama34:
I tried this for registration and post creation and it seems to work.

inc/languages/english/global.lang.php
```diff
@@ -441,13 +441,14 @@
...
+$l['invalid_hcaptcha'] = "Please solve the hCaptcha to verify that you're not a robot.";
 $l['invalid_captcha_verify'] = "The image verification code that you entered was incorrect. Please enter the code exactly how it appears in the image.";
 $l['image_verification'] = "Image Verification";
 $l['human_verification'] = "Human Verification";
 $l['verification_note'] = "Please enter the text contained within the image into the text box below it. This process is used to prevent automated spam bots.";
-$l['verification_note_nocaptcha'] = "Please tick the checkbox that you see below. This process is used to prevent automated spam bots.";
+$l['verification_note_nocaptcha'] = $l['verification_note_hcaptcha'] = "Please tick the checkbox that you see below. This process is used to prevent automated spam bots.";
 $l['verification_subnote'] = "(case insensitive)";
-$l['invalid_nocaptcha_transmit'] = "An error occurred with the human verification. Please try again.";
+$l['invalid_nocaptcha_transmit'] = $l['invalid_hcaptcha_transmit'] = "An error occurred with the human verification. Please try again.";
```
Sama34, line 451: If they use the same string then why bother using two different language strings? Why not simply use `invalid_nocaptcha_transmit`?

Sama34:
If using hidden captcha, the captcha appears after trying to send a reply, but doesn't it warns the user somehow if Javascript is disabled or they fail the test. Also, after sending my reply the captcha doesn't hide (it loads a new step) but when trying to fill it suddenly disappears.
image

Also, I'm sure solving or not solving a hidden captcha doesn't stops the user from posting or registering. I'm not sure how hidden reCaptcha works but this certainly doesn't.

inc/languages/english/global.lang.php
yuliu: That's definitely a language system design short sight. Anything could be different in usage and could be more descriptive, though some of the entries are having the same content.

I would try following:
```
$l['invalid_nocaptcha_transmit'] = "An error occurred with the human verification by noCaptcha. Please try again.";
$l['invalid_hcaptcha_transmit'] = "An error occurred with the human verification by hCaptcha. Please try again.";
```

I think I'll start a discussion about the re-design for the language system, especially on how the entry is named.

n-thumann:
> If using hidden captcha, the captcha appears after trying to send a reply, but doesn't it warns the user somehow if Javascript is disabled or they fail the test.

If the user failed the Captcha, either the Captcha itself says to ("Please try again." in the bottom right) or when the server-side validation fails, MyBB throws an error ("Please solve the hCaptcha to verify that you're not a robot."). Can you try again or specify your environment for me to reproduce?
Disabled JavaScript also affects reCAPTCHA, so it´s more a general problem. I could add a warning in `<noscript>` for both reCAPTCHA and hCaptcha? 😊 
>Also, after sending my reply the captcha doesn't hide (it loads a new step) but when trying to fill it suddenly disappears.

I was able to partially reproduce this error: After solving the hCaptcha, the Captcha briefly reloads, but after that the whole sites reloads anyway. Nevertheless, I solved this issue by removing the Captcha manually from the DOM after submitting.

> Also, I'm sure solving or not solving a hidden captcha doesn't stops the user from posting or registering. I'm not sure how hidden reCaptcha works but this certainly doesn't.

Can you elaborate on that? The Captcha is properly verified on the server, just like with reCAPTCHA

inc/languages/english/global.lang.php
n-thumann: > If they use the same string then why bother using two different language strings? Why not simply use `invalid_nocaptcha_transmit`?

As this is a String for reCAPTCHA, I tried avoiding it in the usage of hCaptcha.

> I would try following:
> 
> ```
> $l['invalid_nocaptcha_transmit'] = "An error occurred with the human verification by noCaptcha. Please try again.";
> $l['invalid_hcaptcha_transmit'] = "An error occurred with the human verification by hCaptcha. Please try again.";
> ```
Sounds good to me 👍

Sama34:
> when the server-side validation fails, MyBB throws an error ("Please solve the hCaptcha to verify that you're not a robot.").

>Can you elaborate on that? The Captcha is properly verified on the server, just like with reCAPTCHA

This isn't the case when using hidden captcha. Hidden captcha doesn't seem to care about validation, I was able to post or register without solving hidden captcha.

I did solve incorrectly the captcha, selecting wrong images and getting the "try again" message, the MyBB code seems to be only checking for captcha submission but not result.

I found no issues with normal captcha.

n-thumann:
> This isn't the case when using hidden captcha. Hidden captcha doesn't seem to care about validation, I was able to post or register without solving hidden captcha.

Can you add `var_dump($response);` [here](https://github.com/n-thumann/mybb/blob/hCaptcha_support/inc/class_captcha.php#L326), then try to post something (or login) while incorrectly solving the Captcha and report back to me with the POST request and response to `https://hcaptcha.com/checkcaptcha/[...]` and `http://foo.bar:8080/newreply.php?tid=[...]`?
> I did solve incorrectly the captcha, selecting wrong images and getting the "try again" message, the MyBB code seems to be only checking for captcha submission but not result.

How did you managed to get hold of the `h-captcha-response` (or `generated_pass_UUID` from the reponse from hCaptcha) token, when the Captcha failed? When the Captcha fails, a new challenge is presented automatically, isn´t it?
> I found no issues with normal captcha.

That´s odd, because both visible and hidden Captcha use the same validation code. 

Is someone else able to reproduce this issue?

Sama34:
@n-thumann I was unable to reproduce but instead I think the explanation was wrong (because I got confused?).

After solving the captcha and successfully sending a quick reply as a guest, the captcha returns and keep asking for solving.

I suppose this was the original issue and not users bypassing the captcha at all as I still had to solve the captcha correctly to register and post.

Eldenroot:
I applied to my test board, just checked few times and works as expected. Maybe I will switch to his one, because I do not like Google at all... :)

Sama34:
@Eldenroot did you try multiple times with the quick reply as a guest? Could you try to figure out if you see the same issue as me?

n-thumann:
@Sama34 The issue should be fixed with my latest commit ✌️

Eldenroot:
Works fine @Sama34 :)

Sama34:
Yes, for me it worked fine, we only need for the rest of pending reviews so it can be merged.

Eldenroot:
I have switched from Google recaptcha to this one... seems to be fine (testing for last 7 days on my board as a direct replacement).

inc/languages/english/global.lang.php
```diff
@@ -442,13 +442,15 @@
 
 $l['invalid_post_code'] = "Authorization code mismatch. Are you accessing this function correctly? Please go back and try again.";
 $l['invalid_nocaptcha'] = "Please solve the reCAPTCHA to verify that you're not a robot.";
+$l['invalid_hcaptcha'] = "Please solve the hCaptcha to verify that you're not a robot.";
 $l['invalid_captcha_verify'] = "The image verification code that you entered was incorrect. Please enter the code exactly how it appears in the image.";
 $l['image_verification'] = "Image Verification";
 $l['human_verification'] = "Human Verification";
 $l['verification_note'] = "Please enter the text contained within the image into the text box below it. This process is used to prevent automated spam bots.";
-$l['verification_note_nocaptcha'] = "Please tick the checkbox that you see below. This process is used to prevent automated spam bots.";
+$l['verification_note_nocaptcha'] = $l['verification_note_hcaptcha'] = "Please tick the checkbox that you see below. This process is used to prevent automated spam bots.";
```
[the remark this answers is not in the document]
n-thumann: Thanks, I´ve overlooked that! 👍

n-thumann:
quick login no captcha support
sendthread invisible
> There are additional templates to be cached should be added:

Done!
> 
> And there's another `build_recaptcha()` needs to be added for hCaptcha for maximum login attempt reached during user login, after noCaptcha/reCaptcha:

Also done!
> Rest part of this PR seems fine. I had a test with _contact_, _registration_, _newthread_, _newreply_ and _quick reply_, using hCaptcha or hCaptcha invisible, all seems fine, for a fresh install and an upgrade.

Thank´s for you review and feedback! ✌️ 
While testing these changes, I noticed that both invisible Captchas are not working when sending a thread (`sendthread.php`) due to a missing `name=submit` attribute in the submit button. Should I fix that as well? 😊

yuliu:
> Done!
> Also done!

Great, thank you!

> While testing these changes, I noticed that both invisible Captchas are not working when sending a thread (`sendthread.php`) due to a missing `name=submit` attribute in the submit button. Should I fix that as well? 😊

Yep I reckon so and lacking of such an attribute will also prevent reCaptcha invisible from working properly, is that correct?

I think it's better to create another issue and open a PR to address it for better issue tracking.

Eldenroot:
@yuliu @n-thumann - I think I fixed partially here: https://github.com/Eldenroot/mybb/commit/41692a5b39380520dc7176395a9384c0569a2841

So it must be a second part which was overlooked?

Ben-MyBB:
> 
> 
> Resolves #3936
> 
> This PR adds hCaptcha as an alternative to Google reCAPTCHA.
> It also resolves issues like [blocked reCAPTCHA in China](https://community.mybb.com/thread-220816.html) and privacy concerns.
> More information on hCaptcha: [hCaptcha.com](https://www.hcaptcha.com/) & [article](https://blog.cloudflare.com/moving-from-recaptcha-to-hcaptcha/) by Cloudflare on moving from reCAPTCHA to hCaptcha.
> I´ll be happy about any feedback 😊

Not sure if you would be able to send us a PR to the docs site with some info on hCaptcha? https://docs.mybb.com/1.8/administration/spam/

n-thumann:
> Not sure if you would be able to send us a PR to the docs site with some info on hCaptcha? https://docs.mybb.com/1.8/administration/spam/

Done: https://github.com/mybb/docs.mybb.com/pull/196 ✌️

Ben-MyBB:
Thanks! Much appreciated

Issue #756, opened 2014-06-16
Remove Hardcoded HTML


Starpaul20:
All hardcoded HTML should be moved to templates.

PR #3933, opened 2020-04-16
Fixes #3931 - language files corrections

### LINKED ISSUE #3931: Language entry issues
In "admin/forum_announcements.lang.php":
update_announcement: Save Announcement seems should be Update Announcement

In "admin/forum_attachments.lang.php":
name_contains_desc: To search by wild card enter *.[file extension]. Example: *.zip.
Looks like the attachment search in AdminCP doesn't support wildcard. Could anyone verify this?

In "/admin/modules/config/attachment_types.php" line 645:
It should be success_deactivated_attachment_type rather than success_activated_attachment_type for deactivating an attachment type.

Original thread: [Language entry issues](https://community.mybb.com/thread-227309.html)

Eldenroot:
Resolves #3931

Anyway I am not sure about the wildcard * - so please any native speaker who can help me to polish it :) thx!


yuliu:
As @lairdshaw [suggested](https://community.mybb.com/thread-227309-post-1347313.html#pid1347313), how about changing
`To search by wild card enter *.[file extension]. Example: *.zip.`
to
`To search by file extension, type <strong>.[file extension]</strong>. Example: <strong>.zip</strong>.`

Or maybe we could implement the feature of searching by wildcard in future.

Eldenroot:
@yuliu - yeah, much better, thx! I have fixed the PR.

Please open a new issue for this wildcard searching support :)

Sama34:
The dot in `.ext` seems irrelevant now that you can only search by extension.

lairdshaw:
> Or maybe we could implement the feature of searching by wildcard in future.

Yep. We could just replace any '*' in the user input with a '%' and (continue to) use the LIKE SQL keyword. But it might be easier just to leave the query as-is - we might want to ask: are many false-positives generally being picked up as-is, such that we need to limit the number of existing returns?

> The dot in `.ext` seems irrelevant now that you can only search by extension.

Am curious as to what you mean by saying that users can "only" search by extension - was that a typo?

But I should probably save all this for any potential new issue for wildcard search support that @yuliu opens on @Eldenroot's suggestion...

Sama34:
Yeah, I got confused. This seems fine to me.

inc/languages/english/admin/forum_attachments.lang.php
```diff
@@ -78,7 +78,7 @@
 $l['index_find_attachments'] = "Attachments - Find Attachments";
 $l['find_where'] = "Find attachments where&hellip;";
 $l['name_contains'] = "File name contains";
-$l['name_contains_desc'] = "To search enter .[file extension]. Example: .zip.";
+$l['name_contains_desc'] = "To search by file extension, type <strong>.[file extension]</strong>. Example: <strong>.zip</strong>.";
```
JordanMussi, line 81: I don't think HTML should be added to the language files after the large effort to move all HTML to templates: #756 #2025

euantorano: I'd agree. I wonder if `To search by file extension, type '.[file extension]'. Example: '.zip'.` would be clear enough?

yuliu: > I don't think HTML should be added to the language files after the large effort to move all HTML to templates: #756 #2025

IMO it's kind of based on the web page designer / language pack author. I personally tend to use `<strong>`/`<em>` for emphasizing.

> I'd agree. I wonder if `To search by file extension, type '.[file extension]'. Example: '.zip'.` would be clear enough?

I'm not sure because someone would probably type exactly `'.[file extension]'`/`'.zip'` into the input area. And it's a little bit common to see some page say `please type "<something_should_be_entered>" (without quotes) in the text box`.

Anyway, it's OK for me to have it written in either way and let it be in MyBB's way.

JordanMussi: Perhaps just rewriting so the `.zip` doesn't fall right before a full stop will avoid confusion enough?

```suggestion
$l['name_contains_desc'] = "Search for attachments that include the given query in the file name. For example, enter .zip to find attachments using the .zip file extension.";
```

yuliu: > Perhaps just rewriting so the `.zip` doesn't fall right before a full stop will avoid confusion enough?

Looks good to me. And that's what @Eldenroot was calling. 🤣

Sama34:
I'm just curious, why didn't you fixed wild card search instead? @Eldenroot suggest to add wild card support, but I don't see why adding a new input instead of fixing this one.

yuliu:
> I'm just curious, why didn't you fixed wild card search instead? @Eldenroot suggest to add wild card support, but I don't see why adding a new input instead of fixing this one.

Either way of fixing issues is valid. Fix the language entry since the code doesn't support wildcard. Fix the search code since the language entry has suggested a wildcard search feature.

JordanMussi:
@Sama34 The fields are using a `LIKE` condition already. The issue is that the description is confusing because it suggests to use a `*` as a wildcard when it is not needed for matching with `.zip` for example.
https://github.com/mybb/mybb/blob/44fc01f723b122be1bc8daaca324e29b690901d6/admin/modules/forum/attachments.php#L646-L654

yuliu:
> @Sama34 The fields are using a `LIKE` condition already. The issue is that the description is confusing because it suggests to use a `*` as a wildcard when it is not needed for matching with `.zip` for example.
> https://github.com/mybb/mybb/blob/44fc01f723b122be1bc8daaca324e29b690901d6/admin/modules/forum/attachments.php#L646-L654

Not really, it searches "*.zip*" and will match `hello.zip.md5`.

JordanMussi:
Well yes, but entering `.zip` matches any file that includes `.zip` but `*.zip` does not which the description gets wrong. If a field for only matching the last part after the last `.` is wanted then a extension field should be added instead of changing the meaning of this one.

Eldenroot:
@JordanMussi - thx for your suggestion, it is implemented!

For others - this is a the simplest temporally solution... it would be better to fix searching at all, but we can create a second issue for this and fix it in future.

Sama34:
Oh, I though it was meant to be used like _keyword*.zip_, to search for _keywork**anything**.zip_. Nevermind if that isn't the case.

However, I noticed there is already an input for file type..
image

I think the first input should allow to search by file name using wildcards and so do the second input for file types. The way this PR suggest we are just _duplicating_ the input.

> Not really, it searches ".zip" and will match hello.zip.md5.

If you use the _File name contains_ input yes, this would be correct. If you use the _File type contains_ input it should be correct when the file extension is `.zip.md5` (which apparently MyBB allows).

>For others - this is a the simplest temporally solution... it would be better to fix searching at all, but we can create a second issue for this and fix it in future.

If I'm not wrong I think you should be reverting the changes in this PR and the search should be fixed altogether in a different issue.

JordanMussi:
> However, I noticed there is already an input for file type..

It seems that is used for MIME type instead of extension.
https://github.com/mybb/mybb/blob/44fc01f723b122be1bc8daaca324e29b690901d6/admin/modules/forum/attachments.php#L899

Which again is missing a clear description - it has no description at all!

> If I'm not wrong I think you should be reverting the changes in this PR and the search should be fixed altogether in a different issue.

I think the description just needs to be clear that the field should be used to find attachments whose file name contains the string entered. (which I think is intention of this PR)

> If a field for only matching the last part after the last . is wanted then a extension field should be added instead of changing the meaning of this one.

Sama34:
Oh, I suppose you are quite right. @Eldenroot could you add in this PR a language string to clarify the _File type contains~ input field as well? Then I'd have nothing to add to the current suggestions for the _File name contains_ string.

> If a field for only matching the last part after the last . is wanted then a extension field should be added instead of changing the meaning of this one.

Yes, I agree.

Eldenroot:
@Sama34 - ok, write me what text do you wanna add?

Sama34:
To be honest I have no idea what that input is for, it seems like it does nothing. Leave it as is.

Eldenroot:
@Sama34 please review :)

PR #3942, opened 2020-04-17
Fix #3937 - utf8 aware keyword cleaning for fulltext searches

### LINKED ISSUE #3836: Search - non-english characters are not counted to the search limit
My minimum search word lenght is set to 3 (Standard search type).

Anyway if you use any "non-english" character - iny my case "á", "é", "í", "ž" etc. (Czech language) it wont count as a character to the limit. So I get the error message that my phrase is too short. (noa vs noá, etc.)

image

### LINKED ISSUE #3937: Fulltext search does not work in non-English language
Hello, 
The forum in Polish language. Fulltext search option is set in the ACP Search options.

The word "tydzień" is "week" in Polish - fairly popular.

Bug: When I type in "tydzień" in the Advanced Search and search only in subject, I receive a message, that there's no search results.

Mysql fulltext seems to work,

SELECT count(*) FROM `mybb_threads` WHERE match(subject) against ('tydzień');
returns 14 entries.

When I switch to "Standard search" I do get 14 results too.

I came across this bug several times on different installations, hosts etc.

Queries with only ASCII chars work just fine between the Standard and Fulltext seach modes. They return sometimes different number of results due to the nature of the fulltext seach versus the Like '%%' query ;)

Original thread: [Fulltext search does not work in non-English language](https://community.mybb.com/thread-227328.html)

euantorano:
Fixed #3937
Fixed #3836

See https://github.com/mybb/mybb/issues/3937#issuecomment-615487837 for context.

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -282,8 +282,18 @@ function clean_keywords_ft($keywords)
 	}
 	$min_word_length -= 1;
 
+	$word_length_regex = '';
+	if($min_word_length <= 1)
+	{
+		$word_length_regex = '{1}';
+	}
+	else
+	{
+		$word_length_regex = "{1,{$min_word_length}}";
+	}
+
 	// Replaces less than 3 characters
-	$keywords = preg_replace("/(\b.{1,{$min_word_length}})(\s)|(\b.{1,{$min_word_length}}$)/u", '$2', $keywords);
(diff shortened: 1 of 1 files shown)


Sama34 (approved):

Eldenroot:
@euantorano - this fixes the mentioned issue (but maybe i nfuture on some other installations it will need some adjustments). On my local board no issue, but I cannot test on my live board.

PR #3944, opened 2020-04-19
Fixes Missing globals/variables and duplicate array_keys

### LINKED ISSUE #3943: Missing globals/variables  and array_key fixes
Reviewing the MyBB code found some areas to fix will submit a PR

Missing global $mybb in function verify_timezone(), function output_custom_profile_fields, upgrade_css_120_to_140

Missing $collation in upgrade12.php
$collation = $db->build_create_table_collation();

cache_stylesheet(1, $stylesheet['attributes']['name'], $stylesheet['value']); in upgrade30.php 

Missing variables for COPY_YEAR and $login_label_width

Duplicate array key fixes

VBGAMER45:
Fixes Missing globals/variables and duplicate array_keys

Signed-off-by: Jonathan <vbgamer45@gmail.com>

Resolves #3943

### DIFF
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -722,7 +722,6 @@ function add_attachments($pid, $forumpermissions, $attachwhere, $action=false)
 		{
 			foreach($fields as $field)
 			{
-				$attach1[$field] = $_FILES['attachments'][$field][$i];
 				$attachments[$i][$field] = $_FILES['attachments'][$field][$i];
 			}
 
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -722,7 +722,7 @@ function add_attachments($pid, $forumpermissions, $attachwhere, $action=false)
 		{
 			foreach($fields as $field)
 			{
-				$attach1[$field] = $_FILES['attachments'][$field][$key];
+				$attach1[$field] = $_FILES['attachments'][$field][$i];
 				$attachments[$i][$field] = $_FILES['attachments'][$field][$i];
 			}
 
--- a/install/resources/upgrade12.php
+++ b/install/resources/upgrade12.php
@@ -1865,6 +1865,8 @@ function upgrade12_redothemes()
 		$db->drop_table("themestylesheets");
 	}
 
+	$collation = $db->build_create_table_collation();
+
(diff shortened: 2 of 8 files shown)


inc/functions_upload.php
```diff
@@ -722,7 +722,7 @@ function add_attachments($pid, $forumpermissions, $attachwhere, $action=false)
 		{
 			foreach($fields as $field)
 			{
-				$attach1[$field] = $_FILES['attachments'][$field][$key];
+				$attach1[$field] = $_FILES['attachments'][$field][$i];
```
yuliu, line 725: Looks like `$attach1` is never used in this function, maybe legacy code.

VBGAMER45: Yeah, that's what I saw as well looking at it.  I see a few other undefined variables in other areas but probably have to someone look at them for the fix.

Sama34: If it has no use then it should be removed.

VBGAMER45: Done removed the line

install/resources/upgrade12.php
```diff
@@ -1865,6 +1865,8 @@ function upgrade12_redothemes()
 		$db->drop_table("themestylesheets");
 	}
 
+	$collation = $db->build_create_table_collation();
+
```
Sama34, line 1869: We no longer support these versions so fixing this seems.

@mybb/developers

install/resources/upgrade30.php
```diff
@@ -2200,7 +2199,7 @@ function upgrade30_updatetheme()
 					$sid = $db->insert_query("themestylesheets", $new_stylesheet);
 					$css_url = "css.php?stylesheet={$sid}";
 
-					$cached = cache_stylesheet($tid, $stylesheet['attributes']['name'], $stylesheet['value']);
```
Sama34, line 2203: Just the same, unsupported version.

euantorano:
We don’t support these old versions, but I don’t see a reason to reject a patch for them when there’s a clear issue.

JordanMussi:
Fixing the problems will only help people upgrade to the latest!

Eldenroot:
Good catch!

PR #3946, opened 2020-04-19
Fixes #3945 Fields to show isn't mandatory for administrative user views

### LINKED ISSUE #3945: Fields to show isn't mandatory for administrative user views
If administrators create views (Home » Users » View Manager) with no Fields to Show selected there is a PHP error.

Administrators shouldn't be able to create views with no fields to show.

Original thread: [Fields to show isn't mandatory for administrative user views](https://community.mybb.com/thread-227381.html)

PaulBender:
Fixes #3945

### DIFF
--- a/admin/inc/functions_view_manager.php
+++ b/admin/inc/functions_view_manager.php
@@ -64,7 +64,7 @@ function view_manager($base_url, $type, $fields, $sort_options=array(), $conditi
 			{
 				$mybb->input['fields'] = explode(",", $mybb->input['fields_js']);
 			}
-			if(is_array($mybb->input['fields']) && count($mybb->input['fields']) <= 0)
+			if(!is_array($mybb->input['fields']) || (is_array($mybb->input['fields']) && count($mybb->input['fields']) <= 0))
 			{
 				$errors[] = $lang->error_no_view_fields;
 			}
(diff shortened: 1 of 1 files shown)


admin/inc/functions_view_manager.php
```diff
@@ -64,7 +64,7 @@ function view_manager($base_url, $type, $fields, $sort_options=array(), $conditi
 			{
 				$mybb->input['fields'] = explode(",", $mybb->input['fields_js']);
 			}
-			if(is_array($mybb->input['fields']) && count($mybb->input['fields']) <= 0)
+			if(!is_array($mybb->input['fields']) || (is_array($mybb->input['fields']) && count($mybb->input['fields']) <= 0))
```
JordanMussi, line 67: I think the second `is_array` check is not needed because if `$mybb->input['fields']`:
- is an array the first condition will be false so the second `is_array` will return true again
- is not an array the first condition will be true and second check short-circuited 

```suggestion
			if(!is_array($mybb->input['fields']) || count($mybb->input['fields']) <= 0)
```

admin/inc/functions_view_manager.php
```diff
@@ -254,7 +254,7 @@ function view_manager($base_url, $type, $fields, $sort_options=array(), $conditi
 				$mybb->input['fields'] = explode(",", $mybb->input['fields_js']);
 			}
 
-			if(is_array($mybb->input['fields']) && count($mybb->input['fields']) <= 0)
+			if(!is_array($mybb->input['fields']) || (is_array($mybb->input['fields']) && count($mybb->input['fields']) <= 0))
```
JordanMussi, line 257: Same here :-)

```suggestion
			if(!is_array($mybb->input['fields']) || count($mybb->input['fields']) <= 0)
```

PR #3950, opened 2020-04-21
ACP Group Join Settings UI Improvement

### LINKED ISSUE #3947: Checkbox in group settings doesn't work
I go to **ACP** > **Users & Groups** > **Groups**, and open any group I have created,
Than i go under "Publicly Joinable Options" and select "Yes, users must be invited in order to join this group" and when I click "Save user group" I get "The selected user group has been updated successfully." message. Until this point everything is okay, but the problem is that this checkbox didn't applied to group settings, so when I came back to chack settings, I saw that checkbox isn't selected. 
I tried this few times with few other groups and the issue is the same.

effone:
Resolves #3947

Added little flavor of js to improve ACP Group Join Settings avoiding confusion.

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -1043,7 +1043,19 @@ if($mybb->input['action'] == "edit")
 	$tabs = $plugins->run_hooks("admin_user_groups_edit_graph_tabs", $tabs);
 	$page->output_tab_control($tabs);
 
-	echo "<div id=\"tab_general\">";
+	echo "<div id=\"tab_general\">
+	<script type=\"text/javascript\">
+		$(function(){
+			$('input[name=\"moderate\"]').parents(\".group_settings_bit\").addClass(\"joinable_dependent\");
+			$('input[name=\"invite\"]').parents(\".group_settings_bit\").addClass(\"joinable_dependent\");
+			if($('input[name=\"joinable\"]').prop(\"checked\") == false){
+				$(\".joinable_dependent\").hide();
+			}
+			$('input[name=\"joinable\"]').on('change', function() {
+				$(\".joinable_dependent\").slideToggle();
+			})
+		});
+	</script>";
 	$form_container = new FormContainer($lang->general);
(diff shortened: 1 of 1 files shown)


Sama34 (approved):
Since _Yes, all new join requests need to be moderated first_ and _Yes, users must be invited in order to join this group_ can't be active at the same time they should probably be radio buttons.

effone:
A very good point @Sama34 but it is also possible for a user to only keep the option `Yes, users can freely join and leave this group` not opting in any of the second and third option.
So, in a scenario if either of the radio is selected already and user wants to **not** opt any of those (2nd and 3rd options) then in order to clear the already selected radio; user needs to first uncheck the first option -> save -> then again check the first option -> save again. Is it OK and acceptable?

Turning those to radio is easy, just a confirmation needed.

dvz:
Group `type` options: https://github.com/mybb/mybb/issues/3818#issuecomment-542721458

Radio for types 2-5 should work well:
> 2: Disable join and leave options
3: Unrestricted
4: Moderate join requests
5: Invitation only

Eldenroot:
Group type would beusable too, it depends what do you prefer :)

effone:
4 radios sound good.

Sama34:
@effoce you could hide one field when the other is selected, that could be enough?

Eldenroot:
The first or the second one... both are better than it is right now :)

Sama34:
Just to clarify, if you want to improve it I will review again but I'm fine with the PR as is.

Eldenroot:
It should be fine, briefly tested.

effone:
I'm gonna push a new PR shortly, non-js radio control only.
We can compare and adapt the best ...

Eldenroot:
OK, that would be great!

Ben-MyBB:
@effone just a reminder about the extra PR when you get a chance

effone:
OK, give me some times now and I'll do it.

Eldenroot:
This one should be finished, rejected or merged.

Sama34:
If someone wants to improve this please send a PR.

PR #3956, opened 2020-04-25
Fixes #3572 - Resend email activation

### LINKED ISSUE #3572: bug Resend email activation
query of do_resendActivation in member.php line 1515 should change to...
```
$query = $db->query("
		SELECT u.uid, u.username, u.usergroup, u.email, a.code, a.type, a.validated
		FROM ".TABLE_PREFIX."users u
		LEFT JOIN ".TABLE_PREFIX."awaitingactivation a ON (a.uid=u.uid AND (a.type='r' OR a.type='b'))
		WHERE u.email='".$db->escape_string($mybb->get_input('email'))."'
	");
```
"just put ```a.type='r' OR a.type='b'``` in parentheses"

what happened:
by change register type from both to email verify, some users that registered before it, has type 'b' and activated field has value '1'
in query of line 1515 in member.php return many records that are type'b' and activated'1'!!!
if we put parentheses, the condition will be correct.

Eldenroot:
Resolves #3572

Untested, please review.

@Ben-MyBB @Sama34

### DIFF
--- a/member.php
+++ b/member.php
@@ -1519,7 +1519,7 @@ if($mybb->input['action'] == "do_resendactivation" && $mybb->request_method == "
 	$query = $db->query("
 		SELECT u.uid, u.username, u.usergroup, u.email, a.code, a.type, a.validated
 		FROM ".TABLE_PREFIX."users u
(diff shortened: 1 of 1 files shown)


Eldenroot:
It should be fine, briefly tested.

PR #3961, opened 2020-04-26
Fixes #3960 - COPPA URL

Eldenroot:
Resolves #3960

PR #3962, opened 2020-04-26
Fixes #3960 - COPPA URL

### LINKED ISSUE #3960: COPPA URL is not valid anymore
As title says... it needs to be replaced (ACP and frontend part).

Eldenroot:
Resolves #3960


Ben-MyBB:
#3961

PR #3969, opened 2020-04-30
Fix spider double counting, use separate query for global guest count on forum index

### LINKED ISSUE #3852: Incorrect guests number counting

dvz:
Resolves #3852

### DIFF
--- a/index.php
+++ b/index.php
@@ -94,7 +94,7 @@ if($mybb->settings['showwol'] != 0 && $mybb->usergroup['canviewonline'] != 0)
 	// Fetch spiders
 	$spiders = $cache->read('spiders');
 
-	// Loop through all users.
+	// Loop through all users and spiders.
 	while($user = $db->fetch_array($query))
 	{
 		// Create a key to test if this user is a search bot.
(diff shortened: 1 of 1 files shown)


yuliu:
Tested locally and on a live board, statistics in the index page looks correct now.

PR #3970, opened 2020-04-30
Send `Cache-Control: no-cache, private` header by default

### LINKED ISSUE #3908: Send safe HTTP cache headers by default
Modify the `nocacheheaders` setting and [`send_page_headers()`](https://github.com/mybb/mybb/blob/mybb_1822/inc/functions.php#L4693) behavior to include `Cache-Control: private` in HTTP headers for dynamic HTML content.

Other headers set in the function may no longer be necessary.

dvz:
Resolves #3908

### DIFF
--- a/install/resources/upgrade50.php
+++ b/install/resources/upgrade50.php
@@ -22,13 +22,15 @@ $upgrade_detail = array(
 
 function upgrade50_dbchanges()
 {
-	global $output, $cache;
+	global $output, $cache, $db;
 
-	$output->print_header("Updating Cache");
+	$output->print_header("Updating Database");
 
 	echo "<p>Updating cache...</p>";
 
 	$cache->delete("banned");
+
+	$db->update_query('settings', array('value' => 1), "name='nocacheheaders'");
 
 	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
 	$output->print_footer("50_done");
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -4736,10 +4736,7 @@ function send_page_headers()
 
 	if($mybb->settings['nocacheheaders'] == 1)
 	{
-		header("Expires: Sat, 1 Jan 2000 01:00:00 GMT");
-		header("Last-Modified: ".gmdate("D, d M Y H:i:s")." GMT");
-		header("Cache-Control: no-cache, must-revalidate");
-		header("Pragma: no-cache");
+		header("Cache-Control: no-cache, private");
 	}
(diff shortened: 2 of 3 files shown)


install/resources/upgrade50.php
```diff
@@ -22,14 +22,16 @@
...
+	global $output, $cache, $db;
 
-	$output->print_header("Updating Cache");
+	$output->print_header("Updating Database");
 
 	echo "<p>Updating cache...</p>";
 
 	$cache->delete("banned");
 
+	$db->update_query('settings', array('value' => 1), "name='nocacheheaders'");
```
euantorano, line 33: I'm not sure we want to force this setting on if a user purposefully wants it disabled for some reason...

Eldenroot: Yeah, I do not why but there could be some users whose wanna do that...

dvz: Note the default value has changed - it wouldn't revert explicit preferences.

It may be the safer option, compared to adding another bullet point to the [security checklist](https://docs.mybb.com/1.8/administration/security/protection/) for existing installations, even if we mentioned it in release notes.

euantorano:
I may be missing something, but won’t setting the setting value in the DB overwrite explicit preferences (though obviously settings.php is there, but isn’t that rebuilt by the upgrade/install process?)?

Perhaps a nice enhancement for the future would be to have a security checklist built in to the home tab of the ACP? I recently did a Phabricator update at work and after upgrade it gives you a list of potential setup issues which is quite a nice feature.

dvz:
> won’t setting the setting value in the DB overwrite explicit preferences (though obviously settings.php is there, but isn’t that rebuilt by the upgrade/install process?)?

It will, but previously the forums with edge cases (needing custom cache headers) didn't have to change the setting from a default value, and the rest should have set it to _Yes_. With this PR, it would be the edge case forums that now have to opt out (look it up and change it, rather than do nothing).

If we added a new setting for `Cache-Control: private` alone, it would be enabled by default as well - including existing installations - but it's currently added on top of the existing one as both should be enabled in most circumstances.

JordanMussi:
Really it was wrong to not be setting the private cache headers by default anyway so I agree with @dvz. Updating the setting to the default is a safe action since all that would happen is any cache infrastructure would be bypassed (as it should have ideally been doing before). A clear message in the release notes is all that is needed.

euantorano:
Alright, I'm convinced. Let's go for it as is.

Issue #999, opened 2014-07-22
Theme Texteditor


euantorano:
I'm not sure if this is just me, but I did a fresh install of 1.8B3 and I'm having this issue on the text editor in the themes &amp; templates.

[attachment=32004]
[Notice the text editor width is off the page, scrolling does nothing.]

Can anyone confirm?

Original thread: [Theme Texteditor](http://community.mybb.com/thread-156391.html)

Issue #555, opened 2014-05-27
More hooks


DiogoParrinha:
Original issue on 1.8 private repo: https://github.com/mybb/MyBB18/issues/39
What needs to be done: check if it's been finished or not.

Other related issues: (or PR's)
https://github.com/mybb/MyBB18/pull/32
https://github.com/mybb/MyBB18/pull/31
## To quote:

http://community.mybb.com/thread-125697.html

PR #3977, opened 2020-05-04
Adds thread start datetime to thread listings in forumdisplay.php.

lairdshaw:
Replaces PR #3891 at @Eldenroot's request.

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -444,6 +444,10 @@ table {
 	background: #ffdde0;
 }
 
+.thread_start_datetime {
+	color: #999;
+}
+
 .post.unapproved_post .post_author {
 	border-bottom-color: #ffb8be;
 }
@@ -3973,7 +3977,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 		{$attachment_count}
 		<div>
           <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']} &raquo; {$thread['start_datetime']}</div>
+			<div><span class="author smalltext">{$thread['profilelink']},</span> <span class="thread_start_datetime smalltext">{$thread['start_datetime']}</span></div>
 		</div>
 	</td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}"><a href="{$mybb->settings['bburl']}/misc.php?action=whoposted&tid={$thread['tid']}" onclick="MyBB.whoPosted({$thread['tid']}); return false;">{$thread['replies']}</a>{$unapproved_posts}</td>
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3977,7 +3977,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 		{$attachment_count}
 		<div>
           <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div><span class="author smalltext">{$thread['profilelink']},</span> <span class="thread_start_datetime smalltext">{$thread['start_datetime']}</span></div>
+			<div><span class="author smalltext">{$thread['profilelink']}{$lang->comma}</span><span class="thread_start_datetime smalltext">{$thread['start_datetime']}</span></div>
 		</div>
 	</td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}"><a href="{$mybb->settings['bburl']}/misc.php?action=whoposted&tid={$thread['tid']}" onclick="MyBB.whoPosted({$thread['tid']}); return false;">{$thread['replies']}</a>{$unapproved_posts}</td>
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3966,14 +3966,14 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 {$forums}
 </table>
 <br />]]></template>
-		<template name="forumdisplay_thread" version="1813"><![CDATA[<tr class="inline_row">
+		<template name="forumdisplay_thread" version="1823"><![CDATA[<tr class="inline_row">
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%"><span class="thread_status {$folder}" title="{$folder_label}">&nbsp;</span></td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%">{$icon}</td>
 	<td class="{$bgcolor}{$thread_type_class}">
 		{$attachment_count}
 		<div>
           <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']}</div>
+			<div class="author smalltext">{$thread['profilelink']} &raquo; {$thread['start_datetime']}</div>
 		</div>
 	</td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}"><a href="{$mybb->settings['bburl']}/misc.php?action=whoposted&tid={$thread['tid']}" onclick="MyBB.whoPosted({$thread['tid']}); return false;">{$thread['replies']}</a>{$unapproved_posts}</td>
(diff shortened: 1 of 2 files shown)


euantorano:
Looks good, I'll just test it this evening before merging to double check.

Eldenroot:
Thx @lairdshaw 

As I see the output would be a little bit different - no title with non-relative time on mouse hover?

install/resources/mybb_theme.xml
```diff
@@ -3966,14 +3966,14 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
...
-		<template name="forumdisplay_thread" version="1813"><![CDATA[<tr class="inline_row">
+		<template name="forumdisplay_thread" version="1823"><![CDATA[<tr class="inline_row">
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%"><span class="thread_status {$folder}" title="{$folder_label}">&nbsp;</span></td>
 	<td align="center" class="{$bgcolor}{$thread_type_class}" width="2%">{$icon}</td>
 	<td class="{$bgcolor}{$thread_type_class}">
 		{$attachment_count}
 		<div>
           <span>{$prefix} {$gotounread}{$thread['threadprefix']}<span class="{$inline_edit_class} {$new_class}" id="tid_{$inline_edit_tid}"><a href="{$thread['threadlink']}">{$thread['subject']}</a></span>{$thread['multipage']}</span>
-			<div class="author smalltext">{$thread['profilelink']}</div>
+			<div class="author smalltext">{$thread['profilelink']} &raquo; {$thread['start_datetime']}</div>
```
JordanMussi, line 3976: I don't understand the use of `&raquo;` as the separator? I think `|` should be used to be consistent with the rest of the templates (eg usercp_forumsubscriptions_forum) or have the date in parentheses.

lairdshaw:
You're welcome, @Eldenroot. And yes, you do get a non-relative time on mouse hover - the `my_date()` function takes care of that.

install/resources/mybb_theme.xml
lairdshaw: > I don't understand the use of `&raquo;` as the separator?

Simply because (1) @Eldenroot used it in the screenshot at the top of his original PR, and (2) the View Unread Posts plugin uses it in similar contexts (e.g., for search result listings), so I figured it was something of a standard or at least an approach with a couple of precedents.

That `|` occurs in a different context: options for the user to click, rather than variables being displayed, so I'm not sure that that's a precedent. It also has the association of "this _or_ that", which seems inapplicable in this context.

Parentheses are an interesting idea but also seem to have an inapplicable association: that the parenthesised content is somehow incidental or less relevant than that which precedes it.

If we're to do away with the `&raquo;`, then maybe we don't need to replace it with anything at all? The username preceding it is already styled as a link, so maybe the datetime is already adequately distinguished from it?

All of that said, I'm not particularly invested in any particular option and am happy to go with whatever you and others think is best.

Eldenroot:
It looks fine but we should add nonrelative time on mouse hover because to follow GUI as we have in postbit - here you will get non relative time in title when you mouse hover over the relative time.

lairdshaw:
> It looks fine but we should add nonrelative time on mouse hover

See [my comment above](https://github.com/mybb/mybb/pull/3977#issuecomment-623403635): the nonrelative time on mouse hover is already present.

install/resources/mybb_theme.xml
JordanMussi: I hadn't noticed it in the other PR before because it wasn't in the code.

A particular plugin using it isn't a strong argument for the core imo but I do agree that `|` is used for a set of links as this or that. My main concern with `&raquo;` is it seems to be used mostly for page links (eg in calendar and pagination `{$lang->next} &raquo;` etc).

I do like the idea of no separator. After further looking through the templates I found `reputation_vote` uses a `-` to separate between reputation and last updated date.

Just made these screenshots for convenience and I think `-` seems the best

image
image
image
image

From seeing how it looks, even with the date in the .smalltext element it feels too over emphasised. Perhaps a colour like `#666` from `.post .post_head span.post_date` would work well here too.

I'm just keen to get the style of this right since the froumdisplay thread listing is quite a key part of the experience of the forum users.

JordanMussi:
@Eldenroot 

https://github.com/mybb/mybb/blob/b81c692c1a7095f131a607b499a8a7f13f256435/inc/functions.php#L466

https://github.com/mybb/mybb/blob/88f12487bf6e61d2728e3b23712d78be005ae7b0/inc/languages/english/global.lang.php#L367

install/resources/mybb_theme.xml
lairdshaw: > My main concern with `&raquo;` is it seems to be used mostly for page links (eg in calendar and pagination `{$lang->next} &raquo;` etc).

Yep, I agree that that's a concern.

> Just made these screenshots for convenience and I think `-` seems the best

While I prefer `-` to `|` and parens, I think I prefer most of all just adding an &nbsp; immediately after the username to add some extra spacing, and otherwise not using any separator at all. Can you try that and see what you think?

> Perhaps a colour like `#666`

Seems to look fine although I'd go with `#555` instead mostly because... connotations...

> I'm just keen to get the style of this right since the froumdisplay thread listing is quite a key part of the experience of the forum users.

Yep, fair enough.

yuliu: A thoughtless suggestion: `{datetime} {$lang->by} {author}`..

Eldenroot:
@lairdshaw @JordanMussi - I am sorry, you are right, thx for pointing me to this :)

JordanMussi:
Don't worry! 😀

install/resources/mybb_theme.xml
JordanMussi: > Seems to look fine although I'd go with `#555` instead mostly because... connotations...

Aha I only suggested it as it's used for the dates on posts. But I agree it is a bit light so `#555` is probably better.

> adding an `&nbsp;` immediately after the username to add some extra spacing

That's what the first screenshot shows and fees too cluttered to me.

> A thoughtless suggestion: `{datetime} {$lang->by} {author}`..

I like the idea but I worry that it makes it harder to scan down the list of threads than it is at the moment with thread topic and author aligned to the left. Again it is too much emphasis on the date...
image

lairdshaw: > > adding an `&nbsp;` immediately after the username to add some extra spacing
> 
> That's what the first screenshot shows and fees too cluttered to me.

Mmm, it seems to me that your first screenshot only shows a _single_ space, whereas I'm suggesting an `&nbsp;` _followed_ by a single space. That adds an extra little bit of differentiation (in the form of spacing) which, for me, makes the difference, and makes that the best option. The fewer bits of information a reader has to process - commas, dashes, parentheses, etc - the easier it is for her/him to skim the listing.



> > A thoughtless suggestion: `{datetime} {$lang->by} {author}`..
> 
> I like the idea but I worry that it makes it harder to scan down the list of threads than it is at the moment with thread topic and author aligned to the left. Again it is too much emphasis on the date...

Agreed. And for a "thoughtless" suggestion, it's very thoughtful (one boggles at the power of thought if this is not it!), but it doesn't quite hit the mark. I sense and believe that the author should be leftmost in the listing, rather than the date.

euantorano: Paging @justinsoltesz is my answer for any questions layout/design related 😉

lairdshaw: Good one, @euantoro. We do need a qualified circuit-breaker!

justinsoltesz: Thinking about this in the context of 1.9's theme, I would suggest a comma:

Screen Shot 2020-05-05 at 10 05 33 AM
Note that the lighter text color there is `#999` with the link color `#666`

This is 1.8 with a comma and `#999`:
Screen Shot 2020-05-05 at 10 07 06 AM

lairdshaw: Thanks, @justinsoltesz. Perhaps, then, we go with a comma and font colour of #999, with that colour set through a class, and selected in global.css? At the very least, this solution eases the transition to 1.9.

euantorano: Yeah, now the only problem is that CSS updates aren't implemented in the upgrade script in 1.8 😢 I wonder if there's some existing CSS class we can use that gets close enough..?

lairdshaw: I can't see one...

Is a temporary inline style (on top of the CSS that won't get updated) too naughty?

euantorano: I think what we should do is add a style to `global.css` and make a note in the release notes for users. That's probably the best way all round.

JordanMussi: It's not a big issue if an administrator doesn't add the style when upgrading so I agree that's best.

justinsoltesz: Agreed, it's not a breaking change if the administrator doesn't add it.

lairdshaw: OK, see what you guys think of my latest commit.

Eldenroot:
OK, now it looks fine, thx!

euantorano:
Looks goo, but I think we should use [`$lang->comma`](https://github.com/mybb/mybb/blob/feature/inc/languages/english/global.lang.php#L529) rather than `,`. I've committed this change directly so I can merge this.

Eldenroot:
We should create an issue for this one :)

euantorano:
Would you mind creating one @Eldenroot and I can link it all up and set the labels.

Eldenroot:
https://github.com/mybb/mybb/issues/3987

PR #3979, opened 2020-05-04
jQuery 3.5.1 update

### LINKED ISSUE #3981: Update jQuery to 3.5.1
Update jQuery from 3.3.1 to [3.5.1](https://blog.jquery.com/2020/05/04/jquery-3-5-1-released-fixing-a-regression/).

effone:
Resolves #3981

...


yuliu:
Just for reference, a newer version 3.5.1 is out yesterday [jQuery 3.5.1 Released: Fixing a Regression](http://blog.jquery.com/2020/05/04/jquery-3-5-1-released-fixing-a-regression/). Don't know if the update is needed..

Eldenroot:
@effone - when you will be ready let me know and I will try to test (at least some parts of mybb to collect feedback)

effone:
Ran Migrate 3.1.0 for a good amount of time. No significant breaking changes found that needs to be addressed other than what I have already. I am marking this ready for review; however deep scrutiny can reveal some area.

Those who wanna test this, the focal point of checking should be:

- Auto closing HTML tags
- AJAX shorthands.

Eldenroot:
I have applied to my test board, of anything I will let you know.

What about other 3rd libraries? Worth try to update tem as well?

jscripts/bbcodes_sceditor.js
```diff
@@ -174,7 +174,7 @@ $(function ($) {
 
 	$.sceditor.command.set('size', {
 		_dropDown: function (editor, caller, callback) {
-			var content = $('<div />'),
+			var content = $('<div></div>'),
```
dvz, line 177: Were these broken after the update?

jscripts/usercp.js
```diff
@@ -199,7 +199,7 @@ var UserCP = {
 			type: 'post',
 			url: 'usercp.php?action=do_editlists&my_post_key='+my_post_key+'&manage='+type,
 			data: { ajax: 1, add_username: type_add_username.val() },
-			async: false,
+			async: true,
```
dvz, line 202: Is this related to the update?

jscripts/bbcodes_sceditor.js
effone: This is the main change that needs to be done with the update.

jscripts/usercp.js
effone: It started throwing console warning after update, hence changed.

dvz: It could be removed completely since `true` is default (https://api.jquery.com/jQuery.ajax/)

jscripts/bbcodes_sceditor.js
dvz: Looks like singular elements should be fine (and can be found in more places in core and dependencies): https://jquery.com/upgrade-guide/3.5/#description-of-the-change

effone: Closing tag is required to be declared separately now. There was a bug running in DOM while auto-closing tag parsing was in place, and the feature is dropped from jQuery 3.5.0 +

I have searched the entire codebase but couldn't find any such implementation that requires attention. Appreciate others to try detecting any existing ...

jscripts/usercp.js
effone: Re declaring the default value doesn't harm a thing. But I will remove those; if you insist.

jscripts/bbcodes_sceditor.js
dvz: Is this a different case? https://github.com/mybb/mybb/blob/mybb_1822/jscripts/usercp.js#L55

effone: Nope, this should also be addressed in similar way.
I'll do it shortly.

dvz: What happens when it's not changed? I don't see any messages or broken functionality on 3.5.1.

effone: It breaks the HTML structure.
Try for the editor code / PHP changes and see what happens in editor.

Note, I have found few more, in acp js that I will fix too.

dvz: The `<textarea>` corrections seem necessary since they're nested, but the simple ones like `$('<div />')` might be fine.

effone: May be some DOM / borwser side correction is correcting the same and you are not experiencing the issue but if you go through the "Description of the change" its clearly stated that not to use such shorthand for closing.

You were right about the nesting as `<div /></span>` now produces `<div><span></span></div>` in place of earlier `<div></div><span></span>` but at the end of the day jQuery.htmlPrefilter closes the tags and make valid HTML. Wherever we have used such shorthands I have changed to open-close just to avoid any unnecessary nesting issues and manually declaring closing tag is completely valid and acceptable to jQuery.

Now suggest me how you want it to be done and I'll modify accordingly.

dvz: > [...] don't use self-closing tags for tags that may have content [...]

It sounds like we'll be fine with these, since they have no content (it's also stated that the `jQuery()` call still transforms tags and closes them).

Even if closing them is best coding practice, it's probably better off with no unnecessary changes, especially if there are no errors or deprecation notices.

effone: OK, I'm reverting that change shortly.

dvz:
Remaining version parameters to update:
https://github.com/mybb/mybb/blob/mybb_1822/install/resources/output.php#L68
https://github.com/mybb/mybb/blob/mybb_1822/admin/inc/class_page.php#L109
https://github.com/mybb/mybb/blob/mybb_1822/admin/inc/class_page.php#L394
https://github.com/mybb/mybb/blob/mybb_1822/admin/inc/class_page.php#L526

https://github.com/mybb/mybb/blob/mybb_1822/install/resources/mybb_theme.xml#L10393
https://github.com/mybb/mybb/blob/mybb_1822/install/resources/mybb_theme.xml#L12214

dvz:
Missed template version bumps for `private_send`, `usercp_editlists`, fixed via https://github.com/mybb/mybb/commit/3061027030f350e91e59be91a94a728f9ee5c006

PR #3983, opened 2020-05-06
Improve query match pattern in DB_PgSQL::insert_id()

### LINKED ISSUE #3224: DB_PgSQL::insert_id() doesn't always return meaningful information
@dvz mentioned this briefly in Discord:

> speaking of core bugs, bumped into [`DB_PgSQL::insert_id()`](https://github.com/mybb/mybb/blob/mybb_1815/inc/db_pgsql.php#L480) not returning any meaningful value a couple of times

In particular, this block looks suspicious:

```php
$this->last_query = str_replace(array("\r", "\t"), '', $this->last_query);
$this->last_query = str_replace("\n", ' ', $this->last_query);
preg_match('#INSERT INTO ([a-zA-Z0-9_\-]+)#i', $this->last_query, $matches);
```

Given a query such as the below, the above code will not yield the expected matches:

```php
$this->write_query("
    INSERT
    INTO {$this->table_prefix}{$table} (".$fields.")
    VALUES (".$values.")
");
```

(Note the indentation here is spaces, not tabs...)

The result for the matches array would be empty. The regexp could be adapted to something like `INSERT\s+INTO\s+([a-zA-Z0-9_\-]+)` and the `str_replace` calls could then be removed.

dvz:
Resolves #3224

### DIFF
--- a/inc/db_pgsql.php
+++ b/inc/db_pgsql.php
@@ -479,9 +479,7 @@ class DB_PgSQL implements DB_Base
 	 */
 	function insert_id()
 	{
-		$this->last_query = str_replace(array("\r", "\t"), '', $this->last_query);
-		$this->last_query = str_replace("\n", ' ', $this->last_query);
-		preg_match('#INSERT INTO ([a-zA-Z0-9_\-]+)#i', $this->last_query, $matches);
+		preg_match('#INSERT\s+INTO\s+([a-zA-Z0-9_\-]+)#i', $this->last_query, $matches);
 
 		$table = $matches[1];
 


euantorano:
Looks good to me. Only problem I can imagine is that the regex doesn't seem to match schema-qualified table names, or quoted table names - but the previous one didn't either...

I don't believe we use either in the core, but who know about plugins.

dvz:
We could shelve that if it's due to have a proper implementation with PDO.

euantorano:
Ideally all the engines should be using PDO in 1.9, which would make maintenance much easier. We can merge this PR as is and worry about other changes in another issue.

PR #3985, opened 2020-05-07
Force return code check upon sending message completion for SMTP mail…

### LINKED ISSUE #2644: smtp fails with zoho.
The smtp mailhandler does not support sending emails via zoho smtp server (smtp.zoho.com, on port 465 with ssl enabled). It fails on line 250 of the smtp mail handler script with the "&#36;this-&gt;send_data('.', '250');" call, with the server responding back with a 553 code instead of a 250 code.

Original thread: [smtp fails with zoho.](https://community.mybb.com/thread-207804.html)

### LINKED ISSUE #3984: SMTP return code upon message sending completion should be checked
Community thread: https://community.mybb.com/thread-224905-post-1349025.html#pid1349025

Relevant code that will trigger errors:
https://github.com/mybb/mybb/blob/4c5451a241fabdfd22eff085a9336e3373deb64d/sendthread.php#L215
`./sendthread.php` - send thread to a friend:
`$from`: `$mybb->input['fromemail']` from `$mybb->user['email']` for logged in user
`$return_email`: `$mybb->input['fromemail']` from `$mybb->user['email']` for logged in user
The mail will be sent from [b]a user-specified email address[/b] which confronts mail SMTP service's requirement, like [Zoho's](https://www.zoho.com/mail/help/zoho-smtp.html#alink4) requiring `from` is the same as sending account's email address.

Still for Zoho, setting a different `Return Email` other than SMTP username will result in no mail delivered and MyBB frequently set Return Email as `from:` field if it's empty.

So in the above circumstances, Zoho's SMTP authentication will pass but when the full message is sent with a final ending `.`, the server will probably return a 553 code `553 Relaying disallowed as xxx@yyy.zzz`, which MyBB doesn't take care of.

yuliu:
… handler, which should be 250.

Resolves #3918.
Resolves #2644.

### DIFF
--- a/inc/mailhandlers/smtp.php
+++ b/inc/mailhandlers/smtp.php
@@ -247,7 +247,10 @@ class SmtpMail extends MailHandler
 				return false;
 			}
 
-			$this->send_data('.', 250);
+			if(!$this->send_data('.', 250))
+			{
+				$this->fatal_error("Mail may not delivered. Reason: ".$this->get_error());
+			}
 
 			if(!$this->keep_alive)
 			{
--- a/inc/mailhandlers/smtp.php
+++ b/inc/mailhandlers/smtp.php
@@ -249,7 +249,7 @@ class SmtpMail extends MailHandler
 
 			if(!$this->send_data('.', 250))
 			{
-				$this->fatal_error("Mail may not delivered. Reason: ".$this->get_error());
+				$this->fatal_error("Mail may not be delivered. Reason: ".$this->get_error());
 			}
 
 			if(!$this->keep_alive)


inc/mailhandlers/smtp.php
```diff
@@ -247,7 +247,10 @@ function send()
 				return false;
 			}
 
-			$this->send_data('.', 250);
+			if(!$this->send_data('.', 250))
+			{
+				$this->fatal_error("Mail may not delivered. Reason: ".$this->get_error());
```
euantorano, line 252: I think the wording for this could be improved:

```
Mail may not be delivered. Reason: {$this->get_error()}
```

Otherwise, this looks like a nice simple enough change.

yuliu:
Setting an additional check may result in false positive reports, if SMTP server doesn't respect RFC or standards and thus doesn't respond a 250 code to clients. Anyway the mailing should be fine as mails get delivered in such circumstance.

inc/mailhandlers/smtp.php
yuliu: Yep it's pretty simple and obvious that a code 250 is expected after the message is accepted by server.

Had a test with Zoho SMTP service and if SMTP configurations and mailing fields are correctly set according to Zoho's requirement, mails deliver flawless; if `from:` is set to a sender other than its required user for authentication, an error with 553 message is logged in System Mail Log.

Eldenroot:
So does it fix ZOHO email issues? Because I had to switched from Zoho to another provider 3 years ago (it was not working fine).

yuliu:
> So does it fix ZOHO email issues? Because I had to switched from Zoho to another provider 3 years ago (it was not working fine).

No it doesn't. What this PR fixes is that no error is logged if mailing using SMTP fails upon message received but rejected by SMTP server.

However I do have some idea for mitigating Zoho's harsh requirement. The barrier for this mitigation at now is MyBB's lacking of hooks intervening with mail handlers. That'll probably be reported in a separated issue or community thread.

PR #3991, opened 2020-05-10
Fixes #3990 - Invisible Captcha is broken sendthread

### LINKED ISSUE #3990: Invisible Captcha is broken sendthread
When using "Send Thread to a Friend" (`sendthread.php`), the invisible Captcha (e.g. reCaptcha) cannot be bound to the submit button.
Basically the same issue & fix as https://github.com/mybb/mybb/issues/3915.

n-thumann:
Resolves #3990

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -14138,7 +14138,7 @@ $(function() {
 </div>
 </body>
 </html>]]></template>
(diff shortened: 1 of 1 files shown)


Eldenroot:
Applied to my board, it works (and it was fine before as well), so if it was bugged I didnt have a chance to see any report from my users... anyway this should be merged :)

PR #3992, opened 2020-05-10
Multipage hooks

### LINKED ISSUE #3982: Add hooks for multipage
We should add hooks where we use the multipage function.
Adding these hooks allows to edit the page link including extra query parameters.

chack1172:
Resolves #3982 

I added the 2 hooks in forumdiplay and showthread. Should I add it before all multipage functions?

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1038,7 +1038,14 @@ function multipage($count, $perpage, $page, $url, $breadcrumb=false)
 		return '';
 	}
 
-	$plugins->run_hooks('multipage', compact('count', 'perpage', 'page', 'url', 'breadcrumb'));
+	$args = array(
+		'count' => &$count,
+		'perpage' => &$perpage,
+		'page' => &$page,
+		'url' => &$url,
+		'breadcrumb' => &$breadcrumb,
+	);
+	$plugins->run_hooks('multipage', $args);
 
 	$page = (int)$page;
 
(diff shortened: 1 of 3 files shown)


JordanMussi:
I think it’s better to add a hook in the multipage function itself https://github.com/mybb/mybb/blob/127418efedffb189151edc1a9a065679a9811e94/inc/functions.php#L1032

Plugins can then use `THIS_SCRIPT` and other variable to determine what page it is being used on. That way the hook is easily available everywhere pagination is used.

euantorano:
I would agree for definite. Most plugins trying to change any one multi page link would likely want to change them all.

chack1172:
You are right, I didn't add it in the function because it is also called in breadcrumb.
Also they can't access the function parameters with the global keyword.

euantorano:
You can always pass the function parameters to the hook, as an array:

```php
$hook_parameters = array(
    'count' => &$count,
    'perpage' => &$perpage,
    'page' => &$page,
    'url' => &$url,
    'breadcrumb' => &$breadcrumb,
);
```

chack1172:
Ok, I pushed the new edits

inc/functions.php
```diff
@@ -1031,13 +1031,22 @@ function redirect($url, $message="", $title="", $force_redirect=false)
...
 		return '';
 	}
 
+	$args = array(
+		'count' => &$count,
+		'perpage' => &$perpage,
+		'page' => &$page,
+		'url' => &$url,
+		'breadcrumb' => &$breadcrumb,
+	);
```
Sama34, line 1047: Maybe one new line here? I can't remember if that is in the coding standards.

euantorano: Yeah, that's not defined in the coding standards, I'm happy without one really.

PR #3994, opened 2020-05-12
Resolves #3889 - username with an apostrophe does not parse correctly in quotes

### LINKED ISSUE #3889: username with an apostrophe does not parse correctly in quotes
as the title states, if a user has an apostrophe in their username, when someone quotes them, mybb does not parse their username correctly.

clean install of mybb 1.8.22:

Original thread: [username with an apostrophe does not parse correctly in quotes](https://community.mybb.com/thread-226389.html)

Ben-MyBB:
Note: I'm not sure if this is the correct way to resolve this but I wanted to have a quick play about and the change suggested does fix the issue but I can't comment on the implications of such a fix.

Resolves #3889

### DIFF
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -227,7 +227,7 @@ $(function ($) {
 				$cite.remove();
 
 				content = this.elementToBbcode(element);
-				author = '=' + author.replace(/(^\s+|\s+$)/g, '');
+				author = '=' + author.replace(/(^\s+|\s+$)/g,);
 
 				$elm.prepend($cite);
 			}
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -227,7 +227,11 @@ $(function ($) {
 				$cite.remove();
 
 				content = this.elementToBbcode(element);
+<<<<<<< Updated upstream
 				author = '=' + author.replace(/(^\s+|\s+$)/g,);
+=======
+				author = '=' + author.replace(/(^\s+|\s+$)/g);
+>>>>>>> Stashed changes
 
 				$elm.prepend($cite);
 			}
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -227,11 +227,7 @@ $(function ($) {
 				$cite.remove();
 
 				content = this.elementToBbcode(element);
-<<<<<<< Updated upstream
-				author = '=' + author.replace(/(^\s+|\s+$)/g,);
-=======
-				author = '=' + author.replace(/(^\s+|\s+$)/g);
->>>>>>> Stashed changes
+				author = '=' + author.replace(/(^\s+|\s+$)/g, '');
 
 				$elm.prepend($cite);
 			}
@@ -259,7 +255,7 @@ $(function ($) {
 			return '<blockquote' + data + '>' + content + '</blockquote>';
 		},
 		quoteType: function (val, name) {
-			return "'" + val.replace("'", "\\'") + "'";
+			return "'" + val.replace("'", "\") + "'";
 		},
 		breakStart: true,
 		breakEnd: true


euantorano:
This is a problem that has bugged me for an extremely long time as it also breaks MyAlerts, it would be very nice to fix it. I'll test this PR ASAP.

jscripts/bbcodes_sceditor.js
```diff
@@ -227,7 +227,7 @@ $(function ($) {
 				$cite.remove();
 
 				content = this.elementToBbcode(element);
-				author = '=' + author.replace(/(^\s+|\s+$)/g, '');
+				author = '=' + author.replace(/(^\s+|\s+$)/g,);
```
Matslom, line 230: That's wont work. First of all now it will replace all catched chars to 'undefined', cuz you don't provide replacement as second argument. Secondly, that regexp matching only whitespaces at begining and end of the word. There is no replecement for apostrophe. I think, that this is wrong place for fix.

Ben-MyBB: Strange, it works fine as a fix on my local host, I guess the , should be removed before merging.

Matslom: There's still issue when you quoting by pressing "reply" button to post. Replying through "quote" seems to work.
image

Matslom: And the issue with undefined when you add space before/after username in quote 
image

Ben-MyBB: @Matslom Thanks. Removing the second arguement solves the undefined issue. Still looking into the issue with the reply button.

Matslom: But now, it will not remove whitespaces on beginning and end. You should revert your entire changes.

Ben-MyBB: @Matslom Do you plan on submitting a new PR?

Ben-MyBB:
I have updated the method used in this PR. White-spaces are now escaped at the beginning and end of the username.

Eldenroot:
Does it affect only SCeditor? I am going to apply to my test board...

jscripts/bbcodes_sceditor.js
Eldenroot: @Matslom - please check the latest changes :)

Matslom: @Eldenroot Removing whitespaces is ok now. I didn't checked if escaped apostrophe is fixed.

Eldenroot:
I can confirm, works fine (already discussed with Ben on Discord)

Ben-MyBB:
@JordanMussi If you have the time to test this one that would be greatly appreciated.

JordanMussi:
This doesn't fix the problem I don't think and it is a syntax error as the `\"` has escaped the `"` so the string is not properly closed.

---

When I quote a post using multiquote everything works as the author is wrapped with `"` instead of `'`

https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/xmlhttp.php#L755

https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/inc/functions_posting.php#L217-L223

It uses `"` by default unless there is a `"` in the username. So there is clearly a problem if there is both a `"` and `'` in a username. I think best to leave that as is because will probably need quite a few changes to the parser.

So all that needs changing to fix the quote for when replying is to use the `"` quotes by default and not do the escaping weirdness.

Looks like this change might be what is needed:
```diff
diff --git a/jscripts/bbcodes_sceditor.js b/jscripts/bbcodes_sceditor.js
index 19426d590..4b4101bdd 100644
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -255,7 +255,9 @@ $(function ($) {
                        return '<blockquote' + data + '>' + content + '</blockquote>';
                },
                quoteType: function (val, name) {
-                       return "'" + val.replace("'", "\\'") + "'";
+                       if (val.indexOf('"') !== -1)
+                               return "'" + val + "'";
+                       return '"' + val + '"';
                },
                breakStart: true,
                breakEnd: true
```

Anyone else have any thoughts?

euantorano:
I think that looks like a good solution.

I thought it would be worth summarising all of the different scenarios we can think of and seeing what the current behaviour is in SCEditor.

| Scenario | Input | Output |
| -------- | ----- | ------ |
| Write quote manually in source mode, then switch to preview, then back to source | ```[quote=euan]test[/quote]``` | ```[quote='euan']test[/quote]``` |
| Write quote manually in source mode, then switch to preview, then back to source | ```[quote="euan"]test[/quote]``` | ```[quote='euan']test[/quote]``` |
| Write quote manually in source mode, then switch to preview, then back to source | ```[quote="euan's test"]test[/quote]``` | ```[quote='euan\'s test']test[/quote]``` |
| Write quote manually in source mode, then switch to preview, then back to source | ```[quote='euan\'s test']test[/quote]``` | ```[quote='euan\'s test']test[/quote]``` |
| `Reply` button on post by `euan's test` | | ```[quote='euan\'s test' pid='3' dateline='1591456452']Reply[/quote]```
| `Reply` button on post by `euan` | | ```[quote='euan' pid='2' dateline='1591453550']Reply[/quote]```
| `Reply` button on post by `euan` | | ```[quote='euan' pid='2' dateline='1591453550']Reply[/quote]```
| Using the multi-quote functionality and inserting the quotes in quick reply | ```[quote="euan's test" pid='3' dateline='1591456452']Reply[/quote]``` | ```[quote='euan\'s test' pid='3' dateline='1591456452']Reply[/quote]```

So it seems SCEditor always uses `'` even when you tap `"`, and if there is a `'` in the username it adds a `\`. The parser doesn't understand the `\` escape and the following input:

```
[quote='euan\'s test']
escaped single quote
[/quote]

[quote="euan's test"]
double quote
[/quote]
```

Results in:

Screenshot 2020-06-06 at 16 23 47

Ben-MyBB:
#4053

PR #4003, opened 2020-05-21
Resolves: #3630 - implementation of reCAPTCHA v3

### LINKED ISSUE #3630: Enhancement: reCAPTCHA v3
Request:
Can Google reCAPTCHA v3 be integrated, please?

Ben-MyBB:
Note: This is a draft since I haven't had a chance to test this yet but would welcome reviews from anyone.

Resolves #3630


install/resources/mybb_theme.xml
```diff
@@ -9511,6 +9511,39 @@ if(use_xmlhttprequest == "1")
 	</tr>
 </table>
 </td>
+</tr>]]></template>
+<template name="post_recaptcha_v3" version="1823"><![CDATA[<tr id="captcha_trow">
```
effone, line 9515: Invalid template name. Template name should start with `post_captcha_`. Thats a requirement.
So, the template name should be `post_captcha_recaptcha_v3`.

member.php
```diff
@@ -2028,7 +2028,7 @@
 				$captcha = $login_captcha->build_hidden_captcha();
 			}
 		}
-		elseif(in_array($login_captcha->type, array(2, 4, 5)))
+		elseif(in_array($login_captcha->type, array(2, 4, 5, 8)))
```
effone, line 2031: Type 2 is long gone. I guess its an unintentional left out. Please change it to `array(4, 5, 8)`

effone (changes requested):
Looking at the codes I believe we don't need a new template for v3. Our `post_captcha_recaptcha_invisible` template is capable enough to handle V3.

So please remove all the template inclusions and place `post_captcha_recaptcha_invisible` for type 8 in class_captcha.

install/resources/mybb_theme.xml
effone: Oh Wait!!! New template is not required, I believe ...

contact.php
```diff
@@ -11,7 +11,7 @@
 define("IN_MYBB", 1);
 define('THIS_SCRIPT', 'contact.php');
 
-$templatelist = "contact,post_captcha,post_captcha_recaptcha_invisible,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha";
+$templatelist = "contact,post_captcha,post_captcha_recaptcha_invisible,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_recaptcha_v3";
```
effone, line 14: Change unnecessary if we reuse old template...

inc/class_captcha.php
```diff
@@ -122,10 +123,14 @@ function __construct($build = false, $template = "")
 			{
 				$this->captcha_template .= "_hcaptcha_invisible";
 			}
+			elseif($this->type == 8)
+			{
+				$this->captcha_template .= "_recaptcha_v3";
```
effone, line 128: Make this:
```

			elseif($this->type == 8)
			{
				$this->captcha_template .= "_recaptcha_invisible";
			}
```
to reuse old template.

newreply.php
```diff
@@ -15,7 +15,7 @@
 $templatelist .= ",codebuttons,post_attachments_new,post_attachments,post_savedraftbutton,newreply_modoptions,newreply_threadreview_more,postbit_online,postbit_pm,newreply_disablesmilies_hidden,post_attachments_update";
 $templatelist .= ",postbit_warninglevel,postbit_author_user,postbit_edit,postbit_quickdelete,postbit_inlinecheck,postbit_posturl,postbit_quote,postbit_multiquote,newreply_modoptions_close,newreply_modoptions_stick";
 $templatelist .= ",post_attachments_attachment_postinsert,post_attachments_attachment_remove,post_attachments_attachment_unapproved,post_attachments_attachment,post_attachments_viewlink,postbit_attachments_attachment,newreply_signature";
-$templatelist .= ",post_captcha_recaptcha_invisible,post_captcha_hidden,post_captcha,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_javascript,postbit_groupimage,postbit_attachments,newreply_postoptions";
+$templatelist .= ",post_captcha_recaptcha_invisible,post_captcha_hidden,post_captcha,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_recaptcha_v3,post_javascript,postbit_groupimage,postbit_attachments,newreply_postoptions";
```
effone, line 18: Change unnecessary if we reuse old template...

newthread.php
```diff
@@ -14,7 +14,7 @@
 $templatelist = "newthread,previewpost,loginbox,changeuserbox,newthread_postpoll,posticons,codebuttons,postbit,post_attachments_attachment_unapproved,newreply_modoptions_close,newreply_modoptions_stick";
 $templatelist .= ",newthread_disablesmilies,post_attachments_new,post_attachments,post_savedraftbutton,post_subscription_method,post_attachments_attachment_remove,postbit_warninglevel_formatted,postbit_icon";
 $templatelist .= ",forumdisplay_rules,forumdisplay_rules_link,post_attachments_attachment_postinsert,post_attachments_attachment,newthread_signature,post_prefixselect_prefix,post_prefixselect_single,posticons_icon";
-$templatelist .= ",post_captcha_hidden,post_captcha_recaptcha_invisible,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_javascript,postbit_gotopost,newthread_postoptions,post_attachments_add,post_attachments_viewlink";
+$templatelist .= ",post_captcha_hidden,post_captcha_recaptcha_invisible,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_recaptcha_v3,post_javascript,postbit_gotopost,newthread_postoptions,post_attachments_add,post_attachments_viewlink";
```
effone, line 17: Change unnecessary if we reuse old template...

sendthread.php
```diff
@@ -11,7 +11,7 @@
 define("IN_MYBB", 1);
 define('THIS_SCRIPT', 'sendthread.php');
 
-$templatelist = "sendthread,sendthread_fromemail,forumdisplay_password_wrongpass,forumdisplay_password,post_captcha,post_captcha_nocaptcha,post_captcha_hcaptcha";
+$templatelist = "sendthread,sendthread_fromemail,forumdisplay_password_wrongpass,forumdisplay_password,post_captcha,post_captcha_nocaptcha,post_captcha_hcaptcha,post_recaptcha_v3";
```
effone, line 14: Change unnecessary if we reuse old template...

showthread.php
```diff
@@ -19,7 +19,7 @@
...
 $templatelist .= ",showthread_moderationoptions_openclose,showthread_moderationoptions_stickunstick,showthread_moderationoptions_delete,showthread_moderationoptions_threadnotes,showthread_moderationoptions_manage,showthread_moderationoptions_deletepoll";
 $templatelist .= ",postbit_userstar,postbit_reputation_formatted_link,postbit_warninglevel_formatted,postbit_quickrestore,forumdisplay_password,forumdisplay_password_wrongpass,postbit_purgespammer,showthread_inlinemoderation_approve,forumdisplay_thread_icon";
-$templatelist .= ",showthread_moderationoptions_softdelete,showthread_moderationoptions_restore,post_captcha,post_captcha_recaptcha_invisible,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,showthread_moderationoptions,showthread_inlinemoderation_standard,showthread_inlinemoderation_manage";
+$templatelist .= ",showthread_moderationoptions_softdelete,showthread_moderationoptions_restore,post_captcha,post_captcha_recaptcha_invisible,post_captcha_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_recaptcha_v3,showthread_moderationoptions,showthread_inlinemoderation_standard,showthread_inlinemoderation_manage";
```
effone, line 22: Change unnecessary if we reuse old template...

Ben-MyBB:
Thanks @effone. All changes applied.

Ben-MyBB:
I'm still testing this in further depth right now but it has so far been working as expected.

effone (approved):

install/resources/settings.xml
```diff
@@ -167,7 +167,8 @@
 4=NoCAPTCHA reCAPTCHA
 5=reCAPTCHA invisible
 6=hCaptcha
-7=hCaptcha invisible]]></optionscode>
+7=hCaptcha invisible
+8=reCAPTCHA V3]></optionscode>
```
dvz, line 171: Lowercase `v3`?

JordanMussi: Also lost a `]` from `]]>`

JordanMussi:
```suggestion
8=reCAPTCHA v3]]></optionscode>
```

install/resources/mybb_theme.xml
```diff
@@ -9512,6 +9512,7 @@ if(use_xmlhttprequest == "1")
 </table>
 </td>
 </tr>]]></template>
+
```
JordanMussi, line 9515:
```suggestion
```

JordanMussi (changes requested):

install/resources/settings.xml
```diff
@@ -167,7 +167,8 @@
 4=NoCAPTCHA reCAPTCHA
 5=reCAPTCHA invisible
 6=hCaptcha
-7=hCaptcha invisible]]></optionscode>
+7=hCaptcha invisible
+8=reCAPTCHA v3></optionscode>
```
JordanMussi, line 171: @Ben-MyBB it does need this to close the `<![CDATA[`

```suggestion
8=reCAPTCHA v3]]></optionscode>
```

JordanMussi (changes requested):

effone:
Is an upgrade script entry required to inject new settings in DB? @euantorano

euantorano:
Nope, they should just be added automatically by the upgrade script.

JordanMussi:
@effone what makes you think that the reCAPTCHA v2 invisible template works for v3 as well? 

I'm struggling to make sense of if that's right or not looking at https://developers.google.com/recaptcha/docs/v3 and https://developers.google.com/recaptcha/docs/invisible

<details>
<summary>The documentation for v3 suggests loading like:</summary>

```html
<script src="https://www.google.com/recaptcha/api.js?render=reCAPTCHA_site_key"></script>
<script>
function onClick(e) {
    e.preventDefault();
    grecaptcha.ready(function() {
        grecaptcha.execute('reCAPTCHA_site_key', {action: 'submit'}).then(function(token) {
            // Add your logic to submit to your backend server here.
        });
    });
}
</script>
```
</details>
<details>
<summary>Whereas the v2 invisible template does:</summary>

```html
<script src="//www.google.com/recaptcha/api.js?onload=captchaOnloadCallback&render=explicit" async="" defer=""></script>
<script type="text/javascript">
	var captchaOnSubmit = function(token) {
		$("#registration_form").trigger('submit');
	};
	var captchaOnloadCallback = function() {

		var target = $('input[name="regsubmit"]');
		target.attr('data-size', 'invisible');

		grecaptcha.render(target.get(0), {
				'sitekey': 'reCAPTCHA_site_key',
				'callback': captchaOnSubmit
			}, true);
	}
</script>
```
</details>

So when I tested this PR I seemed to just be getting the v2 invisible (but I may be wrong and that's how v3 appears on the page too).

Anyway the actual validation for v3 needs to take into account the `score` returned not just `success=true`: https://developers.google.com/recaptcha/docs/v3#site_verify_response

---

I think it would be beneficial to rename the setting options to reflect how the reCAPTCHA site refers to the different versions/methods and simplify the order:
```xml
<optionscode><![CDATA[select
0=No CAPTCHA
1=MyBB Default CAPTCHA
4=reCAPTCHA v2 checkbox
5=reCAPTCHA v2 invisible
8=reCAPTCHA v3
6=hCaptcha
7=hCaptcha invisible]]></optionscode>
```

dvz:
> I think it would be beneficial to rename the setting options to reflect how the reCAPTCHA site refers to the different versions/methods and simplify the order

We can also switch to string keys in 1.9 or 1.10.

Ben-MyBB:
@JordanMussi I think if we do want to change the order that should be done in a seperate PR since it will require other changes and this is aimed at implementing reCaptcha v3. That being said, if someone is willing they are more than welcome to force push to my branch since I think it would be best if someone else made the changes :)

JordanMussi:
I'm not sure it'll need anything else as it will only change the display order of the dropdown none of the behaviour.

Ben-MyBB:
> 
> 
> I'm not sure it'll need anything else as it will only change the display order of the dropdown but not any of the behaviour.

Interesting, I'll have a play about with it then and if successful push the changes.

effone:
@JordanMussi what makes you think that the reCAPTCHA v2 codes will also work for v3?
I have tested the response from google and it is returning the score in case of v3.
`{ "success": true, "challenge_ts": "2020-06-06T16:40:57Z", "hostname": "dev.eff.one", "score": 0.9 }`

I agree that the score is also required to be taken in account. For that simple validation in class_captcha.php will work.

Do you think we need a setting for acceptable score? Or we are gonna accept any score greater than pre-defined value that we are gonna hard-code?

JordanMussi:
Ah that template stuff is fine then - I was just getting confused between the instructions on the documentation for the two versions and when I inspected the output element when testing it was using resources in a `/api2/` path which added to my confusion!

> Do you think we need a setting for acceptable score? Or we are gonna accept any score greater than pre-defined value that we are gonna hard-code?

I think it should be a new setting as the documentation refers to tweaking it after watching the reports on the dashboard.

---

> For this reason, we recommend including reCAPTCHA verification on forms or actions as well as in the background of pages for analytics.
> &mdash; <cite>[reCAPTCHA v3](https://developers.google.com/recaptcha/docs/v3)</cite>

Not sure how we should handle this requirement - should the JS be run on every page for guests so it can get the amount of data it requires?

effone:
>Not sure how we should handle this requirement - should the JS be run on every page for guests so it can get the amount of data it requires?

Days ago when I looked at V3 I thought it is required to include the CAPTCHA in every page, that was the only reason I didn't make a PR for MyBB til now.
I guess including CAPTCHA for required form submissions is sufficient for MyBB, as we are doing ...

effone:
Note: Reconfirmed checking response value of V2 invisible. There is no score.
`{ "success": true, "challenge_ts": "2020-06-06T20:03:17Z", "hostname": "dev.eff.one" }`

Our existing template is good enough for V3.

Ben-MyBB:
@effone the documentation for v3 isn't the best at all. In terms of the acceptable value between 0 - 1 I plan on offering the options in a drop down menu unless there is any objections.

I have pushed a commit for what I think should be enough for sending the score back to Google

effone:
Don't know if I am getting the flow wrong, but why we bother to send the score setting to Google? If Google is returning the score we can simply compare that with MyBB setting internally and decide to allow the user or not.
@JordanMussi ?

Ben-MyBB:
> Don't know if I am getting the flow wrong, but why we bother to send the score setting to Google? If Google is returning the score we can simply compare that with MyBB setting internally and decide to allow the user or not.
> 
> @JordanMussi ?

Seems much easier to handle that way IMHO. The documentation for v3 is pretty bad and doesn’t explain anything properly.

inc/class_captcha.php
```diff
@@ -307,6 +312,40 @@ function validate_captcha()
...
+
+				if($response == false)
+				{
+					$this->set_error($lang->invalid_nocaptcha_transmit);
+				}
+				else
+				{
+					$answer = json_decode($response, true);
+
+					if($answer['success'] != 'true')
```
JordanMussi, line 341: Yeah that’s right - at this point compare the setting with `$answer[‘score’]` to decide the bottyness/humanness :-)

JordanMussi (changes requested):

admin/modules/config/settings.php
```diff
@@ -961,6 +961,8 @@
...
 		   || (in_array($mybb->settings['captchaimage'], array(4, 5)) && (!$mybb->settings['recaptchaprivatekey'] || !$mybb->settings['recaptchapublickey']))
+		   || (isset($mybb->input['upsetting']['captchaimage']) && in_array($mybb->input['upsetting']['captchaimage'], array(8)) && (!$mybb->input['upsetting']['recaptchaprivatekey'] || !$mybb->input['upsetting']['recaptchapublickey'] || !$mybb->input['upsetting']['recaptchascore']))
+		   || (in_array($mybb->settings['captchaimage'], array(8)) && (!$mybb->settings['recaptchaprivatekey'] || !$mybb->settings['recaptchapublickey'] || !$mybb->settings['recaptchascore']))
 		   || (isset($mybb->input['upsetting']['captchaimage']) && in_array($mybb->input['upsetting']['captchaimage'], array(6, 7)) && (!$mybb->input['upsetting']['hcaptchaprivatekey'] || !$mybb->input['upsetting']['hcaptchapublickey']))
 		   || (in_array($mybb->settings['captchaimage'], array(6, 7)) && (!$mybb->settings['hcaptchaprivatekey'] || !$mybb->settings['hcaptchapublickey'])))
 		{
```
euantorano, line 968: Not related to this PR as such, but this is one wild `if` branch. Would be a good idea to try and clean this up in a. future release.

inc/class_captcha.php
```diff
@@ -307,6 +312,40 @@ function validate_captcha()
 				}
 			}
 		}
+		elseif(in_array($this->type, array(8)))
```
euantorano, line 315: Not much point in using `in_array` with a single element IMO.

```suggestion
		elseif($this->type == 8)
```

inc/class_captcha.php
```diff
@@ -307,6 +312,40 @@ function validate_captcha()
...
+				// We have a reCAPTCHA invisible to handle
+				// Contact Google and see if our reCAPTCHA was successful
+				$response = fetch_remote_file($this->verify_server, array(
+					'secret' => $mybb->settings['recaptchaprivatekey'],
+					'score' => $mybb->settings['recaptchascore'],
+					'remoteip' => $session->ipaddress,
+					'response' => $response
+				));
+
+				if($response == false)
```
euantorano, line 333: We'd normally prefer strict equality comparison in this case.

```suggestion
				if($response === false)
```

Ben-MyBB:
I will open a seperate issue shortly to track that

newthread.php
```diff
@@ -1037,7 +1037,7 @@
 				$post_captcha->build_hcaptcha();
 			}
 		}
-		else if($correct && (in_array($post_captcha->type, array(4, 5))))
+		else if($correct && (in_array($post_captcha->type, array(4, 5, 8))))
```
euantorano, line 1040: Not exactly related to this PR, but I think we should introduce some constants/static variables to define these captcha types rather than having these types of checks scattered throughout the codebase.

euantorano (changes requested):
I think we're almost ready to merge. Just a few minor comments' nd I believe the upgrade script needs updating, similar to the `upgrade41.php` file: https://github.com/mybb/mybb/blob/5a836f408162dbf981b94c49e9204ad64280ef7a/install/resources/upgrade41.php#L70

Note that hCaptcha seems to have missed doing this too, so we need to fix that too - IMO would be OK to just include that in this PR to prevent conflicts.

newthread.php
Ben-MyBB: That makes much more sense and I will include in the issue I open to track the other code improvement.

JordanMussi:
> I believe the upgrade script needs updating, similar to the `upgrade41.php` file: https://github.com/mybb/mybb/blob/5a836f408162dbf981b94c49e9204ad64280ef7a/install/resources/upgrade41.php#L70
> Note that hCaptcha seems to have missed doing this too, so we need to fix that too - IMO would be OK to just include that in this PR to prevent conflicts.

Upgrade process worked fine for me without this change.

The `optionscode` gets updated on upgrade using the `settings.xml` file: https://github.com/mybb/mybb/blob/1326eea040ef2637ac58c92512f9b6f7dd7696bb/install/upgrade.php#L954

euantorano:
Interesting. We have several other upgrade scripts that repeat the pattern I linked above - I wonder why we do that then...

Ben-MyBB:
Do you want me to revert that change or leave as is?

euantorano:
I’ll do some testing with and without it to confirm and get back to you.

n-thumann:
I just tested this PR: Both the Captcha itself and the upgrade script worked fine for me 👍

JordanMussi:
Thank you @n-thumann ! 😀 

@Ben-MyBB I think we are good for this PR to land, just waiting for Euan to assess the upgrade script query.

euantorano:
Yep, will be done double checking either tonight or tomorrow and merged tomorrow at the latest.

euantorano (approved):

JordanMussi:
What was your decision ref the optionscode upgrade query?

euantorano:
@JordanMussi I decided that since we're doing it in other upgrade scripts, I prefer to be consistent in that behaviour. It's not required, but I'd rather make sure we keep sticking to one approach for now.

JordanMussi:
Ok happy with that, probably the last time it’ll happen in 1.8.x anyway...

euantorano:
Unless some other big new service comes around, yeah.

PR #4011, opened 2020-05-26
Resolves #3781 - Long avatar URL causes MySQL error

### LINKED ISSUE #3781: Long avatar URL causes MySQL error

Ben-MyBB:
Resolves #3781


usercp.php
```diff
@@ -2485,6 +2485,12 @@
 				}
 			}
 
+			// Limiting URL string to stay within database limit
+			if(strlen($mybb->input['avatarurl']) > 200)
+			{
+					$avatar_error = $lang->error_avatarurltoolong;
```
Eldenroot, line 2491: formatting :)

inc/languages/english/usercp.lang.php
```diff
@@ -299,6 +299,7 @@
 $l['attachstats_quota'] = "Attachment Quota";
 $l['attachstats_totaldl'] = "Total Downloads";
 $l['attachstats_bandwidth'] = "Approximate Bandwidth Usage";
+$l['error_avatarurltoolong'] = "Sorry, but we cannot change your avatar as the new avatar URL specified is too long. You can try to use a shorter URL which doesn't exceed the database field length.";
```
euantorano, line 302: I think this is better worded without mentioning a DB field as that could confuse some users:

```suggestion
$l['error_avatarurltoolong'] = "Sorry, but we cannot change your avatar as the new avatar URL specified is too long. You can try to use a shorter URL which is less than 200 characters long.";
```

Ben-MyBB:
Thanks @euantorano and @Eldenroot . Both updated plus a missing full-stop in another language string I spotted.

Issue #3753, opened 2019-07-22
ICQ Number/Yahoo ID/Skype ID/Google Hangouts ID can exceed database field


Ben-MyBB:
The values of ICQ Number, Yahoo ID, Skype ID and Google Hangouts ID are not validated (beside casting ICQ to integer) so you can easily exceed the length of the associated database fields:
Quote:SQL Error: 1406 - Data too long for column 'yahoo' at row 1

Original thread: [ICQ Number/Yahoo ID/Skype ID/Google Hangouts ID can exceed database field](https://community.mybb.com/thread-223686.html)

Eldenroot:
This can be solved together with #3781

Ben-MyBB:
Seems there is a feature to limit new custom profile fields so this will only apply to the default ones. I plan on taking a look at this.

Eldenroot:
@Ben-MyBB you are on fire 🔥 👍

PR #4012, opened 2020-05-26
Resolves #3771 - Seperate variables for calendar.php

### LINKED ISSUE #3771: add another variable with a unique style
**calendar**
template: calendar_weekrow_day

There is a variable ({$day_class}) which shows 3 styles which are (.trow1, .trow2 and .trow_sep), I ask for a unique variable for the style .trow_sep, which indicates the current day we are on the calendar EXAMPLE:

image

Here's the problem (at least for me, I don't know if there is another solution), I don't want to use the trow1 and trwo2 styles for "x" reason and I just want to use the trow_sep style. But it won't be possible because it will always give me all three styles

Will there be another solution to what I ask? regards

Ben-MyBB:
Marking this as a draft for now until I have had a look at the upgrade script and throughly tested this, but I wanted to open this for feedback on my method. An inital start on resolving #3771. Upgrade script still needs to be considered before merging to insert the additional two templates required.

All feedback is welcomed since I am still learning PHP :)

Resolves #3771

### DIFF
--- a/calendar.php
+++ b/calendar.php
@@ -2548,19 +2548,16 @@ if(!$mybb->input['action'])
 			// Is the current day
 			if($day.$calendar_month.$year == $today && $month == $calendar_month)
 			{
-				$current_day = "trow_sep";
 				eval("\$day_bits .= \"".$templates->get("calendar_weekrow_currentday")."\";");
 			}
 			// Not in this month
 			else if($in_month == 0)
 			{
-				$day_class = "trow1";
 				eval("\$day_bits .= \"".$templates->get("calendar_weekrow_day")."\";");
 			}
 			// Just a normal day in this month
 			else
 			{
-				$this_month = "trow2";
 				eval("\$day_bits .= \"".$templates->get("calendar_weekrow_thismonth")."\";");
 			}
 			$day_birthdays = $day_events = "";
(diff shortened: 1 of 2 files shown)


Ben-MyBB:
Note: I have set the template versions to 1823 but I will need to update them again after the release of .23 if this PR hasn't been merged.

Ben-MyBB:
@Eldenroot can you review?

calendar.php
```diff
@@ -2548,19 +2548,21 @@
 			// Is the current day
 			if($day.$calendar_month.$year == $today && $month == $calendar_month)
 			{
-				$day_class = "trow_sep";
+				$current_day = "trow_sep";
```
euantorano, line 2551: With these separate templates, I wonder if we can get rid of these hard-coded CSS classes in the PHP code and just put the classes in the templates?

Ben-MyBB: I can take a look but I want to get a fix for #3753 first since that's an important one 👍

Ben-MyBB:
@euantorano updated :)

euantorano:
Looks good to me, just need to test.

Eldenroot:
Should be fine, personally I dont use calendar in MyBB, so just tested briefly.

PR #4016, opened 2020-05-28
Resolves #3738 - Allow Link MyCode off still allows links

### LINKED ISSUE #3738: Allow Link MyCode off still allows links
If you turn off " Allow Link MyCode" in Clickable Smilies and BB Code you can still post live links by just typing them in. 

https://community.mybb.com

Surely the point off turning link MyCode off is to disallow links ? 

I have just updated from 1.8.15 to 1.8.21 so I'm not sure how long this has been happening as I was using a MyCode to break links, but that no longer works. 

*I have double checked on a fresh install of 1.8.21 running on PHP 7.2.19 and you can post live links even if Allow Links is turned off. 

Would be grateful if someone could check this out as I disallow live links on my forum. 

Thank you.

Original thread: [Allow Link MyCode off still allows links](https://community.mybb.com/thread-223575.html)

Ben-MyBB:
Checking allow_url setting before automatically parsing URL's

Resolves #3738

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -521,7 +521,10 @@ class postParser
 			}
 		}
 
-		$message = $this->mycode_auto_url($message);
+		if(!empty($this->options['allowautourl']))
+		{
+			$message = $this->mycode_auto_url($message);
+		}
 
 		return $message;
 	}
@@ -1592,6 +1595,11 @@ class postParser
 	{
 		$message = " ".$message;
 
+		if($this->options['allow_url'] != 1)
+		{
+			return $message;
+		}
+		
 		// Links should end with slashes, numbers, characters and braces but not with dots, commas or question marks
 		// Don't create links within existing links (handled up-front in the callback function).
 		$message = preg_replace_callback("#<a\\s[^>]*>.*?</a>|([\s\(\)\[\>])(http|https|ftp|news|irc|ircs|irc6){1}(://)([^\/\"\s\<\[\.]+\.([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))#ius", array($this, 'mycode_auto_url_callback'), $message);
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -13,6 +13,7 @@ options = array(
 	allow_html
 	allow_smilies
 	allow_mycode
+	allow_links
 	nl2br
 	filter_badwords
 	me_username
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -521,7 +521,7 @@ class postParser
 			}
 		}
 
-		if(!empty($this->options['allowautourl']))
+		if($mybb->settings['allowautourl'] == 1)
 		{
 			$message = $this->mycode_auto_url($message);
 		}
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1590,6 +1590,8 @@ class postParser
 	*/
 	function mycode_auto_url($message)
 	{
+		if($this->options['allow_url'] == 1)
+		{
 		$message = " ".$message;
 
 		// Links should end with slashes, numbers, characters and braces but not with dots, commas or question marks
@@ -1599,7 +1601,12 @@ class postParser
 		$message = my_substr($message, 1);
 
 		return $message;
-	}
+		}
+		else
+		{
+			return $message;
+		}
+		}
 
 	/**
 	* Parses URLs automatically.
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -13,7 +13,7 @@ options = array(
 	allow_html
 	allow_smilies
 	allow_mycode
-	allow_links
+	allow_url
 	nl2br
 	filter_badwords
 	me_username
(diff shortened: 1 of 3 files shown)


inc/class_parser.php
```diff
@@ -1590,6 +1590,8 @@ function mycode_parse_video_disabled_callback($matches)
 	*/
 	function mycode_auto_url($message)
 	{
+		if($this->options['allow_url'] == 1)
+		{
 		$message = " ".$message;
```
euantorano, line 1597: Nitpick: this content inside the `if` should be indented.

inc/class_parser.php
```diff
@@ -1599,7 +1601,12 @@ function mycode_auto_url($message)
 		$message = my_substr($message, 1);
 
 		return $message;
-	}
+		}
+		else
+		{
+			return $message;
+		}
+		}
```
euantorano: Nitpick: this closing `}` should be outdented.

```suggestion
	}
```

euantorano (changes requested):
Minor changes to code style.

I'd normally cheat here and simplify this code (and therefore minify the patch, making future rebases easier):

Add this code to the top of the function, and remove the rest of the changes:

```php
if($this->options['allow_url'] != 1)
{
	return $message;
}
```

This results in an early return from the function, and reduces the level of indenting.

Ben-MyBB:
Thanks @euantorano. I will update this PR once I have a clear direction on the way to resolve it. I think that if URL MyCode is disabled then the links shouldn't automatically be parsed, which is what this PR does. @yuliu and @WildcardSearch are of the opinion that a setting should be added for the automatic parsing of URL's. What's your thoughts @mybb/developers ?

Sama34:
Here:
https://github.com/mybb/mybb/blob/feature/inc/class_parser.php#L524

It should be:
```
if(!empty($this->options['allow_auto_url']))
{
	$message = $this->mycode_auto_url($message);
}
```

Then go on to each building of the parser options to update them and add `allow_auto_url`, like in the following line:
https://github.com/mybb/mybb/blob/feature/inc/functions_post.php#L112

Add right after:
```
$parser_options['allow_auto_url'] = $mybb->settings['bbcodesallowautourl'];
```

But this will require either to add a global setting (`filter_badwords` should have a global setting too) or forum specific settings (which will require an update script).

euantorano:
I think:

1. If the URL MyCode is _disabled_, it should never auto-parse them.
2. if the URL MyCode is _enabled_, there should be a setting to disable automatic parsing. I think a site-wide setting would be fine.

Ben-MyBB:
Okay, I can take a look at the comments above and the second point raised.

Sama34:
Are you going to push some changes?

Ben-MyBB:
Yes I am working on them currently

Ben-MyBB:
@Sama34 Note the additional changes pushed, but this still doesn't want to work on my localhost so still trying to figure out why.

P.S now I have finished my PHP course I am shocked by the original commit.

Eldenroot:
@Ben-MyBB is this complete and ready to test?

Sama34:
I don't think this is functional. Maybe you should keep the global setting and check for it just around the following code:
https://github.com/mybb/mybb/blob/864836e3dd5f1ae263e10d8374f482e50f95f266/inc/class_parser.php#L524
```PHP
		if($mybb->settings['allowautourl'] == 1)
		{
			$message = $this->mycode_auto_url($message);
		}
```

I think additional modifications are unnecessary.

Ben-MyBB:
@Eldenroot This PR can be tested after my lastest changes, should be good to go.

inc/class_parser.php
```diff
@@ -1592,6 +1595,11 @@ function mycode_auto_url($message)
 	{
 		$message = " ".$message;
 
+		if($this->options['allow_url'] != 1)
+		{
+			return $message;
+		}
```
Sama34, line 1602: Hi Ben, I think we can keep this, no need to add it to every parser build instance, developers could still use it even if the core doesn't.

However, I think it should be added in the following list:
https://github.com/mybb/mybb/blob/5b6c198abc6a21906363a4a215f075a7e06f2277/inc/class_parser.php#L13-L21

Ideally that piece of code should describe each option, but that is not part of this PR. @euantorano

Sama34 (approved):

euantorano:
Yeah, adding it to the commented option list would be a great idea - good catch!

install/resources/settings.xml
```diff
@@ -2112,66 +2112,74 @@ min=1]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="allowautourl">
+			<title>Automatically parse URL's into links</title>
+			<description><![CDATA[Setting this will disable the Automatic Parsing of URL's.]]></description>
```
dvz:
```suggestion
			<title>Automatically Parse URLs Into Links</title>
			<description><![CDATA[Setting this to Yes will parse URLs as links.]]></description>
```

inc/class_parser.php
Ben-MyBB: Thanks @Sama34 @euantorano I have applied it in the commented note above

euantorano (approved):

PR #4017, opened 2020-05-28
Resolves #3971 - Add "lazy loading" for mycodes

### LINKED ISSUE #3971: Add "lazy loading" for mycodes
Use native lazy-loading for web - https://web.dev/native-lazy-loading/#image-loading in mycodes to improve performance.

Ben-MyBB:
@Eldenroot can you check if this is what you require?

Resolves #3971


Eldenroot:
This should be fine, width and height is defined, Good job!

Eldenroot:
OK, so no far no issue.

dvz:
Should it be hardcoded and enabled by default?

What about the reflow issue (https://github.com/mybb/mybb/issues/3971#issuecomment-622348588)? I imagine most `[img]` images don't have size specified.

euantorano:
It’s not hard coded as such, given that it’s in a template, which is why I merged. I think that in practice reflowing isn’t going to be a major concern unless an image is loading very slowly, and in that case you can easily edit the template.

PR #4019, opened 2020-05-28
Jump Page Input Type Correction

### LINKED ISSUE #4018: Jump Page Input type should be number
The multipage_jump_page input type should be number.
Currently the input type is text that allows to input anything and upon submitting an invalid / out of range value it resets to first page.

Should be changed to number, because:
- We expect only numeric value as page number
- We can define the acceptable range of input to get auto-validated

snap

effone:

Resolves #4018


Eldenroot:
Tested, works fine.

Sama34:
image

PR #4022, opened 2020-05-29
Fix #3376 Change the my_post_key regularly

### LINKED ISSUE #3376: Change the my_post_key regularly
Quote from pre-PR:
>MyBB uses "my_post_key" to prevent forged requests. Ideally this token should be unique for each request, or unique for each session ([see on wikipedia](https://en.wikipedia.org/wiki/Cross-site_request_forgery#Synchronizer_token_pattern)) but in MyBB, the token never changes, ever.
At the same time it's easy for the my_post_key to be leaked. It's not only used in page sourcecode, as a hidden form input field, but also openly in links (Mark Forums Read, Star Rating, etc.) so a simple Copy & Paste operation may inadvertently include it, it's in the browser history, etc.
So here is an attempt to change the my_post_key regularly.
It adds a time-based element to the hash, causing a new code to be generated every 3 hours. By generating older codes in the verification step (It's an older code, sir, but it checks out...) each code may be valid for up to 12 hours. After that, the code is useless.
In terms of usability, this means if you send a post request with a tab that had been open for half a day, it will no longer work since the my_post_key of the old tab had already expired. You have to reload to get a new key.

PR for this:
https://github.com/mybb/mybb/pull/3374

Sama34:
Fix #3376 Change the my_post_key regularly

Original PR: #3374

### Compatibility Note
Use the `verify_post_check()` function to verify anti-CSRF tokens ("`my_post_key`") instead of relying on manual value comparison, which may result in false negatives (_Authorization mismatch_ errors) after this change.


dvz:
- added generation & verification changes from https://github.com/mybb/mybb/pull/3374/files with:
  - interval set to 6 hours (24 hours total - can be adjusted at a later point)
  - changes to the hashed string: `$mybb->settings['internal']['encryption_key']` added to both codes (so that all tokens can be [invalidated during recovery](https://docs.mybb.com/1.8/administration/security/recovery/#invalidate-tokens))

dvz:
This may break static links sent in e-mails and PMs (e.g. https://github.com/mybb/mybb/blob/mybb_1823/inc/datahandlers/post.php#L1249); the language strings will need to be updated to note the links expire, or only point to locations where actions can be initiated.

yuliu:
IMO we could use another hash key for unsubscribing purpose only.

JordanMussi:
`usercp.php?action=addsubscription&tid=<ID>` requires the post key but that isn't necessary as it just displays the form. So if we remove that requirement the email only needs to have a link to addsubscription where the user can select no subscription.

dvz:
Doesn't seem to be implemented with forum subscriptions, and the subscription in not completely removed when selecting _no notification_.
Since it already requires the user to be logged in, it could simply point to `usercp.php?action=forumsubscriptions` and `usercp.php?action=subscriptions`.

JordanMussi:
> Doesn't seem to be implemented with forum subscriptions, and the subscription in not completely removed when selecting _no notification_.

I see your point but it does remove the email notifications.

There's no reason forumsubscriptions could show a form at `usercp.php?action=removesubscription&type=forum&fid=<ID>` when post code isn't specified? (same with threadsubscriptions at `usercp.php?action=removesubscription&tid=<ID>` or adding an unsubscribe option to the addsubscription form if there already is a subscription).


> Since it already requires the user to be logged in, it could simply point to `usercp.php?action=forumsubscriptions` and `usercp.php?action=subscriptions`.

This is probably best but it doesn't immediately show you the right subscription perhaps we'd need to add a filter for the subscription ID?

dvz:
Adding forms with an unsubscribe option would work too, but may need some template and language changes (and forum subscription would apparently need to be a simple on/off radio).

JordanMussi:
If we instead added a new page for both types of subscriptions with a form to submit the removal that would only require adding new templates so there shouldn't be any issues.

The only problem with this is older language packs but we could do a find/replace for `&my_post_key={[0-9]}`

dvz:
Splitting GET and POST at https://github.com/mybb/mybb/blob/mybb_1822/usercp.php#L2088? Might make more sense that modifying `?action=addsubscription` that's designed to adjust notification types, and this approach shouldn't break old links.

Tracking in #4076

JordanMussi:
Yep that's what I getting at before - shall I cook up a PR?

PR #4028, opened 2020-05-29
Fixes #4023 Editing help document/section seems doesn't check input

### LINKED ISSUE #4023: Editing help document/section seems doesn't check input
Seems not a big problem though.

When editing a document without a valid hid= input:
/admin/index.php?module=config-help_documents&amp;action=edit&amp;hid=

or editing a section without a valid sid= input:
/admin/index.php?module=config-help_documents&amp;action=edit&amp;sid=

will show an edit screen without error, whilst in other modules there will normally be shown with an error.

Besides, the page header of editing an existing document has the link of "editing document" without its "hid". Some other modules also have this issue, but I forget which they are exactly, will update this post later.

Original thread: [Editing help document/section seems doesn't check input](https://community.mybb.com/thread-227342.html)

Starpaul20:
Fixes #4023

### DIFF
--- a/admin/modules/config/help_documents.php
+++ b/admin/modules/config/help_documents.php
@@ -441,7 +441,7 @@ if($mybb->input['action'] == "edit")
 
 		$sub_tabs['edit_help_document'] = array(
 			'title'	=> $lang->edit_document,
-			'link'	=> "index.php?module=config-help_documents&amp;action=edit&amp;hid=".$hid,
+			'link'	=> "index.php?module=config-help_documents&amp;action=edit&amp;hid=".$doc['hid'],
 			'description' => $lang->edit_document_desc
 		);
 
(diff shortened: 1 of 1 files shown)


Ben-MyBB:
This works fine. What I did notice is that when updating the Help Documents in the ACP editor the changes aren't actually shown in the front end documents at all. I do recall reading something similiar. I'm going to have a search and I will update this comment.

Ben-MyBB (approved):

yuliu (changes requested):
Looks good and it's worth noting that `$hid` is not defined if request method isn't POST in
https://github.com/mybb/mybb/blob/9feef312dc22abef36dcc1ddb1a703e66d9c2fe1/admin/modules/config/help_documents.php#L428
where `$hid` is defined in
https://github.com/mybb/mybb/blob/9feef312dc22abef36dcc1ddb1a703e66d9c2fe1/admin/modules/config/help_documents.php#L366

I do notice some occurrences like that somewhere else in AdminCP but only remember this one..

PaulBender:
That should fix it

JordanMussi (approved):

PR #4030, opened 2020-05-30
Fixes #3526 ft_min_word_len query every search

### LINKED ISSUE #3526: ft_min_word_len query every search
I'm doing some custom edits in search and found a line is called in the functions_search.php inside the function perform_search_mysql_ft().  

This is the code:

> // Attempt to determine minimum word length from MySQL for fulltext searches
> 	$query = $db->query("SHOW VARIABLES LIKE 'ft_min_word_len';");
> 	$min_length = $db->fetch_field($query, 'Value');
> 	if(is_numeric($min_length))
> 	{
> 		$mybb->settings['minsearchword'] = $min_length;
> 	}
> 	// Otherwise, could not fetch - default back to MySQL fulltext default setting
> 	else
> 	{
> 		$mybb->settings['minsearchword'] = 4;
> 	}

That one query isn't really needed imho and there are better ways to deal with ft_min_word_len setting than a query each time the function is run.  

Consider an alteration where it just uses the setting. Maybe during install of MyBB you ensure that the setting is the minimum value.

Every query saved is nice.

JordanMussi:
Resolves #3526

`perform_search_mysql_ft` now behaves like `perform_search_mysql` by setting the hard coded minimum when `minsearchword` < 0

`minsearchword` has an enforced minimum when updating settings and using fulltext. This minimum is the one provided by the DB and uses the logic that used to be done on every fulltext search.

Upgrade script used to ensure minimum is enforced on upgrade so that searches aren't performed with a value less then the DB minimum.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -968,6 +968,19 @@ if($mybb->input['action'] == "change")
 			$lang->success_settings_updated .= $lang->success_settings_updated_captchaimage;
 		}
 
+		// If using fulltext then enforce minimum word length given by database
+		if(isset($mybb->input['upsetting']['minsearchword']) && $mybb->input['upsetting']['minsearchword'] > 0 && $mybb->input['upsetting']['searchtype'] == "fulltext" && $db->supports_fulltext_boolean("posts") && $db->supports_fulltext("threads"))
+		{
+			// Attempt to determine minimum word length from MySQL for fulltext searches
+			$query = $db->query("SHOW VARIABLES LIKE 'ft_min_word_len';");
+			$min_length = $db->fetch_field($query, 'Value');
+			if(is_numeric($min_length) && $mybb->input['upsetting']['minsearchword'] < $min_length)
+			{
+				$mybb->input['upsetting']['minsearchword'] = $min_length;
+				$lang->success_settings_updated .= $lang->success_settings_updated_minsearchword;
+			}
+		}
+
 		// Get settings which optionscode is a forum/group select, checkbox or numeric
 		// We cannot rely on user input to decide this
 		$checkbox_settings = $forum_group_select = $prefix_select = array();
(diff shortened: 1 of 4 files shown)


yuliu:
A pretty good solution IMO. It'd be better that a message will be given upon user input value being replaced in either upgrade script or AdminCP.

JordanMussi:
I thought so as well but I couldn't find any other specific settings validation, only silently enforcing.

I don't believe there is much need for a specific output in upgrade script as it can be a silent setting update because it will not change any behaviour. The setting value actually had no effect previously and the upgrade would only change it if below `ft_min_word_len`.

For example:
||ft_min_word_len|minsearchword|effective minsearchword|
|---|---|---|---|
|Before upgrade|4|3|4|
|After upgrade|4|4|4|

||ft_min_word_len|minsearchword|effective minsearchword|
|---|---|---|---|
|Before upgrade|4|0|4|
|After upgrade|4|0|4|

||ft_min_word_len|minsearchword|effective minsearchword|
|---|---|---|---|
|Before upgrade|4|5|4|
|After upgrade|4|5|5|

yuliu:
Yeah but inform the administrator that the setting of minimum word length
for fulltext search is replaced with X which matches db system's
configuration, when enforcement happened. For example in admincp:

```
$lang->success_settings_updated .= ...
```

JordanMussi:
Okay @yuliu I've gone with that 😄

yuliu:
Looks good.

In my opinion, upgrade script will also notify users about value changes and probably this setting's description should be changed accordingly: `Enter the minimum number of characters an individual word in a search query can be. Set to 0 to disable (and accept the hard limit default of 3 for standard searching and 4 for MySQL fulltext searching). If you use MySQL fulltext searching and set this lower than the MySQL setting - MySQL will override it.` .

Sorry for being nitpicking. 🤣

JordanMussi:
Will do some output in the installer but likely nobody will read it! The setting description is accurate with this new behaviour imo.

JordanMussi:
@yuliu all done 😀

euantorano:
Looks like there are some conflicts here @JordanMussi - do you want to fix them or shall I?

JordanMussi:
That’s done for you @euantorano. Web editor seemed to muck up indentation hopefully I’ve fixed it properly!

PR #4032, opened 2020-05-30
Fixes #3619 Bugs with advanced database configuration

### LINKED ISSUE #3619: Bugs with advanced database configuration
I've found 3 bugs regarding Advanced Database Configuration:

1. Documentation. Despite stating in inc/config.default.php that Docs for advanced database configuration can be found at https://docs.mybb.com, I've failed to find them. The only documentation that I've found is "https://blog.mybb.com/2008/03/18/mybb-14-admin-cp-the-complete-overview-part-3/" but it is for version 1.4 and definitely not appliсable for 1.8.10.

So, the correct configuration to use 2 database servers (1 for write and 1 for read) for version 1.8.10 is ("inc/config/php"):

&#36;config['database']['type'] = 'mysqli';
&#36;config['database']['database'] = 'mybb';
&#36;config['database']['table_prefix'] = 'mybb_';
      
&#36;config['database']['write']['hostname'] = 'hostname_read_server';
&#36;config['database']['write']['username'] = 'mybbuser';
&#36;config['database']['write']['password'] = 'mybbpassword';

&#36;config['database']['read']['hostname'] = 'hostname_write_server';
&#36;config['database']['read']['username'] = 'mybbuser';
&#36;config['database']['read']['password'] = 'mybbpassword';

2. errors in "inc/db_mysqli.php" - function: connect():

line 193: unset(&#36;connections);

The problem is that it deletes all configuration for other Database servers when it has found the configuration for the first one. So it doesn't matter how many Database servers you have defined in config.php it takes only the first one and then deletes info about other database servers. My solution is to use another temporary variable for the following operations so I've changed lines 194,198,201,247 to use &#36;connections_current variable instead of &#36;connection:

194: &#36;connections_current[&#36;type][] = &#36;details;
198: shuffle(&#36;connections_current[&#36;type]);
201: foreach(&#36;connections_current[&#36;type] as &#36;single_connection)
247: if(!array_key_exists('write', &#36;connections_current))

and commented line 193: //unset(&#36;connections);

Maybe there are similar error for other DB engines. I've checked only for MySQL and only for 1 write and 1 read Database server. Probably some additional modifications to connect() function should be made to use several write and/or several read database servers.

3. error in "inc/function.php":

line 226 is: 
&#36;db-&gt;query(&#36;query);

but should be: 
&#36;db-&gt;write_query(&#36;query); 

The problem is that line 226 is trying to use query() function to run shutdown queries. But using query() without last parameter (&#36;write_query) means that it would run with default value for that parameter which is 0 and it would run against read database server. But at least some of shutdown queries are queries which contain SQL UPDATE command and that's why they should be run against write database server.

Original thread: [Bugs with advanced database configuration](https://community.mybb.com/thread-208910.html)

JordanMussi:
Resolves #3619

The purpose of these small bits of code seems to be moving the connection details to an array element when the details are specified in the top level so that afterwards `$connections[TYPE]` for `TYPE=read/write` is an array.

The issue has been that the whole of `$connections` is unset once the first one is processed which has prevented the multi connection code from working properly! 😢

Note it would have worked previously if specified by the user like this:
```php
$config['database']['read'][] = array(
	'hostname' => 'localhost',
	'username' => 'mybb',
	'password' => 'password',
);
$config['database']['write'][] = array(
	'hostname' => 'localhost',
	'username' => 'mybb',
	'password' => 'password',
);
```

but not like this:
```php
$config['database']['read'] = array(
	'hostname' => 'localhost',
	'username' => 'mybb',
	'password' => 'password',
);
$config['database']['write'] = array(
	'hostname' => 'localhost',
	'username' => 'mybb',
	'password' => 'password',
);
```

With this PR, both work correctly.

### DIFF
--- a/inc/db_mysql.php
+++ b/inc/db_mysql.php
@@ -191,7 +191,7 @@ class DB_MySQL implements DB_Base
 			if(array_key_exists('hostname', $connections[$type]))
 			{
 				$details = $connections[$type];
-				unset($connections);
+				unset($connections[$type]);
 				$connections[$type][] = $details;
 			}
 
--- a/inc/db_mysqli.php
+++ b/inc/db_mysqli.php
@@ -190,7 +190,7 @@ class DB_MySQLi implements DB_Base
 			if(array_key_exists('hostname', $connections[$type]))
 			{
 				$details = $connections[$type];
-				unset($connections);
+				unset($connections[$type]);
 				$connections[$type][] = $details;
 			}
 
--- a/inc/db_pgsql.php
+++ b/inc/db_pgsql.php
@@ -194,7 +194,7 @@ class DB_PgSQL implements DB_Base
 			if(array_key_exists('hostname', $connections[$type]))
 			{
 				$details = $connections[$type];
-				unset($connections);
+				unset($connections[$type]);
 				$connections[$type][] = $details;
 			}
 
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -224,7 +224,7 @@ function run_shutdown()
 		// Loop through and run them all
 		foreach($shutdown_queries as $query)
 		{
-			$db->query($query);
+			$db->write_query($query);
 		}
 	}
 


euantorano:
🤦 Shows how well tested this functionality is I guess...

PR #4034, opened 2020-05-31
Fixes #4021 empty `From:` field in mail letter's header when from addr is given

### LINKED ISSUE #4021: Mail's From: in the header will be empty if a from address is designated.
posting this for noyle...

when using the 'send this thread to a friend' function within threads, it doesn't fill out the From: address, like so:

Date: Thu, 28 May 2020 17:49:58 +0000

To: email1@gmail.com

Subject: some thread

From:

Reply-To: email2@gmail.com

Message-ID: &lt;1234@&gt;

Original thread: [Mail's From: in the header will be empty if a from address is designated.](https://community.mybb.com/thread-228030.html)

yuliu:
Resolves #4021.

Seems a pretty easy fix. Tested with SMTP. PHP mail should be fine as well.

@andrewjs18 's help is much appreciated.

### DIFF
--- a/inc/class_mailhandler.php
+++ b/inc/class_mailhandler.php
@@ -157,6 +157,7 @@ class MailHandler
 		if($from)
 		{
 			$this->from = $from;
+			$this->from_named = $this->from;
 		}
 		else
 		{
--- a/inc/class_mailhandler.php
+++ b/inc/class_mailhandler.php
@@ -172,7 +172,6 @@ class MailHandler
 		}
 		else
 		{
-			$this->return_email = "";
 			$this->return_email = $this->get_from_email();
 		}
 


euantorano:
While reviewing, I noticed this seemingly useless line just below: https://github.com/mybb/mybb/blob/031714d3d2f39d1bc1e35bd301e8958a82468f2c/inc/class_mailhandler.php#L175

Might be worth double checking at the same time 😄 

This looks good otherwise though.

JordanMussi:
> While reviewing, I noticed this seemingly useless line just below:

Yep that line isn't needed

JordanMussi (approved):

yuliu:
Thanks for reviewing. I was planning to give the mail handler some amendment, so will probably neglect some small mistakes.

PR #4035, opened 2020-05-31
Fixes #4036 Referrals not transferred correctly when merging users

### LINKED ISSUE #2020: Referrals Are Lost on Account Merge

### LINKED ISSUE #4036: Referrals not transferred correctly when merging users
*Split from https://github.com/mybb/mybb/issues/2020#issuecomment-636474793*

In #3405 the `referrer` field is updated in the same way as `referrals`. `referrer` holds the uid of the referring user and `referrals` holds the count of users referred.

https://github.com/mybb/mybb/blob/be77219bc70d7a9df2c7c7375d69c682c4aad139/admin/modules/user/users.php#L1979-L1981

Users who were referred by the source user are left hanging and not recorded as referred by the destination user.

If the source or destination user referred the other then the referrals will be incorrect after the merge as well.
If the destination user was referred by the source user then it should be updated to be referred by the user that referred the source user.
If the destination user wasn't referred it should be changed to the user that referred the source user (as long as it was not the destination user).

Unfortunately there isn't a way to recover referrers of users previously merged in this incorrect way except for using the admin log which may not be a complete record. Is it worth attempting to restore some of the lost information using the admin log in the upgrade script?

JordanMussi:
Resolves #4036

See https://github.com/mybb/mybb/issues/2020#issuecomment-636474793

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1996,8 +1996,16 @@ if($mybb->input['action'] == "merge")
 			$db->update_query("banned", array('admin' => $destination_user['uid']), "admin = '{$source_user['uid']}'");
 
 			// Carry over referrals
-			$db->update_query("users", array("referrer" => ((int)$source_user['referrer'] + (int)$destination_user['referrer'])), "uid='{$destination_user['uid']}'");
-			$db->update_query("users", array("referrals" => ((int)$source_user['referrals'] + (int)$destination_user['referrals'])), "uid='{$destination_user['uid']}'");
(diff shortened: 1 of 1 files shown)


Eldenroot:
Works fine, tested 2 times... thx!

PR #4038, opened 2020-06-01
Fixes #1427 Change max recipients

### LINKED ISSUE #1427: Change max. recipients
Quote:You may send this message to a maximum of 5 people.

How to change this number?

Original thread: [Change max. recipients](http://community.mybb.com/thread-158970.html)

JordanMussi:
Resolves #1427

Sum of selections of both `to` and `bcc` must not exceed `maxpmrecipients`.

When one input is taking the all of the allowance then the other is disabled (as cannot limit `maximumSelectionSize` to less than 1 with select2).

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -10558,7 +10558,7 @@ if(use_xmlhttprequest == "1")
 {$autocompletejs}
 </body>
 </html>]]></template>
-		<template name="private_send_autocomplete" version="1818"><![CDATA[<link rel="stylesheet" href="{$mybb->asset_url}/jscripts/select2/select2.css?ver=1807">
+		<template name="private_send_autocomplete" version="1823"><![CDATA[<link rel="stylesheet" href="{$mybb->asset_url}/jscripts/select2/select2.css?ver=1807">
 <script type="text/javascript" src="{$mybb->asset_url}/jscripts/select2/select2.min.js?ver=1806"></script>
 <script type="text/javascript">
 <!--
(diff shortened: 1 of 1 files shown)


install/resources/mybb_theme.xml
```diff
@@ -10568,7 +10568,10 @@ if(use_xmlhttprequest == "1")
 	$("#to").select2({
 		placeholder: "{$lang->search_user}",
 		minimumInputLength: 2,
-		maximumSelectionSize: {$mybb->usergroup['maxpmrecipients']},
+		maximumSelectionSize: function() {
+			max = {$mybb->usergroup['maxpmrecipients']} - $("#bcc").select2('data').length
```
euantorano, line 10572: Perhaps I'm being blind, but where is `max` defined? There's also a missing semicolon at the end. Should this not be:

```suggestion
			var max = {$mybb->usergroup['maxpmrecipients']} - $("#bcc").select2('data').length;
```

install/resources/mybb_theme.xml
```diff
@@ -10601,12 +10604,25 @@ if(use_xmlhttprequest == "1")
...
+		}
 	});
 
-  $("#bcc").select2({
+	$("#bcc").select2({
 		placeholder: "{$lang->search_user}",
 		minimumInputLength: 2,
-		maximumSelectionSize: {$mybb->usergroup['maxpmrecipients']},
+		maximumSelectionSize: function() {
+			max = {$mybb->usergroup['maxpmrecipients']} - $("#to").select2('data').length
```
euantorano, line 10623: Again, as above:

```suggestion
			var max = {$mybb->usergroup['maxpmrecipients']} - $("#to").select2('data').length;
```

install/resources/mybb_theme.xml
JordanMussi: It's just defining it for the local scope so doesn't need `var`?

Perhaps you'd prefer this instead of the temporary variable?
```js
return Math.max(1, {$mybb->usergroup['maxpmrecipients']} - $("#bcc").select2('data').length);
```

euantorano: Yeah, that looks cleaner to me. 

I always use `var` (or `let` or `const`) personally.

JordanMussi:
Hehe it's not 6 years just yet @euantorano :wink:

euantorano:
Still over half a decade, which probably sounds even scarier 😅

JordanMussi:
It does as well!! 😬

PR #4042, opened 2020-06-02
Fixes #2217 Font Size with numbers

### LINKED ISSUE #2217: Font Size with numbers
Currently we have 7 options at the editor toolbar for the font size:

xx-small

x-small

small

medium

large

x-large

xx-large

However, it seems to be possible to use font sizes with numbers in source edit mode. Eg.:

Code:
[size=5]Some text.[/size]
Which results in: Some text.
Problem is, while the MyBB parsing supports this when we click Preview or Post, the editor does not support it, and the size tag is removed after clicking Preview or switching to wysiwyg mode.

Another bizarre thing is that negative numbers are allowed in the size tag. Eg.:

Code:
[size=-5]Some text.[/size]
Which results in: Some text.

Worst of all, the text can even be made invisible when using values such as -10. Eg.:

Code:
[size=-10]Some text.[/size]
Which results in: Some text.

Original thread: [Font Size with numbers](http://community.mybb.com/thread-181241.html)

JordanMussi:
Resolves #2217


jscripts/bbcodes_sceditor.js
```diff
@@ -182,7 +180,7 @@ $(function ($) {
 				};
 
 			for (var i = 1; i <= 7; i++)
-				content.append($('<a class="sceditor-fontsize-option" data-size="' + i + '" href="#"><font size="' + i + '">' + i + '</font></a>').on('click', clickFunc));
+				content.append($('<a class="sceditor-fontsize-option" data-size="' + i + '" href="#"><font style="font-size: ' + mybbCmd.fsStr[i-1] + '">' + i + '</font></a>').on('click', clickFunc));
```
JordanMussi, line 198: This change isn't relevant to the fix of the behaviour, it just displays the fonts in the drop down in the same size as the text in the editor for that size.

euantorano:
When applying this PR, I noticed some strange behaviour with the font chooser in WYSIWYG mode. Best descried with a video I think: https://share.getcloudapp.com/DOuAwd6m

JordanMussi:
Thanks for the video very helpful! I think I've tackled it in that last commit. Was a problem with needing to prefer the `size` attribute over the others (it will only be present when inserting from editor).

This whole thing is very janky I'm afraid - not very proud of the code tbh but it seems to be the only way to get it to emulate the parser accepting the strings and numbers.

euantorano:
Fantastic, thank you. I’ll give it another go shortly and hopefully we can merge.

JordanMussi:
Probably best to get someone else to test it too don't want another issue slipping through because I didn't notice the other problem when reviewing myself!

euantorano:
We can always deploy early on the community forums. We normally deploy a week before release, but can do so earlier if it may be beneficial.

euantorano:
There appears to be a conflict with `jscripts/bbcodes_sceditor.js`. I'll try to resolve these within this PR before merging.

PR #4043, opened 2020-06-02
Continues #4038 Change max recipients

JordanMussi:
Oversight of when there is no limit (max = 0)

My mistake sorry @euantorano !


Ben-MyBB:
@euantorano do we need a new issue to go along with this?

Eldenroot:
I dont think so, mention it in the previous one.

euantorano:
Yeah, just mention it in the previous issue.

PR #4046, opened 2020-06-03
Fixes #1572 Empty PM notification when using different language pack

### LINKED ISSUE #1572: Empty PM notification when using different language pack
- Enable "Alert me with a notice when I receive a Private Message."
- Subscribe to a thread requesting PM notification

Now when somebody posts a new reply in the thread the PM is sent correctly. However the email notification about the PM will be empty when the poster uses a different language pack than the subscriber.

JordanMussi:
Resolves #1572

Several cases for someone testing this:
- users receiving:
  - forum subscription emails
  - thread subscription emails
  - thread subscription PMs
  - PM subscription emails
  - PM subscription emails (for PM for thread subscription PM)
  - notification for PM quota reached
  - member registration emails
  - group join request notification emails
- admins receiving member registration notification emails


inc/class_session.php
```diff
@@ -186,11 +186,7 @@ function load_user($uid, $loginkey='')
 			$db->shutdown_query("UPDATE ".TABLE_PREFIX."users SET lastactive='$time', timeonline=timeonline+$timespent{$lastip_add} WHERE uid='{$mybb->user['uid']}'");
 		}
 
-		// Sort out the language and forum preferences.
-		if($mybb->user['language'] && $lang->language_exists($mybb->user['language']))
-		{
-			$mybb->settings['bblanguage'] = $mybb->user['language'];
-		}
+		// Choose date format.
```
yuliu, line 189: Maybe we can still have the existence check here.

JordanMussi: I don't think it is necessary as it is done further down:
https://github.com/mybb/mybb/pull/4046/files#diff-73b6280088a2b814ef3608849f74d75eR592

yuliu:
I'll leave some thoughts here, might not applicable either in 1.8 or any feature version.

It's good to see using a separated variable for session-specific ones, rather than overriding variables in board settings `$mybb->settings`, but not sure if it's OK to use `$mybb->user` as this variable holds mostly raw properties coming from database.

I think it's best, at least for me, to use separated session variables rather than messing up `$mybb->settings` and `$mybb->user`. Probably we can use `$session->user`.

Another though is to get all possible options ready for current user in the session class. For example in this PR, choose an available language for current user/guest/spider and save as `$session->user['language']` and directly set up `$lang` using this language.

Anyway, current way or this PR's way or my way, in my opinion, can't really beat each other to win.

--

For this PR, I prefer still using `$mybb->settings['language']` so that the consistency in session-specific settings/attributes will be maintained:
```
$mybb->settings['language']
$mybb->settings['dateformat']
$mybb->settings['timeformat']
$mybb->settings['threadsperpage']
$mybb->settings['postsperpage']
$mybb->settings['postlayout']
```

JordanMussi:
Fair comments @yuliu I was just trying to make the least invasive changes to follow current behaviour as much as possible for plugins.

> For this PR, I prefer still using `$mybb->settings['language']` so that the consistency in session-specific settings/attributes will be maintained:

I'm not sure what you mean here `bblanguage` is the name of the setting already:
https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/install/resources/settings.xml#L152-L159

yuliu:
> Fair comments @yuliu I was just trying to make the least invasive changes to follow current behaviour as much as possible for plugins.

I had this thought before, perhaps plugins that do something depending on checking/updating `$mybb->settings['bblanguage']` will fail.

> > For this PR, I prefer still using `$mybb->settings['language']` so that the consistency in session-specific settings/attributes will be maintained:
> 
> I'm not sure what you mean here `bblanguage` is the name of the setting already:
> 
> https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/install/resources/settings.xml#L152-L159

Oh, sorry my mistake, should be `$mybb->settings['bblanguage']`. And by saying the consistency I mean those variables are fetched from `inc/settings.php` but could be overriding by session-dependent values.
https://github.com/mybb/mybb/blob/9feef312dc22abef36dcc1ddb1a703e66d9c2fe1/inc/class_session.php#L205-L249

JordanMussi:
I've found that the `bblanguage` setting is actually backed up before being overwritten with the user's language:
https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/inc/init.php#L199

So I've updated to use that instead but still extracted some of the repeated logic into a function. So this now has the original behaviour that `$mybb->settings['bblanguage']` is the current language, `$mybb->settings['orig_bblanguage']` is the `bblanguage` as stored in the db and `$mybb->user['language']` is the user language as stored in the db.

---

I have also updated group join request emails to use the get alternative language method.

Eldenroot:
I can confirm that this bug happens (first time when I saw it by myself and not just got report from users).

@JordanMussi - there is a merge conflict

Ben-MyBB:
Can this be force pushed and merged or should it be closed? @Sama34 @dvz

Sama34:
I went ahead and fixed the conflict, but I did not test as of now.

Regards.

Ben-MyBB:
@arcalys / @Eldenroot would you have capacity to test this fix?

Sama34:
If PRs for 1.8 cannot be fixed or do not comply it is likely that they will never get merged, and the issues may be dragged on to 1.9 instead

PR #4049, opened 2020-06-05
Fixes #4048 Send thread uses non admin email to send mail

### LINKED ISSUE #4048: Send thread uses non admin email to send mail
https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/sendthread.php#L201-L215

The user's email shouldn't be used since it is not owned by the forum. The admin email should be used instead with the user's email as the return address (as it currently is).

After a quick search this is the only place where `my_mail` is called with an email address that isn't the `adminemail` setting.

JordanMussi:
Resolves #4048


Eldenroot:
Works fine. thank you!

PR #4051, opened 2020-06-05
Fixes #3733 PMs lost when moved to Inbox

### LINKED ISSUE #3733: Private messsages not shown after moving them into the "Inbox" folder
I see  there's been a change in the private message folders since version 1820 with commit [# 8e26c5b](https://github.com/mybb/mybb/commit/8e26c5bbe69ae736fe6044021bf212da6f8fd88f#diff-3dc0b085bc730d898e3b650e05355834) which reads:

> // Old folder structure. Need to update
> // Since MyBB 1.8.20 fid[0] represents 'Inbox' and fid[1] represents 'Unread'

It seems it's a feature update that would be allowing more functions on private message, but I quite don't get the idea. However, there will be a problem of showing private messages if they've been moved into the "Inbox" folder. Although neither the template nor the script denies a user to do so, the `private.php` script surely denies users from visiting their "Inbox" folder whose ID is 0:

`$folder = !$fid ? 1 : $fid;`

The only way for them to get back these moved messages is to search their messages with some string sequences in the messages subjects, if they could remember any.

I don't get the clear idea of making the "Inbox" folder kind of obsolete. Just having an ID for the "Inbox" folder and read messages still reside in a folder with ID=1, the "Unread" folder. I'm not sure if I make this issue clear, I do has this problem in my MyBB installation. Also, without knowing the purpose of doing so, I really can't patch the code.

It's my first time contributing to MyBB :)

effone:

Resolves #3733

### DIFF
--- a/private.php
+++ b/private.php
@@ -1694,16 +1694,16 @@ if($mybb->input['action'] == "do_stuff" && $mybb->request_method == "post")
 				$mybb->input['fid'] = 1;
 			}
 
-			foreach($mybb->input['check'] as $key => $val)
+			if(array_key_exists($mybb->input['fid'], $foldernames))
 			{
-				$sql_array = array(
-					"folder" => $mybb->input['fid']
-				);
-				$db->update_query("privatemessages", $sql_array, "pmid='".(int)$key."' AND uid='".$mybb->user['uid']."'");
+				$db->update_query("privatemessages", array("folder" => $mybb->input['fid']), "pmid IN (".implode(",", array_keys($mybb->input['check'])).") AND uid='".$mybb->user['uid']."'");
+				update_pm_count();
+			}
+			else
+			{
+				error($lang->error_invalidmovefid);
 			}
 		}
-		// Update PM count
-		update_pm_count();
 
 		if(!empty($mybb->input['fromfid']))
 		{
--- a/private.php
+++ b/private.php
@@ -1686,8 +1686,8 @@ if($mybb->input['action'] == "do_stuff" && $mybb->request_method == "post")
 	}
 	elseif(!empty($mybb->input['moveto']))
 	{
-		$mybb->input['check'] = $mybb->get_input('check', MyBB::INPUT_ARRAY);
-		if(!empty($mybb->input['check']))
+		$pms = array_map('intval', array_keys($mybb->get_input('check', MyBB::INPUT_ARRAY)));
+		if(!empty($pms))
 		{
 			if(!$mybb->input['fid'])
 			{
@@ -1696,7 +1696,7 @@ if($mybb->input['action'] == "do_stuff" && $mybb->request_method == "post")
 
 			if(array_key_exists($mybb->input['fid'], $foldernames))
 			{
-				$db->update_query("privatemessages", array("folder" => $mybb->input['fid']), "pmid IN (".implode(",", array_keys($mybb->input['check'])).") AND uid='".$mybb->user['uid']."'");
+				$db->update_query("privatemessages", array("folder" => $mybb->input['fid']), "pmid IN (".implode(",", $pms).") AND uid='".$mybb->user['uid']."'");
 				update_pm_count();
 			}
 			else
--- a/private.php
+++ b/private.php
@@ -89,14 +89,15 @@ foreach($foldersexploded as $key => $folders)
 	$folder_name = $folderinfo[1];
 
 	eval("\$folderjump_folder .= \"".$templates->get("private_jump_folders_folder")."\";");
-	eval("\$folderoplist_folder .= \"".$templates->get("private_jump_folders_folder")."\";");
-	// Manipulate search folder selection to omit "Unread"
+
+	// Manipulate search folder selection & move selector to omit "Unread"
 	if($folder_id != 1)
 	{
 		if($folder_id == 0)
 		{
 			$folder_id = 1;
 		}
+		eval("\$folderoplist_folder .= \"".$templates->get("private_jump_folders_folder")."\";");
 		eval("\$foldersearch_folder .= \"".$templates->get("private_jump_folders_folder")."\";");
 	}
 }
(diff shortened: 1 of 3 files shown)


yuliu:
I think we need an upgrade script to move accidentally moved PMs back.

effone:
>I think we need an upgrade script to move accidentally moved PMs back.

Hmm. OK. Will do that as well.

lairdshaw:
> However missed from #3515 was changing the pmfolders update in usercp.php like was done in private.php

Good catch. I'd noticed that too just recently, and would have created a new issue at some point if you (or somebody else) hadn't picked it up first. I do wonder though why the `pmfolders` setting isn't being initialised when the user first registers. Wouldn't it make more sense to do it there, once, rather than to have the initialisation occur in these two separate places, which are - as your comment demonstrates - they are prone to getting out of sync?

euantorano:
Yes, I think it would make sense to do on registration too. Never known why it isn't done that way really.

effone:
This should be a separate issue, really.
Along with moving / wiping the old structure correction code:
https://github.com/mybb/mybb/blob/feature/private.php#L51-L57

This PR's scope is to fix the issue with moving PM to correct folder.

lairdshaw:
> This should be a separate issue, really.

Agreed.

> Along with moving / wiping the old structure correction code:

Maybe that could be moved to the post-login code?

JordanMussi:
Ok I've opened #4054 

@euantorano I think it is probably done this way because `pmfolders` isn't actually needed until the user views the UserCP and then the folder structure needs to be known. Maybe to save db space since the space isn't wasted for users who never went to usercp/activated 🤷 

Looking back through the history of the lines in usercp.php the blame shows the "Adding 1.x" commit so the decision for this was a long time ago 😂  https://github.com/mybb/mybb/blame/12c579c4cae259d4ebdb4a03d60a5d576ad0ed48/usercp.php#L45-L49

effone:
Default folder structure is a little string. No idea how it helps saving DB space.
No reason can define the purpose of blanking out default folder structure during registration.

private.php
```diff
@@ -1688,6 +1689,11 @@
 		$mybb->input['check'] = $mybb->get_input('check', MyBB::INPUT_ARRAY);
 		if(!empty($mybb->input['check']))
 		{
+			if(!$mybb->input['fid'])
+			{
+				$mybb->input['fid'] = 1;
+			}
```
JordanMussi, line 1695: I've noticed that this is actually the only check for moving a PM. While this is fine for this ticket as it redirects PMs moving to Unread to end up in Inbox, you can move a PM to any other number for a folder even if it does not exist. (negative numbers result in an SQL error)

Shouldn't be that big a change so are you happy to fix in this PR or would you rather split out into a new issue?

effone: Why do you expect an out-of-range number from a $_POST call?
Even this check is a fallback and never gonna get hit as we have removed `Unread` option from select control.

JordanMussi: The simplest case is having a custom folder already, loading the PMs page then later on delete the folder and after that move a PM to the folder from the original page which still has the form loaded from when the folder existed. I’m not sure there’s an argument for allowing invalid input to be accepted, at the very least negative numbers should be rejected.

effone:
I took a step ahead to:
- Validate target folder
- Move all messages in a single query
- Update PM Count if only changes occur.

Let me know if its fine now. @JordanMussi

private.php
```diff
@@ -1688,16 +1689,21 @@
 		$mybb->input['check'] = $mybb->get_input('check', MyBB::INPUT_ARRAY);
```
JordanMussi, line 1689:
```suggestion
		$pms = array_map('intval', array_keys($mybb->get_input('check', MyBB::INPUT_ARRAY)));

```

private.php
```diff
@@ -1688,16 +1689,21 @@
...
-				$sql_array = array(
-					"folder" => $mybb->input['fid']
-				);
-				$db->update_query("privatemessages", $sql_array, "pmid='".(int)$key."' AND uid='".$mybb->user['uid']."'");
+				$mybb->input['fid'] = 1;
+			}
+
+			if(array_key_exists($mybb->input['fid'], $foldernames))
+			{
+				$db->update_query("privatemessages", array("folder" => $mybb->input['fid']), "pmid IN (".implode(",", array_keys($mybb->input['check'])).") AND uid='".$mybb->user['uid']."'");
```
JordanMussi, line 1699:
```suggestion
				$db->update_query("privatemessages", array("folder" => $mybb->input['fid']), "pmid IN (".implode(",", $pms).") AND uid='".$mybb->user['uid']."'");
```

JordanMussi:
> * Combine "Move" action & "Delete" action to optimize code, since "Delete" is nothing but just a special move (if not deleting from Trash). On the other hand Moving messages to "Trash" literally deletes (soft) the messages but doesn't update delete timestamp in database.

I think it is fine as is for 1.8 and would require template changes. Deleting when in the trash permanently deletes. It is a bit confusing for the user for the delete button to mean different things. Perhaps best to only show delete button when in the trash folder.

> * Omit current folder from move dropdown (as well as jump dropdown), that doesn't make sense.
> * Omit "Trash" folder from move dropdown, since separate "Delete" button is present.

Maybe make a new issue for these as I think they can be handled without altering the templates :-)

private.php
effone: If we are taking filtered PM IDs in `$pms` shouldn't we check this variable's emptiness in next line?
In place of `if(!empty($mybb->input['check']))`?

Why don't we overwrite same var as it was? Why introduce a new var?

JordanMussi: Yes should probably change the `if` but it's functionally equivalent. Just made a new variable because it is representing in a different way to prevent any possible confusion.
`$pms` is an array of PM IDs, `$mybb->input['check']` has the PM IDs as keys and the values as `1`.

JordanMussi:
@effone please see my latest comments 😃

PR #4052, opened 2020-06-06
Fixes #3853 Missing plugin language file for AdminCP prevents from using/entering AdminCP

### LINKED ISSUE #3853: Missing plugin language file for AdminCP prevents from using/entering AdminCP
There could be different circumstances where a missing plugin language file for AdminCP would make AdminCP inaccessible.

1. If a plugin has a language file for AdminCP in English, and the language file is not correctly uploaded, the AdminCP's plugin page will say `/inc/languages/english/admin/....lang.php does not exist`. **Yes, the forum owner should really have that file uploaded.**

2. If a plugin has a language file for AdminCP only in English, and all files are correctly uploaded:
   - but the user's AdminCP preference is set to another language, maybe Swedish, the AdminCP's plugin page will say `/inc/languages/swedish/admin/....lang.php does not exist` before activating this plugin.
   - or Administrator A who uses English for AdminCP activated the plugin first, and then Administrator B who uses Swedish visiting the plugin page will get a `XXX does not exist` error.
   - or worse for the previous circumstance, if the plugin has registered a header menu link at AdminCP, this Administrator B who uses Swedish will never make it to the AdminCP page. Any visits to the AdminCP will result in a blank page with `XXX does not exist` message.

Here are my thoughts:
- If a plugin is using any language file for AdminCP, a non-officially supported language file missing should trigger AdminCP's resorting to use a default language file for correctly displaying for AdminCP. I.e., AdminCP code should detect missing language files and use a default one upon that.

- If a plugin is using any language file for AdminCP, it might not always be English. The plugin should declare some property of a default AdminCP language, such as `English`, `Japanese`, etc, which used by AdminCP for the last resort.

However, I doubt the feasibility of the above-mentioned method under current MyBB's language handling way. WordPress's way seems good but its language file packaging is not trivial as a plain PHP file.

JordanMussi:
Resolves #3853

The extra checks in admin/index.php are not necessary as the same is performed in the `set_language` method.

Note this only falls back to `english` since `$lang->fallback` is never changed in the core.
It could be worth exploring for 1.9 to use a fallback array and step through each fallback until giving up if the file doesn't exist in any.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -45,10 +45,6 @@ $mybb->settings['cookiepath'] = substr($loc, 0, strrpos($loc, "/{$config['admin_
 
 if(!isset($cp_language))
 {
-	if(!file_exists(MYBB_ROOT."inc/languages/".$mybb->settings['cplanguage']."/admin/home_dashboard.lang.php"))
-	{
-		$mybb->settings['cplanguage'] = "english";
-	}
 	$lang->set_language($mybb->settings['cplanguage'], "admin");
 }
 
@@ -665,7 +661,7 @@ if(!empty($admin_options['authsecret']) && $admin_session['authenticated'] != 1)
 }
 
 // Now the user is fully authenticated setup their personal options
-if(!empty($admin_options['cplanguage']) && file_exists(MYBB_ROOT."inc/languages/".$admin_options['cplanguage']."/admin/home_dashboard.lang.php"))
+if(!empty($admin_options['cplanguage']))
 {
 	$cp_language = $admin_options['cplanguage'];
 	$lang->set_language($cp_language, "admin");
--- a/inc/class_language.php
+++ b/inc/class_language.php
@@ -19,18 +19,29 @@ class MyLanguage
 	public $path;
 
 	/**
-	 * The language we are using.
+	 * The language we are using and the area (if admin).
+	 * 
+	 * For example 'english' or 'english/admin'.
 	 *
 	 * @var string
 	 */
 	public $language;
 
 	/**
-	 * The fallback language we are using.
+	 * The fallback language we are using and the area (if admin).
+	 * 
+	 * For example 'english' or 'english/admin'.
 	 *
 	 * @var string
 	 */
 	public $fallback = 'english';
+
+	/**
+	 * The fallback language we are using.
+	 *
+	 * @var string
+	 */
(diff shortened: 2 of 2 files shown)


yuliu:
Looks good and probably the `$fallback` needs to strip `/admin` from it if a front-end language file (usually `messages.lang.php`) is requested in https://github.com/mybb/mybb/pull/4052/files#diff-900a97abade145488f2d8a5dc80a8eb0R160

Or is it feasible to add another member variable `$fallback_admin` for admin area equivalently to `$fallback` for front-end area? Hmm.. seems not really necessary.

JordanMussi:
I was leaving the behaviour as it was even though it is odd to me too that `$lang->language` and `$lang->fallback` aren't just the language name and the area stored in `$lang->area` for example.

For the messages it is loaded with `$isdatahandler=true`:
https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/admin/index.php#L57
https://github.com/mybb/mybb/blob/76bd4568a4cbbf35b6b3922b8605e40809d7d6d2/inc/class_language.php#L132-L143

So it removes the /admin but not for fallback so I'll fix that now.

yuliu:
> I was leaving the behaviour as it was even though it is odd to me too that `$lang->language` and `$lang->fallback` aren't just the language name and the area stored in `$lang->area` for example.

Yeah I was thinking to make the variable usage more clearly by using:
```
$lang->language     // Current language, could be "english", "english/admin".
$lang->language_frontend     // Current language, could be "english", "japanese".
$lang->language_admin     // Current language, could be "english", "chinese".
```
And similarly to `$lang->fallback`. But it seems not worth to do this in 1.8.

> For the messages it is loaded with `$isdatahandler=true`:

And perhaps it's a good time to change the ambiguous parameter name `$isdatahandler` to another one, probably `$isfrontend` (?). How do you think?

JordanMussi:
I think leaving behaviour as close to what it was as possible is best for 1.8. For 1.9 definitely would look towards doing the language stuff with an array of languages to back off to.

Will rename that parameter as I agree that's a good idea.

Ben-MyBB:
@JordanMussi Seems there is a conflict here

JordanMussi:
Conflict resolved @Ben-MyBB 😄

PR #4056, opened 2020-06-06
Fixes #4054 Default PM Folder Structure

### LINKED ISSUE #4054: `pmfolders` initialisation and inconsistent updates
*Split from https://github.com/mybb/mybb/pull/4051#pullrequestreview-425739893*

`pmfolders` logic should be done by the userhandler.

effone:
Resolves #4054 

Task List:
- [x] Add default folder structure in user datahandler to be created during registration
- [x] Add default folder structure for superadmin account during installation
- [x] Drop old folder structure creation from usercp.php
- [x] Drop new folder structure correction codes from private.php
- [x] Add an upgrade script entry to check for / correct / create new folder structure once for all.

### DIFF
--- a/install/resources/upgrade50.php
+++ b/install/resources/upgrade50.php
@@ -29,7 +29,11 @@ function upgrade50_dbchanges()
 	echo "<p>Updating cache...</p>";
 
 	$cache->delete("banned");
-	
+
+	// Moved PM wrong folder correction
+	$db->update_query("privatemessages", array('folder' => 1), "folder='0'");
+
+	// PM folder structure conversion
 	$db->update_query('users', array('pmfolders' => "0**$%%$1**$%%$2**$%%$3**$%%$4**"), "pmfolders = ''");
 	switch($db->type)
 	{
@@ -41,10 +45,8 @@ function upgrade50_dbchanges()
 			$update = "CONCAT('0**$%%$', pmfolders)";
 	}
 	$db->write_query("UPDATE ".TABLE_PREFIX."users SET pmfolders=".$update." WHERE pmfolders NOT LIKE '0%'");
-	$db->update_query('settings', array('value' => 1), "name='nocacheheaders'");
 
-	// Moved PM wrong folder correction
-	$db->update_query("privatemessages", array('folder' => 1), "folder='0'");
+	$db->update_query('settings', array('value' => 1), "name='nocacheheaders'");
 
 	// Add hCaptcha support
 	echo "<p>Updating settings...</p>";
--- a/install/resources/upgrade50.php
+++ b/install/resources/upgrade50.php
@@ -30,7 +30,17 @@ function upgrade50_dbchanges()
 
 	$cache->delete("banned");
 	
-	$db->update_query('users', array('pmfolders' => "0**$%%$1**$%%$2**$%%$3**$%%$4**"), "pmfolders NOT LIKE '0%'");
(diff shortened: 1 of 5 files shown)


install/resources/upgrade50.php
```diff
@@ -29,7 +29,8 @@ function upgrade50_dbchanges()
 	echo "<p>Updating cache...</p>";
 
 	$cache->delete("banned");
-
+	
+	$db->update_query('users', array('pmfolders' => "0**$%%$1**$%%$2**$%%$3**$%%$4**"), "pmfolders NOT LIKE '0%'");
```
JordanMussi, line 33: A couple of things this doesn't address:
- users who have registered but not visited usercp/private.php `pmfolders=''`
- users who don't have the latest structure will loose their custom folders eg `pmfolders='1**$%%$2**$%%$3**$%%$4**$%%$5**Custom1Folder'`

effone: **It will cover all.**
- Users who have registered but not visited UCP must be having blank folder string in DB (which is not starting with 0), this will add new default structure for them.
 - Users who have custom folders must have visited `private.php` and a single call to that script has already corrected the structure which is starting with `0`, hence ignored by this upgrade line.
- Users who have visited will have old _default_ structure (starting with 1) with no custom folder. This will make that new _default_ structure.

JordanMussi: Yep sorry didn’t think through the empty one!

But for a user that has customised their folders by visiting private.php before the new structure will loose their customised folders. I tested it out earlier, reproduction steps:
1. Start with pre new structure copy of mybb
1. Customise folders as user (and then don’t load private.php or usercp.php)
1. Switch to this PR code
1. Run upgrade script
1. Go to private.php now and users folders have been lost

effone: I know, but I have no idea if we really need to care about the users created custom folders years back and never visited since then. If required at all; I will change the update script.
@euantorano ?

JordanMussi: I don't think that's very responsible. Anyone upgrading from 1.8.19 or older will have all custom folders lost!

JordanMussi: Something like this should do the trick
```suggestion
	$db->update_query('users', array('pmfolders' => "0**$%%$1**$%%$2**$%%$3**$%%$4**"), "pmfolders = ''");
	switch($db->type)
	{
		case "pgsql":
		case "sqlite":
			$update = "'0**$%%$' || pmfolders";
			break;
		default:
			$update = "CONCAT('0**$%%$', pmfolders)";
	}
	$db->write_query("UPDATE ".TABLE_PREFIX."users SET pmfolders=".$update." WHERE pmfolders NOT LIKE '0%'");
```

effone:
I get what you said. Done.
Thanks for taking care of pgsql / sqlite, which I always forget.

JordanMussi:
I only tested quickly so would be worth someone else making sure the upgrade query works as intended! :smiley:

JordanMussi:
*Note: will also require a change for the merge system*
https://github.com/mybb/merge-system/blob/8dd884842b9cab505b8ab8ba1ccb590f0904853a/resources/modules/users.php#L75

JordanMussi:
Just need someone else to confirm the query works for postgres

effone:
Postgres @kawaii

PR #4058, opened 2020-06-10
Fixes #4057 Hooks to allow plugins to extend custom moderation tools

### LINKED ISSUE #4057: Hooks to allow plugins to extend custom moderation tools
Plugins are currently not able to extend custom moderation tools to provide additional options because of a lack of useful hooks:

There are no hooks in `inc/class_custommoderation.php` whereas there are in `inc/class_moderation.php`.

The hooks that are in `admin/modules/config/mod_tools.php` aren't quite sufficient for adding extra options.

Although a very nasty solution is to use the `admin_form_output_submit_wrapper` hook and filter on the current modules loaded to add extra form fields but that doesn't resolve the need for hooks in the custommoderation class.

JordanMussi:
Resolves #4057 

- New hooks in Admin CP Moderator Tools to allow additional options for tools by plugins
- New hooks in CustomModeration class to
  - allow plugins to run their moderation actions
  - perform replacements on subject/message of replies/PMs like the core does by default replacing `{username}`
- Add additional `{author}` replacement for the username of thread's first post

Available replacements shown to administrator in Mod Tools:
image
Plugins can add more by appending to the language string using the hooks that already exist at the start of the pages.


MaxXor:
Thanks for your work on this! I'm eagerly wanting to add some plugin for custom moderation tools. I'm trying to understand your changes. When the hook is called, how does the plugin know which custom moderation tool is being executed?

Sama34:
Who is able to fix the conflicts here?

Sama34:
> Thanks for your work on this! I'm eagerly wanting to add some plugin for custom moderation tools. I'm trying to understand your changes. When the hook is called, how does the plugin know which custom moderation tool is being executed?

I think we should pass `$this` on as well. @mybb/developers

PR #4063, opened 2020-06-11
Contact Redirect

### LINKED ISSUE #4060: contact.php page redirection
when a user submits the contact form, rather than redirecting back to the contact form again, it should redirect them back to the forum index OR their last location prior to going to the contact page.

small UX improvement, I think..

effone:

Resolves #4060

### DIFF
--- a/contact.php
+++ b/contact.php
@@ -244,15 +244,18 @@ if($mybb->request_method == "post")
 			);
 			$db->insert_query("maillogs", $log_entry);
 		}
-
-		if($mybb->usergroup['emailfloodtime'] > 0 || (isset($sent_count) && $sent_count + 1 >= $mybb->usergroup['maxemails']))
+		
+		$mybb->input['from'] = $mybb->get_input('from');
+		if(isset($mybb->input['from']) && !empty($mybb->input['from']))
 		{
-			redirect('index.php', $lang->contact_success_message, '', true);
+			$redirect_url = $mybb->input['from'];
 		}
 		else
 		{
-			redirect('contact.php', $lang->contact_success_message, '', true);
+			$redirect_url = 'index.php';
 		}
+		
+		redirect($redirect_url, $lang->contact_success_message, '', true);
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


contact.php
```diff
@@ -244,14 +244,15 @@
 			);
 			$db->insert_query("maillogs", $log_entry);
 		}
-
-		if($mybb->usergroup['emailfloodtime'] > 0 || (isset($sent_count) && $sent_count + 1 >= $mybb->usergroup['maxemails']))
+		
+		$mybb->input['from'] = $mybb->get_input('from');
+		if(isset($mybb->input['from']) && !empty($mybb->input['from']))
```
euantorano: This can be simplified:

```suggestion
		if(!empty($mybb->input['from']))
```

euantorano (changes requested):
Looks good, just a minor comment.

contact.php
```diff
@@ -244,14 +244,15 @@
 			);
 			$db->insert_query("maillogs", $log_entry);
 		}
-
-		if($mybb->usergroup['emailfloodtime'] > 0 || (isset($sent_count) && $sent_count + 1 >= $mybb->usergroup['maxemails']))
```
JordanMussi, line 248: Is this line intended to be lost? I don’t see how it relates to the ticket...

euantorano: I hadn't even noticed that somehow.

effone: Yes it is intended. That like was for specifically making a decision whether we can / should return to contact.php, on the contrary of the current scenario as we not at all wanna return on a successful submission.

effone: ~~Wait wait, these conditions need to be placed somewhere, I need to revise it.~~

effone: Well, I have rechecked it further. Email flood time as well as Sent limit is taken care in the beginning already.
This check is redundant and not applicable as we are not redirecting to `contact.php` further.
I believe what I have done here is correct. @JordanMussi @euantorano

effone:
@euantorano please have a look on my last comment. I guess this PR is finished.

euantorano:
> @euantorano please have a look on my last comment. I guess this PR is finished.

I'll have a look this afternoon and hopefully get this merged then, thanks!

Eldenroot:
Works fine :)

Eldenroot:
@euantorano - tested again, works fine.

install/resources/mybb_theme.xml
```diff
@@ -14120,7 +14120,7 @@ $(function() {
 	{$sourcemode}
 });
 </script>]]></template>
-		<template name="contact" version="1816"><![CDATA[<html>
+		<template name="contact" version="1824"><![CDATA[<html>
```
```suggestion
		<template name="contact" version="1825"><![CDATA[<html>
```

euantorano (approved):

Issue #654, opened 2014-05-31
Warning points per type can be 0 and 10


GiantCrocodile:
The warning points per type can be set 0 even if it's explicitly stated that it must be greater than 0 and smaller than 10. I think it was meant to be greater than 0 and smaller or equal to 10.

I will push a fix for this.

PR #4064, opened 2020-06-11
Fixes #3470 Small bug in warning log

### LINKED ISSUE #3470: Warning System - small bug in warning log
In my forum, I have a warning type that is basically a "first warning". It is just logged as such but doesn't add warning points (set to 0). That works nicely everywhere BUT in one place: In ACP &gt; Tools &amp; Maintenance &gt; User Warning Log, it lists these warnings with "(+50)" as if the user had been given the Maximum Warning Points (100%) set in configuration &gt; warning system settings. 
It is merely a cosmetic issue, but it's still bugging me because it looks in the logs as if the least offense got the highest penalty.

Original thread: [Warning System - small bug in warning log](https://community.mybb.com/thread-220210.html)

JordanMussi:
Resolves #3470, #654

### DIFF
--- a/admin/modules/config/warning.php
+++ b/admin/modules/config/warning.php
@@ -542,7 +542,7 @@ if($mybb->input['action'] == "edit_type")
 			$errors[] = $lang->error_missing_type_title;
 		}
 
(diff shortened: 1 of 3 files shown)


Ben-MyBB (approved):
Looks good to me, relatively small changes.

PR #4065, opened 2020-06-11
Resolves #3327 - Usergroup permission for invisible

### LINKED ISSUE #3327: Usergroup permission for being invisible
Usergroup settings to implement for allowing users to set themselves as invisible.

Original thread:
https://community.mybb.com/thread-217188.html

Ben-MyBB:
Note: If a user is currently invisible and the Administrator removes the group's ability to hide their online status the * will no longer show and the option will disappear from the UserCP. So this is probably worth mentioning in the blog post for the release about these changes.

Resolves #3327


Ben-MyBB:
Note: I have set the template versions to 1824, assuming this will be included in the next release but just a reminder they will need to be changed again if not merged for .24.

JordanMussi (changes requested):
A few things that you've missed or "GitHub lost the commit" :wink:

- A field for `canbeinvisible` on the usergroups table in each of the `*_db_tables.php` for install
- Upgrade script change to add the field on upgrade
- Add the field to [usergroups.xml](https://github.com/mybb/mybb/blob/feature/install/resources/usergroups.xml)

In each case I would default the value to 1 to reflect current behaviour.

JordanMussi:
After a quick review there's mixed spaces/tabs indentation in upgrade51.php

I see in the upgrade script you've gone with default `1`, but in the table schema a default `0`. I still think the default should be `1` everywhere so that administrators have to opt in to disallowing a group from setting themselves invisible.

Also in the usergroups you've gone with `0` for banned but in upgrade banned won't have it set to `0`. I think it should be consistent between this. By giving banned `0` by default, when a user becomes banned they will not be hidden from online even if previously set to hide online status. For this reason I think banned should also default to `1` to not change the functionality without the administrator changing it for themselves.

What are your thoughts?

install/resources/pgsql_db_tables.php
```diff
@@ -976,6 +976,7 @@
   canratemembers smallint NOT NULL default '0',
   canchangename smallint NOT NULL default '0',
   canbereported smallint NOT NULL default '0',
+  canbeinvisible smallint NOT NULL default '0',
```
JordanMussi:
```suggestion
  canbeinvisible smallint NOT NULL default '1',
```

install/resources/sqlite_db_tables.php
```diff
@@ -898,6 +898,7 @@
 	canratemembers tinyint(1) NOT NULL default '0',
 	canchangename tinyint(1) NOT NULL default '0',
 	canbereported tinyint(1) NOT NULL default '0',
+	canbeinvisible tinyint(1) NOT NULL default '0',
```
JordanMussi:
```suggestion
	canbeinvisible tinyint(1) NOT NULL default '1',
```

install/resources/upgrade51.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    global $output, $cache, $db, $mybb;
+
+    $output->print_header("Updating Database");
+
+    echo "<p>Performing necessary upgrade queries...</p>";
+    flush();
+
+    // Add new setting for new usergroup permission if group members can hide online status
+    $db->add_column("usergroups", "canbeinvisible", "smallint NOT NULL default '1' AFTER canusercp");
```
JordanMussi: This will need a switch for mysql,sqlite/postgresql (tinyint vs smallint)

For example:
https://github.com/mybb/mybb/blob/bdd1f65378685f164c2b47ec2adc81b807bdc454/install/resources/upgrade43.php#L48-L56

JordanMussi (changes requested):
Thanks for doing the indentation but I think it should be all tabs not all spaces 🙈 

A few other suggestions and I'll give it a test tomorrow 😃

install/resources/upgrade51.php
euantorano: Note: MariaDB supports the `BOOLEAN` column type. This is a synonym for `TINYINT(1)`. Not sure about MySQL.

Ben-MyBB: Seems MySQL also accepts tinyint(1) so i've pushed the upgrade changes.

install/resources/mybb_theme.xml
```diff
@@ -12874,7 +12874,7 @@ if(use_xmlhttprequest == "1")
 <input type="submit" class="button" name="submit" value="{$lang->update_notepad}" />
 </div>
 </form>]]></template>
-		<template name="usercp_options" version="1809"><![CDATA[<html>
+		<template name="usercp_options" version="1824"><![CDATA[<html>
```
```suggestion
		<template name="usercp_options" version="1825"><![CDATA[<html>
```

Sorry for the slow review, causing this bump.

install/resources/mybb_theme.xml
```diff
@@ -13078,6 +13069,16 @@ if(use_xmlhttprequest == "1")
 {$footer}
 </body>
 </html>]]></template>
+		<template name="usercp_options_invisible" version="1824"><![CDATA[<fieldset class="trow2">
```
```suggestion
		<template name="usercp_options_invisible" version="1825"><![CDATA[<fieldset class="trow2">
```

usercp.php
```diff
@@ -878,13 +878,18 @@
...
-		$invisiblecheck = "";
+		if(isset($user['invisible']) && $user['invisible'] == 1)
+		{
+			$invisiblecheck .= "checked=\"checked\"";
+		}
+		elseif($user['invisible'] != 0)
+		{
+			$invisiblecheck = "";
+		}
+		eval('$canbeinvisible = "'.$templates->get("usercp_options_invisible")."\";");
```
euantorano, line 894: I believe we want a default for `$canbeinvisible` to ensure the variable is always set - it should be before the `if` on line 882:

```php
$canbeinvisible = '';
```

euantorano (changes requested):
Looks good to me, just some minor changes - mostly template versions, and a possibly uninitialized variable.

euantorano:
Also looks like a conflict - likely because `upgrade51.php` should become `upgrade52.php` for 1.8.24 -> 1.8.25.

usercp.php
```diff
@@ -878,13 +878,18 @@
 		$allownoticescheck = "";
 	}
 
-	if(isset($user['invisible']) && $user['invisible'] == 1)
+	// Check usergroup permission before showing invisible check box
+	if($mybb->usergroup['canbeinvisible'] == 1)
```
euantorano, line 884:
```suggestion
	$canbeinvisible = '';

	// Check usergroup permission before showing invisible check box
	if($mybb->usergroup['canbeinvisible'] == 1)
```

euantorano (approved):

PR #4067, opened 2020-06-12
Resolves #3560 - Missing captcha v2 in "error_nopermission"

### LINKED ISSUE #3560: Missing captcha v2 in "error_nopermission"
Okay,

This is the third time I'm posting and I'm sorry that I haven't looked through all places.

This should be a final missing captcha place. Error template for not logged in users.

Ben-MyBB:
Note: This PR removes the login form from the error_nopermission template and leaves a single "Login" button which takes you to the full login form.

Resolves #3560

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -3639,7 +3639,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 </div>
 <br />]]></template>
 		<template name="error_inline_item" version="1818"><![CDATA[<li>{$error}</li>]]></template>
-		<template name="error_nopermission" version="1818"><![CDATA[{$lang->error_nopermission_guest_1}
+		<template name="error_nopermission" version="1824"><![CDATA[{$lang->error_nopermission_guest_1}
 <ol>
 <li>{$lang->error_nopermission_guest_2}</li>
 <li>{$lang->error_nopermission_guest_3}</li>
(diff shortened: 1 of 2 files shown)


install/resources/mybb_theme.xml
```diff
@@ -3639,36 +3639,15 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 </div>
 <br />]]></template>
 		<template name="error_inline_item" version="1818"><![CDATA[<li>{$error}</li>]]></template>
-		<template name="error_nopermission" version="1818"><![CDATA[{$lang->error_nopermission_guest_1}
+		<template name="error_nopermission" version="1824"><![CDATA[{$lang->error_nopermission_guest_1}
 <ol>
 <li>{$lang->error_nopermission_guest_2}</li>
```
JordanMussi, line 3644:
```suggestion
<li>{$lang->error_nopermission_guest_2} <a href="member.php?action=login">{$lang->login}</a> | <a href="member.php?action=register">{$lang->need_reg}</a></li>
```

I think it's best to add some links here and change the language line:
https://github.com/mybb/mybb/blob/36e8e7167c693ad0ac78dd05406e4cfa8fe0a9eb/inc/languages/english/messages.lang.php#L54

Ben-MyBB:
Thanks @JordanMussi . I have updated the PR. I'm wondering if the links we have added are actually sufficent, the buttons look out of place imo and the links fit much better. 

The buttons work fine on my local host, and action is set. If we decide to keep the buttons I can look into it further tomorrow.

Ben-MyBB:
@JordanMussi any thoughts?

JordanMussi:
Yeah I think the links inline are better than the text saying something like "use the buttons at the bottom of the page" :-)

Irrelevant if we are dropping the buttons but odd that it didn't work for me when testing because I needed a `<input type="hidden" name="action" value="login">` to get it working.

Ben-MyBB:
@JordanMussi Thanks, changes pushed to use links only. 

Strange, your probably right though :P

Ben-MyBB:
@euantorano Just with the various Captcha changes being made in .23, I wonder if we should include this relatively small PR?

euantorano:
Might make sense to do so, yeah.

JordanMussi:
Bump down lol

Ben-MyBB:
Same thing!

PR #4071, opened 2020-06-15
Fixes #4025 Insert image width & height uses incorrect MyCode

### LINKED ISSUE #4025: Inserting image in editor with width & height uses incorrect MyCode
From Discord #18-support:

Quote:@Omnipotent 09/01/2019
Nice, MyBB image buttons are messed up for me because it translates into [img height=* width=*] instead of [img=*x*] so it doesn't get parsed

Note if you switch out of source mode it will fix the syntax but if you post it without switching the mode then it will give an invalid img tag which won't be parsed.

Anyway I am sure someone will open a report on the github and it will get fixed.

Test code:

[img width=1 height=2]aaaa[/img]

Original thread: [Editor's source mode of inserting an image with width/height](https://community.mybb.com/thread-224668.html)

demtor:
Resolves #4025 

The format part  can be omitted for an actual bug of SCEditor: it should format to the new syntax `[img width="" height=""]`  but it is stuck to an old syntax (I guess) which matches to the mybb current one:

https://github.com/samclarke/SCEditor/blob/7fd14aecc6f0c68f1e8bb98b12f6f0da769c671a/src/formats/bbcode.js#L578-L579

So I added it in case they are gonna resolve it one day.

### DIFF
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -522,6 +522,76 @@ $(function ($) {
 		tooltip: 'Insert a video'
 	});
 
+	// Update image command to support MyBB syntax
+	$.sceditor.formats.bbcode.set('img', {
+		format: function (element, content) {
+			if ($(element).data('sceditor-emoticon'))
+				return content;
+
+			var url = $(element).attr('src'),
+				width = $(element).attr('width'),
+				height = $(element).attr('height')
+			;
+
+			var attrs = width && height
+				? '=' + width + 'x' + height
+				: ''
+			;
+
+			return '[img' + attrs + ']' + url + '[/img]';
+		},
+	})
+
+	$.sceditor.command.set('image', {
+		_dropDown: function (editor, caller) {
+			var $content;
+
+			$content = $(
+				'<div>' +
+				'<div>' +
+				'<label for="image">' + editor._('URL') + ':</label> ' +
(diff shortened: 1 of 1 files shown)


jscripts/bbcodes_sceditor.js
```diff
@@ -522,6 +522,76 @@ $(function ($) {
...
+				'</div>' +
+				'<div>' +
+				'<label for="width">' + editor._('Width (optional)') + ':</label> ' +
+				'<input type="text" id="width" size="2" />' +
+				'</div>' +
+				'<div>' +
+				'<label for="height">' + editor._('Height (optional)') + ':</label> ' +
+				'<input type="text" id="height" size="2" />' +
+				'</div>' +
+				'<div><input type="button" class="button" value="' + editor._('Insert') + '" /></div>' +
```
JordanMussi, line 563:
```suggestion
				'<div>' +
				'<input type="button" class="button" value="' + editor._('Insert') + '" />' +
				'</div>' +
```

JordanMussi (changes requested):
Great stuff this is all working for me :-)

One minor thing for consistency 👇

JordanMussi:
*Note to self: cache busting will need bumping when merging*

JordanMussi (approved):
This could probably sneak into 1.8.23 along with #4082...

JordanMussi:
Thank you @demtor!

PR #4072, opened 2020-06-16
Resolves #2058/#1758 - Can view/reply only own threads and guests

### LINKED ISSUE #1758: Can view/reply only own threads and guests
As @ATofighi mentioned here: https://github.com/mybb/mybb/pull/1574 if you turn that permission on for guests, they will be able to view/reply to all threads regardless of it due to 0 UID matching the `$thread['uid']`.

Not sure what we should do here. Probably either:
a) check the IP of the 1st post
b) hide all threads for guests when the 1st permission is enabled for them. And the reply buttons if the 2nd is enabled.
c) hide these permissions for the guest group

@mybb/developers @JN-Jones

### LINKED ISSUE #2058: Guests permissions to edit/delete posts/threads
Hello, I'm trying to allow guests on the forums to edit their own posts and delete their threads.
When I visited the ACP -&gt; Users &amp; Groups -&gt; Groups -&gt; Guests -&gt; Forums &amp; Posts and I scrolled down, I noticed there were options for this and they were activated ( http://i.imgur.com/Fxg8VdA.png ) But for some reason when I visit the forum as a guest I don't see the delete/edit buttons in postbits for posts I made as a guest.

Original thread: [Guests permissions to edit/delete posts/threads](http://community.mybb.com/thread-163512.html)

Ben-MyBB:
This PR will remove the following options from the guest usergroup since they currently have no effect if they are either disabled or enabled.

Editing/Deleting Options
Can edit own posts?
Can delete own posts?
Can delete own threads?
Can update own attachments?

Resolves https://github.com/mybb/mybb/issues/2058 & Resolves https://github.com/mybb/mybb/issues/1758


Ben-MyBB:
Actually, the whole Editing/Deleting Options section can be removed with the other two options:
 Can view deletion notices?
Edit Time Limit
The number of minutes until regular users cannot edit their own posts (if they have the permission). Enter 0 (zero) for no limit.


I'll push the extra changes shortly.

admin/modules/user/groups.php
```diff
@@ -1132,15 +1132,19 @@
...
-	$editing_options = array(
-		$form->generate_check_box("caneditposts", 1, $lang->can_edit_posts, array("checked" => $mybb->input['caneditposts'])),
-		$form->generate_check_box("candeleteposts", 1, $lang->can_delete_posts, array("checked" => $mybb->input['candeleteposts'])),
-		$form->generate_check_box("candeletethreads", 1, $lang->can_delete_threads, array("checked" => $mybb->input['candeletethreads'])),
-		$form->generate_check_box("caneditattachments", 1, $lang->can_edit_attachments, array("checked" => $mybb->input['caneditattachments'])),
-		$form->generate_check_box("canviewdeletionnotice", 1, $lang->can_view_deletion_notices, array("checked" => $mybb->input['canviewdeletionnotice'])),
-		"{$lang->edit_time_limit}<br /><small class=\"input\">{$lang->edit_time_limit_desc}</small><br />".$form->generate_numeric_field('edittimelimit', $mybb->input['edittimelimit'], array('id' => 'edittimelimit', 'class' => 'field50', 'min' => 0))
+	// Remove these options if the group being editied is Guest (GID=1)
+	if($usergroup['gid'] != 1) 
+	{
```
JordanMussi:
```suggestion
	{
```

admin/modules/user/groups.php
```diff
@@ -1132,15 +1132,19 @@
...
+		$editing_options = array(
+			$form->generate_check_box("caneditposts", 1, $lang->can_edit_posts, array("checked" => $mybb->input['caneditposts'])),
+			$form->generate_check_box("candeleteposts", 1, $lang->can_delete_posts, array("checked" => $mybb->input['candeleteposts'])),
+			$form->generate_check_box("candeletethreads", 1, $lang->can_delete_threads, array("checked" => $mybb->input['candeletethreads'])),
+			$form->generate_check_box("caneditattachments", 1, $lang->can_edit_attachments, array("checked" => $mybb->input['caneditattachments'])),
+			$form->generate_check_box("canviewdeletionnotice", 1, $lang->can_view_deletion_notices, array("checked" => $mybb->input['canviewdeletionnotice'])),
+			"{$lang->edit_time_limit}<br /><small class=\"input\">{$lang->edit_time_limit_desc}</small><br />".$form->generate_numeric_field('edittimelimit', $mybb->input['edittimelimit'], array('id' => 'edittimelimit', 'class' => 'field50', 'min' => 0))
 	);
 	$form_container->output_row($lang->editing_deleting_options, "", "<div class=\"group_settings_bit\">".implode("</div><div class=\"group_settings_bit\">", $editing_options)."</div>");
+	};
```
JordanMussi:
```suggestion
		);
		$form_container->output_row($lang->editing_deleting_options, "", "<div class=\"group_settings_bit\">".implode("</div><div class=\"group_settings_bit\">", $editing_options)."</div>");
	}
```

JordanMussi (changes requested):
These options are still available for the forum-management permission options:

https://github.com/mybb/mybb/blob/1326eea040ef2637ac58c92512f9b6f7dd7696bb/admin/modules/forum/management.php#L722-L726

Also need to work out what to do about `canonlyviewownthreads` as mentioned in #1758

Ben-MyBB:
> 
> 
> These options are still available for the forum-management permission options:
> 
> https://github.com/mybb/mybb/blob/1326eea040ef2637ac58c92512f9b6f7dd7696bb/admin/modules/forum/management.php#L722-L726
> 
> Also need to work out what to do about `canonlyviewownthreads` as mentioned in #1758

Since those are in a different place, I wonder if we should treat them as a seperate issue or handle them here?

In terms of `canonlyviewownthreads` we could just check if you are editing group ID=1 and remove that option if so?

JordanMussi:
I think best to bundle as it’s basically the same setting just on the per forum basis.

That’s what I was thinking too. It might be worth doing a review to see what effects these options have at the moment and whether we should set them all to 0 on upgrade for guests...

Ben-MyBB:
I'll update this PR to fix both of those points raised. I think it's actually just a small inconsistency with the options being shown because there's no way to validate which guest has posted without using the questionable method of checking IP addresses but we obviously don't have that implemented.

JordanMussi:
Was thinking it might be worth a check to make sure everywhere is ensuring uid>0 before checking those settings but we can get to that a bit later.

JordanMussi:
> In terms of `canonlyviewownthreads` we could just check if you are editing group ID=1 and remove that option if so?

Please could you do this too then?

Also I think we should 0 out these options in the upgrade script since at the moment guests *can* see other guest's threads when `canonlyviewownthreads` is enabled (I'm sure the same is true for attachments...)

Ben-MyBB:
> > In terms of `canonlyviewownthreads` we could just check if you are editing group ID=1 and remove that option if so?
> 
> 
> 
> Please could you do this too then?
> 
> 
> 
> Also I think we should 0 out these options in the upgrade script since at the moment guests *can* see other guest's threads when `canonlyviewownthreads` is enabled (I'm sure the same is true for attachments...)

Yep, that option should be removed along with one other in my latest changes. (Hopefully github hasn’t removed them as well?) 

I can take a look at the upgrade script when I get on my laptop in a bit.

JordanMussi:
> Yep, that option should be removed along with one other in my latest changes. (Hopefully github hasn’t removed them as well?)

Quite right - I didn't see it sneak in there :wink:

Eldenroot:
@Ben-MyBB - please can you finish this one?

Ben-MyBB:
I will try and get this finished for .27.

PR #4079, opened 2020-06-18
Fixes #4076 Remove anti-CSRF token from PMs and e-mails

### LINKED ISSUE #4076: Remove anti-CSRF token from PMs and e-mails
Adjust subscription management links and forms to allow links with no post code attached (that may become invalid as a result of #3376).

Previous discussion: https://github.com/mybb/mybb/pull/4022#issuecomment-644268664

JordanMussi:
Resolves #4076


euantorano:
Looks like this is targeted at 1.8.23, I'll try and review it as soon as I can this week.

usercp.php
```diff
@@ -2102,7 +2102,7 @@
 	}
 }
 
-if($mybb->input['action'] == "removesubscription")
+if($mybb->input['action'] == "removesubscription" && ($mybb->request_method == "post" || verify_post_check($mybb->get_input('my_post_key'), true)))
```
dvz, line 2105: Is that to allow old links (GET) with `my_post_key` to work?

usercp.php
```diff
@@ -2168,6 +2168,83 @@
 	}
 }
 
+// Show remove subscription form when GET method and without valid my_post_key
+if($mybb->input['action'] == "removesubscription")
```
dvz, line 2172: Can the GET/POST sections be merged to remove duplicated code?

JordanMussi: I mainly did this to avoid a huge diff when indenting the section above but I'm more than happy to combine both sections under one if statement.

usercp.php
JordanMussi: Yes. The "remove forum subscription" link in forumdisplay still uses this.

usercp.php
euantorano: IMO that should be don separately in a future PR for 1.8.24, in the interest of keeping this easy to review.

PR #4081, opened 2020-06-18
Fixes #3290 Inserting link issue

### LINKED ISSUE #3290: Inserting link issue
Hi,

I don't know if this has already been posted.....
If one is using the MyBB wysiwyg-editor and uses the "insert a link" mycode button, the linke won't be closed after the mycode-window has been closed. This happens in Firefox, Edge and IE but not with Chrome.

This issues has been posted here and here.... and you will see the issue here in text which follows the links. Generally, the text should look like.....

It should look like:
This issues has been posted here and here.... and you will see the issue here in text which follows the links. Generally, the text should look like.....

The url-tags do not close properly when using Firefox, Edge or IE..... 

Original thread: [inserting link issue](https://community.mybb.com/thread-218244.html)

JordanMussi:
⚠️ Depends on https://github.com/mybb/SCEditor/pull/7

Resolves #3290

Custom email setup can be removed as it duplicates upstream which now has support for description in the email command:
https://github.com/mybb/SCEditor/blob/4d7c6b3a0a30d2bff93aaf55c83febe5853bf15a/src/lib/defaultCommands.js#L517-L566

URL fix can be removed as it is now done upstream:
https://github.com/mybb/SCEditor/blob/4d7c6b3a0a30d2bff93aaf55c83febe5853bf15a/src/formats/bbcode.js#L635-L641

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -14054,9 +14054,9 @@ if(use_xmlhttprequest == "1")
 <td class="trow1" width="20%"><strong>{$lang->username}</strong></td>
 <td class="trow1">{$mybb->user['username']} <span class="smalltext">[<strong><a href="member.php?action=logout&amp;logoutkey={$mybb->user['logoutkey']}">{$lang->change_user}</a></strong>]</span></td>
 </tr>]]></template>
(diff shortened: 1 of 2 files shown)


Sama34:
So after merging mybb/SCEditor#7 we do merge this?

PR #4089, opened 2020-06-23
While soft deleting threads, now its posts are also removed.

### LINKED ISSUE #4083: Posts in unviewable threads appear in search
In my forum users recently noticed that they can see user's messages in removed thread when they search for their posts. I could confirm this happening even after I moved thread to another forum - to see if it's permission issue or not. It seems that post search feature doesn't check if that post is located in removed thread.

Original thread: [Post search doesn't check if thread is removed](https://community.mybb.com/thread-224436.html)

rajat315315:
Resolves #4083 

For soft_delete_thread requests, posts are also deleted now, so they won't appear in search results.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -970,6 +970,19 @@ if($mybb->input['action'] == "change")
 			$lang->success_settings_updated .= $lang->success_settings_updated_captchaimage;
 		}
 
+		// If using fulltext then enforce minimum word length given by database
+		if(isset($mybb->input['upsetting']['minsearchword']) && $mybb->input['upsetting']['minsearchword'] > 0 && $mybb->input['upsetting']['searchtype'] == "fulltext" && $db->supports_fulltext_boolean("posts") && $db->supports_fulltext("threads"))
+		{
+			// Attempt to determine minimum word length from MySQL for fulltext searches
+			$query = $db->query("SHOW VARIABLES LIKE 'ft_min_word_len';");
+			$min_length = $db->fetch_field($query, 'Value');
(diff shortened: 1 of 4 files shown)


yuliu:
Hmm, I don't think the root cause of #4083 is how soft delete works.

rajat315315:
When a thread is soft deleted, it should delete its replies also.
It was not happening earlier.

Ben-MyBB:
This would mean that if a thread is soft deleted and then restored, all replies to the thread will be hidden unless they are also restored. I don't believe that could be coded in to this fix because you may intentionally deleted one post on the thread already.

I haven't looked at the code yet for the search, but I was thinking that it wasn't checking if the post is soft deleted in that location?

rajat315315:
Yeah, that's true. I agree. Haven't thought about deleting and restoring part.

Eldenroot:
Also you have two fixes in one PR, you should create branches for every PR (fix).

For this issue, you have to check if the thread is deleted or not and then apply to posts...

yuliu:
> I think when a thread is soft deleted, it should delete its replies also.

Hmm, that's not true IIRC. Currently when a thread is soft-deleted only the thread is marked and none its posts will be marked, and that's why it's called thread soft-deleted not posts soft-deleted. Other modules should check if a thread is marked as soft-deleted but not by its posts.

> But what was happening is :
> Variable "visible" remained set to 1 for the replies of soft_deleted threads, so even though they didn't appear to be deleted, they appeared in search.

That's because the code in search.php isn't doing right. Please see https://github.com/mybb/mybb/issues/4083#issuecomment-648744802 .

PR #4095, opened 2020-06-28
PHP 8.0 fixes

### LINKED ISSUE #4138: Fix PHP 8.0 compatibility problems
- [x] Various compatibility issues: #4095
- [x] Common _undefined_ errors — Admin CP: #4207
- [x] Common _undefined_ errors — front-end

JordanMussi:
I gave MyBB a spin with PHP 8.0.0alpha1 and decided to give a first pass at making some changes. Most of this is fixing warnings found by browsing and using various parts of MyBB.

See [UPGRADING](https://github.com/php/php-src/blob/php-8.0.0alpha1/UPGRADING) and [NEWS](https://github.com/php/php-src/blob/php-8.0.0alpha1/NEWS) for changes introduced in PHP 8.

Summary of how this PR gives PHP 8 compatibility:

1. <details><summary>Notices converted into warnings.</summary>

    ```text
        A number of notices have been converted into warnings:

        * Attempting to read an undefined variable.
        * Attempting to read an undefined property.
        * Attempting to read a property of a non-object.
        * Attempting to access an array index of a non-array.
        * Attempting to convert an array to string.
        * Attempting to use a resource as an array key.
        * Attempting to use null, a boolean or a float as a string offset.
        * Attempting to read an out-of-bounds string offset.
        * Attempting to assign an empty string to a string offset.
    ```

    </details>

    Of particular interest is:
      - `Attempting to read an undefined variable` - eg some variables used in templates which are conditionally set which need `$example = '';` set before the condition block
      - `Attempting to read an undefined property` - various missing language strings
      - `Attempting to access an array index of a non-array` - for example in `fetch_field` implementations when `null` is returned from `fetch_array` and `$array[$field]` is returned
        I initially used `num_rows` in several places before changing the `fetch_field` method. I can go back and remove these if wanted but they are equivalent and in some cases more expressive of the intention.

2. <details><summary>A naming conflict is introduced with a new builtin <code>XmlParser</code> class.</summary>

    ```text
    - XML:
      . xml_parser_create(_ns) will now return an XmlParser object rather than a
        resource. Return value checks using is_resource() should be replaced with
        checks for `false`. The xml_parser_free() function no longer has an effect,
        instead the XmlParser instance is automatically destroyed if it is no longer
        referenced.
    ```

    </details>

    To fix this:
    - `XMLParser` is moved to `inc/class_xmlparser.php` and renamed to `MyBBXMLParser`
    - `inc/class_xml.php` contains an `XMLParser` class that extends `MyBBXMLParser`
    - all `new XMLParser(...)` have been changed to use a wrapper function `create_xml_parser(...)` to hide following logic:
      - PHP < 8.0 require `class_xml.php` and use `new XMLParser`
      - PHP >= 8.0 require `class_xmlparser.php` and use `new MyBBXMLParser`

    This will be backwards compatible with earlier PHP versions and plugins previously written to use `inc/class_xml.php` but these plugins will not support PHP 8.0.

    It may be helpful to review this by checking the individual commits to see the file move rather than whole file removed and whole file added...

3. <details><summary><code>each()</code> was deprecated in PHP 7.2 and has been removed in PHP 8.</summary>

    ```text
      . Removed each(). foreach or ArrayIterator should be used instead.
    ```

    </details>

    Updated the horde `Text_Diff` 3rd party dependency (and `String.php` from horde Util) to latest version. These updated versions state compatibility with PHP 7.

4. <details><summary>Declaring a required parameter after an optional one is deprecated.</summary>

    ```text
      . Declaring a required parameter after an optional one is deprecated. As an
        exception, declaring a parameter of the form "Type $param = null" before
        a required one continues to be allowed, because this pattern was sometimes
        used to achieve nullable types in older PHP versions.

            function test($a = [], $b) {}       // Deprecated
            function test(Foo $a = null, $b) {} // Allowed
    ```

    </details>

    For example `function user_search_conditions($input=array(), &$form)`: `$input` is optional as default specified and `$form` is required as no default specified.

5. <details><summary>Curly braces can no longer be used for offset access</summary>

    ```text
      . Removed support for deprecated curly braces for offset access
        RFC: https://wiki.php.net/rfc/deprecate_curly_braces_array_access
    ```

    </details>

    For example `$next_char = $text{$region[1]+8};` in `inc/functions_posting.php`

6. <details><summary>Calling <code>implode()</code> with arguments in incorrect order is no longer supported.</summary>

    ```text
      . Calling implode() with parameters in a reverse order ($pieces, $glue) is no
        longer supported.
    ```

    </details>

    For example `implode($posts, ",")` in `moderation.php`.


Eldenroot:
Great! Thx!

dvz:
> Updated the horde Text_Diff 3rd party dependency (and String.php from horde Util) to latest version. These updated versions state compatibility with PHP 7.

Any problems on PHP 5.2?

>A naming conflict is introduced with a new builtin XmlParser class.

Looks like we could go ahead with the name and apply [`class_alias()`](https://www.php.net/manual/en/function.class-alias.php) for backward compatibility.

JordanMussi:
> Any problems on PHP 5.2?
None that I saw. They only test as far back as 5.4 https://github.com/horde/Text_Diff/blob/v2.2.1/.travis.yml
I’m happy to install 5.2 and give it a test before merging.

> Looks like we could go ahead with the new name and apply [`class_alias()`](https://www.php.net/manual/en/function.class-alias.php) for backward compatibility.

That’s what I initially went for but class_alias was introduced in 5.3. I had an if where it used the old name for pre 5.3 and the new one with alias for later (if class wasn’t already set) but I simplified to the method used now in the PR.

dvz:
It may be acceptable to skip the alias for when running on 5.2 for the sake of simplicity.

Sama34:
Why should be care about 5.2? MyBB 1.8 doesn't officially support PHP 8 and it shouldn't change. IIRC 1.9 isn't going to support 5.2, or am I confusing this?

JordanMussi:
> It may be acceptable to skip the alias for when running on 5.2 for the sake of simplicity.

I don't see a benefit of doing it this way over using a `XMLParser` class that extends `MyBBXMLParser`.

> Why should be care about 5.2? MyBB 1.8 doesn't officially support PHP 8 and it shouldn't change. IIRC 1.9 isn't going to support 5.2, or am I confusing this?

MyBB 1.8 supports 5.2 as a minimum. Imo there's no reason to not include this work in 1.8 to improve compatibility when it doesn't introduce problems for earlier PHP versions.

euantorano:
> MyBB 1.8 supports 5.2 as a minimum. Imo there's no reason to not include this work in 1.8 to improve compatibility when it doesn't introduce problems for earlier PHP versions.

I agree.

Eldenroot:
I know that we should not change the minimum supported version of PHP in MyBB 1.8.x branch... but, we should respect minimum officially maintained PHP version, so we should bump the version, because it is a drawback and we should encourage users to update... we are fixing security issues in MyBB core, but php 5.2 is not supported anymore and it is not secure to use on live board...

Eldenroot:
And please check this:
https://www.php.net/supported-versions.php

image

We should really support PHP versions which have security support by PHP team... not older!

euantorano:
In an ideal world, yes. In the real world, bumping supported PHP versions in a minor point release is a recipe for disaster. Just look at how many people have problems with the minor theme change for the authorisation code - now think about how many support requests there would be if we changed supported PHP versions!

JordanMussi:
> I know that we should not change the minimum supported version of PHP in MyBB 1.8.x branch... but, we should respect minimum officially maintained PHP version, so we should bump the version, because it is a drawback and we should encourage users to update... we are fixing security issues in MyBB core, but php 5.2 is not supported anymore and it is not secure to use on live board...

I agree with all of this and it looks like the plan was for 1.9 to support at minimum PHP 7.1 https://community.mybb.com/thread-215212-post-1294239.html#pid1294239

I think the consensus was to keep in step with the PHP versions supported by the PHP team. We do recommend boards use the latest versions of PHP but we don't stop users from using old.

A significant number of installs are still using PHP versions that are EOL. Less than half are using PHP with active support. Since 1.9 is not released yet we have the ability to specify the minimum we want to set without worrying about users already using that series that will be affected.

euantorano:
At the moment it's looking like 19 will be PHP 7.3+.

Eldenroot:
You can always announce this in advance, but yes it is a problem (you should again reconsider which minimum version will be supported in MyBB 1.9 - I would say that 7.3 will be the bottom line if you will release final in 2021). 

We should speed up releasing minor versions, you wrote in past that many users are complaining about that, but it is only a minority. If everything is well documented (template changes), it should not be a problem for 99,9 % users.

JordanMussi:
8.0.0alpha2 released today.

<details>
<summary>Diff of <code>UPGRADING</code> for <code>8.0.0alpha1...8.0.0alpha2</code></summary>

```patch
diff --git a/UPGRADING-8.0.0alpha1.txt b/UPGRADING-8.0.0alpha2.txt
index b18ff4a..1b54be9 100755
--- a/UPGRADING-8.0.0alpha1.txt
+++ b/UPGRADING-8.0.0alpha2.txt
@@ -201,6 +201,11 @@ PHP 8.0 UPGRADE NOTES
     com.autoregister_casesensitive may no longer be disabled; case-insensitive
     markers in com.typelib_file are ignored.
 
+- Curl:
+  . CURLOPT_POSTFIELDS no longer accepts objects as arrays. To interpret an
+    object as an array, perform an explicit (array) cast. The same applies to
+    other options accepting arrays as well.
+
 - Date:
   . mktime() and gmmktime() now require at least one argument. time() can be
     used to get the current timestamp.
@@ -270,6 +275,8 @@ PHP 8.0 UPGRADE NOTES
 
 - LDAP:
   . The deprecated function ldap_sort has been removed.
+  . The interface of ldap_set_rebind_proc has changed; the $callback parameter
+    does not accept empty string anymore; null value shall be used instead.
 
 - Mbstring:
   . The mbstring.func_overload directive has been removed. The related
@@ -308,6 +315,10 @@ PHP 8.0 UPGRADE NOTES
   . The legacy behaviour of passing the encoding as the third argument instead
     of an offset for the mb_strrpos() function has been removed, provide an
     explicit 0 offset with the encoding as the fourth argument instead.
+  . The ISO_8859-* character encoding aliases have been replaced by ISO8859-*
+    aliases for better interoperability with the iconv extension. The mbregex
+    ISO 8859 aliases with underscores (ISO_8859_* and ISO8859_*) have also been
+    removed.
 
 - PCRE:
   . When passing invalid escape sequences they are no longer interpreted as
@@ -437,6 +448,19 @@ PHP 8.0 UPGRADE NOTES
     respect the inherited locale without an explicit setlocale() call. An
     explicit setlocale() call is now always required if you wish to change any
     locale component from the default.
+  . Removed deprecated DES fallback in crypt(). If an unknown salt format is
+    passed to crypt(), the function will fail with *0 instead of falling back
+    to a weak DES hash now.
+  . The result of sorting functions may have changed, if the array contains
+    equal-comparing elements.
+  . Sort comparison functions return true/false will now throw a deprecation
+    warning and should be replaced with an implementation return an integer
+    smaller than, equal to, or greater than zero.
+
+        // Replace
+        usort($array, fn($a, $b) => $a > $b);
+        // With
+        usort($array, fn($a, $b) => $a <=> $b);
 
 - Sysvmsg:
   . msg_get_queue() will now return an SysvMessageQueue object rather than a
@@ -457,6 +481,12 @@ PHP 8.0 UPGRADE NOTES
   . The $use_include_path parameter, which was not used internally, has been
     removed from tidy_repair_string().
 
+- Tokenizer:
+  . T_COMMENT tokens will no longer include a trailing newline. The newline will
+    instead be part of a following T_WHITESPACE token. It should be noted that
+    T_COMMENT is not always followed by whitespace, it may also be followed by
+    T_CLOSE_TAG or end-of-file.
+
 - XML:
   . xml_parser_create(_ns) will now return an XmlParser object rather than a
     resource. Return value checks using is_resource() should be replaced with
@@ -535,6 +565,7 @@ PHP 8.0 UPGRADE NOTES
     RFC: https://wiki.php.net/rfc/mixed_type_v2
   . Added support for Attributes
     RFC: https://wiki.php.net/rfc/attributes_v2
+    RFC: https://wiki.php.net/rfc/attribute_amendments
   . Added support for constructor property promotion (declaring properties in
     the constructor signature).
     RFC: https://wiki.php.net/rfc/constructor_promotion
@@ -543,14 +574,18 @@ PHP 8.0 UPGRADE NOTES
   . Added DateTime::createFromInterface() and
     DateTimeImmutable::createFromInterface().
 
+- Dom:
+  . Introduce DOMParentNode and DOMChildNode with new traversal and
+    manipulation APIs.
+    RFC: https://wiki.php.net/rfc/dom_living_standard_api
+
 - Enchant:
   . enchant_dict_add()
   . enchant_dict_is_added()
   . LIBENCHANT_VERSION macro
 
-- dom:
-  . Introduce DOMParentNode and DOMChildNode with new traversal and manipulation APIs
-    RFC: https://wiki.php.net/rfc/dom_living_standard_api
+- Hash:
+  . HashContext objects can now be serialized.
 
 - Opcache:
   . If the opcache.record_warnings ini setting is enabled, opcache will record
@@ -565,7 +600,7 @@ PHP 8.0 UPGRADE NOTES
     OPENSSL_ENCODING_PEM.
 
 - Standard:
-  . printf() and friends how support the %h and %H format specifiers. These
+  . printf() and friends now support the %h and %H format specifiers. These
     are the same as %g and %G, but always use "." as the decimal separator,
     rather than determining it through the LC_NUMERIC locale.
   . printf() and friends now support using "*" as width or precision, in which
@@ -581,6 +616,10 @@ PHP 8.0 UPGRADE NOTES
 
         $proc = proc_open($command, [['pty'], ['pty'], ['pty']], $pipes);
 
+  . Sorting functions are now stable, which means that equal-comparing elements
+    will retain their original order.
+    RFC: https://wiki.php.net/rfc/stable_sorting
+
 - Zip:
   . Extension updated to version 1.19.0
   . New ZipArchive::lastId property to get index value of last added entry.
@@ -595,6 +634,9 @@ PHP 8.0 UPGRADE NOTES
 3. Changes in SAPI modules
 ========================================
 
+- Apache:
+  . The PHP module has been renamed from php7_module to php_module.
+
 - CGI and FPM will now use CONTEXT_DOCUMENT_ROOT to scan for .user.ini files,
   if it is defined. Otherwise, DOCUMENT_ROOT will be used as before. This
   improves support for Apache mod_userdir and mod_alias.
@@ -669,6 +711,10 @@ PHP 8.0 UPGRADE NOTES
   . Added get_resource_id($resource) function, which returns the same value as
     (int) $resource. It provides the same functionality under a clearer API.
 
+- LDAP:
+  . Added ldap_count_references(), which returns the number of reference
+    messages in a search result.
+
 - OpenSSL:
   . Added openssl_cms_encrypt() encrypts the message in the file with the
     certificates and outputs the result to the supplied file.
@@ -763,6 +809,17 @@ PHP 8.0 UPGRADE NOTES
     longer referenced.
   . The deprecated parameter `$version` of curl_version() has been removed.
 
+- Date:
+  . DatePeriod now implements IteratorAggregate (instead of Traversable).
+
+- DOM:
+  . DOMNamedNodeMap now implements IteratorAggregate (instead of Traversable).
+  . DOMNodeList now implements IteratorAggregate (instead of Traversable).
+
+- Intl:
+  . IntlBreakIterator now implements IteratorAggregate (instead of Traversable).
+  . ResourceBundle now implements IteratorAggregate (instead of Traversable).
+
 - Enchant:
   . The enchant extension now uses libenchant-2 by default when available.
     libenchant version 1 is still supported but is deprecated and could
@@ -783,13 +840,29 @@ PHP 8.0 UPGRADE NOTES
 - MBString:
   . The Unicode data tables have been updated to version 13.0.0.
 
+- PDO:
+  . PDOStatement now implements IteratorAggregate (instead of Traversable).
+
 - MySQLi / PDO MySQL:
   . When mysqlnd is not used (which is the default and recommended option),
     the minimum supported libmysqlclient version is now 5.1.
+  . mysqli_result now implements IteratorAggregate (instead of Traversable).
 
 - PGSQL / PDO PGSQL:
   . The PGSQL and PDO PGSQL extensions now require at least libpq 9.1.
 
+- SimpleXML:
+  . SimpleXMLElement now implements RecursiveIterator and absorbed the
+    functionality of SimpleXMLIterator. SimpleXMLIterator is an empty extension
+    of SimpleXMLElement.
+
+- Shmop:
+  . shmop_open() will now return a Shmop object rather than a resource. Return
+    value checks using is_resource() should be replaced with checks for `false`.
+    The shmop_close() function no longer has an effect, and is deprecated;
+    instead the Shmop instance is automatically destroyed if it is no longer
+    referenced.
+
 ========================================
 10. New Global Constants
 ========================================
```

</details>

There doesn't seem to be any significant changes affecting MyBB. Also gave MyBB a quick spin with no problems ✨

showthread.php
```diff
@@ -903,10 +905,7 @@
             $visible
             ORDER BY p.dateline
         ");
-        if(!is_array($postsdone))
-        {
-            $postsdone = array();
-        }
+		$postsdone = array();
```
dvz, line 888: Indentation here and other areas of the file is off - we can correct that too

inc/functions_upload.php
```diff
@@ -62,7 +62,8 @@ function remove_attachment($pid, $posthash, $aid)
 		}
 
 		$date_directory = explode('/', $attachment['attachname']);
-		if(@is_dir($uploadpath."/".$date_directory[0]))
+		$query = $db->simple_select("attachments", "COUNT(aid) as indir", "attachname LIKE '".$db->escape_string_like($date_directory[0])."/%'");
```
dvz: Are these changes related to compatibility?

inc/functions.php
```diff
@@ -347,11 +347,6 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
 			$offset = (float)$mybb->user['timezone'];
 			$dstcorrection = $mybb->user['dst'];
 		}
-		elseif(defined("IN_ADMINCP"))
-		{
-			$offset = (float)$mybbadmin['timezone'];
-			$dstcorrection = $mybbadmin['dst'];
```
dvz, line 353: `$mybbadmin` can be removed from `global`

inc/3rdparty/diff/README.txt
```diff
@@ -0,0 +1,5 @@
+## Versions
```
dvz, line 1: It may be fine to just mention versions in commits or related PRs

wlisikpl:
According to [this](https://www.php.net/archive/2020.php#2020-07-23-1), php8 is in **alpha** version, so its highly not recommended to be used in `production-scope` projects. If I were you (team), I'd wait till php8 reaches stability and gains `stable` version

euantorano:
Hi,

Yes, that is correct. The purpose of a PR like this is so that when PHP 8 is released as a stable version, we are ready to support it. We prefer to start preparing before the release is stable, rather than rushing to support it straight after release and possibly having users try to run the software on a version that isn't yet supported.

We definitely don't recommend trying to run PHP 8 in production for the time being! 😄

inc/functions_upload.php
JordanMussi: I was getting the following warning as MyBB attempts to delete the directory every time an attachment is deleted.

image

It is not strictly related to compatibility with PHP 8.0 but I made the change thinking it was at the time.

I can extract into a separate PR if you prefer. I have also realised that since there is the `index.html` file in the dir then the delete directory will actually never succeed...

inc/3rdparty/diff/README.txt
JordanMussi: I thought having the versions explicitly in the source control would help with documenting and upgrading as it is easier to find the current version used.

inc/functions_upload.php
JordanMussi: I've removed the changes from this PR and opened a new PR against #4147.

JordanMussi:
Since I last looked at this PR there have been a few further releases.

<details>
<summary>Diff of <code>UPGRADING</code> for <code>8.0.0alpha2...8.0.0alpha3</code></summary>

```patch
diff --git a/UPGRADING-8.0.0alpha2.txt b/UPGRADING-8.0.0alpha3.txt
index 1b54be9..f233cd6 100755
--- a/UPGRADING-8.0.0alpha2.txt
+++ b/UPGRADING-8.0.0alpha3.txt
@@ -194,6 +194,11 @@ PHP 8.0 UPGRADE NOTES
     RFC: https://wiki.php.net/rfc/locale_independent_float_to_string
   . Removed support for deprecated curly braces for offset access
     RFC: https://wiki.php.net/rfc/deprecate_curly_braces_array_access
+  . Applying the final modifier on a private method will now produce a warning
+    unless that method is the constructor.
+    RFC: https://wiki.php.net/rfc/inheritance_private_methods
+  . If an object constructor exit()s, the object destructor will no longer be
+    called. This matches the behavior when the constructor throws.
 
 - COM:
   . Removed the ability to import case-insensitive constants from type
@@ -320,6 +325,10 @@ PHP 8.0 UPGRADE NOTES
     ISO 8859 aliases with underscores (ISO_8859_* and ISO8859_*) have also been
     removed.
 
+- OCI8:
+  . Several alias functions have been marked as deprecated.
+  . oci_internal_debug() and its alias ociinternaldebug() have been removed.
+
 - PCRE:
   . When passing invalid escape sequences they are no longer interpreted as
     literals. This behaviour previously required the X modifier - which is
@@ -371,6 +380,8 @@ PHP 8.0 UPGRADE NOTES
   . ReflectionMethod::isConstructor() and ReflectionMethod::isDestructor() now
     also return true for `__construct` and `__destruct` methods of interfaces.
     Previously, this would only be true for methods of classes and traits.
+  . ReflectionType::isBuiltin() method has been moved to ReflectionNamedType.
+    ReflectionUnionType does not have it.
 
 - Socket:
   . The deprecated AI_IDN_ALLOW_UNASSIGNED and AI_IDN_USE_STD3_ASCII_RULES
@@ -451,6 +462,8 @@ PHP 8.0 UPGRADE NOTES
   . Removed deprecated DES fallback in crypt(). If an unknown salt format is
     passed to crypt(), the function will fail with *0 instead of falling back
     to a weak DES hash now.
+  . Specifying out of range rounds for sha256/sha526 crypt() will now fail with
+    *0 instead of clamping to the closest limit. This matches glibc behavior.
   . The result of sorting functions may have changed, if the array contains
     equal-comparing elements.
   . Sort comparison functions return true/false will now throw a deprecation
@@ -462,6 +475,11 @@ PHP 8.0 UPGRADE NOTES
         // With
         usort($array, fn($a, $b) => $a <=> $b);
 
+  . Any functions accepting callbacks that are not explicitly specified to
+    accept parameters by reference will now warn if a callback with reference
+    parameters is used. Examples include array_filter() and array_reduce().
+    This was already the case for most, but not all, functions previously.
+
 - Sysvmsg:
   . msg_get_queue() will now return an SysvMessageQueue object rather than a
     resource. Return value checks using is_resource() should be replaced with
@@ -569,6 +587,12 @@ PHP 8.0 UPGRADE NOTES
   . Added support for constructor property promotion (declaring properties in
     the constructor signature).
     RFC: https://wiki.php.net/rfc/constructor_promotion
+  . Added support for `match` expression.
+    RFC: https://wiki.php.net/rfc/match_expression_v2
+  . Private methods declared on a parent class no longer enforce any
+    inheritance rules on the methods of a child class. (with the exception of
+    final private constructors)
+    RFC: https://wiki.php.net/rfc/inheritance_private_methods
 
 - Date:
   . Added DateTime::createFromInterface() and
@@ -616,6 +640,14 @@ PHP 8.0 UPGRADE NOTES
 
         $proc = proc_open($command, [['pty'], ['pty'], ['pty']], $pipes);
 
+  . proc_open() now supports socket pair descriptors. The following attaches
+    a distinct socket pair to stdin, stdout and stderr:
+
+        $proc = proc_open(
+            $command, [['socket'], ['socket'], ['socket']], $pipes);
+
+    Unlike pipes, sockets do not suffer from blocking I/O issues on Windows.
+    However, not all programs may work correctly with stdio sockets.
   . Sorting functions are now stable, which means that equal-comparing elements
     will retain their original order.
     RFC: https://wiki.php.net/rfc/stable_sorting
@@ -884,6 +916,9 @@ PHP 8.0 UPGRADE NOTES
     shell now consistently execute `%comspec% /s /c "$commandline"`, which has
     the same effect as executing `$commandline` (without additional quotes).
 
+- GD:
+  . php_gd2.dll has been renamed to php_gd.dll.
+
 - php-test-pack:
   . The test runner has been renamed from run-test.php to run-tests.php, to
     match its name in php-src.
```

</details>

<details>
<summary>Diff of <code>UPGRADING</code> for <code>8.0.0alpha3...8.0.0beta1</code></summary>

```patch
diff --git a/UPGRADING-8.0.0alpha3.txt b/UPGRADING-8.0.0beta1.txt
index f233cd6..15e19f2 100755
--- a/UPGRADING-8.0.0alpha3.txt
+++ b/UPGRADING-8.0.0beta1.txt
@@ -21,6 +21,8 @@ PHP 8.0 UPGRADE NOTES
 ========================================
 
 - Core:
+  . Assertion failures now throw by default. If the old behavior is desired,
+    then set `assert.exception=0` in INI settings.
   . Methods with the same name as the class are no longer interpreted as
     constructors. The __construct() method should be used instead.
   . Removed ability to call non-static methods statically.
@@ -47,7 +49,7 @@ PHP 8.0 UPGRADE NOTES
     parameter consistent with isset() and normal array access. All key types now
     use the usual coercions and array/object keys throw a TypeError.
   . Any array that has a number n as its first numeric key will use n+1 for
-    its next implicit key. Even if n is negative.
+    its next implicit key, even if n is negative.
     RFC: https://wiki.php.net/rfc/negative_array_index
   . The default error_reporting level is now E_ALL. Previously it excluded
     E_NOTICE and E_DEPRECATED.
@@ -78,6 +80,10 @@ PHP 8.0 UPGRADE NOTES
     Additionally, care should be taken that error messages are not displayed in
     production environments, which can result in information leaks. Please
     ensure that display_errors=Off is used in conjunction with error logging.
+  . Adding more than one @ operator to an expression is no longer supported,
+    since this syntax is now used for attributes (previously extra @ operators
+    had no effect).
+    RFC: https://wiki.php.net/rfc/shorter_attribute_syntax
   . Inheritance errors due to incompatible method signatures (LSP violations)
     will now always generate a fatal error. Previously a warning was generated
     in some cases.
@@ -85,7 +91,7 @@ PHP 8.0 UPGRADE NOTES
   . The precedence of the concatenation operator has changed relative to
     bitshifts and addition as well as subtraction.
     RFC: https://wiki.php.net/rfc/concatenation_precedence
-  . Arguments with a default-value that resolves to null at run-time will no
+  . Arguments with a default value that resolves to null at runtime will no
     longer implicitly mark the argument type as nullable. Either use an explicit
     nullable type, or an explicit null default value instead.
 
@@ -110,11 +116,12 @@ PHP 8.0 UPGRADE NOTES
 
     * Attempting to read an undefined variable.
     * Attempting to read an undefined property.
+    * Attempting to read an undefined array key.
     * Attempting to read a property of a non-object.
     * Attempting to access an array index of a non-array.
     * Attempting to convert an array to string.
     * Attempting to use a resource as an array key.
-    * Attempting to use null, a boolean or a float as a string offset.
+    * Attempting to use null, a boolean, or a float as a string offset.
     * Attempting to read an out-of-bounds string offset.
     * Attempting to assign an empty string to a string offset.
 
@@ -127,7 +134,7 @@ PHP 8.0 UPGRADE NOTES
   . Uncaught exceptions now go through "clean shutdown", which means that
     destructors will be called after an uncaught exception.
   . Compile time fatal error "Only variables can be passed by reference" has
-    been delayed until run-time and converted to "Cannot pass parameter by
+    been delayed until runtime and converted to "Cannot pass parameter by
     reference" exception.
   . Some "Only variables should be passed by reference" notices have been
     converted to "Cannot pass parameter by reference" exception.
@@ -143,13 +150,6 @@ PHP 8.0 UPGRADE NOTES
 
     The name shown above is still followed by a null byte and and a unique
     suffix.
-  . Declaring a required parameter after an optional one is deprecated. As an
-    exception, declaring a parameter of the form "Type $param = null" before
-    a required one continues to be allowed, because this pattern was sometimes
-    used to achieve nullable types in older PHP versions.
-
-        function test($a = [], $b) {}       // Deprecated
-        function test(Foo $a = null, $b) {} // Allowed
   . Non-absolute trait method references in trait alias adaptations are now
     required to be unambiguous:
 
@@ -199,6 +199,58 @@ PHP 8.0 UPGRADE NOTES
     RFC: https://wiki.php.net/rfc/inheritance_private_methods
   . If an object constructor exit()s, the object destructor will no longer be
     called. This matches the behavior when the constructor throws.
+  . Non-strict comparisons between numbers and non-numeric strings now work by
+    casting the number to string and comparing the strings. Comparisons between
+    numbers and numeric strings continue to work as before. Notably, this means
+    that `0 == "not-a-number"` is considered false now.
+    RFC: https://wiki.php.net/rfc/string_to_number_comparison
+  . Namespaced names can no longer contain whitespace: While `Foo\Bar` will be
+    recognized as a namespaced name, `Foo \ Bar` will not. Conversely, reserved
+    keywords are now permitted as namespace segments.
+    RFC: https://wiki.php.net/rfc/namespaced_names_as_token
+  . Nested ternaries now require explicit parentheses.
+    RFC: https://wiki.php.net/rfc/ternary_associativity
+  . debug_backtrace() and Exception::getTrace() will no longer provide
+    references to arguments. It will not be possible to change function
+    arguments through the backtrace.
+  . Numeric string handling has been altered to be more intuitive and less
+    error-prone. Trailing whitespace is now allowed in numeric strings for
+    consistency with how leading whitespace is treated. This mostly affects:
+     - The is_numeric() function
+     - String-to-string comparisons
+     - Type declarations
+     - Increment and decrement operations
+    The concept of a "leading-numeric string" has been mostly dropped; the
+    cases where this remains exist in order to ease migration. Strings which
+    emitted an E_NOTICE "A non well-formed numeric value encountered" will now
+    emit an E_WARNING "A non-numeric value encountered" and all strings which
+    emitted an E_WARNING "A non-numeric value encountered" will now throw a
+    TypeError. This mostly affects:
+     - Arithmetic operations
+     - Bitwise operations
+    This E_WARNING to TypeError change also affects the E_WARNING
+    "Illegal string offset 'string'" for illegal string offsets. The behavior
+    of explicit casts to int/float from strings has not been changed.
+    RFC: https://wiki.php.net/rfc/saner-numeric-strings
+  . Magic Methods will now have their arguments and return types checked if
+    they have them declared. The signatures should match the following list:
+
+      __call(string $name, array $arguments): mixed
+      __callStatic(string $name, array $arguments): mixed
+      __clone(): void
+      __debugInfo(): ?array
+      __get(string $name): mixed
+      __invoke(mixed $arguments): mixed
+      __isset(string $name): bool
+      __serialize(): array
+      __set(string $name, mixed $value): void
+      __set_state(array $properties): object
+      __sleep(): array
+      __unserialize(array $data): void
+      __unset(string $name): void
+      __wakeup(): void
+
+    RFC: https://wiki.php.net/rfc/magic-methods-signature
 
 - COM:
   . Removed the ability to import case-insensitive constants from type
@@ -267,7 +319,7 @@ PHP 8.0 UPGRADE NOTES
     gmp_random_bits() should be used instead.
 
 - Iconv:
-  . iconv() implementations which do not properly set errno in case of errors,
+  . iconv() implementations which do not properly set errno in case of errors
     are no longer supported.
 
 - Intl:
@@ -317,8 +369,8 @@ PHP 8.0 UPGRADE NOTES
     now be empty.
   . The $is_hex parameter, which was not used internally, has been removed from
     mb_decode_numericentity().
-  . The legacy behaviour of passing the encoding as the third argument instead
-    of an offset for the mb_strrpos() function has been removed, provide an
+  . The legacy behavior of passing the encoding as the third argument instead
+    of an offset for the mb_strrpos() function has been removed; provide an
     explicit 0 offset with the encoding as the fourth argument instead.
   . The ISO_8859-* character encoding aliases have been replaced by ISO8859-*
     aliases for better interoperability with the iconv extension. The mbregex
@@ -329,9 +381,26 @@ PHP 8.0 UPGRADE NOTES
   . Several alias functions have been marked as deprecated.
   . oci_internal_debug() and its alias ociinternaldebug() have been removed.
 
+- OpenSSL:
+  . openssl_x509_read() and openssl_csr_sign() will now return an
+    OpenSSLCertificate object rather than a resource. Return value checks using
+    is_resource() should be replaced with checks for `false`.
+  . The openssl_x509_free() function is deprecated and no longer has an effect,
+    instead the OpenSSLCertificate instance is automatically destroyed if it is no
+    longer referenced.
+  . openssl_csr_new() will now return an OpenSSLCertificateSigningRequest object
+    rather than a resource. Return value checks using is_resource() should be
+    replaced with checks for `false`.
+  . openssl_pkey_new() will now return an OpenSSLAsymmetricKey object rather than a
+    resource. Return value checks using is_resource() should be replaced with
+    checks for `false`.
+  . The openssl_pkey_free() function is deprecated and no longer has an effect,
+    instead the OpenSSLAsymmetricKey instance is automatically destroyed if it is no
+    longer referenced.
+
 - PCRE:
   . When passing invalid escape sequences they are no longer interpreted as
-    literals. This behaviour previously required the X modifier - which is
+    literals. This behavior previously required the X modifier - which is
     now ignored.
 
 - PDO:
@@ -339,13 +408,20 @@ PHP 8.0 UPGRADE NOTES
     "exceptions". See https://www.php.net/manual/en/pdo.error-handling.php
     for details of behavior changes and how to explicitly set this attribute.
     RFC: https://wiki.php.net/rfc/pdo_default_errmode
-  . The method PDOStatement::setFetchMode() now accepts the following signature:
+  . The signatures of some PDO methods have changed:
 
-        PDOStatement::setFetchMode($mode, $classname, $params)
+        PDO::query(
+            string $statement, ?int $fetch_mode = null, ...$fetch_mode_args)
+        PDOStatement::setFetchMode(int $mode, ...$params)
 
 - PDO_ODBC:
   . The php.ini directive pdo_odbc.db2_instance_name has been removed
 
+- Phar:
+  . Metadata associated with a phar will no longer be automatically unserialized,
+    to fix potential security vulnerabilities due to object instantiation, autoloading, etc.
+    RFC: https://wiki.php.net/rfc/phar_stop_autoloading_metadata
+
 - Reflection:
   . The method signatures
 
@@ -383,9 +459,16 @@ PHP 8.0 UPGRADE NOTES
   . ReflectionType::isBuiltin() method has been moved to ReflectionNamedType.
     ReflectionUnionType does not have it.
 
-- Socket:
+- Sockets:
   . The deprecated AI_IDN_ALLOW_UNASSIGNED and AI_IDN_USE_STD3_ASCII_RULES
     flags for socket_addrinfo_lookup() have been removed.
+  . socket_create(), socket_create_listen(), socket_accept(),
+    socket_import_stream(), socket_addrinfo_connect(), socket_addrinfo_bind(),
+    and socket_wsaprotocol_info_import() will now return a Socket object rather
+    than a resource. Return value checks using is_resource() should be replaced
+    with checks for `false`.
+  . socket_addrinfo_lookup() will now return an array of AddressInfo objects
+    rather than resources.
 
 - SPL:
   . SplFileObject::fgetss() has been removed.
@@ -465,10 +548,10 @@ PHP 8.0 UPGRADE NOTES
   . Specifying out of range rounds for sha256/sha526 crypt() will now fail with
     *0 instead of clamping to the closest limit. This matches glibc behavior.
   . The result of sorting functions may have changed, if the array contains
-    equal-comparing elements.
-  . Sort comparison functions return true/false will now throw a deprecation
-    warning and should be replaced with an implementation return an integer
-    smaller than, equal to, or greater than zero.
+    elements that compare as equal.
+  . Sort comparison functions that return true or false will now throw a
+    deprecation warning, and should be replaced with an implementation
+    that returns an integer less than, equal to, or greater than zero.
 
         // Replace
         usort($array, fn($a, $b) => $a > $b);
@@ -479,6 +562,14 @@ PHP 8.0 UPGRADE NOTES
     accept parameters by reference will now warn if a callback with reference
     parameters is used. Examples include array_filter() and array_reduce().
     This was already the case for most, but not all, functions previously.
+  . The HTTP stream wrapper as used by functions like file_get_contents()
+    now advertises HTTP/1.1 rather than HTTP/1.0 by default. This does not
+    change the behavior of the client, but may cause servers to respond
+    differently. To retain the old behavior, set the 'protocol_version'
+    stream context option, e.g.
+
+        $ctx = stream_context_create(['http' => ['protocol_version' => '1.0']]);
+        echo file_get_contents('http://example.org', false, $ctx);
 
 - Sysvmsg:
   . msg_get_queue() will now return an SysvMessageQueue object rather than a
@@ -504,6 +595,11 @@ PHP 8.0 UPGRADE NOTES
     instead be part of a following T_WHITESPACE token. It should be noted that
     T_COMMENT is not always followed by whitespace, it may also be followed by
     T_CLOSE_TAG or end-of-file.
+  . Namespaced names are now represented using the T_NAME_QUALIFIED (Foo\Bar),
+    T_NAME_FULLY_QUALIFIED (\Foo\Bar) and T_NAME_RELATIVE (namespace\Foo\Bar)
+    tokens. T_NS_SEPARATOR is only used for standalone namespace separators,
+    and only syntactially valid in conjunction with group use declarations.
+    RFC: https://wiki.php.net/rfc/namespaced_names_as_token
 
 - XML:
   . xml_parser_create(_ns) will now return an XmlParser object rather than a
@@ -517,8 +613,8 @@ PHP 8.0 UPGRADE NOTES
     objects instead of resources.
 
 - Zip:
-  . ZipArchive::OPSYS_Z_CPM have been removed, should use ZipArchive::OPSYS_CPM
-    (this name was a typo).
+  . ZipArchive::OPSYS_Z_CPM has been removed (this name was a typo). Use
+    ZipArchive::OPSYS_CPM instead.
 
 - Zlib:
   . gzgetss() has been removed.
@@ -584,6 +680,7 @@ PHP 8.0 UPGRADE NOTES
   . Added support for Attributes
     RFC: https://wiki.php.net/rfc/attributes_v2
     RFC: https://wiki.php.net/rfc/attribute_amendments
+    RFC: https://wiki.php.net/rfc/shorter_attribute_syntax
   . Added support for constructor property promotion (declaring properties in
     the constructor signature).
     RFC: https://wiki.php.net/rfc/constructor_promotion
@@ -593,10 +690,16 @@ PHP 8.0 UPGRADE NOTES
     inheritance rules on the methods of a child class. (with the exception of
     final private constructors)
     RFC: https://wiki.php.net/rfc/inheritance_private_methods
+  . Added support for nullsafe operator (`?->`).
+    RFC: https://wiki.php.net/rfc/nullsafe_operator
+  . Added support for named arguments.
+    RFC: https://wiki.php.net/rfc/named_params
 
 - Date:
   . Added DateTime::createFromInterface() and
     DateTimeImmutable::createFromInterface().
+  . Added the DateTime format specifier "p" which is the same as "P" but
+    returning "Z" for UTC.
 
 - Dom:
   . Introduce DOMParentNode and DOMChildNode with new traversal and
@@ -608,6 +711,11 @@ PHP 8.0 UPGRADE NOTES
   . enchant_dict_is_added()
   . LIBENCHANT_VERSION macro
 
+- FPM:
+  . Added a new option pm.status_listen that allows getting status from
+    different endpoint (e.g. port or UDS file) which is useful for getting
+    status when all children are busy with serving long running requests.
+
 - Hash:
   . HashContext objects can now be serialized.
 
@@ -678,6 +786,13 @@ PHP 8.0 UPGRADE NOTES
 ========================================
 
 - Core:
+  . Declaring a required parameter after an optional one is deprecated. As an
+    exception, declaring a parameter of the form "Type $param = null" before
+    a required one continues to be allowed, because this pattern was sometimes
+    used to achieve nullable types in older PHP versions.
+
+        function test($a = [], $b) {}       // Deprecated
+        function test(Foo $a = null, $b) {} // Allowed
   . Calling get_defined_functions() with $exclude_disabled explicitly set to
     false is deprecated. get_defined_functions() will never include disabled
     functions.
@@ -690,6 +805,11 @@ PHP 8.0 UPGRADE NOTES
   . enchant_broker_free and enchant_broker_free_dict, unset the object instead
   . ENCHANT_MYSPELL and ENCHANT_ISPELL constants
 
+- LibXML:
+  . libxml_disable_entity_loader() has been deprecated. As libxml 2.9.0 is now
+    required, external entity loading is guaranteed to be disabled by default,
+    and this function is no longer needed to protect against XXE attacks.
+
 - PGSQL / PDO PGSQL:
   . The constant PG_VERSION_STR has now the same value as PG_VERSION, and thus
     is deprecated.
@@ -731,7 +851,7 @@ PHP 8.0 UPGRADE NOTES
    . env_method
    . enc_password
  . ZipArchive::addEmptyDir, ZipArchive::addFile and aZipArchive::addFromString
-   methods have a new "flags" argument. This allow to manage name encoding
+   methods have a new "flags" argument. This allows managing name encoding
    (ZipArchive::FL_ENC_*) and entry replacement (ZipArchive::FL_OVERWRITE)
  . ZipArchive::extractTo now restore file modification time.
 
@@ -875,6 +995,11 @@ PHP 8.0 UPGRADE NOTES
 - PDO:
   . PDOStatement now implements IteratorAggregate (instead of Traversable).
 
+- LibXML:
+  . The minimum required libxml version is now 2.9.0. This means that external
+    entity loading is now guaranteed to be disabled by default, and no extra
+    steps need to be taken to protect against XXE attacks.
+
 - MySQLi / PDO MySQL:
   . When mysqlnd is not used (which is the default and recommended option),
     the minimum supported libmysqlclient version is now 5.1.
@@ -883,6 +1008,12 @@ PHP 8.0 UPGRADE NOTES
 - PGSQL / PDO PGSQL:
   . The PGSQL and PDO PGSQL extensions now require at least libpq 9.1.
 
+- Readline:
+  . Calling readline_completion_function() before the interactive prompt starts
+    (e.g. in auto_prepend_file) will now override the default interactive prompt
+	completion function. Previously, readline_completion_function() only worked
+	when called after starting the interactive prompt.
+
 - SimpleXML:
   . SimpleXMLElement now implements RecursiveIterator and absorbed the
     functionality of SimpleXMLIterator. SimpleXMLIterator is an empty extension
@@ -907,6 +1038,10 @@ PHP 8.0 UPGRADE NOTES
 11. Changes to INI File Handling
 ========================================
 
+- zend.exception_string_param_max_len
+  . New INI directive to set the maximum string length in an argument of a
+    stringified stack strace.
+
 ========================================
 12. Windows Support
 ========================================
```

</details>

<details>
<summary>Diff of <code>UPGRADING</code> for <code>8.0.0beta1...8.0.0beta2</code></summary>

```patch
diff --git a/UPGRADING-8.0.0beta1.txt b/UPGRADING-8.0.0beta2.txt
index 15e19f2..4cb91bc 100755
--- a/UPGRADING-8.0.0beta1.txt
+++ b/UPGRADING-8.0.0beta2.txt
@@ -21,6 +21,7 @@ PHP 8.0 UPGRADE NOTES
 ========================================
 
 - Core:
+  . `match` is now a reserved keyword.
   . Assertion failures now throw by default. If the old behavior is desired,
     then set `assert.exception=0` in INI settings.
   . Methods with the same name as the class are no longer interpreted as
@@ -206,7 +207,9 @@ PHP 8.0 UPGRADE NOTES
     RFC: https://wiki.php.net/rfc/string_to_number_comparison
   . Namespaced names can no longer contain whitespace: While `Foo\Bar` will be
     recognized as a namespaced name, `Foo \ Bar` will not. Conversely, reserved
-    keywords are now permitted as namespace segments.
+    keywords are now permitted as namespace segments, which may also change the
+    interpretation of code: `new\x` is now the same as `constant('new\x')`, not
+    `new \x()`.
     RFC: https://wiki.php.net/rfc/namespaced_names_as_token
   . Nested ternaries now require explicit parentheses.
     RFC: https://wiki.php.net/rfc/ternary_associativity
```

</details>

<details>
<summary>Diff of <code>UPGRADING</code> for <code>8.0.0beta3...8.0.0beta3</code></summary>

```patch
diff --git a/UPGRADING-8.0.0beta2.txt b/UPGRADING-8.0.0beta3.txt
index 4cb91bc..2412b60 100755
--- a/UPGRADING-8.0.0beta2.txt
+++ b/UPGRADING-8.0.0beta3.txt
@@ -149,7 +149,7 @@ PHP 8.0 UPGRADE NOTES
         new class {};
         // -> class@anonymous
 
-    The name shown above is still followed by a null byte and and a unique
+    The name shown above is still followed by a null byte and a unique
     suffix.
   . Non-absolute trait method references in trait alias adaptations are now
     required to be unambiguous:
@@ -381,6 +381,9 @@ PHP 8.0 UPGRADE NOTES
     removed.
 
 - OCI8:
+  . The OCI-Lob class is now called OCILob, and the OCI-Collection class is now
+    called OCICollection for name compliance enforced by PHP 8 arginfo
+    type annotation tooling.
   . Several alias functions have been marked as deprecated.
   . oci_internal_debug() and its alias ociinternaldebug() have been removed.
 
@@ -1045,6 +1048,10 @@ PHP 8.0 UPGRADE NOTES
   . New INI directive to set the maximum string length in an argument of a
     stringified stack strace.
 
+- com.dotnet_version
+  . New INI directive to choose the version of the .NET framework to use for
+    dotnet objects.
+
 ========================================
 12. Windows Support
 ========================================
```

</details>


The only thing being of note here as far as I can see is the following now being a warning instead of a notice
> Attempting to read an undefined array key.

This generates a number of warnings all around the codebase which I forsee as being a pain for development but can be avoided by using PHP 7.4 during development. As we move to a more modern codebase the problematic areas will be refactored and not have this problem.

dvz:
Is is feasible to also address the _undefined array key_ issues in 1.8? We can initially support PHP 8 and keep fixing warnings that pop up along the way. If its' too prevalent, we may able to filter these warnings globally.

dvz (approved):

PR #4096, opened 2020-06-28
Fix - Inefficient thread moderation code - threads queried one by one

### LINKED ISSUE #2114: Inefficient thread moderation code - threads queried one by one
As the title says:
https://github.com/mybb/mybb/blob/feature/inc/class_moderation.php#L3650
https://github.com/mybb/mybb/blob/feature/inc/class_moderation.php#L538
https://github.com/mybb/mybb/blob/feature/inc/class_moderation.php#L3496
https://github.com/mybb/mybb/blob/feature/inc/class_moderation.php#L388

All `tids`/`pids` are known so I'm not sure why a single query with a left join to the `forums` table + a while loop for results weren't used in the first place.

rajat315315:
Closes #2114

### DIFF
--- a/inc/class_moderation.php
+++ b/inc/class_moderation.php
@@ -384,9 +384,18 @@ class Moderation
 
 		$tid_list = $forum_counters = $user_counters = $posts_to_approve = array();
 
+		$tids_list = implode(",", $tids);
+		$threads = array();
+		$query = $db->simple_select("threads", "*", "tid IN ($tids_list)");
+
+		while($thread = $db->fetch_array($query))
+		{
+			$threads[$thread['tid']] = $thread;
+		}
+
 		foreach($tids as $tid)
 		{
-			$thread = get_thread($tid);
+			$thread = $threads[$tid];
 			if(!$thread || $thread['visible'] == 1 || $thread['visible'] == -1)
 			{
 				continue;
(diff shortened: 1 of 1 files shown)


inc/class_moderation.php
```diff
@@ -534,9 +543,18 @@ function unapprove_threads($tids)
...
+			$threads[$thread['tid']] = $thread;
+		}
+
 		foreach($tids as $tid)
 		{
-			$thread = get_thread($tid);
+			$thread = $threads[$tid];
 			$forum = get_forum($thread['fid']);
 
 			if($thread['visible'] == 1 || $thread['visible'] == -1)
```
Sama34:
```suggestion
			if(!$thread || $thread['visible'] == 1 || $thread['visible'] == -1)
```

inc/class_moderation.php
```diff
@@ -3649,9 +3675,17 @@ function soft_delete_threads($tids)
...
+			$threads[$thread['tid']] = $thread;
+		}
+
 		foreach($tids as $tid)
 		{
-			$thread = get_thread($tid);
+			$thread = $threads[$tid];
 			$forum = get_forum($thread['fid']);
 
 			if($thread['visible'] == 1 || $thread['visible'] == 0)
```
Sama34:
```suggestion
			if(!$thread || $thread['visible'] == 1 || $thread['visible'] == 0)
```

Sama34 (changes requested):

PR #4102, opened 2020-06-29
Fixes #4085 Thread View Mode cannot be set to Use Default

### LINKED ISSUE #4085: Thread View Mode cannot be set to Use Default but selects Threaded/Linear depending on default set by administrator
In UserCP &gt; Edit Options
If someone opts for "Thread View Mode" as "Use Default"; MyBB renders the select option for "Thread View Mode" as per current setting; resulting the user setting gets updated with default value, rather than "Use Default".

Let me explain with steps.
1. Open ACP » Settings » Show Thread Options, set "Usenet Style Thread View" to "Yes".
2. Open UCP &gt; Edit Options. Select "Thread View Mode" as "Use Default". Save settings. Notice, after saving the settings and page load the select dropdown appears as "Threaded" selected, not "Use Default".
3. Again open ACP  » Settings » Show Thread Options, set "Usenet Style Thread View" to "No".
4. Open UCP &gt; Edit Options. Watch, the setting still shows "Threaded" whereas current default is "Linear". User setting got updated with last default, in place of "Use Default".

Original thread: [Thread View Mode setting dishonors default](https://community.mybb.com/thread-228297.html)

rajat315315:
Closes #4085

This omitted code checks if the User Thread Mode is User Default (empty string) then sets it to Admin Default.
Now, this creates problem.
If Admin has set to "Threaded" or "Linear" it sets User Thread Mode instead of leaving the field unset.
Now, if admin re-changes the field, it will not change the User Thread Mode again as it is already been set.

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -690,7 +690,7 @@ class UserDataHandler extends DataHandler
 		{
 			$options['dst'] = 0;
 		}
-		
+
 		if($this->method == "insert" || (isset($options['threadmode']) && $options['threadmode'] != "linear" && $options['threadmode'] != "threaded" && $options['threadmode'] != ''))
 		{
 			$options['threadmode'] = '';
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -690,6 +690,11 @@ class UserDataHandler extends DataHandler
 		{
 			$options['dst'] = 0;
 		}
+		
+		if($this->method == "insert" || (isset($options['threadmode']) && $options['threadmode'] != "linear" && $options['threadmode'] != "threaded" && $options['threadmode'] != ''))
+		{
+			$options['threadmode'] = '';
+		}
 
 		// Verify the "threads per page" option.
 		if($this->method == "insert" || (array_key_exists('tpp', $options) && $mybb->settings['usertppoptions']))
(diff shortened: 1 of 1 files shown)


JordanMussi (changes requested):
Thanks for the PR!

I don't think we should be removing the check for a valid value since we still want the field to only be `''`, `'threaded'`, or `'linear'`. What do you think about the following?

```php
		if($this->method == "insert" || (isset($options['threadmode']) && $options['threadmode'] != "linear" && $options['threadmode'] != "threaded" && $options['threadmode'] != ''))
		{
			$options['threadmode'] = '';
		}
```

rajat315315:
@JordanMussi Thanks for the help. Totally missed checking the input.

inc/datahandlers/user.php
```diff
@@ -690,17 +690,10 @@ function verify_options()
 		{
 			$options['dst'] = 0;
 		}
-
-		if($this->method == "insert" || (isset($options['threadmode']) && $options['threadmode'] != "linear" && $options['threadmode'] != "threaded"))
+		
+		if($this->method == "insert" || (isset($options['threadmode']) && $options['threadmode'] != "linear" && $options['threadmode'] != "threaded" && $options['threadmode'] != ''))
```
JordanMussi, line 694:
```suggestion

		if($this->method == "insert" || (isset($options['threadmode']) && $options['threadmode'] != "linear" && $options['threadmode'] != "threaded" && $options['threadmode'] != ''))
```

JordanMussi (approved):

JordanMussi:
@rajat315315 thank you for your first contribution to be merged into MyBB! 😀

PR #4105, opened 2020-07-01
Added error handling in (and,or) keyword search

### LINKED ISSUE #4033: Search fails to throw errors if keywords is met with specific criteria
For standard search, keywords like `or`, `orand`, `and`, etc.: after `clean_keywords()` on input, the variable holding the keywords is an empty string.

And for fulltext search, keywords whose length is less than `minsearchword`: after `clean_keywords_ft()` on input, the variable holding the keywords is an empty string.

Both `perform_search_mysql()` and `perform_search_mysql_ft()` will neither process empty strings as keywords nor throw any error. And finally all records matching other search criteria will be returned.

rajat315315:
Closes #4033

    1) Removes starting "and"/"or" and appends a " " space.
    2) Throws error if keyword is empty.

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -199,7 +199,7 @@ function get_password_protected_forums($fids=array())
  */
 function clean_keywords($keywords)
 {
-	global $db;
+	global $db, $lang;
 
 	$keywords = my_strtolower($keywords);
 	$keywords = $db->escape_string_like($keywords);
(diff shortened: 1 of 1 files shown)


inc/functions_search.php
```diff
@@ -213,11 +213,18 @@ function clean_keywords($keywords)
...
+		$keywords = " ".$keywords;
 	}
 
 	if(my_strpos($keywords, "and") === 0)
 	{
 		$keywords = substr_replace($keywords, "", 0, 3);
+		$keywords = " ".$keywords;
+	}
+
+	if(!$keywords)
```
euantorano, line 225: I think this check would be better in the `perform_search_mysql` function, based on the following usages:

- [`privatemessage_perform_search_mysql`](https://github.com/mybb/mybb/blob/f538e4403bd76036af5f6f0dc4be86c84563df60/inc/functions_search.php#L492) - already checks if there are keywords and does the same error.
- [`helpdocument_perform_search_mysql`](https://github.com/mybb/mybb/blob/f538e4403bd76036af5f6f0dc4be86c84563df60/inc/functions_search.php#L809) - already checks if there are keywords and does the same error.

Otherwise, this looks good.

Eldenroot: @euantorano - is possible to finish this without @rajat315315 if he will not respond? Quite useful fix 👍

euantorano:
Yes, I believe the PR is open to contributors.

PR #4114, opened 2020-07-10
Bug in calculating merged reputation

### LINKED ISSUE #3509: Reputation is lost on account merge
Related to #2020, was decided that a separate PR would be good in #3405.

rajat315315:
Resolves #3509

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -2019,7 +2019,7 @@ if($mybb->input['action'] == "merge")
 			);
 
 			$to_remove = array();
-			$query = $db->simple_select("reputation", "*", "adduid = '".$destination_user['uid']."'");
+			$query = $db->simple_select("reputation", "*", "adduid = '".$destination_user['uid']."'", $options);
 			while($rep = $db->fetch_array($query))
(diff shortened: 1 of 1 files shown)


Sama34:
https://github.com/mybb/mybb/issues/3509#issuecomment-656851884

> I have created a PR which resolves a major BUG in reputation merging, which might have caused almost all merges untill now to be incorrect.

Is this the case? I don't see how the options should affect  the merge at all.

JordanMussi:
It’s because the order is significant due to the comparisons with`$last_result`

rajat315315:
@Sama34 Consider a case, when a third user has increased reputation of both merged_user_1 & merged_user_2.
It would be double counted in current scenario, right?
The only way to get rid of it is sorting by `uid` using `$options` and then using `$last_result` to see if he has already upvoted either merged_user_1 & merged_user_2.

Other way you can think is that why is the `$options` array is there at the first place?

rajat315315:
Any possibility of merging? updates?

euantorano:
Apologies, I'll review this weekend.

Issue #4084, opened 2020-06-19
Allow users to manage own orphan attachments they have uploaded


Ben-MyBB:
A feature will be good to allow users to manage own orphaned attachments they have uploaded. This is helpful for the usergroups with restricted attachment quota, so that someone don't land up to a situation of having no attachment attached with any post while exhausted allotted quota.

This can be driven with a setting per usergroup basis.
Currently user can freely delete their attachments. Proposed change will allow Admins to define which usergroup can delete own attachment as well as whether they are allowed to delete orphaned attachments or not.

Original thread: [Allow users to manage own orphan attachments they have uploaded](https://community.mybb.com/thread-228135.html)

Ben-MyBB:
@euantorano Given the ongoing rebase, I'm not sure if you want this to be targeted for 1.9 or are happy with 1.8? (If someone is willing to work on it)

euantorano:
I think the first step to deciding that is determining how much effort it will take. If it’s going to involve a bunch of new templates or template edits, then my preference would be 1.9.

effone:
As much I can see a maximum of 2 templates will be added, if we at all can't modify existing one.
If acceptance provided I can try working on it.

Ben-MyBB:
@euantorano any further thoughts? If your happy for this to be added could @effone you take a look at #4117 since it is related please?

Azareal:
The existence of orphaned attachments seems like a bug in and of itself. Perhaps, the software could detect these and delete them automatically?

Ben-MyBB:
> 
> 
> The existence of orphaned attachments seems like a bug in and of itself. Perhaps, the software could detect these and delete them automatically?

Did you get a chance to read through the thread on the community forums? It was discussed about deleting automatically but thought that it may be useful in some scenarios for them to remain.

Azareal:
Let's say a user uploads a file as an attachment. It fails.

Six months later, they discover it lingered around on the server. This isn't really something a user would expect and it complicates reasoning about attachments.

PR #4118, opened 2020-07-11
Delete attachments button removal when 0 attachments (JS fix)

### LINKED ISSUE #4117: Delete Selected Attachments button is visible if i haven't got any attachments
Hello,

In /usercp.php?action=attachments

We see this message: You currently do not have any files attached to your posts. and despite that we have this button Delete Selected Attachments
I just want to remind you that we don't have any attachments let alone selected attachments

I think this button could be visible only if we have available attachments (but best if we chose some attachments)

Original thread: [Delete Selected Attachments button is visible if i haven't got any attachments](https://community.mybb.com/thread-228515.html)

rajat315315:
Resolves #4117

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -14356,5 +14356,10 @@ $(function() {
 </tr>]]></template>
 		<template name="task_image" version="1800"><![CDATA[<img src="{$mybb->settings['bburl']}/task.php" width="1" height="1" alt="" />]]></template>
 		<template name="smilie" version="1800"><![CDATA[<img src="{$smilie['image']}" alt="{$smilie['name']}" title="{$smilie['name']}" class="smilie smilie_{$smilie['sid']}{$extra_class}"{$onclick} />]]></template>
+	<template name="delete_attachments_button" version="1800"><![CDATA[		
+		<div align="center">
+						<input type="hidden" name="action" value="do_attachments" />
+						<input type="submit" class="button" value="{$lang->delete_attachments}" />
+					</div>]]></template>
 	</templates>
 </theme>
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -14356,7 +14356,7 @@ $(function() {
 </tr>]]></template>
 		<template name="task_image" version="1800"><![CDATA[<img src="{$mybb->settings['bburl']}/task.php" width="1" height="1" alt="" />]]></template>
 		<template name="smilie" version="1800"><![CDATA[<img src="{$smilie['image']}" alt="{$smilie['name']}" title="{$smilie['name']}" class="smilie smilie_{$smilie['sid']}{$extra_class}"{$onclick} />]]></template>
-	<template name="delete_attachments_button" version="1800"><![CDATA[		
+	<template name="delete_attachments_button" version="1825"><![CDATA[		
 		<div align="center">
 						<input type="hidden" name="action" value="do_attachments" />
 						<input type="submit" class="button" value="{$lang->delete_attachments}" />
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -12051,6 +12051,14 @@ if(use_xmlhttprequest == "1")
 <head>
 	<title>{$mybb->settings['bbname']} - {$lang->attachments_manager}</title>
 	{$headerinclude}
+		<script>
+		$( document ).ready(function() {
+			var numAttach = $('input[type=checkbox][name^=attachments]').length
+			if(numAttach == 0){
+				$('input[type=submit][class=button]').hide();			
+			}
+		});
+	</script>
 </head>
 <body>
 	{$header}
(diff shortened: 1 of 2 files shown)


Eldenroot:
Better to get rid of js in templates, use as an external source.

Ben-MyBB:
Thanks for another PR! I haven’t reviewed this one but we do prefer JS to kept out of the templates

rajat315315:
> Better to get rid of js in templates, use as an external source.

I also thought of using external js but we do not load js on all pages. (atleast on delete attachments page.)
Also, it is not a good idea to make a separate js containing only this code for this particular page

lairdshaw:
@rajat315315, have you thought about including relevant Javascript in `jscripts/post.js`? It looks to me that you might be able to slot appropriate code similar to that which you've provided immediately after [line 146](https://github.com/mybb/mybb/blob/548b6949ff2985f99c140cbff59bf1804026a2f6/jscripts/post.js#L146), although I haven't tested this.

rajat315315:
@lairdshaw Delete attachments doesn't use ajax as opposed to remove attachments (in preview post).
User gets a page refresh after he clicks delete attachments button. Had it been using an Ajax request, I would have already put my js code there.

rajat315315:
One way out can be to include `jscripts/post.js` in this page and call my javascript code from there.
Any suggestions?

lairdshaw:
> @lairdshaw Delete attachments doesn't use ajax as opposed to remove attachments (in preview post).

Apologies, I had misunderstood the page on which this code was to be active. Without checking the code itself, I wonder whether this could be handled by conditional templates: that is, the "Delete Selected Attachments" button is eval'd to be empty when there are no attachments to be deleted, and otherwise to its current, default value?

rajat315315:
@lairdshaw yeah, that is also one nice way of doing it.
Is it preferred over including JS in templates? If yes, then I would make the corresponding change.

lairdshaw:
That looks a lot better to me at least, @rajat315315. One suggestion I do have though is that the setting of `$delete_button` on lines 4026-4030 of `usercp.php` would be better as the evaluation of some sort of `delete_button` **_template_** rather than hard-coded.

rajat315315:
@lairdshaw I too thought of same but adding another template just for that button seemed too much. Although that is a nice and clean way of doing it.

lairdshaw:
Yep, @rajat315315, I suspect that your PR won't be merged unless you move that hard-coded HTML+$lang variable out into a template, but I'm just an occasional contributor, not a MyBB team member, so I could be wrong about that.

Eldenroot:
I have same feeling

install/resources/mybb_theme.xml
```diff
@@ -14359,5 +14356,10 @@ $(function() {
 </tr>]]></template>
 		<template name="task_image" version="1800"><![CDATA[<img src="{$mybb->settings['bburl']}/task.php" width="1" height="1" alt="" />]]></template>
 		<template name="smilie" version="1800"><![CDATA[<img src="{$smilie['image']}" alt="{$smilie['name']}" title="{$smilie['name']}" class="smilie smilie_{$smilie['sid']}{$extra_class}"{$onclick} />]]></template>
+	<template name="delete_attachments_button" version="1800"><![CDATA[
```
euantorano: Can we change this template version to match the next release? I'm happy to merge once that's done 😄 

```suggestion
	<template name="delete_attachments_button" version="1825"><![CDATA[		
```

euantorano (changes requested):
Looks good, thanks for the contribution - we just need to update the template version due to how long it's taken to review (sorry!).

euantorano (approved):

install/resources/mybb_theme.xml
```diff
@@ -14359,5 +14356,10 @@ $(function() {
 </tr>]]></template>
 		<template name="task_image" version="1800"><![CDATA[<img src="{$mybb->settings['bburl']}/task.php" width="1" height="1" alt="" />]]></template>
 		<template name="smilie" version="1800"><![CDATA[<img src="{$smilie['image']}" alt="{$smilie['name']}" title="{$smilie['name']}" class="smilie smilie_{$smilie['sid']}{$extra_class}"{$onclick} />]]></template>
+	<template name="delete_attachments_button" version="1825"><![CDATA[
```
effone, line 14359: Sorry to be late to the hall, but this button belongs to `User Control Panel Templates` Group and hence should be named  with prefix `usercp_` otherwise this template will land into some other place / group.

effone (commented):
If we care about the unwanted delete button then why not the unwanted select-all check box at the same time which is located under `usercp_attachments` template?

install/resources/mybb_theme.xml
euantorano: Please open a PR to fix this @effone

euantorano:
> If we care about the unwanted delete button then why not the unwanted select-all check box at the same time which is located under `usercp_attachments` template?

Yes, we should probably resolve that too, but that can be done in another PR.

install/resources/mybb_theme.xml
effone: OK. I will address this with the Attachment PR for #4084 I am currently working on.

effone: Making a sidenote for self...
This new template should be cached in `usercp.php` as well.

PR #4121, opened 2020-07-13
use cookieless domain for yt video embed

### LINKED ISSUE #4149: Allow use of cookieless domain for YouTube video embeds
Via #4121

Azareal:
Resolves #4149

Context: https://community.mybb.com/thread-228558.html

It might need a bit more testing before being merged.


euantorano:
I think the `video_youtube_embed` template should also be updated: https://github.com/mybb/mybb/blob/feature/install/resources/mybb_theme.xml#L13757

Azareal:
> I think the `video_youtube_embed` template should also be updated: https://github.com/mybb/mybb/blob/feature/install/resources/mybb_theme.xml#L13757

It should be updated.

euantorano:
Oh, somehow I looked at the PR and managed to miss the fact it was already. Doh!

JordanMussi:
I think we possibly need to accommodate the fact some forums may be relying on the fact users are logged in to YouTube and have only shared the videos with select users. The privacy focused embed would fail because no user would be logged in for this scenario.

I think we should probably either have a setting or decide that the scenario I have described is sufficiently unlikely.

euantorano:
> I think we should probably either have a setting or decide that the scenario I have described is sufficiently unlikely.

Administrators could just edit the template to support this, I'm not sure if we need to over-complicate it with a setting? I'd be interested to know how many forums are currently relying on this...

Eldenroot:
Just merge it, 99,99% of admins do not know that.

PR #4131, opened 2020-07-19
1.9 Development issues, style issues fix

### LINKED ISSUE #3785: showteam page doesn't work with group leaders

rajat315315:
Resolves #3785

### DIFF
--- a/inc/views/base/usercp/attachments.twig
+++ b/inc/views/base/usercp/attachments.twig
@@ -46,7 +46,7 @@
             </li>
             <li class="list__item">
                 <span class="list__field">{{ lang.attachstats_spaceused }}</span>
-                <span class="list__value">{{ friendly_usage }} ({{ percent }})</span>
+                <span class="list__value">{{ friendly_usage }} {{ (percent) }}</span>
             </li>
             <li class="list__item">
                 <span class="list__field">{{ lang.attachstats_quota }}</span>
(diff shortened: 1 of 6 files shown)


inc/views/base/usercp/attachments.twig
```diff
@@ -46,7 +46,7 @@
             </li>
             <li class="list__item">
                 <span class="list__field">{{ lang.attachstats_spaceused }}</span>
-                <span class="list__value">{{ friendly_usage }} ({{ percent }})</span>
+                <span class="list__value">{{ friendly_usage }} {{ (percent) }}</span>
```
euantorano, line 49: I'm not sure that this change is correct - I believe the intention was to wrap the output in `()`, but this just wraps the variable and won't be shown in the output.

euantorano (changes requested):
This looks good, thanks. Just a minor comment.

inc/views/base/usercp/attachments.twig
rajat315315: The issue was that when percent is zero, empty brackets were shown.
This change will show `()` only when percent is NOT NULL ??

euantorano: Ah, ok that makes sense.

rajat315315 (commented):
Changes requested resolved.

euantorano (approved):

rajat315315:
Hello everyone,
Can somebody please nominate me, if I have done some meaningful contribution at the community?
[https://stars.github.com/nominate/](https://stars.github.com/nominate/)

PR #4135, opened 2020-07-23
Fix #4134 - add mail handler hook

### LINKED ISSUE #4134: Add hook for custom mail handler registration
The core doesn't currently have any way to create a custom mail handler from a plugin. It checks for SMTP, then defaults to PHP mail.

The use case is a plugin to send emails using the AWS SES API (and other APIs for similar services such as MailGun, Sendgrid, Postmark, etc.).

https://github.com/mybb/mybb/blob/1b25406ba6b6586f060547dc41e7af653f182276/inc/functions.php#L577:L586

euantorano:
Resolves #4134.

This adds a new hook to `my_mail` called `my_mail_mailhandler_init`. The hook is passed a reference to the `$mail` variable and should set up a new mail handler instance.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -631,7 +631,7 @@ function my_mail($to, $subject, $message, $from="", $charset="", $headers="", $k
 	$mail = &get_my_mailhandler();
 
 	// If MyBB's built-in SMTP mail handler is used, set the keep alive bit accordingly.
-	if(isset($mybb->settings['mail_handler']) && $mybb->settings['mail_handler'] == 'smtp' && $keep_alive == true && !empty($mail->keep_alive))
+	if($keep_alive == true && isset($mail->keep_alive) && isset($mybb->settings['mail_handler']) && $mybb->settings['mail_handler'] == 'smtp')
 	{
 		require_once MYBB_ROOT . "inc/class_mailhandler.php";
 		require_once MYBB_ROOT . "inc/mailhandlers/smtp.php";
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -550,6 +550,64 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
 }
 
 /**
+ * Get a mail handler instance, a MyBB's built-in SMTP / PHP mail hander or one created by a plugin.
+ * @param bool $use_buitlin Whether to use MyBB's built-in mail handler.
+ *
+ * @return object A MyBB's built-in mail handler or one created by plugin(s).
+ */
+function &get_mybb_mailhandler($use_buitlin = false)
+{
+	global $mybb, $plugins;
+	static $mybb_mailhandler;
+	static $mybb_mailhandler_builtin;
+
+	if($use_buitlin)
+	{
+		// If our built-in mail handler doesn't exist, create it.
+		if(!is_object($mybb_mailhandler_builtin))
+		{
+			require_once MYBB_ROOT . "inc/class_mailhandler.php";
+
+			// Using SMTP.
+			if(isset($mybb->settings['mail_handler']) && $mybb->settings['mail_handler'] == 'smtp')
+			{
+				require_once MYBB_ROOT . "inc/mailhandlers/smtp.php";
+				$mybb_mailhandler_builtin = new SmtpMail();
+			}
+			// Using PHP mail().
+			else
+			{
+				require_once MYBB_ROOT . "inc/mailhandlers/php.php";
+				$mybb_mailhandler_builtin = new PhpMail();
+				if(!empty($mybb->settings['mail_parameters']))
+				{
+					$mybb_mailhandler_builtin->additional_parameters = $mybb->settings['mail_parameters'];
+				}
+			}
+		}
+
+		return $mybb_mailhandler_builtin;
+	}
+
+	// If our mail handler doesn't exist, create it.
+	if(!is_object($mybb_mailhandler))
+	{
+		require_once MYBB_ROOT . "inc/class_mailhandler.php";
+
+		$plugins->run_hooks('mybb_mailhandler_init', $mybb_mailhandler);
+
+		if(!is_object($mybb_mailhandler) || !($mybb_mailhandler instanceof MailHandler))
+		{
+			$mybb_mailhandler = &get_mybb_mailhandler(true);
+		}
+
+		$plugins->run_hooks('mybb_mailhandler_after_init', $mybb_mailhandler);
+	}
+
+	return $mybb_mailhandler;
+}
+
+/**
  * Sends an email using PHP's mail function, formatting it appropriately.
  *
  * @param string $to Address the email should be addressed to.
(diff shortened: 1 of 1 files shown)


yuliu:
I prefer in this way of forcing custom mail handlers to inherit from class `MailHandler` to assure the function's using of `->build_message()` and `->send()` methods. Not sure if it'll work:

```php
	$plugins->run_hooks('my_mail_mailhandler_init', $mail);

	// If our mail handler still doesn't exist or isn't a mail handler, create it
	if(!isset($mail) || !($mail instanceof MailHandler))
	{
		require_once MYBB_ROOT."inc/class_mailhandler.php";

		if($mybb->settings['mail_handler'] == 'smtp')
		{
			require_once MYBB_ROOT."inc/mailhandlers/smtp.php";
			$mail = new SmtpMail();
		}
		else
		{
			require_once MYBB_ROOT."inc/mailhandlers/php.php";
			$mail = new PhpMail();
		}
	}
```

But the `->send()` method isn't declared in the base class..

Anyway, your way works and we may do more comprehensive upgrading to mailing in 1.9, such as to move it to the MyBB namespace.

euantorano:
I definitely want to see an overhaul in the future for definite.

euantorano:
Please don't merge this until discussion here is resolved: https://github.com/euantorano/mybb/pull/1

euantorano:
Merged @yuliu's changes into my branch, so this is ready for review.

If it is not accepted by this weekend, the change will unfortunately have to be postponed until 1.9.

inc/functions.php
```diff
@@ -566,47 +624,73 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
-		{
-			require_once MYBB_ROOT."inc/mailhandlers/php.php";
-			$mail = new PhpMail();
-		}
-	}
-
-	// Using SMTP based mail
-	if($mybb->settings['mail_handler'] == 'smtp')
+	// If MyBB's built-in SMTP mail handler is used, set the keep alive bit accordingly.
+	if(isset($mybb->settings['mail_handler']) && $mybb->settings['mail_handler'] == 'smtp' && $keep_alive == true && !empty($mail->keep_alive))
```
Sama34: What if the custom handler set `$mail->keep_alive` to `false`?

inc/functions.php
```diff
@@ -566,47 +624,73 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
 	{
-		if($keep_alive == true)
+		require_once MYBB_ROOT . "inc/class_mailhandler.php";
+		require_once MYBB_ROOT . "inc/mailhandlers/smtp.php";
+		if($mail instanceof MailHandler && $mail instanceof SmtpMail)
```
Sama34, line 638: Isn't `instanceof MailHandler` already checked above?

inc/functions.php
```diff
@@ -566,47 +624,73 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
 	$mail->build_message($to, $subject, $message, $from, $charset, $headers, $format, $message_text, $return_email);
-	return $mail->send();
+
+	// Following variables will help sequential plugins to determine how to process the hook.
+	// Mark this variable true if the hooked plugin has sent the mail, otherwise don't modify it.
+	$is_mail_sent = false;
+	// Mark this variable false if the hooked plugin doesn't suggest sequential plugins to continue processing.
+	$continue_process = true;
+
+	$my_mail_parameters = compact("to", "subject", "message", "from", "charset", "headers", "keep_alive", "format", "message_text", "return_email", "is_mail_sent", "continue_process");
```
Sama34: Why aren't we passing by reference here?

inc/functions.php
```diff
@@ -549,6 +549,64 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
 	return $date;
 }
 
+/**
+ * Get a mail handler instance, a MyBB's built-in SMTP / PHP mail hander or one created by a plugin.
+ * @param bool $use_buitlin Whether to use MyBB's built-in mail handler.
+ *
+ * @return object A MyBB's built-in mail handler or one created by plugin(s).
+ */
+function &get_mybb_mailhandler($use_buitlin = false)
```
Sama34: I think the `mybb_` prefix within this function is superfluous, specially in the function and hook names.

inc/functions.php
yuliu: No it isn't.

inc/functions.php
yuliu: Sorry I don't quite understand you here. Did you mean passing the variable to the following hook by reference?

inc/functions.php
yuliu: Hmm, I didn't think about that before. Now I have two perspectives on behalf of your comment:
1. How can a custom mail handler set keep_alive?
2. Since here keep_alive is set to true, what if a custom mail handler is going to set it to false?

I guess the second one is what you mean.

I can't think of a good way to set keep_alive here for custom handlers, as we may not be sure if the custom handler could have a similar member variable `$keep_alive` (so is `build_message` method below before actual sending). Anyway, it can be set via any one of the following two hooks just before sending the mail, though it's not quite intuitive.

inc/functions.php
yuliu: Oh, yes! Using `my_` as prefix would be better which also consists with the current function naming convention.

inc/functions.php
euantorano: I believe this is the intention:

```suggestion
	$my_mail_parameters = array(
		'to' => &$to,
		'subject' => &$subject,
		'message' => &$message,
		'from' => &$from,
		'charset' => &$charset,
		'headers' => &$headers,
		'keep_alive' => &$keep_alive,
		'format' => &$format,
		'message_text' => &$message_text,
		'return_email' => &$return_email,
		'is_mail_sent' => &$is_mail_sent,
		'continue_process' => &$continue_process,
	);
```

yuliu: The suggestion has been made, and the variable is much simpler and the code looks much cleaner now. Had a quick test with no plugin hooked and the pre send hook, all went well.

inc/functions.php
```diff
@@ -562,51 +620,64 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
+		'headers' => &$headers,
+		'keep_alive' => &$keep_alive,
+		'format' => &$format,
+		'message_text' => &$message_text,
+		'return_email' => &$return_email,
+		'is_mail_sent' => &$is_mail_sent,
+		'continue_process' => &$continue_process,
+	);
+
+	$my_mail_parameters = $plugins->run_hooks('my_mail_pre_build_message', $my_mail_parameters);
```
```suggestion
	$plugins->run_hooks('my_mail_pre_build_message', $my_mail_parameters);
```

inc/functions.php
```diff
@@ -562,51 +620,64 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
  * @param string $message_text The text message of the email if being sent in html format, for email clients that don't support html
  * @param string $return_email The email address to return to. Defaults to admin return email address.
- * @return bool
+ * @return bool True if the mail is sent, false otherwise.
  */
 function my_mail($to, $subject, $message, $from="", $charset="", $headers="", $keep_alive=false, $format="text", $message_text="", $return_email="")
 {
-	global $mybb;
+	global $mybb, $plugins;
 	static $mail;
```
Sama34: Is there a need for this than `static $my_mailhandler;` above doesn't accomplish?

inc/functions.php
```diff
@@ -549,6 +549,64 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
+			// Using PHP mail().
+			else
+			{
+				require_once MYBB_ROOT . "inc/mailhandlers/php.php";
+				$my_mailhandler_builtin = new PhpMail();
+				if(!empty($mybb->settings['mail_parameters']))
+				{
+					$my_mailhandler_builtin->additional_parameters = $mybb->settings['mail_parameters'];
+				}
+			}
```
Sama34, line 586: I won't be sure, but perhaps a hook here could be handy? Maybe not much.

inc/functions.php
```diff
@@ -549,6 +549,64 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
+ * Get a mail handler instance, a MyBB's built-in SMTP / PHP mail hander or one created by a plugin.
+ * @param bool $use_buitlin Whether to use MyBB's built-in mail handler.
+ *
+ * @return object A MyBB built-in mail handler or one created by plugin(s).
+ */
+function &get_my_mailhandler($use_buitlin = false)
+{
+	global $mybb, $plugins;
+	static $my_mailhandler;
+	static $my_mailhandler_builtin;
```
Sama34, line 562: Seems like in a default install both `$my_mailhandler_builtin` and `$my_mailhandler` would store the same data, is there a real need to keep both variables?

inc/functions.php
Sama34: I'm a little confused, here, `!empty($mail->keep_alive)` seems to check for not false, , which makes it redundant to set to `true` within it (`$mail->keep_alive = true;` below). Perhaps an `else` statement is missing?

Perhaps this block should be moved into `get_my_mailhandler()` ?

yuliu: > I'm a little confused, here, `!empty($mail->keep_alive)` seems to check for not false, , which makes it redundant to set to `true` within it (`$mail->keep_alive = true;` below). Perhaps an `else` statement is missing?

Oh, yes, that's my fault. I meant `isset()` actually. Will fix it soon.


> Perhaps this block should be moved into `get_my_mailhandler()` ?

I'm not sure about that. There are two reasons for me to keep setting `keep_alive` here, one is to maintain the compatibility with the function calling `->build()` below and the other is, maybe the same reason as it was done here before, to make it possible to control `keep_alive` upon sending each mail.

I think it'll be better to get it moved/removed in future.

inc/functions.php
yuliu: Perhaps a custom mail handler is a wrapper of builtin mail handler? So when creating a custom one, it could get the default builtin one first.

inc/functions.php
yuliu: What's the purpose of this hook? We may add it later if we're not so sure about it at present time.

inc/functions.php
yuliu: Oh, that's right!

inc/functions.php
yuliu: And by the way, if a custom mail handler is not created, `$my_mailhandler` contains a reference to the object stored in `$my_mailhandler_builtin`.

inc/functions.php
yuliu: Hmm, on a second thought, the two hooks below may be useful here as well. It was that in Euan's original PR.

inc/functions.php
```diff
@@ -549,6 +549,65 @@ function my_date($format, $stamp=0, $offset="", $ty=1, $adodb=false)
...
+
+		$plugins->run_hooks('my_mailhandler_init', $my_mailhandler);
+
+		// If no plugin creates the mail handler, resort to use the built-in one.
+		if(!is_object($my_mailhandler) || !($my_mailhandler instanceof MailHandler))
+		{
+			$my_mailhandler = &get_my_mailhandler(true);
+		}
+
+		$plugins->run_hooks('my_mailhandler_after_init', $my_mailhandler);
```
Sama34: I'm now being picky, but we usually use the `_start` and `_end` suffixes for hooks that need it. Shouldn't be something to stop this from being merged of course.

inc/functions.php
Sama34: I think this assumes `get_my_mailhandler(true)` will never be used by third parties, which might not be the case.

inc/functions.php
Sama34: I will concede on this one then.

inc/functions.php
yuliu: OK, I'll rename them accordingly shortly. However, maybe just like you talked about using `get_my_mailhandler(true)`, here the latter hook may be there right after the former hook initialized a custom handler with the in-between `if-block` unsatisfied. I just don't know what exactly should be done here right now.

yuliu: Well, I think it's here to discuss about this comment:

> I think this assumes get_my_mailhandler(true) will never be used by third parties, which might not be the case.

Yes, that's right, maybe.

If no plugin hooks to create a custom handler before calling my_mail, then nothing can be done to create a custom handler. So I'm thinking about returning the builtin one when `use_builtin` is `false other than assigning it to `$my_mailhandler`, which means that just leave it as empty.

Sama34: I think you might be right about the hook mentioned above being unnecessary at this moment, I have  resolved the conversation. Developers could use `get_my_mailhandler()` and get the built in handler overwritten that way.

> So I'm thinking about returning the builtin one when use_builtin is false other than assigning it to $my_mailhandler`, which means that just leave it as empty.

Having resolved my opinion on the above issue, I think this hook should stay as is (to assign it to and return `$my_mailhandler`).

If you think any of my points is still valid for further consideration feel free to redo the resolves.

Sama34 (approved):

yuliu:
Thank you @Sama34 for the review. I have some further commits to address the mail handler issue you mentioned above. I'll push them today and another review may be needed before merging it

yuliu:
@Sama34,

I've done those changes about my_mailhandler hooks. I think that would suffice. Feel free to review it.

By the way, the hook `my_mail_pre_build_message` is supposed to change `$my_mail_parameters`. Shouldn't we use the assignment expression `$my_mail_parameters = $plugins->runhooks('my_mail_pre_build_message', ....`? Current method suggested by you works, too. However, if you ever want to change your mind, you may suggest edits and commit it directly, otherwise, we leave it as is.

Sama34 (approved):
@yuliu I'm fine with the latest changes you pushed (hook rearrangement).

If I recall correctly the usage of `$var = $plugins->run_hooks('foo', $var);` is a legacy for 1.4 when passing by reference worked different, but is no longer required, thus the suggested change.

PR #4140, opened 2020-08-07
Picture extensions

### LINKED ISSUE #4139: Hardcoded picture extensions limits the creation of thumbnails
in functions_upload.php, the extensions for pictures are hard coded, so adding new pictures extensions (webp, jfif, ...) won't work correctly:
- new formats won't be accepted as avatars (regexp line 214)
- when attached, no thumbnail will be created (condition line 548)

Related: https://community.mybb.com/thread-228933.html and https://community.mybb.com/thread-228935.html

crazycatdevs:
Resolves #4139 
Uses attachment cache to create the list of allowed extensions for avatar and for thumbnail generation


Stefan-MyBB:
First of all avatar types should not depend on attachment type settings. Also current code doesn't not support avatars and thumbnails from some image/* types users may add as attachment types. Webm is most likely not supported by older GD versions and SVG doesn't seem to be supported at all. Given the limited number of common and supported file extensions hard-coding them isn't that bad: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#Image_types

crazycatdevs:
Ok, so if hard-coding isn't bad, why aren't all _common and supported file extensions_ listed in the code ?

Stefan-MyBB:
Can you provide some prove that .jfif or .pjp are actually popular file extensions? Most information about .jfif is about a rar bug in Windows 10 accidentally using that file extension for jpeg files. Other projects like symfony/mime don't recognizes these file extensions either (https://github.com/symfony/mime/blob/master/MimeTypes.php#L1323).

crazycatdevs:
Sorry, you talk about common extensions and give a link to a page where jfif and webp are listed. I just ask why, if this page was a reference, every extension listed in it isn't allowed in MyBB.
Now, I know that some are really exotic, but when a mimetype is really common, why don't acept all its extensions ? Even if there is 0.000000001% of users which use it.
Not allowing ico,svg or tiff, I can understand. but any bmp, gif, jpg, png and webp extensions isn't stupid in my mind.

And if a forum admin want to allow avatar and attachment really exotic, why don't offer this possibility ?

BTW, I did this modification because someone ask for it, you can reject the PR, I don't really care. My "job" is done, I won't decide if it must be accepted or not.

Stefan-MyBB:
I'm just telling you how you could improve the PR. I'm not part of the team and do not decide anything.
The Mozilla link was just meant to be a fancy mime type list (illustrating why image/* is not a good idea) not a list of common or popular file extensions. Also your PR isn't even trying to support webp.

My opinion:
* Hard-coded extensions for avatars are reasonable. The files are stored on the server with that extension and need to be supported on all browsers and server environments. Also imagine adding .php with an image/ mime type, this would be a huge risk.
* Thumbnail generation could use mime types instead (but not image/*). The developers need to decide this.

Sama34:
I'm confused why we have an `Avatar File` setting for attachments yet it is not used there. Something like the following should work:
```PHP
	$attachtypes = (array)$cache->read('attachtypes');

	$allowed_mime_types = $allowed_file_types = array();
	foreach($attachtypes as $attachtype)
	{
		if(defined('IN_ADMINCP') || is_member($attachtype['groups']) && $attachtype['avatarfile'])
		{
			$allowed_mime_types[$attachtype['mimetype']] = $attachtype['maxsize'];
			$allowed_file_types[] = my_strtolower($attachtype['extension']);
		}
	}

	// Check we have a valid extension
    	$ext = get_extension(my_strtolower($avatar['name']));
    	if(!$allowed_file_types || !preg_match("#^(".implode('|', $allowed_file_types).")$#i", $ext))
```

You can move the following to avoid code duplication:
https://github.com/mybb/mybb/blob/feature/inc/functions_upload.php#L297-L307

For the thumbnail generation, I think we can go with a global setting in the attachments setting group, simple add all valid extensions from the source provided above avoiding invalid formats (SVG, as pointed by Stefan). Make sure only supported file types are selected by default, so I'd go with all current supported formats to be selected by default specially to avoid issues with old PHP versions if any.

Azareal:
I have encountered jfif files on occasion. You can use them identically to how you would use a jpeg.

effone:
What is the point of making 2 rarely used one liner functions?

PR #4141, opened 2020-08-07
No comma sign `,`

wlisikpl:
No comma sign `,` after last element in array[]
Resolves #ISSUE_ID

### DIFF
--- a/inc/class_session.php
+++ b/inc/class_session.php
@@ -42,7 +42,7 @@ class session
 	 */
 	public $ignore_parameters = array(
 		'my_post_key',
-		'logoutkey',
+		'logoutkey'
 	);
 
 	/**


euantorano:
Hi,

Including a trailing comma in an array declaration is perfectly valid, and some coding formats even require it (such as [the Zend coding standard](https://framework.zend.com/manual/2.4/en/ref/coding.standard.html#arrays)) actually require it for multi-line arrays.

Including a trailing comma makes it easier to add additional items in the future and doesn't cause any kind of overhead either computationally or in terms of developer effort.

wlisikpl:
@euantorano when I use my phpstorm (btw: excellent ide) and do put comma after last array element, I get warming, so it needs investigation.....

euantorano:
@wlisikpl What warning do you get in PHPStorm? I use it too and don't get any such warning. If you paste the warning message here, it might help.

wlisikpl:
@euantorano No specific textual one. IDE just underlines array definition and displays yellow bulb on the left margin

euantorano:
> @euantorano No specific textual one. IDE just underlines array definition and displays yellow bulb on the left margin

If you click the bubble, what are the available actions? I've just tried it, and the only thing I see is an option to convert the array to short syntax. We can't use the short syntax in MyBB 1.8 as it supports PHP 5.2, and PHP 5.2 doesn't support the short syntax for arrays.

wlisikpl:
@euantorano indeed thats the only option available, but, after some digging, I have found [this]https://stackoverflow.com/questions/6146713/no-comma-after-last-element-in-array/30438165).

Just adding to what Ive linked, I must say that the usage of comma `,` after last element (of an array) is discussable. Ive heard many times (and been tought) **not to** use it.

euantorano:
In my opinion, the benefits outweigh the downsides. Some of the benefits include:

- It's easier to add another entry onto the end of the array - no need to remember to add a comma to the previous entry.
- When adding an entry to an array in a PR, the generated diff in Git doesn't include an alteration to the previous entry, making the diff easier to read.

Trailing commas are [also supported in JavaScript](https://jsbin.com/rocorimewa/edit?html,js,output), which is the only other language that MyBB uses and they're likely spread throughout multiple files by now.

My normal practice when writing PHP is to always include a trailing comma.

wlisikpl:
> My normal practice when writing PHP is to always include a trailing comma.
Mine is just opposite, but if yours works for you, than why not? :)

Sama34:
I have always thought myself adding such trailing comma to be old-fashioned and to add unnecessary cluttering (which in turn makes less fancy coding standards).

However, I consider this to be unnecessary to fix and @euantorano points seem fair enough to not even bother.

In the event that anyone decides to contribute with this _change_ I'd say it would be necessary to apply it to all necessary places at once.

PR #4154, opened 2020-09-14
Resolves mybb#4153 - disable Purge Spammer IP ban (mybb#4153)

### LINKED ISSUE #4153: Purge Spammer should have a setting to skip adding the IP to the ban filters
Spammers are issued IPs from ISPs, which commonly match the regular users' IP addresses because of dynamic IPs. Often, admins who use Purge Spammer need to delete the IP ban filters otherwise their users (or even themselves) get IP banned from the board because the ISPs issue the same IP temporarily.

Should be a matter of adding a new setting for Purge Spammer (default true for non-breaking change) and in moderation.php executing this block if the setting is enabled:
```
				// Add the IP's to the banfilters
				foreach(array($user['regip'], $user['lastip']) as $ip)
				{
					$ip = my_inet_ntop($db->unescape_binary($ip));
					$query = $db->simple_select("banfilters", "type", "type = 1 AND filter = '".$db->escape_string($ip)."'");
					if($db->num_rows($query) == 0)
					{
						$insert = array(
							"filter" => $db->escape_string($ip),
							"type" => 1,
							"dateline" => TIME_NOW
						);
						$db->insert_query("banfilters", $insert);
					}
				}
```

DiogoParrinha:
Add a new purgespammerbanip setting (default is enabled for non-breaking change) to enable/disable adding the banned user's IP to the banned IP filters.

My git is configured to perform some identation conversions - not sure if this is alright for you but I use Linux line endings.

### DIFF
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -2418,6 +2418,16 @@ delete=Delete]]></optionscode>
 			<isdefault>1</isdefault>
 			<helpkey></helpkey>
 		</setting>
+		</setting>
+		<setting name="purgespammerbanip">
+			<title>Add banned IP filter</title>
+			<description><![CDATA[Do you want to ban the spammer's IP using this tool? If other users are issued the same banned IP, they will not be able to use the forums.]]></description>
+			<disporder>6</disporder>
+			<optionscode><![CDATA[yesno]]></optionscode>
+			<settingvalue><![CDATA[1]]></settingvalue>
+			<isdefault>1</isdefault>
+			<helpkey></helpkey>
+		</setting>
 	</settinggroup>
 	<settinggroup name="stopforumspam" title="Stop Forum Spam" description="This section allows you to change the settings used for the integration with StopForumSpam.com" disporder="27" isdefault="1">
 		<setting name="enablestopforumspam_on_register">
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -2422,7 +2422,7 @@ delete=Delete]]></optionscode>
 		<setting name="purgespammerbanip">
 			<title>Add banned IP filter</title>
 			<description><![CDATA[Do you want to ban the spammer's IP using this tool? If other users are issued the same banned IP, they will not be able to use the forums.]]></description>
-			<disporder>6</disporder>
+			<disporder>7</disporder>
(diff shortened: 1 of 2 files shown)


moderation.php
```diff
@@ -3063,20 +3063,22 @@
...
-					{
-						$insert = array(
-							"filter" => $db->escape_string($ip),
-							"type" => 1,
-							"dateline" => TIME_NOW
-						);
-						$db->insert_query("banfilters", $insert);
-					}
-				}
+                if($mybb->settings['purgespammerbanip'] == 1) {
```
euantorano: Nit-pick: Indentation should be tabs rather than spaces.

euantorano (changes requested):
Looks good, just a minor nit-pick regarding code style.

moderation.php
dvz: Also the `{` brace should be on a separate line; the PR includes some unrelated whitespace changes that can be dropped.

euantorano: I'm happy enough with the unrelated whitespace changes personally - they're mostly improvements to drop un-needed whitespace which won't cause any harm.

dvz: Generally, it can clutter up git history and code comparison - IMO better to introduce code style Checks and fix all files in one go (presumably 1.9).

DiogoParrinha: I think it's resolved, can you check?

install/resources/settings.xml
```diff
@@ -2418,6 +2418,16 @@ delete=Delete]]></optionscode>
 			<isdefault>1</isdefault>
 			<helpkey></helpkey>
 		</setting>
+		</setting>
+		<setting name="purgespammerbanip">
+			<title>Add banned IP filter</title>
+			<description><![CDATA[Do you want to ban the spammer's IP using this tool? If other users are issued the same banned IP, they will not be able to use the forums.]]></description>
+			<disporder>6</disporder>
```
Sama34: This should be 7, after this it should be fine to merge. Eventually we should get rid of these unnecessary elements.
```suggestion
			<disporder>7</disporder>
```

Sama34 (changes requested):

Sama34 (approved):

euantorano (approved):

PR #4159, opened 2020-09-27
Add mybb_posts.pid to ORDER BY clauses in posts-related queries

### LINKED ISSUE #4158: Posts with same dateline may be out of order
Processing and listing of posts depends on the `mybb_posts.dateline` timestamp only, and the intended order during insertion (especially when automated) may not be always preserved.

`ORDER BY` clauses should take advantage of `mybb_posts.pid`.

Verified on PostgreSQL.

Entities other than posts may be similarly affected.

dvz:
Resolves #4158

### DIFF
--- a/archive/index.php
+++ b/archive/index.php
@@ -144,7 +144,7 @@ switch($action)
 
 		$pids = array();
 		// Fetch list of post IDs to be shown
-		$query = $db->simple_select("posts", "pid", "tid='{$id}' AND visible='1'", array('order_by' => 'dateline', 'limit_start' => $start, 'limit' => $perpage));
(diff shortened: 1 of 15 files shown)


yuliu:
All seems good to me and there's just a small discussion/question that may be a little off-topic.

IMO, it'd be better to check the threads table to find if a post is the first post of a thread for a definite result. Anyway, checking posts table would return the same result, it's just that that's less intuitive. Is it for avoiding potential data racing problem?

And well.. for example, in `editpost.php`, why not using `$thread['firstpost']` (`$thread` defined at https://github.com/mybb/mybb/blob/mybb_1824/editpost.php#L59) for comparison when checking?

dvz:
Yes, it's possible individual locations have better alternatives.

> Is it for avoiding potential data racing problem?

Inserting multiple posts in the same thread at once causes them to be displayed in unexpected order.

PR #4164, opened 2020-10-11
Fixes #4123: Removes warning link if user cannot warn

### LINKED ISSUE #4123: The level of warnings in your profile leads to usercp.php
Hello,

in your profile e.g. 0% leads to usercp.php - i think for your own profile level of warnings could be not a link

Edit:
same in postbit_author_user warning level link

Original thread: [The level of warnings in your profile leads to usercp.php](https://community.mybb.com/thread-228511.html)

n-thumann:
Resolves #4123

This PR not only removes the check, if a user is his own profile (as [suggested](https://community.mybb.com/thread-228511.html)), but removes the link altogether, if the user isn´t permitted to warn:
Now users that are allowed to warn will always end up on `warnings.php`when clicking the percentage (even when they´re on their own profile) and users that are not won´t see a link at all (instead of a link to their own profile, so their currently opened site).

### DIFF
--- a/member.php
+++ b/member.php
@@ -17,7 +17,7 @@ $nosession['avatar'] = 1;
 
 $templatelist = "member_register,member_register_hiddencaptcha,member_register_coppa,member_register_agreement_coppa,member_register_agreement,member_register_customfield,member_register_requiredfields,member_profile_findthreads";
 $templatelist .= ",member_loggedin_notice,member_profile_away,member_register_regimage,member_register_regimage_recaptcha_invisible,member_register_regimage_nocaptcha,post_captcha_hcaptcha_invisible,post_captcha_hcaptcha,post_captcha_hidden,post_captcha,member_register_referrer";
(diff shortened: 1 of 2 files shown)


member.php
```diff
@@ -2520,16 +2520,17 @@
...
 		}
 
-		$warn_user = '';
-		$warning_link = 'usercp.php';
 		$warning_level = get_colored_warning_level($warning_level);
-		if($mybb->usergroup['canwarnusers'] != 0 && $memprofile['uid'] != $mybb->user['uid'])
+		if($mybb->usergroup['canwarnusers'] != 0)
 		{
 			eval("\$warn_user = \"".$templates->get("member_profile_warn")."\";");
 			$warning_link = "warnings.php?uid={$memprofile['uid']}";
```
Sama34: I see no reason to hard code this, could you move the link directly to the template?

Sama34 (changes requested):

Sama34:
Also, please do add `member_profile_warninglevel_link` to the template cache list.
https://github.com/mybb/mybb/blob/aacebaecb49d0cb951096134277250192a2cedba/member.php#L28

member.php
n-thumann: Good point, thanks :)

n-thumann:
Both comments addressed ✌️

Sama34 (approved):

PR #4170, opened 2020-11-15
Minor optimisation to thread moderation code

euantorano:
This is a small optimisation to the code merged in 5358ea2 - rather than building an array then immediately reading out of it, simply loop through the query results.


euantorano:
Any input @mybb/developers before I merge?

Sama34:
Seems fine to me. ~~Only issue I can predict is plugin backwards compatibility but I know no plugin for sure that would be affected so I wont stand on it.~~ Never mind I think this has never been on production code.

PR #4176, opened 2020-11-18
Fix 4175 Profile fields aren't loaded in quick replies

### LINKED ISSUE #4175: Profile fields aren't loaded in quick replies
Because of the following line profile fields are not parsed in quick replies.

https://github.com/mybb/mybb/blob/7c1c632771720181b0f0b5ea1bc2ff21bdecebea/inc/functions_post.php#L422

As $profile_fields is defined in showthread.php but not newreply.php.

Original thread: [Profile fields aren't loaded in quick replies](https://community.mybb.com/thread-230104.html)

Sama34:
Resolves #4175


euantorano:
The only thing to consider here is whether this may break any plugins by changing how profile fields are being read for post bit... I hope nobody was altering the global or anything weird like that.

Sama34:
I thought about that but ultimately decided to go with it. We can also keep it at the global scope rather than on local scope.

Keeping it at the global scope should be enough as the first hook that is executed right after defining the variable is apparently the postbit hooks anyways.

https://github.com/mybb/mybb/blob/48e107bc1a0a617e7c3fcbd38dce4c1c18b18f43/showthread.php#L821-L921

PR #4185, opened 2020-11-21
Fixes #4181 AdminCP > User Email Log should work with other log types

### LINKED ISSUE #4181: AdminCP > User Email Log should work with other log types
Currently AdminCP &gt; User Email Log can only correctly show type 1 (email users), 2 (send thread mails), 3(contact mails) for mail logs and that's very restrictive for the using of mail loggings.

https://github.com/mybb/mybb/blob/mybb_1...#L303-L399

I think we could display log entries of other mail types as general ones. And further, we could roll out a list of log types for recommended usages.

Original thread: [AdminCP > User Email Log should work with other log types](https://community.mybb.com/thread-229936.html)

Starpaul20:
Resolves #4181 

Emails which aren't of the three default types will be shown just as a generic email but will still display all relevent fields. Reduced almost all code duplication in the log.

Also fixed minor issue where find emails sent by user icon was being displayed on user 0.

### DIFF
--- a/admin/modules/tools/maillogs.php
+++ b/admin/modules/tools/maillogs.php
@@ -300,7 +300,12 @@ if(!$mybb->input['action'])
 		$log['subject'] = htmlspecialchars_uni($log['subject']);
 		$log['dateline'] = my_date('relative', $log['dateline']);
 
-		if($log['type'] == 2)
+		if($log['type'] == 1)
+		{
+			$table->construct_cell("<img src=\"styles/{$page->style}/images/icons/maillogs_user.png\" title=\"{$lang->email_sent_to_user}\" alt=\"\" />", array("width" => 1));
+			$table->construct_cell("<a href=\"javascript:MyBB.popupWindow('index.php?module=tools-maillogs&amp;action=view&amp;mid={$log['mid']}', null, true);\">{$log['subject']}</a>");
+		}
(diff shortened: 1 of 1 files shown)


Sama34 (approved):

Sama34:
Perhaps we can get a hook within that while loop?

admin/modules/tools/maillogs.php
```diff
@@ -300,7 +300,12 @@
 		$log['subject'] = htmlspecialchars_uni($log['subject']);
 		$log['dateline'] = my_date('relative', $log['dateline']);
 
-		if($log['type'] == 2)
+		if($log['type'] == 1)
```
euantorano, line 303: More magic numbers 😢 We seem to really like these in MyBB. A nice quality of life improvement for anyone interested would be a PR to replace these at some point in the future...

euantorano:
> Perhaps we can get a hook within that while loop?

A hook might be a nice idea, but doesn't necessarily need to be included with this PR - it could be done as a later PR. I'll merge this this weekend if there's no further movement.

Sama34:
I will merge this as is.

As future notes we should probably do a mass replacement of magic numbers and probably add a hook around the code provided here.

Eldenroot:
Would be nice to add a hook to extend the usability.

PR #4186, opened 2020-11-25
Fix #4184 and (partially) #4182 - SCEditor-related bugs

### LINKED ISSUE #4182: The 'double quotes in font lists' browser-independent SCEditor bug
As opposed to my previous two recent SCEditor bug reports, the 'redundant tags on paste of unformatted text' browser-dependent SCEditor bug and the 'reduced size on paste of unformatted text' browser-dependent SCEditor bug, this bug report references solely Hovatek's thread MyCode parse error ( [font="droid sans", sans-serif] ). Along with those two prior bug reports, this one is heavily based on my post #9 in Hovatek's thread, from the first horizontal line onwards.

Specific to this bug report (and note that as opposed to the previous two bugs which were browser-dependent, this bug is browser-independent):

The 'double quotes in font lists' browser-independent SCEditor bug

Description: Sometimes, in scenarios the criteria of which I haven't yet rigorously identified, when pasting text into the WYSIWYG editor, its specified font is actually a comma-separated list of two or more fonts, and at least one of the fonts in that comma-separated list is enclosed in double quotes. When the post is posted, the associated MyBB font tag(s) is/are not rendered by MyBB's PHP parser, because that parser doesn't support double-quotes within the comma-separated list of fonts of a MyCode font tag.

How to reproduce: In any browser, bring up a new editor window through a page such as a New Reply page. Select Source mode. Paste the following MyCode into the editor, after first replacing the curly braces with square braces (since I am posting this on a MyBB forum, I can't paste valid MyCode tags as they would then be rendered - or not, as per this bug!):

{font=Test, "Test two"}Blah{/font}

Switch out of Source mode and then back into it. Confirm that the code has not changed. Then submit or preview the post. Notice that in the rendered post, the raw font tag shows up unrendered.

Impact: Moderate-severe. There can at times be so many of these unrendered font tags that the post once posted is effectively unreadable.

A working solution: Simply replace in the file jscript/bbcodes_sceditor.js the following code from within the "format" method relating to the "font" bbcode:

				return '[font=' + this.stripQuotes(font) + ']' + content + '[/font]';
with this:

				if (font[0] == '&quot;' &amp;&amp; font[font.length-1] == '&quot;' &amp;&amp; (font.match(/&quot;/g) || []).length == 2) {
					font = font.substr(1, font.length-2);
				}
				font = font.replace(/&quot;/g, &quot;'&quot;);
				return '[font=' + font + ']' + content + '[/font]';

Now, when text is pasted into the editor in WYSIWYG mode, or when switching into Source mode, double quotes within lists of fonts are replaced with single quotes. Problem solved.

Additionally, it might be worth integrating into core code the code from my basic plugin attached to post #4 in Hovatek's thread. That code handles posts that have already been posted with double-quotes in their font tags, such that even with the above fix, those font tags display unrendered in the absence of the plugin's code.

Original thread: [The 'double quotes in font lists' browser-independent SCEditor bug](https://community.mybb.com/thread-229727.html)

### LINKED ISSUE #4184: The 'redundant tags on paste of unformatted text' browser-dependent SCEditor bug
This SCEditor bug and the next of the other two I plan to report (and now have reported) reference Hovatek's thread MyCode parse error ( [font="droid sans", sans-serif] ) and nixer55's post #29 in the SCEditor Issues and Resolutions [1.8.21 upgrade] thread (the remaining bug references only Hovatek's thread).

All three bug reports are heavily based on my post #9 in Hovatek's thread, from the first horizontal line onwards.

Specific to this bug report:

The 'redundant tags on paste of unformatted text' browser-dependent SCEditor bug.

Description: In WYSIWYG mode, on pasting unformatted text into the editor which was cut/copied from elsewhere within the editor, some browsers (e.g., Chrome and Microsoft Edge) generate redundant font+size+color attributes in the paste data whose values reflect the editor's default style (based on the default stylesheet jscripts/sceditor/styles/jquery.sceditor.mybb.css), causing redundant MyCode tags with those default font+size+color attributes to be generated (except for the font tag, to which a different bug applies, to be reported next). Other browsers (e.g., Firefox) don't generate redundant attributes, and are not subject to this bug.

How to reproduce: In Chrome or Microsoft Edge, bring up a new editor window through a page such as a New Reply page. Deselect Source mode if it is selected, such that the editor is in WYSIWYG mode. Now, type some text into it, such as "Blah". Do not style or format that text in any way. Now, select that text, cut or copy it, then paste it, and switch to Source mode. You will see that the pasted text has had font, color, and size MyCode tags added to it even though it was unformatted and unstyled prior to being pasted, and thus should not have had any MyCode tags applied to it.

Impact: Mild-moderate. Not detectable in WYSIWYG mode (except in the form of the next bug I plan to report, the 'reduced size on paste of unformatted text' browser-dependent SCEditor bug); irritating when switching to Source mode.

Potential but flawed solution: Because this bug stems from some browsers (but not others) redundantly setting values for style attributes in pasted text that are already at their default values, the "obvious" fix is to edit appropriate sections of jscript/bbcodes_sceditor.js such that default values of those size/font/color attributes are detected and ignored, and so that no tag is thus added in those scenarios. I have tested this approach - for size, font, and color tags - and it does work on a default MyBB installation, resolving both this bug and the next to be reported. The problem, however, is that it involves hard-coding those default style values into this Javascript file based on those in the default stylesheet - again, jscripts/sceditor/styles/jquery.sceditor.mybb.css - whereas, as Hovatek's scenario demonstrates, those default stylesheet values might be changed (customised) by themes. That is to say that, for example, the default font-family specified in (again) jscripts/sceditor/styles/jquery.sceditor.mybb.css is 'Tahoma, Verdana, Arial, Sans-Serif', whereas for Hovatek's forum it appears instead to have been customised to '"droid sans", sans-serif'. Detecting and ignoring in Javascript the default font-family for the editor as specified in (again) jscripts/sceditor/styles/jquery.sceditor.mybb.css would thus fail on Hovatek's forum, because his theme has overridden that default font-family.

Original thread: [The 'redundant tags on paste of unformatted text' browser-dependent SCEditor bug](https://community.mybb.com/thread-229725.html)

lairdshaw:
Fix #4184 - redundant tags on paste of unformatted text - by ignoring
fonts, colors, and sizes that match the editor's default.

Partially fix (client-side only) #4182 - double quotes in font lists -
by replacing double quotes in fonts with single quotes.

Commit both fixes at once because doing so separately would be likely to
cause a confusing conflict that would then need to be resolved (due to
the change to the condition in the 'font' method required by the fix for
#4184).

Resolves #4184.
Resolves #4182.

### DIFF
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -151,9 +151,17 @@ $(function ($) {
 			var fontsize = 1,
 				scefontsize = $($elm).data('scefontsize'),
 				parsed = parseInt(scefontsize, 10),
-				size = parseInt($($elm).attr('size'), 10);
+				size = parseInt($($elm).attr('size'), 10),
+				iframe = $('.sceditor-container iframe'),
+				editor_body = $('body', iframe.contents());
 
-			if (!isNaN(size) && size >= 1 && size <= mybbCmd.fsStr.length) {
+			if ($($elm).css('font-size') == editor_body.css('font-size')) {
+				// Eliminate redundant [size] tags for unformatted text.
+				// Part of the fix for the browser-dependent bug of issue #4184.
+				// Also fixes the browser-dependent bug described here:
+				//   <https://community.mybb.com/thread-229726.html>
+				fontsize = -1;
+			} else if (!isNaN(size) && size >= 1 && size <= mybbCmd.fsStr.length) {
 				fontsize = mybbCmd.fsStr[size - 1];
 			} else if ($.inArray(scefontsize, mybbCmd.fsStr) !== -1) {
 				fontsize = scefontsize;
@@ -161,7 +169,7 @@ $(function ($) {
 				fontsize = parsed;
 			}
 
-			return '[size=' + fontsize + ']' + content + '[/size]';
+			return fontsize != -1 ? '[size=' + fontsize + ']' + content + '[/size]' : content;
 		},
 		html: function (token, attrs, content) {
 			var size = 0,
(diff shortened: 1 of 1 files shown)


Eldenroot:
Thank you for your contribution!

PR #4187, opened 2020-11-25
Server-side fix for #4182 - double quotes in font lists

### LINKED ISSUE #4182: The 'double quotes in font lists' browser-independent SCEditor bug

lairdshaw:
Back-fix #4182 server-side after PR #4186 forward-fixes it client-side.

Resolves #4182.

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -368,8 +368,8 @@ class postParser
 
 		if($mybb->settings['allowfontmycode'] == 1)
 		{
-			$nestable_mycode['font']['regex'] = "#\[font=(\"?)([a-z0-9 ,\-_']+)\\1\](.*?)\[/font\]#si";
-			$nestable_mycode['font']['replacement'] = "<span style=\"font-family: $2;\" class=\"mycode_font\">$3</span>";
+			$callback_mycode['font']['regex'] = "#\[font=\\s*(\"?)([a-z0-9 ,\-_'\"]+)\\1\\s*\](.*?)\[/font\]#si";
+			$callback_mycode['font']['replacement'] = array($this, 'mycode_parse_font_callback');
 
 			++$nestable_count;
 		}
@@ -1137,6 +1137,23 @@ class postParser
 
 		eval("\$mycode_url = \"".$templates->get("mycode_url", 1, 0)."\";");
 		return $mycode_url;
+	}
+
+	/**
+	* Parses font MyCode.
+	*
+	* @param array $matches Matches.
+	* @return string The HTML <span> tag with styled font.
+	*/
+	function mycode_parse_font_callback($matches)
+	{
+		// Replace any occurrence(s) of double quotes in fonts with single quotes.
+		// A back-fix for double-quote-containing MyBB font tags in existing
+		// posts prior to the client-side aspect of this fix for the
+		// browser-independent SCEditor bug of issue #4182.
+		$fonts = str_replace('"', "'", $matches[2]);
+
+		return "<span style=\"font-family: {$fonts};\" class=\"mycode_font\">{$matches[3]}</span>";
 	}
 
 	/**


yuliu:
So, I guess this PR is trying to fix the font-parsing issue, if any, for any existing post. Am I right?

Replacing the font parser from `$nestable_mycode` to `$callback_mycode` might break some plugins that rely on or modify `$this->mycode_cache['nestable']['font']`. Just don't know if there's any 😆 .

lairdshaw:
> So, I guess this PR is trying to fix the font-parsing issue, if any, for any existing post. Am I right?

Yes, exactly right.

> Replacing the font parser from `$nestable_mycode` to `$callback_mycode` might break some plugins that rely on or modify `$this->mycode_cache['nestable']['font']`. Just don't know if there's any :laughing: .

You're very diligent when it comes to backwards-compatibility for plugins. Kudos.

Yes, that's a possible harm to weigh against the benefit of the fix. Another possibility is to leave the font parser as `$nestable_mycode` and then pre-parse font tags to replace double quotes with single quotes, which is how my little [fixunrenderedfonttags.php](https://community.mybb.com/attachment.php?aid=43322) fix-it plugin performs this task. And there, the benefit of guaranteeing backwards-compatibility has to be weighed against the inelegance of that solution.

My weighings? The likelihood of a plugin relying on `$this->mycode_cache['nestable']['font']` is too low, and the alternative solution too inelegant, to justify rejecting or modifying this PR.

lairdshaw:
Just to add to that: looking at all of the calls to `$plugins->run_hooks` in the parser, it doesn't appear that any of them provide access to the parser object anyway, so - unless I'm missing something - plugins currently do not even have access to `$this->mycode_cache['nestable']['font']` in the first place.

yuliu:
> Just to add to that: looking at all of the calls to `$plugins->run_hooks` in the parser, it doesn't appear that any of them provide access to the parser object anyway, so - unless I'm missing something - plugins currently do not even have access to `$this->mycode_cache['nestable']['font']` in the first place.

Yes, but [`$mycode_cache` is declared public](https://github.com/mybb/mybb/blob/mybb_1824/inc/class_parser.php#L33) and `$parser` is a global variable in [`showthread.php`](https://github.com/mybb/mybb/blob/mybb_1824/showthread.php#L32) or is defined earlier in [`global.php`](https://github.com/mybb/mybb/blob/mybb_1824/global.php#L854) where it can be read and modified by plugin hooks.

> My weighings? The likelihood of a plugin relying on `$this->mycode_cache['nestable']['font']` is too low, and the alternative solution too inelegant, to justify rejecting or modifying this PR.

I have the same thought, fixing this issue is bigger than breaking things that are not in core MyBB and we should encourage plugin authors to make their work compatible with MyBB's new releases.

I just have another concern. I hope doing so won't suggest people saving fonts with double quoting or some weird combinations like `"Font Name'` in posts. I think we'd like make things simple and controllable when it comes to saving something into databases.

lairdshaw:
> Yes, but [`$mycode_cache` is declared public](https://github.com/mybb/mybb/blob/mybb_1824/inc/class_parser.php#L33) and `$parser` is a global variable in [`showthread.php`](https://github.com/mybb/mybb/blob/mybb_1824/showthread.php#L32) or is defined earlier in [`global.php`](https://github.com/mybb/mybb/blob/mybb_1824/global.php#L854) where it can be read and modified by plugin hooks.

I was missing something then. Thanks for setting me straight. It's useful information.

> I have the same thought, fixing this issue is bigger than breaking things that are not in core MyBB and we should encourage plugin authors to make their work compatible with MyBB's new releases.

Good, it seems we agree.

> I just have another concern. I hope doing so won't suggest people saving fonts with double quoting in posts. I think we'd like make things simple and controllable when it comes to saving something into databases.

I agree - the thing is though that the horse has already bolted, and shutting the gate now would not achieve anything much for existing forums (at least, not for those which have already experienced this problem): it can only keep things simple for new forums and those existing forums which have not already experienced this problem.

If we do think it's worth achieving that, then might we perhaps add a little magic to `PostDataHandler::verify_message()` to replace double-quotes with single-quotes in font tags?

Sama34:
Hi, I was reading this post, if I understand correctly this bug was fixed in the user side by _fixing_ the editor.

While I agree with the attempt to mitigate this issue from the browser side (editor), I think we should avoid fixing it from the script side, as it is not an intended feature and dropping the editor in the future is almost set on stone (parser rewrite already is). Accepting this change could potentially complicate stuff when the time comes, considering this is not a feature but a bug.

lairdshaw:
Hi @Sama34. Here's how I see things:

Before these PRs, the editor was producing corrupt data (font tags with double-quotes within the list of fonts) which was then saved to the database.

PR #4186 does fix, as you note, the editor such that it no longer produces this corrupt data, and thus this corrupt data is no longer saved to the database going forward.

That PR does not, however, have any impact on the corrupt data _already_ saved to the database.

That's where this PR comes in: it "massages" the saved corrupt data such that when pulled out of the database for display, the corrupted font tags render as intended, rather than remaining unrendered.

In other words, the client-side fix of #4186 is _insufficient_ as a fix: it only eliminates _future_ corruption; it does not deal with _existing_ corruption. This PR is needed to handle that existing corruption.

Does that give you greater confidence in its utility?

euantorano:
I fully agree we need both a frontend and backend fix, ad my testing of this fix has gone smoothly so I'm going to merge this.

Sama34:
> it does not deal with existing corruption. This PR is needed to handle that existing corruption.

If this is the case this should have been handled by the upgrade script IMO.

From now on that data is no longer corrupted but complies with the core features.

lairdshaw:
> If this is the case this should have been handled by the upgrade script IMO.

Fair enough. That's a legitimate opinion.

> From now on that data is no longer corrupted but complies with the core features.

Well, there's a principle: "be lax in what you accept, strict in what you emit". Here, we conform to the first part of the principle.

PR #4193, opened 2020-12-13
Implement PDO based DB drivers for MySQL and PostgreSQL

### LINKED ISSUE #4192: DB_MySQLi::modify_column does not set default value
The code for `modify_column` in `DB_MySQLi` does not use the `$default` variable when writing the query to perform the modification, [as can be seen here](https://github.com/mybb/mybb/blob/feature/inc/db_mysqli.php#L1396).

euantorano:
This PR adds two new database drivers for MySQL and PostgreSQL that make use of [PDO](https://www.php.net/manual/en/book.pdo.php).

It also amends the SQLite driver to not just die if the database file cannot be opened.

It also fixes #4192.

### DIFF
--- a/inc/AbstractPdoDbDriver.php
+++ b/inc/AbstractPdoDbDriver.php
@@ -300,10 +300,10 @@ abstract class AbstractPdoDbDriver implements DB_Base
 				$portSeparator = strpos($hostname, ':', $closingSquareBracket);
 
 				// there is no port specified
-				if ($portSeparator === false && strlen($hostname) > $portSeparator + 1) {
+				if ($portSeparator === false) {
 					return array($hostname, null);
 				} else {
-					$host = substr($hostname, $openingSquareBracket, $closingSquareBracket);
+					$host = substr($hostname, $openingSquareBracket, $closingSquareBracket + 1);
 					$port = (int) substr($hostname, $portSeparator + 1);
 
 					return array($host, $port);
@@ -315,7 +315,7 @@ abstract class AbstractPdoDbDriver implements DB_Base
 
 		// either an IPv4 address or a hostname
 		$portSeparator = strpos($hostname, ':', 0);
-		if ($portSeparator === false && strlen($hostname) > $portSeparator + 1) {
+		if ($portSeparator === false) {
 			return array($hostname, null);
 		} else {
 			$host = substr($hostname, 0, $portSeparator);
--- a/inc/AbstractPdoDbDriver.php
+++ b/inc/AbstractPdoDbDriver.php
@@ -417,7 +417,7 @@ abstract class AbstractPdoDbDriver implements DB_Base
 	 *
 	 * @return PDOStatement|null The result of the query, or null if an error occurred and {@see $hideErrors} was set.
 	 */
-	function query($string, $hideErrors = false, $writeQuery = 0)
+	public function query($string, $hideErrors = false, $writeQuery = false)
 	{
 		global $mybb;
 
@@ -474,9 +474,9 @@ abstract class AbstractPdoDbDriver implements DB_Base
 	 *
 	 * @return PDOStatement|null The result of the query, or null if an error occurred and {@see $hideErrors} was set.
 	 */
-	function write_query($query, $hideErrors = false)
+	public function write_query($query, $hideErrors = false)
 	{
-		return $this->query($query, $hideErrors, 1);
+		return $this->query($query, $hideErrors, true);
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


inc/AbstractPdoDbDriver.php
```diff
@@ -0,0 +1,713 @@
...
+	 *
+	 * @return array Array of host and port.
+	 *
+	 * @throws InvalidArgumentException Thrown if {@see $hostname} is an IPv6 address which lacks a closing square bracket.
+	 */
+	private static function parseHostname($hostname)
+	{
+		// first, check for an IPv6 address - IPv6 addresses always start with `[`
+		$openingSquareBracket = strpos($hostname, '[');
+		if ($openingSquareBracket === 0) {
```
euantorano, line 295: Note: This logic is not yet tested. I believe that other drivers doing much the same thing would never have worked with IPv6 addressed DB hosts...

inc/AbstractPdoDbDriver.php
```diff
@@ -0,0 +1,713 @@
...
+	 * @param PDOStatement $query The query data.
+	 * @return int|bool The number of rows in the result, or false on failure.
+	 */
+	public function num_rows($query)
+	{
+		if (is_null($query) || !($query instanceof PDOStatement)) {
+			return false;
+		}
+
+		if (preg_match('/^\\s*SELECT\\b/i', $query->queryString) === 1) {
```
euantorano, line 601: Note: using `num_rows` with PDO is not optimal at all. We need to review usages within the core and see when/why we use `num-rows` and if they can be replaced so that we don't hit this path too often.

This works better with the old `mysql_` type functions as those functions buffer results internally, which PDO does not do.

inc/db_pgsql_pdo.php
```diff
@@ -0,0 +1,684 @@
...
+		foreach($fields as $key => $field)
+		{
+			$fieldstring .= "{$comma}{$append}{$field['Field']}";
+			$comma = ',';
+		}
+
+		return $fieldstring;
+	}
+
+	public function __set($name, $value)
```
euantorano, line 667: Note: This special handling is important to ensure compatibility.

euantorano:
I haven't yet tested the MySQL PDO driver, but have tested the Postgres one. I'd like to land this in 1.8.25 but not make too much noise about it. Then during the cycle for 1.8.26 I plan to add some more methods to all of the DB drivers to support prepared statements (which will be emulated for some drivers).

euantorano:
I've now done some testing of this with MySQL too and ran through an install and a few key pages.

I'd very much appreciate it if others could test this PR. Upgraded boards can opt-in to the PDO based drivers by modifying `inc/config.php` to alter the DB engine to one of the following options:

```php
$config['database']['type'] = 'mysql_pdo';
```

```php
$config['database']['type'] = 'pgsql_pdo';
```

Once altered, there shouldn't be any visible change if all goes well.

lairdshaw:
> I'd very much appreciate it if others could test this PR.

I haven't done any systematic testing, but have been using the `mysql_pdo` driver for about two weeks on my local dev board whilst working on a plugin, and have had no problems so far. Seems good.

awslabspl:
> > I'd very much appreciate it if others could test this PR.
> 
> I haven't done any systematic testing, but have been using the `mysql_pdo` driver for about two weeks on my local dev board whilst working on a plugin, and have had no problems so far. Seems good.

Second this. Also no problems.

euantorano:
Awesome, thanks Laird!

inc/AbstractPdoDbDriver.php
```diff
@@ -0,0 +1,717 @@
...
+		$openingSquareBracket = strpos($hostname, '[');
+		if ($openingSquareBracket === 0) {
+			// find ending `]`
+			$closingSquareBracket = strpos($hostname, ']', $openingSquareBracket);
+
+			if ($closingSquareBracket !== false) {
+				$portSeparator = strpos($hostname, ':', $closingSquareBracket);
+
+				// there is no port specified
+				if ($portSeparator === false && strlen($hostname) > $portSeparator + 1) {
```
dvz: Does the second condition make sense (also on line 318)?

euantorano: Looking at it again, no I don't think it does.

This is the kind of thing that would be best tested by some unit tests, if we had any...

inc/db_pgsql.php
```diff
@@ -451,6 +451,8 @@ function fetch_field($query, $field, $row=false)
 		{
 			return pg_fetch_result($query, $row, $field);
 		}
+
+		return pg_fetch_result($query, $row, $field);
```
dvz, line 451: Unreachable statement

euantorano:
Seems some changes were missing for this. I think I’d simply dropped the else but must not have staged it. Will fix later.

Sama34 (approved):

euantorano:
I'm going ahead and merging this, it's had a fair bit of testing now I believe.

PR #4194, opened 2020-12-16
Mass mail fixes

### LINKED ISSUE #4183: Mass mail: lost radio button selection & illegal string offset warnings
Mass mail at AdminCP:

1. If the scheduled sending date is in past, the error message screen loses that radio button selection.

Creating:

[attachment=43244]

Editing:

[attachment=43243]

2. When creating a new mass mailing, navigating away before submitting recipients criteria would result in blank in conditions in table massemails for this mass mailing. And warnings of Illegal string offset will be given when editing this mass mailing with PHP 7.4+.

One example of the warnings:
Warning [2] Illegal string offset 'username' - Line: 540 - File: \admin\modules\user\mass_mail.php PHP 7.4.1 (WINNT)
File 	Line 	Function
\admin\modules\user\mass_mail.php 	540 	errorHandler-&gt;error
\admin\index.php 	824 	require

Original thread: [Mass mail: lost radio button selection & illegal string offset warnings](https://community.mybb.com/thread-229293.html)

effone:
Resolves #4183

### DIFF
--- a/admin/modules/user/mass_mail.php
+++ b/admin/modules/user/mass_mail.php
@@ -62,7 +62,7 @@ if($mybb->input['action'] == "edit")
 	else
 	{
 		// Fill the conditions with default values
-		$email['conditions'] = Array(
+		$email['conditions'] = array(
 			"username" => "",
 			"email" => "",
 			"postnum_dir" => "greater_than",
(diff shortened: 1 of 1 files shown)


admin/modules/user/mass_mail.php
```diff
@@ -59,6 +59,22 @@
 	{
 		$email['conditions'] = my_unserialize($email['conditions']);
 	}
+	else
+	{
+		// Fill the conditions with default values
+		$email['conditions'] = Array(
```
Sama34: Capitalized `Array` word might need to be changed.

admin/modules/user/mass_mail.php
```diff
@@ -675,6 +691,7 @@ function insertText(value, textarea)
 		{
 			$page->output_inline_error($errors);
 			$input = $mybb->input;
+			$delivery_type_checked[$input['delivery_type']] = " checked=\"checked\"";
```
Sama34, line 701: I think we need to check if `$input['delivery_type']` is set. Maybe just `$mybb->get_input('delivery_type')`

Sama34 (changes requested):

PR #4200, opened 2020-12-20
Add GitHub workflow to lint PHP files for syntax errors

euantorano:
This adds a new GitHub Actions workflow to lint all PHP files for syntax errors.

It also fixes a syntax error found in `upgrade52.php` by the lint tool, and a couple of warnings in `upgrade30.php` all found by the lint.

Note that linting is currently done using 3 PHP versions:

- 5.3 (the oldest available version...)
- 7.4 (the latest available version in the 7.x series)
- 8.0 (the latest available version in the 8.x series)

This could be expanded in the future.

This should catch any syntax errors in pushes and PRs, which makes life slightly easier.

### DIFF
--- a/.github/workflows/php-syntax-check.yml
+++ b/.github/workflows/php-syntax-check.yml
@@ -1,0 +1,21 @@
+name: PHP Syntax Check
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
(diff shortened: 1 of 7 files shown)


euantorano:
Note: The action is currently failing. This is due to syntax errors with PHP 8, which don't seem to have been covered by #4095, including in `class_custommoderation.php`.

These syntax errors should be fixed in a separate PR.

PR #4203, opened 2020-12-21
Force Download Attachments

### LINKED ISSUE #4163: [Feature Suggestion] Force Attachment Download
This suggestion is related to this thread: https://community.mybb.com/thread-229749.html
In future, it would be nice to add option in Attachments Types that will allow us to choose will that attachment will be downloadable or will it be opened in a browser (For example, .txt files).
I am running forum where users can upload their extensions for games, so when they upload language files that are in .txt extension or settings files in .ini, those files are opened in browser.

This option should't affect types that are already selected as downloadable in attachment.php file, so my idea is to add function  in this line https://github.com/mybb/mybb/blob/aacebaecb49d0cb951096134277250192a2cedba/attachment.php#L182
that will check is file extension is marked as downloadable or not. If it's marked as downloadable, it will be downloaded, if not it will be opened in the browser. That way, other file types won't be affected.

effone:
Resolves #4163


effone:
Note:
HTML5 has a property `download` to declare with `a` tag to force download which can be considered further; I went with the suggested old header disposition way.

install/resources/upgrade52.php
```diff
@@ -31,13 +31,15 @@ function upgrade51_dbchanges()
...
-		case "pgsql":
-			$db->add_column("usergroups", "canbeinvisible", "smallint NOT NULL default '1' AFTER canusercp");
-			break;
-
-		default:
-			$db->add_column("usergroups", "canbeinvisible", "tinyint(1) NOT NULL default '1' AFTER canusercp");
-			break;
-	}
+        // Add new settings for (a) new usergroup permission if group members can hide online status & (b) attachments force download
+        case "pgsql":
```
euantorano: Nitpick: indentation should be tabs, not spaces here - that's why the diff looks so odd.

PR #4210, opened 2020-12-28
Fix #4209 - send the mail queue using a task that runs every minute

### LINKED ISSUE #4209: Move mail queue processing to a task rather than a shutdown function
The mail queue is currently ran using a shutdown function, which is registered [in `init.php`](https://github.com/mybb/mybb/blob/aacebaecb49d0cb951096134277250192a2cedba/inc/init.php#L243).

This can be problematic when using SMTP to send mail and the SMTP server is either slow to respond or offline, as it leads to slow page load times.

We should instead use a task to send the mail queue.

euantorano:
Fixes #4209.

This adds a new task, `sendmailqueue`, which defaults to running every minute.

It also adds a new setting to manage the maximum number of messages to send from the mail queue when the task runs - previously only the default value of 10 emails would be used.

### DIFF
--- a/install/resources/upgrade52.php
+++ b/install/resources/upgrade52.php
@@ -40,4 +40,8 @@ function upgrade51_dbchanges()
 			$db->add_column("usergroups", "canbeinvisible", "tinyint(1) NOT NULL default '1' AFTER canusercp");
 			break;
 	}
+
+	$added_tasks = sync_tasks();
+
+	echo "<p>Added {$added_tasks} new tasks.</p>";
 }
--- a/install/resources/upgrade52.php
+++ b/install/resources/upgrade52.php
@@ -46,4 +46,7 @@ function upgrade51_dbchanges()
 	$added_tasks = sync_tasks();
 
 	echo "<p>Added {$added_tasks} new tasks.</p>";
+
+	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
+	$output->print_footer("52_done");
 }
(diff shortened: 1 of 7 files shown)


Ben-MyBB (approved):
Aside from the conflicts on this PR it all looks good. @mybb/developers given that I'm not overly experienced at the minute could someone else give this a look over please?

euantorano:
I’ll sort the conflicts this evening.

euantorano:
Conflicts resolved.

Eldenroot:
I have applied to my board, seems to be fine, if anything I will add a note here.

install/resources/upgrade52.php
```diff
@@ -42,4 +42,8 @@ function upgrade51_dbchanges()
 			$db->add_column("attachtypes", "forcedownload", "tinyint(1) NOT NULL default '0' AFTER enabled");
 			break;
 	}
+  
+	$added_tasks = sync_tasks();
+
+	echo "<p>Added {$added_tasks} new tasks.</p>";
```
Sama34, line 48:
```suggestion
	echo "<p>Added {$added_tasks} new tasks.</p>";

	$output->print_contents("<p>Click next to continue with the upgrade process.</p>");
	$output->print_footer("52_done");
```

Sama34 (approved):
I think you should add the suggested code to the upgrade script.

install/resources/upgrade52.php
Sama34: @euantorano I did test this change, so you could either accept it or left it for later and merge.

PR #4213, opened 2020-12-29
Normalization of Strings in Language Files

Xazin:
Some strings start with a citation mark, others with a quotation mark; thus I normalized all strings that I could find, so they all use quotation marks ("). All language files have been normalized. Multiple double spaces _(redundant/inconsistent)_ have been made into a single space.

A review of all edits to ensure escaped strings and similar situations should be considered for all 40 files (41 edits), though it will be quick since most edits pertain to few lines.

_Note: Since this is not based on an issue nor a suggestion, I did not know whether to submit a suggestion or do anything beforehand, thus going forward with the pull request. I sincerely hope I do not cause anyone distress._


inc/languages/english/admin/config_settings.lang.php
```diff
@@ -90,24 +90,24 @@
 $l['error_ajax_search'] = "There was a problem searching for settings:";
 $l['error_ajax_unknown'] = "An unknown error occurred while searching for settings.";
 $l['error_chmod_settings_file'] = "The settings file \"./inc/settings.php\" isn't writable. Please CHMOD to 666.<br />For more information on CHMODing, see the <a href=\"https://docs.mybb.com/1.8/administration/security/file-permissions\" target=\"_blank\" rel=\"noopener\">MyBB Docs</a>.";
-$l['error_admin_email_settings_empty'] = '<div class="smalltext" style="font-weight: normal;">Please note that the <strong>Admin Email</strong> was not saved as the setting can not be empty, and must be a valid email address.</div>';
+$l['error_admin_email_settings_empty'] = "<div class=\"smalltext\" style=\"font-weight: normal;\">Please note that the <strong>Admin Email</strong> was not saved as the setting can not be empty, and must be a valid email address.</div>";
```
euantorano, line 93: In this case, changing to double quotes actually makes the string less readable in my opinion.

inc/languages/english/admin/config_settings.lang.php
```diff
@@ -90,24 +90,24 @@
...
-$l['success_settings_updated_allowmultipleemails'] = '<div class="smalltext" style="font-weight: normal;">Please note that the <b>Allow emails to be registered multiple times?</b> setting can\'t be enabled because the <b>Allowed Login Methods</b> setting allows users to login by e-mail address.</div>';
-$l['success_settings_updated_captchaimage'] = '<div class="smalltext" style="font-weight: normal;">Please note that the <strong>CAPTCHA Images for Registration &amp; Posting</strong> setting was reverted to <strong>MyBB Default Captcha</strong> due to the lack of public/private key(s).</div>';
-$l['success_settings_updated_minsearchword'] = '<div class="smalltext" style="font-weight: normal;">Please note that the <strong>Minimum Search Word Length</strong> setting was set to match the database system configuration.</div>';
+$l['success_settings_updated_hiddencaptchaimage'] = "<div class=\"smalltext\" style=\"font-weight: normal;\">Please note that the <strong>Hidden CAPTCHA field</strong> setting was reverted to <strong>{1}</strong> due to a conflict with the <strong>{2}</strong> field in the registration form.</div>";
```
euantorano, line 99: Same here - single quotes prevent the need for escaping HTML attributes, which makes comprehension far easier for me.

euantorano (changes requested):
I think the changes to remove double spaces are great, but changing the quoting is actually a step backwards in some cases - especially language strings which embed HTML.

inc/languages/english/admin/config_settings.lang.php
Xazin: It is done in this way in many other places in the language files where the quotation mark is used when embedding HTML and escaping many characters.

_Like in:
archive.lang.php
calendar.lang.php
datahandler_login.lang.php
editpost.lang.php
global.lang.php
helpdocs.lang.php
managegroup.lang.php
member.lang.php
..._

I also do get that it actually helps with comprehension, and I totally agree. Still, since more than 2-300 occurrences were built like this, I thought the correct way was to correct the lesser part instead of correcting the larger part. To have it as uniform as possible.

If you want, I can begin correcting the language files such that they're more comprehensible instead of generalized?

lairdshaw:
I'm not sure to what extent these are still (expected to be) followed, but some of these changes are inconsistent with the [Strings](https://docs.mybb.com/1.8/development/standards/#strings) section of the MyBB 1.8 Development Standards.

Xazin:
> I'm not sure to what extent these are still (expected to be) followed, but some of these changes are inconsistent with the [Strings](https://docs.mybb.com/1.8/development/standards/#strings) section of the MyBB 1.8 Development Standards.

If you speak of this
`When a string is literal (contains no variable substitutions), the apostrophe or ‘single quote’ must be used to demarcate the string:` [Strings](https://docs.mybb.com/1.8/development/standards/#strings)

You are right. But the approach mentioned hasn't been followed for most of the source code.

In an effort to normalize the variables in the language file, I opted to pursue the generic example derived from the existing layout. It can be seen by looking through the language files that largely whether it is a quotation mark or a single quote is dependant on who wrote that line and at what point in time. But more than 80 % of the lines consisted of double quotes _(although it is against the standards in those cases - since most language variables are literally 'string literals'')_

I am willing to reformat all the language files so that they can be properly generalized.

Currently, the only aspect in which these normalizations adhere to are:
- If a string contains a literal or single quote, it should be escaped by double quotes
_(Due to them all being escaped by double quotes - Which were the case for more than 300 existing occurrences of escaping multiple double quotes inside a string.)_

lairdshaw:
> You are right. But the approach mentioned hasn't been followed for most of the source code.

Yes, you're right. As I said, I'm not sure whether the Standard is expected to be followed in this respect - because a lot of the time, it isn't / hasn't been. I guess it's up to the core developers.

euantorano:
I believe the standard is primarily aimed at new code, rather than existing code (as there are a **lot** of strings!).

Interested to hear input from other @mybb/developers and @mybb/sqa.

inc/languages/english/admin/config_settings.lang.php
euantorano: It would certainly be good to have a standard, as the current code has been contributed over many years by many different developers with different approaches.

For an example of this difference in approaches, [I write HTML differently to others](https://github.com/mybb/mybb/pull/4193/files#diff-bad7941be860c04bf31b5406e617741e0d90e0530436b1a1c9917c2be85a83b0R51).

Hopefully others will chime in with their thoughts.

dvz:
Language strings might be an additional exception to generally preferring single quotes (`'`; to prevent accidental interpretation of various characters), given the common usage of the apostrophe in the default English language.

Note that the ACP currently [uses `"` for generated files](https://github.com/mybb/mybb/blob/mybb_1824/admin/modules/config/languages.php#L442).

inc/languages/english/admin/config_settings.lang.php
Sama34: I think it is not worth to do the larger part at the cost of comprehension.

Edit: I think, eventually, we should be dropping the style and class elements usage from language strings.

inc/languages/english/online.lang.php
```diff
@@ -37,7 +37,7 @@
 $l['viewing_whoposted2'] = "Viewing Who Posted in Thread <a href=\"{1}\">{2}</a>";
 $l['marking_read'] = "<a href=\"misc.php?action=markread&amp;my_post_key={1}\">Marking Forums as Read</a>";
 $l['viewing_helpdocs'] = "Viewing <a href=\"misc.php?action=help\">Help Documents</a>";
-$l['viewing_buddylist'] = 'Viewing <a href="#" onclick="MyBB.popupWindow(\'/misc.php?action=buddypopup&amp;modal=1\'); return false;">Buddy List</a>';
+$l['viewing_buddylist'] = "Viewing <a href=\"#\" onclick=\"MyBB.popupWindow('/misc.php?action=buddypopup&amp;modal=1'); return false;\">Buddy List</a>";
```
Sama34, line 40: Maybe we should even drop these onclick elements as well.

inc/languages/english/online.lang.php
```diff
@@ -66,7 +66,7 @@
 $l['editing_signature'] = "<a href=\"usercp.php?action=editsig\">Editing Signature</a>";
 $l['changing_avatar'] = "<a href=\"usercp.php?action=avatar\">Changing Avatar</a>";
 $l['viewing_subscriptions'] = "Viewing <a href=\"usercp.php?action=subscriptions\">Thread Subscriptions</a>";
-$l['editing_pad'] = 'Editing <a href="usercp.php">Personal Pad</a>';
+$l['editing_pad'] = "Editing <a href=\"usercp.php\">Personal Pad</a>";
```
Sama34, line 69: Probably even anchor tags altogether but not so sure about these. Maybe all HTML tags should be in templates, which might hard to decide because of how the current system works, but once the new template system (more optimized) is functional maybe we can review this.

Sama34 (approved):

Sama34:
Additionally, the coding standards should probably be updated as this kind of contributions and discussions arise.

@Ben-MyBB :P

Sama34:
I think unrelated commits are now being pushed? I really doubt this will make it to 1.8 @mybb/developers

Xazin:
> I think unrelated commits are now being pushed? I really doubt this will make it to 1.8 @mybb/developers

I am pretty sure this should be scrapped for now. Since a specific convention to follow has not been established yet.

PR #4214, opened 2020-12-30
Fix #3804 Post Merge Verification

### LINKED ISSUE #3804: Post merge doesn't do check on video/image count
Well, it's pretty obvious through this thread: https://community.mybb.com/thread-224651.html , which breaks the intention of limiting maximum videos/images per post if post merge is available.

Original thread: [Post merge doesn't do check on video/image count](https://community.mybb.com/thread-224655.html)

Xazin:
Due to missing check on whether a post is a merge before a post is validated, it was possible to double post and merge multiple images and videos despite the limitations imposed through settings in AdminCP (Post Settings).

Resolves #3804

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -790,20 +790,7 @@ class PostDataHandler extends DataHandler
 				$this->first_post = true;
 			}
 		}
-
-		// Make sure there isn't multiple posts (due to merge) that need to be verified
-		$old_message = $post['message'];
-
-		if ($this->method == "insert") 
-		{
-			$double_post = $this->verify_post_merge();
-		}
-
-		if($double_post !== true && $double_post['visible'] == 1)
-		{
-			$post['message'] = $double_post['message'] .= "\n".$mybb->settings['postmergesep']."\n".$post['message'];
-		}
-
+		
 		// Verify all post assets.
 		if($this->method == "insert" || array_key_exists('uid', $post))
 		{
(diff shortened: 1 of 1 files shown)


Xazin:
The solution is rough but gives insight into a possible solution for issue #3804 .

Xazin:
I performed some tests but hope that someone can review the changes and see if there are any problems with the logic.

inc/datahandlers/post.php
```diff
@@ -1002,52 +1002,60 @@ function insert_post()
...
-						"posthash" => ''
+						"visible" => $visible,
+						"merge" => true
 					);
-					$db->update_query("attachments", $attachmentassign, "posthash='{$post['posthash']}' AND pid='0'");
+					
+					$plugins->run_hooks("datahandler_post_insert_merge", $this);
+					
+					return $this->return_values;
+				} else {
```
Sama34:
```suggestion
				}
				else
				{
```

Sama34 (changes requested):
https://github.com/mybb/mybb/issues/3804#issuecomment-752590036
> * if the limits are verified on the resulting posts, like currently proposed in #4214, it may be impossible to post more items within _Post Merge Time_, which may be unexpected.

Is this already accounted for in this current code? It doesn't seem like so to me.

Xazin:
> [#3804 (comment)](https://github.com/mybb/mybb/issues/3804#issuecomment-752590036)
> 
> > * if the limits are verified on the resulting posts, like currently proposed in #4214, it may be impossible to post more items within _Post Merge Time_, which may be unexpected.
> 
> Is this already accounted for in this current code? It doesn't seem like so to me.

Yes, this fixes the issues from #3804 and the issues @dvz mentioned on that issue regarding the original pr that was made.

euantorano:
This looks good to me and seems to work based on my testing.

PR #4215, opened 2020-12-31
Partial Fix to #3540

Xazin:
This is a partial fix to #3540, and more specifically, see this comment:
https://github.com/mybb/mybb/issues/3540#issuecomment-643508337

When there are no threads in a forum, there is no need to sort the threads. This hides the sorting for the case where there are 0 threads in a threadlist.

_This is more visual/redundant than needed. There might be a need to rename the template 'forumdisplay_forumsort' and variable '$forumsort' to something more logical._

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1494,6 +1494,12 @@ if($foruminfo['type'] != "c")
 
 	$prefixselect = build_forum_prefix_select($fid, $tprefix);
 
+	// Populate Forumsort
+	if($threadcount > 0)
+	{
+		eval("\$forumsort = \"".$templates->get("forumdisplay_forumsort")."\";");
+	}
+	
 	$plugins->run_hooks("forumdisplay_threadlist");
 
 	$lang->rss_discovery_forum = $lang->sprintf($lang->rss_discovery_forum, htmlspecialchars_uni(strip_tags($foruminfo['name'])));
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1495,6 +1495,8 @@ if($foruminfo['type'] != "c")
 	$prefixselect = build_forum_prefix_select($fid, $tprefix);
 
 	// Populate Forumsort
+	$forumsort = '';
+	
 	if($threadcount > 0)
 	{
 		eval("\$forumsort = \"".$templates->get("forumdisplay_forumsort")."\";");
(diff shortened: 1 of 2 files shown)


forumdisplay.php
```diff
@@ -1494,6 +1494,12 @@
 
 	$prefixselect = build_forum_prefix_select($fid, $tprefix);
 
+	// Populate Forumsort
+	if($threadcount > 0)
+	{
+		eval("\$forumsort = \"".$templates->get("forumdisplay_forumsort")."\";");
```
Matslom, line 1502: I think that `$forumsort` need initialization with empty string, above this if. Otherwise there will be a notice for case that threadcount is 0.

Xazin: Hmm, I never stumbled into that problem, but taking precaution is a good idea. I'll make the changes.

install/resources/mybb_theme.xml
```diff
@@ -3823,6 +3823,39 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 <td class="{$bgcolor} forumdisplay_announcement" style="white-space: nowrap; text-align: right"><span class="smalltext">{$postdate}</span></td>
 {$modann}
 </tr>]]></template>
+		<template name="forumdisplay_forumsort" version="1824"><![CDATA[<tr>
```
```suggestion
		<template name="forumdisplay_forumsort" version="1825"><![CDATA[<tr>
```

install/resources/mybb_theme.xml
```diff
@@ -4007,7 +4040,7 @@ var announcement_quickdelete_confirm = "{$lang->announcement_quickdelete_confirm
 	</td>]]></template>
 		<template name="forumdisplay_thread_rating_moved" version="1800"><![CDATA[<td class="{$bgcolor}" style="text-align: center;">-</td>]]></template>
 		<template name="forumdisplay_thread_unapproved_posts" version="1800"><![CDATA[ <span title="{$unapproved_posts_count}">({$thread['unapprovedposts']})</span>]]></template>
-		<template name="forumdisplay_threadlist" version="1820"><![CDATA[<div class="float_left">
+		<template name="forumdisplay_threadlist" version="1824"><![CDATA[<div class="float_left">
```
```suggestion
		<template name="forumdisplay_threadlist" version="1825"><![CDATA[<div class="float_left">
```

euantorano (changes requested):

euantorano (approved):

PR #4218, opened 2021-01-04
Fix usage of class constants

### LINKED ISSUE #4070: CAPTCHA code improvements
Not exactly related to this PR, but I think we should introduce some constants/static variables to define these captcha types rather than having these types of checks scattered throughout the codebase.

_Originally posted by @euantorano in https://github.com/mybb/mybb/pull/4003#discussion_r439869586_

<hr>

Not related to this PR as such, but this is one wild `if` branch. Would be a good idea to try and clean this up in a. future release.

_Originally posted by @euantorano in https://github.com/mybb/mybb/pull/4003#discussion_r439869231_

Xazin:
Missing self:: or captcha:: keyword when using class constants

Class constants are being called as if they are global constants, this commit adds the captcha:: keyword to all class constant calls so they are valid.

### DIFF
--- a/inc/class_captcha.php
+++ b/inc/class_captcha.php
@@ -119,30 +119,30 @@ class captcha
 		{
 			$this->captcha_template = $template;
 
-			if($this->type == NOCAPTCHA_RECAPTCHA)
+			if($this->type == captcha::NOCAPTCHA_RECAPTCHA)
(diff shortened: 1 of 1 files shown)


euantorano:
Thanks for the contribution, and good catch!

PR #4225, opened 2021-01-07
Fixes #4224 - functions.php - add a new hook

### LINKED ISSUE #4224: Functions.php - add a new hook
A proposal to add a new hook for plugin developers around line https://github.com/mybb/mybb/blob/feature/inc/functions.php#L1809

Eldenroot:
Resolves #4224

@Supryk

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1817,7 +1817,8 @@ function is_moderator($fid=0, $action="", $uid=0)
 
 	$plugins->run_hooks("is_moderator", $hook_args);
 	
-	if (isset($hook_args['is_moderator'])) {
+	if(isset($hook_args['is_moderator']))
+	{
 		return true;
 	}
 
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1812,7 +1812,6 @@ function is_moderator($fid=0, $action="", $uid=0)
 		'fid' => $fid,
 		'action' => $action,
 		'uid' => $uid,
-		'user_perms' => &$user_perms',
 	);
 
 	$plugins->run_hooks("is_moderator", $hook_args);
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1818,7 +1818,7 @@ function is_moderator($fid=0, $action="", $uid=0)
 	
 	if(isset($hook_args['is_moderator']))
 	{
-		return (bool) $hook_args['is_moderator'];
+		return (boolean) $hook_args['is_moderator'];
 	}
 
 	if(!empty($user_perms['issupermod']) && $user_perms['issupermod'] == 1)
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1819,7 +1819,7 @@ function is_moderator($fid=0, $action="", $uid=0)
 	
 	if(isset($hook_args['is_moderator']))
 	{
-		return true;
+		return (bool) $hook_args['is_moderator'];
 	}
 
 	if(!empty($user_perms['issupermod']) && $user_perms['issupermod'] == 1)
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1794,7 +1794,7 @@ function get_moderator_permissions($fid, $uid=0, $parentslist="")
  */
 function is_moderator($fid=0, $action="", $uid=0)
 {
-	global $mybb, $cache;
+	global $mybb, $cache, $plugins;
 
 	if($uid == 0)
 	{
@@ -1807,6 +1807,9 @@ function is_moderator($fid=0, $action="", $uid=0)
 	}
 
 	$user_perms = user_permissions($uid);
+
+	$plugins->run_hooks("is_moderator");
+
 	if(!empty($user_perms['issupermod']) && $user_perms['issupermod'] == 1)
 	{
 		if($fid)
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1808,7 +1808,18 @@ function is_moderator($fid=0, $action="", $uid=0)
 
 	$user_perms = user_permissions($uid);
 
-	$plugins->run_hooks("is_moderator");
+	$hook_args = array(
+		'fid' => $fid,
+		'action' => $action,
+		'uid' => $uid,
+		'user_perms' => &$user_perms',
+	);
+
+	$plugins->run_hooks("is_moderator", $hook_args);
+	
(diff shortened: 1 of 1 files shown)


lairdshaw (commented):
How about passing the `is_moderator()` arguments to the hook function?

lairdshaw:
Also, it could be useful to allow the hook function to override the return of `is_moderator()`.

inc/functions.php
```diff
@@ -1807,6 +1807,9 @@ function is_moderator($fid=0, $action="", $uid=0)
 	}
 
 	$user_perms = user_permissions($uid);
+
+	$plugins->run_hooks("is_moderator");
```
euantorano: It would probably be good to pass at least the function arguments and the user permissions.

As suggested by @lairdshaw I've also added a way to allow plugins to return a result that will short-circuit the rest of the function.

```suggestion
	$hook_args = array(
		'fid' => $fid,
		'action' => $action,
		'uid' => $uid,
		'user_perms' => &$user_perms',
	);

	$plugins->run_hooks("is_moderator", $hook_args);
	
	if (isset($hook_args['is_moderator'])) {
		return true;
	}
```

lairdshaw:
That's very close to my complete suggestion, @euantorano. I offer only a couple of other possibilities:

1. That $user_perms is passed by reference.
2. That we allow for the possibility of a forced false return just as much as we allow for the possibility of a forced true return, i.e., that we change your suggested:
```php
	if (isset($hook_args['is_moderator']) && $hook_args['is_moderator'] === true) {
		return true;
	}
```
to (untested):
```php
	if (isset($hook_args['is_moderator'])) {
		return $hook_args['is_moderator'];
	}
```

lairdshaw:
Also, do we perhaps need to pass the 'is_moderator' array entry explicitly and by reference, perhaps initially set to null, or otherwise store the return of `run_hooks` (to `$hook_args`)? I'd have to test these possibilities to be sure that I'm talking sense.

euantorano:
> That's very close to my complete suggestion, @euantorano. I offer only a couple of other possibilities:
> 
> 1. That $user_perms is passed by reference.
> 2. That we allow for the possibility of a forced false return just as much as we allow for the possibility of a forced true return, i.e., that we change your suggested:
> 
> ```
> 	if (isset($hook_args['is_moderator']) && $hook_args['is_moderator'] === true) {
> 		return true;
> 	}
> ```
> 
> to (untested):
> 
> ```
> 	if (isset($hook_args['is_moderator'])) {
> 		return $hook_args['is_moderator'];
> 	}
> ```

1. Yep, that would make sense.
2. Again, I can definitely see the use case here so agreed.

I've updated my suggestion with these changes so it's easier to just click and merge.

Eldenroot:
@euantorano thank you for your reply and code review. I accepted yout suggestion above, it is ready to merge. 

@lairdshaw thank you for your input and help with improving the PR.

inc/functions.php
```diff
@@ -1808,7 +1808,18 @@ function is_moderator($fid=0, $action="", $uid=0)
...
+		'fid' => $fid,
+		'action' => $action,
+		'uid' => $uid,
+		'user_perms' => &$user_perms',
+	);
+
+	$plugins->run_hooks("is_moderator", $hook_args);
+	
+	if (isset($hook_args['is_moderator'])) {
+		return true;
```
lairdshaw: Oh, my suggestion was (is) that rather than `return true;` this line be `return $hook_args['is_moderator'];`, so that the hook can force a false return too.

inc/functions.php
```diff
@@ -1807,6 +1807,20 @@ function is_moderator($fid=0, $action="", $uid=0)
...
+	$hook_args = array(
+		'fid' => $fid,
+		'action' => $action,
+		'uid' => $uid,
+		'user_perms' => &$user_perms',
+	);
+
+	$plugins->run_hooks("is_moderator", $hook_args);
+	
+	if (isset($hook_args['is_moderator'])) {
```
```suggestion
	if(isset($hook_args['is_moderator']))
	{
```

inc/functions.php
Sama34: I think that is a good idea.

inc/functions.php
```diff
@@ -1807,6 +1807,20 @@ function is_moderator($fid=0, $action="", $uid=0)
 	}
 
 	$user_perms = user_permissions($uid);
+
+	$hook_args = array(
+		'fid' => $fid,
+		'action' => $action,
+		'uid' => $uid,
+		'user_perms' => &$user_perms',
```
Sama34: It is unnecessary to pass this over as you already passed `$uid` and `user_permissions()` handles caching.

Plus with the suggested change to usage of `$hook_args['is_moderator']` it is superfluous to pass as developers can already affect the returned value of `is_moderator()`.

Sama34 (approved):

inc/functions.php
Sama34: @lairdshaw do you have a use scenario where passing `$user_perms` is necessary considering my message above?

Sama34 (changes requested):
I approved by mistake..

inc/functions.php
lairdshaw: I don't think so, no. It seems reasonable to exclude it, especially considering that later in the function there are similar variables which wouldn't be passed to the hook, e.g., `$forumpermissions` and `$modcache`.

inc/functions.php
euantorano:
```suggestion
		return (bool) $hook_args['is_moderator'];
```

inc/functions.php
euantorano:
```suggestion
```

Eldenroot:
Thank you for review and code improvements!

inc/functions.php
```diff
@@ -1807,6 +1807,20 @@ function is_moderator($fid=0, $action="", $uid=0)
...
+		'fid' => $fid,
+		'action' => $action,
+		'uid' => $uid,
+	);
+
+	$plugins->run_hooks("is_moderator", $hook_args);
+	
+	if(isset($hook_args['is_moderator']))
+	{
+		return (bool) $hook_args['is_moderator'];
```
euantorano:
```suggestion
		return (boolean) $hook_args['is_moderator'];
```

PR #4229, opened 2021-01-08
Fixes #4228 - maillogs.php -add a new hook

### LINKED ISSUE #4228: maillogs.php - add a new hook for logs
Check this comment:

https://github.com/mybb/mybb/pull/4185#issuecomment-745270580

The plugin hook is missing.

Eldenroot:
Resolves #4228 

Add a hook for easier plugin development :)

@Supryk

### DIFF
--- a/admin/modules/tools/maillogs.php
+++ b/admin/modules/tools/maillogs.php
@@ -301,6 +301,8 @@ if(!$mybb->input['action'])
 		$log['subject'] = htmlspecialchars_uni($log['subject']);
 		$log['dateline'] = my_date('relative', $log['dateline']);
 
+		$plugins->run_hooks("admin_tools_maillogs_log");
+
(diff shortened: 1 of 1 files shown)


admin/modules/tools/maillogs.php
```diff
@@ -301,6 +301,8 @@
 		$log['subject'] = htmlspecialchars_uni($log['subject']);
 		$log['dateline'] = my_date('relative', $log['dateline']);
 
+		$plugins->run_hooks("admin_tools_maillogs_log");
```
euantorano, line 304: We might want to pass `$log` as a parameter to this hook I think?

Sama34: @euantorano this hook is in the global scope, so there is no need.

Sama34 (approved):

PR #4233, opened 2021-01-11
Detect missing avatar image file and show correct error message

### LINKED ISSUE #4232: Avatar Upload - Wrong Error Message
First disable the "Allow Remote Avatars" option from ACP.

Now go to UCP &gt; Change Avatar : The input field for remote avatar URL will be hidden. Only the local file input will be available.

Don't select any image file to upload, just hit "Change Avatar" button. You will get this error message in place of missing image alert:

Please correct the following errors before continuing:Remote avatar URLs have been disabled by the forum administrator.

Original thread: [Avatar Upload - Wrong Error Message](https://community.mybb.com/thread-230812.html)

effone:
Resolves #4232

### DIFF
--- a/inc/languages/english/usercp.lang.php
+++ b/inc/languages/english/usercp.lang.php
@@ -305,6 +305,7 @@ $l['error_avatarurltoolong'] = "Sorry, but we cannot change your avatar as the n
 $l['error_avatartoobig'] = "Sorry, but we cannot change your avatar as the new avatar you specified is too big. The maximum dimensions are {1}x{2} (width x height).";
 $l['error_invalidavatarurl'] = "The URL you entered for your avatar does not appear to be valid. Please ensure you enter a valid URL.";
 $l['error_remote_avatar_not_allowed'] = "Remote avatar URLs have been disabled by the forum administrator.";
+$l['error_avatarimagemissing'] = "Avatar image file is missing, please select one before attempting to upload.";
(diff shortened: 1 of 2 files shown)


dvz:
We could, alternatively, ignore the request in this case.

effone:
> We could, alternatively, ignore the request in this case.

Thats statement doesn't make a decision. Sir. Let me know what exactly you want and I will change my PR accordingly.

IMO, mal-operation should be intimated that we know you did wrong.

dvz:
The added error may help when `!$_FILES['avatarupload']['name']` (for whatever reason) and remote avatars are disabled, but when they're enabled, it still shows an URL-related error (instead of informing the user that neither method seemed to work, but we don't know which one).
There also seems to be a related, unused string: https://github.com/mybb/mybb/blob/mybb_1824/inc/languages/english/usercp.lang.php#L340 (it can be removed or updated and used to address this),

By not doing anything we can reduce the number of language strings, but either approach seems OK.

Ben-MyBB:
For what it's worth, this approach seems best to me.

Sama34 (approved):

awslabspl:
> We could, alternatively, ignore the request in this case.

I disagree. It would create confusion amongst less experienced users.

euantorano:
I much prefer an error message to silently failing. This seems like the best approach to me.

PR #4234, opened 2021-01-13
Fix #4231: client-side support for rgba() color specs

### LINKED ISSUE #4231: [color] tags with rgba() colour spec are unrendered
e.g., [color=rgba(224, 224, 224, 0.8)]The color tag surrounding this text will not render[/color].

This bug was reported on Discord.

I'm not sure how the rgba() colour specification got there in the first place; presumably it was due to a copy-n-paste operation of rich text.

Neither the editor (as part of its colour normalisation) nor the MyBB parser currently handle rgba() colour specifications.

Original thread: [[color] tags with rgba() colour spec are unrendered](https://community.mybb.com/thread-230808.html)

lairdshaw:
Fix #4231 by converting rgba() in the editor to a 6-digit hex color
(discarding the transparency component).

Resolves #4231

### DIFF
--- a/jscripts/bbcodes_sceditor.js
+++ b/jscripts/bbcodes_sceditor.js
@@ -753,12 +753,19 @@ $(function ($) {
 		colorStr = colorStr || '#000';
 
 		// rgb(n,n,n);
-		if ((match =
-			colorStr.match(/rgb\((\d{1,3}),\s*?(\d{1,3}),\s*?(\d{1,3})\)/i))) {
+		if ((match = colorStr.match(/rgb\((\d{1,3}),\s*?(\d{1,3}),\s*?(\d{1,3})\)/i))) {
 			return '#' +
 				toHex(match[1]) +
 				toHex(match[2]) +
 				toHex(match[3]);
+		}
+
+		// rgba(n,n,n,f.p);
+		if ((match = colorStr.match(/rgba\((\d{1,3}),\s*?(\d{1,3}),\s*?(\d{1,3}),\s*?(\d*\.?\d+\s*)\)/i))) {
+			return '#' +
+			toHex(match[1]) +
+			toHex(match[2]) +
+			toHex(match[3]);
 		}
 
 		// expand shorthand


Xazin:
Does this take into consideration once alpha is 1 or above? Seems to me like it's not able to correctly find the match once alpha isn't a float/double such as `0.1`.

For some reason, once I try using alpha 1 or above (100%) then it is unable to match the color. Might be worth trying to replicate.

lairdshaw:
> Does this take into consideration once alpha is 1 or above?

Yep. The relevant part of the regex is `\d*\.?\d+`, meaning "zero or more digits optionally followed by a dot, followed by one or more digits". This matches 1, 1.0, 1.1, etc.

> Seems to me like it's not able to correctly find the match once alpha isn't a float/double such as `0.1`.

Huh. That's odd. It works for me (example matches above tested). Note that a scenario (the only one?) in which it _won't_ work is when you put the editor into source mode, directly enter something like `[color=rgba(100, 100, 100, 1.0)]Test[/color]`, and then click _Post Reply_ - but in that scenario, it won't work for _any_ alpha value, not just those in which alpha is 1+.

> For some reason, once I try using alpha 1 or above (100%) then it is unable to match the color. Might be worth trying to replicate.

I can't replicate it, except in the scenario described above.

Xazin:
> > Does this take into consideration once alpha is 1 or above?
> 
> Yep. The relevant part of the regex is `\d*\.?\d+`, meaning "zero or more digits optionally followed by a dot, followed by one or more digits". This matches 1, 1.0, 1.1, etc.
> 
> > Seems to me like it's not able to correctly find the match once alpha isn't a float/double such as `0.1`.
> 
> Huh. That's odd. It works for me (example matches above tested). Note that a scenario (the only one?) in which it _won't_ work is when you put the editor into source mode, directly enter something like `[color=rgba(100, 100, 100, 1.0)]Test[/color]`, and then click _Post Reply_ - but in that scenario, it won't work for _any_ alpha value, not just those in which alpha is 1+.
> 
> > For some reason, once I try using alpha 1 or above (100%) then it is unable to match the color. Might be worth trying to replicate.
> 
> I can't replicate it, except in the scenario described above.

After further checking, I can see that you're right, I also think there might have been a problem with my sceditor, thus I reuploaded necessary files and since then I haven't been able to replicate the issue. Only the scenario you describe where a person does not leave sourcemode before posting their reply containing the rgba values.

PR #4235, opened 2021-01-14
Fix #4176 Profile fields aren't loaded in quick replies

Sama34:
Fix #4176 Profile fields aren't loaded in quick replies

This was already discussed, just need a different pair of eyes for a quick merge.


Eldenroot:
@Sama34 - you tagged wrong issue, not?

PR #4239, opened 2021-01-14
Fix #4238 Add commit hook to attachments forum module (ACP)

### LINKED ISSUE #4238: Add commit hook to attachments forum module (ACP)

Sama34:
Fix #4238 Add commit hook to attachments forum module (ACP)

### DIFF
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -643,6 +643,8 @@ if(!$mybb->input['action'])
 	{
 		$search_sql = '1=1';
 
+		$plugins->run_hooks("admin_forum_attachments_commit_start");
+
 		// Build the search SQL for users
 
 		// List of valid LIKE search fields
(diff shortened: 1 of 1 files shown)


Ben-MyBB:
Note: We would need to add this onto the hooks page https://docs.mybb.com/1.8/development/plugins/hooks/

euantorano:
The hooks page is automatically generated, so it should just be a case of regenerating it.

Sama34:
I think I will add more hooks to this module, just spare me a couple of days for reviewing so.

awslabspl:
@euantorano If the page is automatically generated than there should be no need to trigger regeneration as new version will trigger regeneration automatically, am I right?

Sama34:
It is currently complicate to tweak the filter options. There are other places where hooks could be useful which I will add.

euantorano:
> @euantorano If the page is automatically generated than there should be no need to trigger regeneration manually as new version will trigger regeneration automatically, am I right?

I believe that part of our release workflow is to regenerate the page, but I don’t think it gets triggered automatically.

euantorano:
I’m going to merge this as is for now, and we can open more PRs for any other hooks.

PR #4241, opened 2021-01-24
Fix for #4219

### LINKED ISSUE #4219: Password Length Settings
In ACP > Login and Registration Options, Minimum Password Length (`minpasswordlength`) can be set greater than Maximum Password Length (`maxpasswordlength`).
There is no validation in place. Resulting user impossible to register.

Xazin:
Resolves #4219

- minnamelength cannot be larger than maxnamelength
- minpasswordlength cannot be larger than maxpasswordlength
- If complex password is required minpasswordlength cannot be less than 3 (1 uppercase, 1 lowercase & 1 number = 3 chars)


admin/modules/config/settings.php
```diff
@@ -965,6 +965,28 @@
 			}
 		}
 
+		// Validate minnamelength, maxnamelength, minpasswordlength (complex and regular) and maxpasswordlength
+		if ($gid == 9)
```
Sama34, line 969: We should avoid magic numbers on the future. @mybb/developers

PR #4242, opened 2021-01-24
Fix for #4240

### LINKED ISSUE #4240: Images are unviewable but rendered
Under Settings -&gt; Attachments -&gt; Show Attached Thumbnails in Posts -&gt; Full Size Image

If the user can't download images these are rendered anyways.

If the user can't download the full image it might be better to render the thumbnail.

Original thread: [Images are unviewable but rendered](https://community.mybb.com/thread-230845.html)

Xazin:
Resolves #4240 

As @Sama34 (Omar) said, it should show the thumbnail of an image in the attachment area if the user does not have permission to download the image.

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -1019,7 +1019,7 @@ function get_post_attachments($id, &$post)
 					}
 					elseif((($attachment['thumbnail'] == "SMALL" && $forumpermissions['candlattachments'] == 1) || $mybb->settings['attachthumbnails'] == "no") && $isimage)
 					{
-						if (forum_permissions($post['fid'])['candlattachments'])
+						if ($forumpermissions['candlattachments'])
 						{
 							eval("\$post['imagelist'] .= \"".$templates->get("postbit_attachments_images_image")."\";");
 						} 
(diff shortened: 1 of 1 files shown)


inc/functions_post.php
```diff
@@ -1019,7 +1019,20 @@ function get_post_attachments($id, &$post)
...
+						if ($forumpermissions['candlattachments'])
+						{
+							eval("\$post['imagelist'] .= \"".$templates->get("postbit_attachments_images_image")."\";");
+						} 
+						else 
+						{
+							eval("\$post['thumblist'] .= \"".$templates->get("postbit_attachments_thumbnails_thumbnail")."\";");
+							if($tcount == 5)
+							{
+								$thumblist .= "<br />";
```
euantorano, line 1031: I can't seem to work out where this `$thumblist` variable actually gets used (yes it's used a couple of lines above too), it doesn't seem to be used in a template or anything and is only ever written to and not read. I think it's supposed to be:

```suggestion
								$post['thumblist'] .= "<br />";
```

Any thoughts @mybb/developers @mybb/designers? This shouldn't hold this PR back, but it also affects here: https://github.com/mybb/mybb/blob/ad721421d55f3feac633a44f89c151c380c0c007/inc/functions_post.php#L1015

This is probably a bug in the existing code that's not been noticed until now, so nice work! 😄

PR #4248, opened 2021-02-07
Fix #4247 - thumblist variable

### LINKED ISSUE #4247: Attachment thumbnails aren't properly split into multiple lines
As mentioned [here](https://github.com/mybb/mybb/pull/4242#discussion_r571623405), attachment thumbnails are meant to be split into 5 per line, but the code isn't correct for doing so - rather than appending to `$post['thumblist']`, a variable called `$thumblist` is sued:

- https://github.com/mybb/mybb/blob/feature/inc/functions_post.php#L1013
- https://github.com/mybb/mybb/blob/feature/inc/functions_post.php#L1029
- https://github.com/mybb/mybb/blob/1db7d2d5b38bf2d8f04c6b2fe23b32f1ed777731/portal.php#L718

euantorano:
Resolves #4247

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -1010,7 +1010,7 @@ function get_post_attachments($id, &$post)
 						eval("\$post['thumblist'] .= \"".$templates->get("postbit_attachments_thumbnails_thumbnail")."\";");
 						if($tcount == 5)
 						{
-							$thumblist .= "<br />";
+							$post['thumblist'] .= "<br />";
 							$tcount = 0;
 						}
 						++$tcount;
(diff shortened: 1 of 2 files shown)


Sama34:
https://github.com/mybb/mybb/issues/4247#issue-803036105
> attachment thumbnails are meant to be split into 5 per line

Please note that this _bug_ caused a _feature_; for instance, in my board I would lose a feature because of this.

To avoid this I think at least one of the following changes need to be added:
* Add a setting for the number of thumbnails per line.
* Add the `<br />` tag to a template.

euantorano:
Yes, I was thinking the same but kept the initial PR simple for now. I’ll add a setting, and also look at the best way to add it to a template.

Sama34:
Will you delay this for later @euantorano ?

euantorano:
> Will you delay this for later @euantorano ?

Yes, as the fix is slightly larger (setting + template changes), I thought it might be easier to delay to 1.9.

For an example, doing this in 1.9's template system is __much__ easier:

```twig

{% for thumb in thumbs %}
    {# output thumb... #}
    {% if loop.index is divisible by(3) %}
        <br />
    {% endif %}
{% endfor %}
```

PR #4255, opened 2021-02-19
Use forums table counters for calculating thread count in forum view

### LINKED ISSUE #4253: Use cached values for calculating thread count in forum view
The number of threads counted using a single query at https://github.com/mybb/mybb/blob/mybb_1824/forumdisplay.php#L610 can be optimized by taking advantage of visible, soft-deleted, and unapproved thread counts available in the `mybb_forums` table, depending on the user's permissions.

Similar solution was removed in https://github.com/mybb/mybb/pull/3190/files

Regression after #3190

dvz:
Resolves #4253

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -609,8 +609,39 @@ if(isset($fpermissions['canonlyviewownthreads']) && $fpermissions['canonlyviewow
 if($fpermissions['canviewthreads'] != 0)
 {
 	// How many threads are there?
-	$query = $db->simple_select("threads t", "COUNT(tid) AS threads", "fid = '$fid' $tuseronly $tvisibleonly $datecutsql2 $prefixsql2");
-	$threadcount = $db->fetch_field($query, "threads");
+	if ($useronly === "" && $datecutsql === "" && $prefixsql === "")
+	{
+		$threadcount = 0;
+
+		$query = $db->simple_select("forums", "threads, unapprovedthreads, deletedthreads", "fid=".(int)$fid);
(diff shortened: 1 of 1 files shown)


dragonexpert:
I tested on my site and the code if you don't have any special parameters set like canonlyviewownthreads it passes.  It does manually load the template: forumdisplay_forumsort.

forumdisplay.php
```diff
@@ -609,8 +609,39 @@
 if($fpermissions['canviewthreads'] != 0)
 {
 	// How many threads are there?
-	$query = $db->simple_select("threads t", "COUNT(tid) AS threads", "fid = '$fid' $tuseronly $tvisibleonly $datecutsql2 $prefixsql2");
-	$threadcount = $db->fetch_field($query, "threads");
+	if ($useronly === "" && $datecutsql === "" && $prefixsql === "")
+	{
+		$threadcount = 0;
+
+		$query = $db->simple_select("forums", "threads, unapprovedthreads, deletedthreads", "fid=".(int)$fid);
```
Sama34, line 616: Why do you cast here but not below? It isn't necessary.

Sama34 (approved):

forumdisplay.php
dvz: Just a good habit when rewriting queries (easier to spot unescaped input, even though currently most queries don't include it)

Sama34 (approved):
Seems to work for me.

Edit: Performance wise was meaningless for me, but I only have 13.5k rows in the threads table.

PR #4259, opened 2021-02-20
Fixes #4258 Not possible to remove signature length limit

### LINKED ISSUE #4258: Not possible to remove signature length limit
"_The maximum number of characters a user can place in a signature. Set to 0 to disable the limitation._"

**If you set 0:** "_You cannot update your signature because it is too long. The maximum length for signatures is 0 characters._"

But wasn't 0 to remove the limit?

Starpaul20:
Resolves #4258

Also fixed the max characters in the User CP to say unlimited when the setting is set to 0.

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -1858,7 +1858,7 @@ class UserDataHandler extends DataHandler
 		$parsed_sig = preg_replace("#\s#", "", $parsed_sig);
 		$sig_length = my_strlen($parsed_sig);
 
-		if($sig_length > $mybb->settings['siglength'])
+		if($sig_length > $mybb->settings['siglength'] && $mybb->settings['siglength'] > 0)
 		{
 			$this->set_error('sig_too_long', array($mybb->settings['siglength']));
 
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -1859,19 +1859,19 @@ class UserDataHandler extends DataHandler
 		{
 			$parsed_sig = preg_replace("#\s#", "", $parsed_sig);
 			$sig_length = my_strlen($parsed_sig);
-		}
 
-		if($sig_length > $mybb->settings['siglength'] && $mybb->settings['siglength'] > 0)
-		{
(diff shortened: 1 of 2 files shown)


Eldenroot:
Thank you!

inc/datahandlers/user.php
```diff
@@ -1858,7 +1858,7 @@ public function verify_signature()
 		$parsed_sig = preg_replace("#\s#", "", $parsed_sig);
 		$sig_length = my_strlen($parsed_sig);
```
Sama34: This should be set only if `$mybb->settings['siglength'] > 0` is true. I will accept anyways in case I'm the only one.

Sama34 (approved):

Xazin (approved):
Would be great to have the change that Omar pointed out.

Didn't find further issues when testing.

inc/datahandlers/user.php
Eldenroot: Good idea, others?

inc/datahandlers/user.php
```diff
@@ -1855,10 +1855,13 @@ public function verify_signature()
...
-		$parsed_sig = preg_replace("#\s#", "", $parsed_sig);
-		$sig_length = my_strlen($parsed_sig);
+		if($mybb->settings['siglength'] > 0)
+		{
+			$parsed_sig = preg_replace("#\s#", "", $parsed_sig);
+			$sig_length = my_strlen($parsed_sig);
+		}
 
-		if($sig_length > $mybb->settings['siglength'])
+		if($sig_length > $mybb->settings['siglength'] && $mybb->settings['siglength'] > 0)
```
Sama34: Sorry to be a bother but I think `$mybb->settings['siglength'] > 0` should be the first check in case `$sig_length` isn't set.

@dvz Am I on the right track or I'm just saying nonsense and we could get this merged?

Sama34 (approved):

inc/datahandlers/user.php
dvz: It looks like this block can be moved inside the one above to solve this issue and simplify the code.

Sama34 (approved):

PR #4261, opened 2021-02-22
Suggestion for PM Folders

Xazin:
This PR creates a 'delete' button below removable folders inside _Edit Folders_ in the Private Messaging Panel. 

The function rebuilds the folders without the _folder id_ in question, and deletes all private messages in that folder.

Added a preview to show how the button looks.
HyT6sX6


Xazin:
Also, inside compose, tracking and edit folders, the folder specific navigation is missing. Might add it to this PR.

Xazin:
I've injected `$folders` such that folders are in navigation whilst viewing compose, tracking and edit folders.

euantorano:
Note: I'm not ignoring this, but will merge it at a later date.

Xazin:
Will resolve merge conflicts.

@euantorano , can you take a quick look and let me know if there are any changes you want here as well?

Xazin:
@euantorano Resolved conflicts.

I saw that there are predefined 5 empty text fields to add new folders, seem quite redundant? Why not just have an add more option to add a new text field option? Opinion?

PR #4266, opened 2021-02-23
Fixes #4262 - Exclude bots, spiders and thread authors from thread views

### LINKED ISSUE #4262: Add options to prevent spiders and thread autor from increasing thread view count
I found a solution here: https://community.mybb.com/post-882217.html

But I ask why is there not an option in the settings to prevent this or even a standard configuration like the one above?

Eldenroot:
Resolves #4262 

Thank you buddy @lairdshaw - for your mentoring, I really appreciate it!

### DIFF
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -882,14 +882,14 @@ min=0]]></optionscode>
 			<disporder>15</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
-			<isdefault>1</isdefault>
+			<isdefault>0</isdefault>
 		</setting>
 		<setting name="threadviewcountexcludeguests">
 			<title>Exclude views by guests in thread view count?</title>
 			<description><![CDATA[Here you can exclude views by guests from the thread view counter.]]></description>
 			<disporder>16</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
-			<settingvalue><![CDATA[1]]></settingvalue>
+			<settingvalue><![CDATA[0]]></settingvalue>
 			<isdefault>0</isdefault>
 		</setting>
 		<setting name="threadviewcountexcludethreadauthors">
@@ -897,7 +897,7 @@ min=0]]></optionscode>
 			<description><![CDATA[Here you can exclude views by thread author from the thread view counter.]]></description>
 			<disporder>17</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
-			<settingvalue><![CDATA[1]]></settingvalue>
+			<settingvalue><![CDATA[0]]></settingvalue>
 			<isdefault>0</isdefault>
 		</setting>
 	</settinggroup>
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -878,7 +878,7 @@ min=0]]></optionscode>
 		</setting>
 		<setting name="threadviewcountexcludespiders">
 			<title>Exclude views by spiders in thread view count?</title>
-			<description><![CDATA[Here you can turn off excluding views by spiders in the thread view counter.]]></description>
+			<description><![CDATA[Here you can exclude views by spiders from the thread view counter.]]></description>
 			<disporder>15</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
@@ -886,7 +886,7 @@ min=0]]></optionscode>
 		</setting>
 		<setting name="threadviewcountexcludeguests">
 			<title>Exclude views by guests in thread view count?</title>
-			<description><![CDATA[Here you can turn off excluding views by guests in the thread view counter.]]></description>
+			<description><![CDATA[Here you can exclude views by gueststhread author from the thread view counter.]]></description>
 			<disporder>16</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
@@ -894,7 +894,7 @@ min=0]]></optionscode>
 		</setting>
 		<setting name="threadviewcountexcludethreadauthors">
 			<title>Exclude views by thread author in thread view count?</title>
-			<description><![CDATA[Here you can turn off excluding views by thread author in the thread view counter.]]></description>
+			<description><![CDATA[Here you can exclude views by thread author from the thread view counter.]]></description>
 			<disporder>17</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -882,7 +882,7 @@ min=0]]></optionscode>
 			<disporder>15</disporder>
 			<optionscode><![CDATA[yesno]]></optionscode>
 			<settingvalue><![CDATA[0]]></settingvalue>
-			<isdefault>0</isdefault>
+			<isdefault>1</isdefault>
 		</setting>
 		<setting name="threadviews_countguests">
 			<title>Increment Thread View Count on Guest Visits?</title>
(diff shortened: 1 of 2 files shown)


lairdshaw:
Happy to have helped, buddy. I think you ended up with a good outcome.

showthread.php
```diff
@@ -734,15 +734,32 @@
...
+	if($mybb->settings['threadviewcountexcludeguests'] == 1 && $mybb->user['uid'] == 0)
+	{
+		$count_view = false;
+	}
+	if($mybb->settings['threadviewcountexcludethreadauthor'] == 1 && $mybb->user['uid'] == $thread['uid'])
+	{
+		$count_view = false;
+	}
+	
+	if($count_view == true)
```
dvz: It might be better to use a single `if` condition for the block below

dvz:
It may be more intuitive to flip the meaning of these settings and use _include_ instead of _exclude_.

showthread.php
lairdshaw: Yes, you're right, it might be. @Eldenroot and I were going for an "Easy enough for a total noob to grok" control flow.

lairdshaw:
> It may be more intuitive to flip the meaning of these settings and use _include_ instead of _exclude_.

Doh. It was me who advised the inversion. I still stand by that choice. I think inclusion is more ambiguous in this scenario. e.g., The user is a guest, and the setting says "include guests" so s/he is included, but the user is also *not* a spider, and the setting says "include spiders", so on that basis should s/he _not_ be included...?... It's ambiguous. A principle of exclusion is not (IMO) *as* ambiguous: if *any* of the exclusion criteria are applicable, then the user is excluded. If *none* of them apply, then the user is included.

(I do realise that from a strictly logical point of view, this whole argument could be turned around on me, so I guess all I'm saying is that I find exclusion more intuitive - but if the consensus is that it's not, then that's OK with me).

showthread.php
Eldenroot: @dvz - Done, simplified!

showthread.php
```diff
@@ -734,15 +734,18 @@
 	}
 
 	// Increment the thread view.
-	if($mybb->settings['delayedthreadviews'] == 1)
+	if(!($mybb->settings['threadviewcountexcludespiders'] == 1 && $session->is_spider == true || $mybb->settings['threadviewcountexcludeguests'] == 1 && $mybb->user['uid'] == 0 || $mybb->settings['threadviewcountexcludethreadauthor'] == 1 && $mybb->user['uid'] == $thread['uid']))
```
dvz:
```suggestion
	if(!(
		($mybb->settings['threadviewcountexcludespiders'] == 1 && $session->is_spider == true) ||
		($mybb->settings['threadviewcountexcludeguests'] == 1 && $mybb->user['uid'] == 0) ||
		($mybb->settings['threadviewcountexcludethreadauthor'] == 1 && $mybb->user['uid'] == $thread['uid'])
	))
```
Split + parentheses for clarity

Eldenroot:
@dvz thank you, now it should be fine.

dvz:
Currently _No_ values would result in double negatives - skimming through descriptions and spotting `[...] guests [...]` with a _Yes / No_ setting may be interpreted as _include/exclude guests_.

Would `include X` be mistaken for `[only] include X`, rather than `[also] include X`?

We can also use alternatives like _Increment Thread View Count on Guest Visits_.

install/resources/settings.xml
```diff
@@ -876,6 +876,30 @@ min=0]]></optionscode>
...
+			<optionscode><![CDATA[yesno]]></optionscode>
+			<settingvalue><![CDATA[1]]></settingvalue>
+			<isdefault>0</isdefault>
+		</setting>
+		<setting name="threadviewcountexcludeguests">
+			<title>Exclude views by guests in thread view count?</title>
+			<description><![CDATA[Here you can exclude views by guests from the thread view counter.]]></description>
+			<disporder>16</disporder>
+			<optionscode><![CDATA[yesno]]></optionscode>
+			<settingvalue><![CDATA[0]]></settingvalue>
```
Sama34: Should be On by default.

install/resources/settings.xml
```diff
@@ -876,6 +876,30 @@ min=0]]></optionscode>
...
+			<isdefault>0</isdefault>
+		</setting>
+		<setting name="threadviewcountexcludethreadauthors">
+			<title>Exclude views by thread author in thread view count?</title>
+			<description><![CDATA[Here you can exclude views by thread author from the thread view counter.]]></description>
+			<disporder>17</disporder>
+			<optionscode><![CDATA[yesno]]></optionscode>
+			<settingvalue><![CDATA[0]]></settingvalue>
```
Sama34: Should be On by default too.

install/resources/settings.xml
```diff
@@ -876,6 +876,30 @@ min=0]]></optionscode>
 			<settingvalue><![CDATA[0]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="threadviewcountexcludespiders">
+			<title>Exclude views by spiders in thread view count?</title>
+			<description><![CDATA[Here you can exclude views by spiders from the thread view counter.]]></description>
```
Sama34:
```suggestion
		<setting name="threadviews_countspiders">
			<title>Increment Thread View Count on Spider Visits</title>
			<description><![CDATA[Turn this setting On to increment the thread view count on spiders visits.]]></description>
```

Take this how as guide for the rest. Note the wording as this setting (as per suggestion) should be the one that is Off by default.

Sama34 (changes requested):

Sama34:
I think a _include_ approach with a _Increment Thread View Count on Guest Visits_ wording is the best approach to avoid confusion.

Eldenroot:
@Sama34 - please check this comment from @dvz - https://github.com/mybb/mybb/issues/4262#issuecomment-784451016

It was changed to off by default because of the request.

lairdshaw:
I defer to your stronger position, @dvz and @Sama34. Exclusion was not, as I'd initially thought, as good a framing as inclusion. You guys are correct...

install/resources/settings.xml
Eldenroot: Done!

install/resources/settings.xml
Eldenroot: Done!

install/resources/settings.xml
Eldenroot: Done!

Eldenroot:
Thank you all for your valuable feedback! Now it should be fixed and all your suggestions were implemented.

showthread.php
```diff
@@ -734,15 +734,22 @@
 	}
 
 	// Increment the thread view.
-	if($mybb->settings['delayedthreadviews'] == 1)
+	if(
+		($mybb->settings['threadviews_countspiders'] == 1 && $session->is_spider == true) ||
+		($mybb->settings['threadviews_countguests'] == 1 && $mybb->user['uid'] == 0) ||
+		($mybb->settings['threadviews_countthreadauthor'] == 1 && $mybb->user['uid'] == $thread['uid'])
+	)
```
dvz:
```suggestion
	if(
		(
			$mybb->user['uid'] == 0 &&
			(
				($session->is_spider == true && $mybb->settings['threadviews_countspiders'] == 1) ||
				($session->is_spider == false && $mybb->settings['threadviews_countguests'] == 1)
			)
		) ||
		(
			$mybb->user['uid'] != 0 &&
			($mybb->settings['threadviews_countthreadauthor'] == 1 || $mybb->user['uid'] != $thread['uid'])
		)
	)
```

Addressed:
- missed visitors from none of the groups from settings
- confusing guests and spiders (`uid == 0`)
- confusing guest thread authors with other guests

install/resources/settings.xml
```diff
@@ -876,6 +876,30 @@ min=0]]></optionscode>
 			<settingvalue><![CDATA[0]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="threadviews_countspiders">
+			<title>Increment Thread View Count on Spider Visits?</title>
+			<description><![CDATA[Here you can include views by spiders in the thread view counter.]]></description>
+			<disporder>15</disporder>
+			<optionscode><![CDATA[yesno]]></optionscode>
+			<settingvalue><![CDATA[0]]></settingvalue>
+			<isdefault>0</isdefault>
```
dvz: I believe `isdefault` should be `1` for all core settings

Eldenroot: I will fix in 9 hours. I was confused with this, not properly documented.

Thx

Eldenroot: Changed.

PR #4273, opened 2021-02-24
Resolves #4270 - Correct assignment operators

### LINKED ISSUE #4270: Timezone select, UCP's Thread Subscriptions list output overwritten
Some variable concatenating assignment operators `.=` were incorrectly changed to assignment operators `=` in #4095, resulting in overwriting of loop iteration data.

- timezone select input: https://github.com/mybb/mybb/blob/105efd28476ed23a84d111026eaec314e6c583b8/usercp.php#L4452, modified via https://github.com/mybb/mybb/commit/1db7d2d5b38bf2d8f04c6b2fe23b32f1ed777731#diff-f7fdbf357a321ef157867099148616d4fda48504a761d7db5aca5d3076cd293cL4438-R4446
- User CP's _Thread Subscriptions With New Posts_: https://github.com/mybb/mybb/blob/105efd28476ed23a84d111026eaec314e6c583b8/inc/functions.php#L7156, modified via https://github.com/mybb/mybb/commit/1db7d2d5b38bf2d8f04c6b2fe23b32f1ed777731#diff-f3a2689b39e6a1a9532d63ed0bc25e99cafc51ae5d7486c07b1e68e5a9681a3fL7134-R7142

Regression from #4095

Ben-MyBB:
Resolves #4270

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -7153,7 +7153,7 @@ function build_timezone_select($name, $selected=0, $short=false)
 			$label = $lang->sprintf($lang->timezone_gmt_short, $label." ", $time_in_zone);
 		}
 
-		eval("\$timezone_option = \"".$templates->get("usercp_options_timezone_option")."\";");
+		eval("\$timezone_option .= \"".$templates->get("usercp_options_timezone_option")."\";");
 	}
 
 	eval("\$select = \"".$templates->get("usercp_options_timezone")."\";");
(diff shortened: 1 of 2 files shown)


dvz:
We'll also need to initialize these variables before the loops to avoid errors

Sama34 (approved):
Please fix what is mentioned above and it should be good to go.

PR #4281, opened 2021-02-26
Fix Wrong PHP Directive

### LINKED ISSUE #4280: Incorrect name used for `upload_max_filesize` PHP directive
It started in this thread, because something was not clear to me.
https://community.mybb.com/thread-231391.html

In pursuing it, I notice code is trying to pull a php directive, but it is not right.

Lots of people have viewed the thread but no one has replied. The thread will fall out of sight with no replies, so I am pointing out the error here.

Original thread: [incorrect php directive called](https://community.mybb.com/thread-231406.html)

effone:
Resolves #4280

### DIFF
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -8,6 +8,29 @@
  *
  */
 
+/**
+ * Get maximum upload filesize limit set in PHP
+ * @since MyBB 1.8.27
+ * @return int maximum allowed filesize
+ */
+function get_phpuploadlimit()
+{	
+	$maxsize = array(return_bytes(ini_get('upload_max_filesize')), return_bytes(ini_get('post_max_size')));
+	$maxsize = array_filter($maxsize); // Remove empty values
+
+	if(empty($maxsize))
+	{
+		return 0;
+	}
+	else if(count($maxsize) == 1)
+	{
+		return (int)$maxsize[0];
+	}
+	else
+	{
+		return (int)min($maxsize);
+	}
+}
 
 /**
  * Remove an attachment from a specific post
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -13,7 +13,7 @@
  * @since MyBB 1.8.27
  * @return int maximum allowed filesize
  */
-function get_phpuploadlimit()
+function get_php_upload_limit()
 {	
 	$maxsize = array(return_bytes(ini_get('upload_max_filesize')), return_bytes(ini_get('post_max_size')));
 	$maxsize = array_filter($maxsize); // Remove empty values
@@ -21,10 +21,6 @@ function get_phpuploadlimit()
 	if(empty($maxsize))
 	{
 		return 0;
-	}
-	else if(count($maxsize) == 1)
-	{
-		return (int)$maxsize[0];
 	}
 	else
 	{
--- a/newthread.php
+++ b/newthread.php
@@ -1125,7 +1125,7 @@ if($mybb->input['action'] == "newthread" || $mybb->input['action'] == "editdraft
 		}
 	}
 
-	$php_max_upload_size = get_phpuploadlimit();
+	$php_max_upload_size = get_php_upload_limit();
 	$php_max_file_uploads = (int)ini_get('max_file_uploads');
 	eval("\$post_javascript = \"".$templates->get("post_javascript")."\";");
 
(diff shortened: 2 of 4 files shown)


lairdshaw:
Working your usual one-line array-function wizardry, I see, @effone. :+1: 

Just a bit of a meta question:

I wonder why we are combining `upload_max_filesize` and `post_max_size` into a single variable in the first place? Is there a reason why it might too challenging/awkward/otherwise-unreasonable for us to make use of and respond independently to the values of those two variables in `post.js`?

You have put the most effort into this area of the code of late which is why I'm asking you in particular.

lairdshaw:
Nevermind. I get it now. No need to answer.

effone:
I'll put a note here that the modification may throw warning if `upload_max_filesize` and `post_max_size` both are falsy (or 0), which impossible scenario I didn't consider while we are talking about actually **uploading** a file.

For the height of the satisfaction I can change it to `(int)@min( ... )` to force it to return a 0, but will it help to upload anything actually?

newthread.php
```diff
@@ -1125,18 +1125,7 @@
...
-	if ($php_max_upload_filesize != 0 && $php_post_max_size != 0)
-	{
-		$php_max_upload_size = min($php_max_upload_filesize, $php_post_max_size);
-	}
-	else
-	{
-		$php_max_upload_size = max($php_max_upload_filesize, $php_post_max_size);
-	}
-
+	$php_max_upload_size = min(array_filter(array(return_bytes(ini_get('upload_max_filesize')), return_bytes(ini_get('post_max_size')))));
```
euantorano: As this logic is spread across 3 places (at the minute - it might spread further with plugins and such), it would make a lot of sense to create a function to do this.

It might also be nice to split this into a few lines to make reading easier if it is moved to a function:

```suggestion
	$php_max_upload_size = @min(
		array_filter(
			array(
				return_bytes(ini_get('upload_max_filesize')), 
				return_bytes(ini_get('post_max_size'))
			)
		)
	);
```

effone:
Although the point of readability is accepted (didn't want to introduce unnecessary variables) but I can't agree with the point of moving it to a function. As we know MyBB is no way DRY at this moment the processing of attachments is just copy paste in all three scripts. No point to isolate this when there is a huge chunk of code is repeated.

euantorano:
> Although the point of readability is accepted (didn't want to introduce unnecessary variables) but I can't agree with the point of moving it to a function. As we know MyBB is no way DRY at this moment the processing of attachments is just copy paste in all three scripts. No point to isolate this when there is a huge chunk of code is repeated.

I know it's currently split all over the place, but if we're making changes I think it's best to do it right the first time round rather than having to go back and fix it later. It should be easy enough to add to `inc/functions_upload.php` which I blieve should already be included in all 3 files.

dvz:
It may be better to deal with `0`s explicitly, rather than using `@`.

effone:
Will modify shortly and will move to function.

newthread.php
awslabspl: While multilining makes code3 more readable, I honestly think this is not needed here. MyBB is written in `php` and - I hope ) no dev has any trouble reading plain-php code.........

effone:
Modified as suggested.

inc/functions_upload.php
```diff
@@ -8,6 +8,29 @@
  *
  */
 
+/**
+ * Get maximum upload filesize limit set in PHP
+ * @since MyBB 1.8.27
+ * @return int maximum allowed filesize
+ */
+function get_phpuploadlimit()
```
dvz:
```suggestion
function get_php_upload_limit()
```

inc/functions_upload.php
```diff
@@ -8,6 +8,29 @@
...
+	$maxsize = array(return_bytes(ini_get('upload_max_filesize')), return_bytes(ini_get('post_max_size')));
+	$maxsize = array_filter($maxsize); // Remove empty values
+
+	if(empty($maxsize))
+	{
+		return 0;
+	}
+	else if(count($maxsize) == 1)
+	{
+		return (int)$maxsize[0];
```
dvz: The `0` key may not be set as `array_filter()` can leave gaps in keys - we can use `min()` even if there is only one result

dvz (changes requested):

inc/functions_upload.php
effone: I forgot that `min()` accepts single value through array.

dvz (changes requested):
The function calls may currently fail in `newthread.php` and `newreply.php`, since the `inc/functions_upload.php` file doesn't get included.

dvz (approved):

PR #4283, opened 2021-02-26
Fixes #4277 - Who is online - add users per page settings

### LINKED ISSUE #4277: Who is online limit (per page) setting is taken from thread settings
The who is online list limit (per page) setting is taken from the show thread setting group, which makes no sense whatsoever  🤪

https://github.com/mybb/mybb/blob/66dca7...hp#L47-L53

It should have its own setting.

Original thread: [Who is online limit (per page) setting is taken from thread settings](https://community.mybb.com/thread-231404.html)

Eldenroot:
Resolves #4277

### DIFF
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -1834,7 +1834,7 @@ activity=Order By Last Activity (DESC)]]></optionscode>
 			<isdefault>1</isdefault>
 		</setting>
 		<setting name="wolusersperpage">
-			<title>Users per Page</title>
+			<title>Users Per Page</title>
 			<description><![CDATA[The number of users displayed per page.]]></description>
 			<disporder>4</disporder>
 			<optionscode><![CDATA[numeric
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -1833,6 +1833,15 @@ activity=Order By Last Activity (DESC)]]></optionscode>
 			<settingvalue><![CDATA[username]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="wolusersperpage">
+			<title>Users per Page</title>
+			<description><![CDATA[The number of users displayed per page.]]></description>
+			<disporder>4</disporder>
+			<optionscode><![CDATA[numeric
+min=0]]></optionscode>
+			<settingvalue><![CDATA[20]]></settingvalue>
+			<isdefault>1</isdefault>
+		</setting>
 	</settinggroup>
 	<settinggroup name="userpruning" title="User Pruning" description="User Pruning allows you to remove users from your forum meeting certain criteria. Here you can configure that criteria." disporder="19" isdefault="1">
 		<setting name="enablepruning">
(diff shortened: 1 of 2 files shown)


online.php
```diff
@@ -44,13 +44,8 @@
 	$query = $db->simple_select("users", "COUNT(uid) AS users", "lastactive > '{$threshold}' AND invisible = '1'");
 	$invis_count = $db->fetch_field($query, "users");
 
-	if(!$mybb->settings['threadsperpage'] || (int)$mybb->settings['threadsperpage'] < 1)
-	{
-		$mybb->settings['threadsperpage'] = 20;
-	}
-
```
Starpaul20: This block should be kept, only the setting name should be changed.

Starpaul20 (changes requested):

install/resources/settings.xml
```diff
@@ -1833,6 +1833,15 @@ activity=Order By Last Activity (DESC)]]></optionscode>
 			<settingvalue><![CDATA[username]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="wolusersperpage">
+			<title>Users Per Page</title>
+			<description><![CDATA[The number of users displayed per page.]]></description>
+			<disporder>4</disporder>
+			<optionscode><![CDATA[numeric
+min=0]]></optionscode>
```
Sama34, line 1841: What will happen if it is `0`= Maybe the min should be `1`

Sama34 (changes requested):

install/resources/settings.xml
lairdshaw: That was my initial thought as an alternative to the suggestion by @Starpaul20 above. Doing a bit of hunting through the code for "perpage" though, it seems that it is general practice for there to be a zero minimum for per-page settings which is interpreted as "use the default", which itself seems to be generally (universally?) hard-coded as 20.

So, setting a minimum of 1 is possible and would avoid the need for @Starpaul20's recommended change, but it would be a break with current standard practice.

euantorano: I would agree that if there is a convention used by other settings, we should follow that convention.

Eldenroot:
Ok, will edit it today.

Starpaul20 (approved):

install/resources/settings.xml
```diff
@@ -1833,10 +1833,18 @@ activity=Order By Last Activity (DESC)]]></optionscode>
...
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="wolusersperpage">
+			<title>Users Per Page</title>
+			<description><![CDATA[The number of users displayed per page. Set to 0 to use the default.]]></description>
+			<disporder>4</disporder>
+			<optionscode><![CDATA[numeric
+min=0]]></optionscode>
+			<settingvalue><![CDATA[20]]></settingvalue>
+			<isdefault>1</isdefault>
```
Starpaul20, line 1843: You forgot the closing `</setting>`

Starpaul20 (changes requested):

install/resources/settings.xml
Eldenroot: Good spot, it was because resolving merge conflict. I will fix this, thx

Starpaul20 (approved):

Sama34 (approved):

PR #4288, opened 2021-03-04
Fixes #3565 - Thread Subject maxchar settings

### LINKED ISSUE #3565: Enhancement: $lastpost_subject > 25
The number of characters shouldn't be hard coded. It should be set in Forumhome or Forumdisplay settings.

Different templates, not to mention mobile screens, require this to be adaptable.

inc/functions_forumlist.php
Line 312
```
if(my_strlen($lastpost_subject) > 25)
{
	$lastpost_subject = my_substr($lastpost_subject, 0, 25)."...";
}
```

----------

Edit: Infact, can there be no hardcoded limit and use css text-overflow: ellipsis; in the template?

Eldenroot:
Resolves #3565 - CSS is quite difficult to add during updates, so this solution is not perfect but still better that use hardcoded values...

https://github.com/mybb/mybb/issues/3565

### DIFF
--- a/inc/functions_forumlist.php
+++ b/inc/functions_forumlist.php
@@ -312,10 +312,10 @@ function build_forumbits($pid=0, $depth=1)
 					$lastpost_link = get_thread_link($lastpost_data['lastposttid'], 0, "lastpost");
 					$lastpost_subject = $full_lastpost_subject = $parser->parse_badwords($lastpost_data['lastpostsubject']);
 					
-					// Max amount of characters before the rest of the thread subject is truncated
-					if(my_strlen($lastpost_subject) > $mybb->settings['maxthreadsubjectlenght'] && $mybb->settings['maxthreadsubjectlenght'] != 0)
+					// Max number of characters before the rest of the thread subject is truncated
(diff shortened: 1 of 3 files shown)


Ben-MyBB:
A solution like this may be the best way to tackle the issue in 1.8 but should also be useable in 1.9 unless @justinsoltesz has implemented a CSS solution for 1.9?

@mybb/developers What's your thoughts on this solution?

Eldenroot:
This should be added at least for 1.8 because hardcoded values are not good at all.... for MyBB 1.9 we can combine with css or leave this.

Sama34:
I think this should be fixed at the CSS level, breaking the subject to fit a set (and unjustified) value doesn't seem right.

I think we should deprecate this approach and move on to let CSS handle stuff it should handle. Lets add a CSS fix and use `{$full_lastpost_subject}` in the default theme after that point (to avoid breaking themes, `{$lastpost_subject}` should be left in core in 1.8).

Eldenroot:
So CSS solution is the final move?

euantorano:
Yes.

PR #4291, opened 2021-03-04
Fixes #4287 - Inconsistency with notices in header

### LINKED ISSUE #4287: Inconsistency with notices in header
For example, the awaiting moderation header will show like such when you hover over it:

Awaiting Moderation: 1 unapproved thread.

The group leader notice will show like this when you hover over it:

Group Leader Notice: You have 1 pending group membership join request.

Original thread: [Inconsistency with notices in header](https://community.mybb.com/thread-231493.html)

Eldenroot:
Resolves #4287

### DIFF
--- a/global.php
+++ b/global.php
@@ -647,8 +647,7 @@ if($mybb->user['uid'] != 0 && is_array($groupleaders) && array_key_exists($mybb-
 			}
 			else
 			{
-				$total_joinrequests = my_number_format($total_joinrequests);
-				$lang->pending_joinrequests = $lang->sprintf($lang->pending_joinrequests, $total_joinrequests);
+				$lang->pending_joinrequests = $lang->sprintf($lang->pending_joinrequests, my_number_format($total_joinrequests));
 			}
 
 			eval('$pending_joinrequests = "'.$templates->get('global_pending_joinrequests').'";');
--- a/inc/languages/english/global.lang.php
+++ b/inc/languages/english/global.lang.php
@@ -328,8 +328,8 @@ $l['unapproved_attachment'] = "1 unapproved attachment";
 $l['unapproved_attachments'] = "{1} unapproved attachments";
 $l['unread_report'] = "1 unread report";
 $l['unread_reports'] = "{1} unread reports";
-$l['pending_joinrequest'] = "Group Leader Notice: You have 1 pending group membership join request.";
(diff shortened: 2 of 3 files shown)


global.php
```diff
@@ -647,8 +647,7 @@
 			}
 			else
 			{
-				$total_joinrequests = my_number_format($total_joinrequests);
-				$lang->pending_joinrequests = $lang->sprintf($lang->pending_joinrequests, $total_joinrequests);
+				$lang->pending_joinrequests = $lang->sprintf($lang->pending_joinrequests, my_number_format($total_joinrequests));
```
Sama34, line 650: What is the reasoning for this change?

Sama34 (approved):

global.php
Eldenroot: Maybe I should revert it back if it is necessary. I wanted to solve the mentioned issue in a different way but then I found easier way without a lot of changes... this should be fine, but no problem to change it to the default code.

Sama34: I don't mind, but it would probably be better to keep has it was? @mybb/developers

Ben-MyBB: If it's no longer being used to resolve the issue we should probably revert this part.

Ben-MyBB:
@Eldenroot we'll need to bump the template here as well

Eldenroot:
Done!

inc/languages/english/global.lang.php
```diff
@@ -328,8 +328,9 @@
 $l['unapproved_attachments'] = "{1} unapproved attachments";
 $l['unread_report'] = "1 unread report";
 $l['unread_reports'] = "{1} unread reports";
-$l['pending_joinrequest'] = "Group Leader Notice: You have 1 pending group membership join request.";
-$l['pending_joinrequests'] = "Group Leader Notice: You have {1} pending group membership join requests.";
+$l['groupleader_notice'] = "Group Leader Notice:";
```
effone, line 331: What about "Pending Group Leader Action(s)"?

Sama34: How many _actions_ are there for group leaders?

Sama34 (approved):

Eldenroot:
@Sama34 - I wanted to use same approach how we have for moderators notice... if it is not good, I can change it.

Eldenroot:
I think this is done and can be merged :)

Sama34:
I did test.

PR #4295, opened 2021-03-10
Fix #4294: Searching for members with underscores in their name via…

### LINKED ISSUE #4294: Searching for members with underscores in their name via memberlist.php fails
As reported by Cloud on Discord.

The problem is line #225 of memberlist.php, which uses the variable $username_like_query, which has been escaped for a LIKE condition, whereas we need it to be escaped as an ordinary string.

A fix is to change that line from:            $search_query .= &quot; AND u.username='{$username_like_query}'&quot;;

to:			$username_esc = $db-&gt;escape_string($search_username); 
			$search_query .= &quot; AND u.username='{$username_esc}'&quot;;

Original thread: [Searching for members with underscores in their name via memberlist.php fails](https://community.mybb.com/thread-231647.html)

lairdshaw:
…memberlist.php fails

Resolves #4294

### DIFF
--- a/memberlist.php
+++ b/memberlist.php
@@ -222,7 +222,8 @@ else
 		// Exact
 		else
 		{
-			$search_query .= " AND u.username='{$username_like_query}'";
+			$username_esc = $db->escape_string($search_username); 
+			$search_query .= " AND u.username='{$username_esc}'";
 		}
 
 		$search_url .= "&username=".urlencode($search_username);


Sama34 (approved):

Sama34:
Now that I think about it, does this handle upper case characters? Usually the core does convert usernames to lower case for queries, but I'm unsure why this isn't followed in all places.

lairdshaw:
> Now that I think about it, does this handle upper case characters? Usually the core does convert usernames to lower case for queries, but I'm unsure why this isn't followed in all places.

On MySQL/MariaDB, the comparison is case-insensitive. I'm not sure about the other DB engines.

euantorano:
> > Now that I think about it, does this handle upper case characters? Usually the core does convert usernames to lower case for queries, but I'm unsure why this isn't followed in all places.
> 
> On MySQL/MariaDB, the comparison is case-insensitive. I'm not sure about the other DB engines.

Postgres is case sensitive by default - you can use `ILIKE` to get round it or convert to lower. Not sure about SQLite.

Sama34:
I think the following is on point with your comments (both):
https://github.com/mybb/mybb/blob/34dc6258cde31903df8dea9676c1813e07ec1253/usercp.php#L2955
https://github.com/mybb/mybb/blob/34dc6258cde31903df8dea9676c1813e07ec1253/inc/functions.php#L6384-L6395

Plus the one below (which doesn't match your comments but still valid enough):
https://github.com/mybb/mybb/blob/34dc6258cde31903df8dea9676c1813e07ec1253/private.php#L578

How should we proceed here @euantorano ?

Note: Please note that this isn't arisen by this issue.

PR #4296, opened 2021-03-11
Fixes #4293 - "Click to edit" added when post content is empty

### LINKED ISSUE #4293: "Click to edit" added when post content is empty
A _Click to edit_ message is added to posts when their content is empty.

With the [Minimum Message Length](https://github.com/mybb/mybb/blob/mybb_1825/install/resources/settings.xml#L1285-L1293), empty posts may suggest a server-side problem, but the unnecessary client-side message may point in the wrong direction when troubleshooting.

Details: https://community.mybb.com/thread-231510-post-1364051.html

Eldenroot:
Resolves #4293

### DIFF
--- a/inc/languages/english/global.lang.php
+++ b/inc/languages/english/global.lang.php
@@ -312,6 +312,7 @@ $l['confirm_title'] = "Please Confirm";
 $l['redirect'] = "You will now be redirected";
 $l['unknown_error'] = "An unknown error has occurred.";
 $l['post_fetch_error'] = 'There was an error fetching the posts.';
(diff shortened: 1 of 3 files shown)


Eldenroot:
Just a note: I will update template versions (due to JS update) once I get the feedback. :)

Eldenroot:
@Sama34 - OK, done!

Sama34 (approved):
I would like these settings to be defined through the templates, but I'm fine with this fix.

Eldenroot:
I think it is fine... we can avoid adding more template changes for this simple fix.

PR #4301, opened 2021-03-12
Fixes #4136 - Add style variable in survey template

### LINKED ISSUE #4136: Add style variable in survey template
Hello good morning, I wanted to make a request to the MyBB developers which is to add a style variable in the survey template that indicates which of the survey options you voted for. I ask is asking for more options on customizing the survey in CSS. I think it would be a great idea for MyBB 1.9. The other way to do what I ask is to create a new template specifically where the user voted and thus add a unique class (html).

Example:

**polls_showresults_resultbit**

image

Thanks for your help and for your amazing MyBB software.

Eldenroot:
Resolves #4136

We can also make more changes - I just wanted to add easier way for customization without core changes.

### DIFF
--- a/showthread.php
+++ b/showthread.php
@@ -561,7 +561,7 @@ if($mybb->input['action'] == "thread")
 			// Mark the option the user voted for.
 			if(!empty($votedfor[$number]))
 			{
-				$optionbg = "poll_votedfor";
+				$optionbg = "trow2 poll_votedfor";
 				$votestar = "*";
 			}
 			else
--- a/showthread.php
+++ b/showthread.php
@@ -561,7 +561,7 @@ if($mybb->input['action'] == "thread")
 			// Mark the option the user voted for.
 			if(!empty($votedfor[$number]))
 			{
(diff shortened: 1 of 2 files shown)


showthread.php
```diff
@@ -561,7 +561,7 @@
 			// Mark the option the user voted for.
 			if(!empty($votedfor[$number]))
 			{
-				$optionbg = "trow2";
+				$optionbg = "poll_votedfor";
```
euantorano: Can we change this to:

```suggestion
				$optionbg = "trow2 poll_votedfor";
```

The upgrade system doesn't manage CSS stylesheet upgrades, so this would just use an unspecified CSS class until users manually updated their themes. If we just add an extra class, it should still allow the same level of flexibility.

euantorano (changes requested):

Eldenroot:
@euantorano - thx!

Sama34 (approved):
I really dislike hard coded CSS classes names.

PR #4313, opened 2021-03-13
Fix #4015 - User CP input validation

### LINKED ISSUE #4015: User CP input validation - move to datahandler
The user CP `do_profile` action has some validation done in` usercp.php` that then uses `error()`. This validation should be moved to the user data handler:

- [ ] [Away return date](https://github.com/mybb/mybb/blob/feature/usercp.php#L184)
- [ ] [ICQ number](https://github.com/mybb/mybb/blob/feature/usercp.php#L249)
- [ ] [Skype/Google hangouts details](https://github.com/mybb/mybb/blob/feature/usercp.php#L258)

Eldenroot:
Resolves #4015

Thank you @lairdshaw for another lesson :)


inc/datahandlers/user.php
```diff
@@ -324,14 +324,78 @@ function verify_website()
 	 */
 	function verify_icq()
 	{
+		global $mybb;
+
 		$icq = &$this->data['icq'];
 
+		if($mybb->settings['allowicqfield'] == '' || !is_member($mybb->settings['allowicqfield'], getuser($this->data['uid'])))
+		{
+			$icq = '';
```
Sama34, line 333: These three fields are still displayed in the ACP when editing a user. You either need to remove this from the data handler or hide them from all areas in the front end.

I'm unsure if disabling at the data handler level is the best approach. I'm also unsure if the settings to disable these 
fields make it clear that they will stop being editable from the ACP. @mybb/developers

lairdshaw: Well, just recall that this PR was created to address issue #4015, which is concerned with input _validation_, and I think that it _is_ appropriate as suggested in that issue and implemented in this PR that the validation in question occurs in the data handler. The issue of inhibiting the UI _display_ of disabled settings is a separate, though of course related, one, which I don't think this PR should address. I have no problem with a new issue and/or PR being created to address that distinct but related matter.

inc/datahandlers/user.php
```diff
@@ -324,14 +324,78 @@ function verify_website()
 	 */
 	function verify_icq()
 	{
+		global $mybb;
+
 		$icq = &$this->data['icq'];
 
+		if($mybb->settings['allowicqfield'] == '' || !is_member($mybb->settings['allowicqfield'], getuser($this->data['uid'])))
```
Sama34: I think the first `$mybb->settings['allowicqfield'] == ''` check is superfluous.

Also, isn't it `get_user()`?

inc/datahandlers/user.php
```diff
@@ -324,14 +324,78 @@ function verify_website()
...
+	 */
+	function verify_skype()
+	{
+		global $mybb;
+
+		$user = &$this->data;
+
+		if($mybb->settings['allowskypefield'] == '' || !is_member($mybb->settings['allowskypefield'], getuser($this->data['uid'])))
+		{
+			$this->data['skype'] = '';
```
Sama34, line 390: What about `$user['skype'] = '';`? You are being inconsistent in the same PR. Just pointing it out, it isn't something that would stop me from accepting a PR.

inc/datahandlers/user.php
Sama34: I agree that this validation should be in the data handler, as suggested. What I'm against is for the data handler to reset and ignore values based on settings that were checked outside the data handler before. Before this patch, changes from the ACP worked just fine, but with this patch changes will be ignored, thus this issue should be taken care of here,

I would be fine with the ACP input fields being hidden just as in the forums (keeping the group permission check in the data handler), but then I would suggest to update the settings to be clear that disabling for anyone also means disabling ACP capabilities to bypass such setting for updating user data.

I tagged the team here because although I'm sure there is an issue raised with these changes, I'm unsure about how we would want to deal with it, if as I suggest or something different. We could keep the settings check outside the data handler as before for example, but for me that looks somehow uglier.

inc/datahandlers/user.php
lairdshaw: That function name typo slipped me by as I helped @Eldenroot with this PR. Looks like we need to add testing to the "lessons", because that wouldn't have made it through. And, without investigating, you're probably right about the superfluous check. We just based the condition on the existing code which we moved from outside the data handler. We should have been more cautious about that code.

inc/datahandlers/user.php
lairdshaw: OK. Understood. My preference would be for the solution you suggest. Will be interested in what the team thinks.

inc/datahandlers/user.php
Sama34:
```suggestion
		if(!is_member($mybb->settings['allowicqfield'], get_user($this->data['uid'])))
```

inc/datahandlers/user.php
dvz: Yes, it looks like the _Allow * Field To Usergroups_ settings are taken care of outside the datahandler: https://github.com/mybb/mybb/blob/a4225c8456ee73341e880ce043f7ef553433a069/usercp.php#L226-L257

Eldenroot:
@Sama34 - thank you for your commits and fixes :)

inc/datahandlers/user.php
```diff
@@ -324,14 +324,78 @@ function verify_website()
...
+	 *
+	 * @return boolean True when valid, false when invalid.
+	 */
+	function verify_google()
+	{
+		global $mybb;
+
+		$user = &$this->data;
+
+		if($mybb->settings['allowgooglefield'] == '' || !is_member($mybb->settings['allowgooglefield'], getuser($this->data['uid'])))
```
Sama34:
```suggestion
		if($mybb->settings['allowgooglefield'] == '' || !is_member($mybb->settings['allowgooglefield'], get_user($this->data['uid'])))
```

inc/datahandlers/user.php
```diff
@@ -324,14 +324,78 @@ function verify_website()
...
+	 *
+	 * @return boolean True when valid, false when invalid.
+	 */
+	function verify_skype()
+	{
+		global $mybb;
+
+		$user = &$this->data;
+
+		if($mybb->settings['allowskypefield'] == '' || !is_member($mybb->settings['allowskypefield'], getuser($this->data['uid'])))
```
Sama34:
```suggestion
		if($mybb->settings['allowskypefield'] == '' || !is_member($mybb->settings['allowskypefield'], get_user($this->data['uid'])))
```

Sama34 (changes requested):

Eldenroot:
Thank you

Sama34:
I was rebasing this, but just noticed these are meant to be removed altogether. See #4987

I apologize for the delay !

PR #4320, opened 2021-03-16
Rewrite auto URL regex, prevent consuming partial HTML entities

### LINKED ISSUE #4303: Auto URL parsing may break HTML entities
The `postParser` automatic URL parsing may incorrectly consume partial HTML entities.

MyCode:
```
http://domain.example/&
```

Expected output:
```
<a href="http://domain.example/" target="_blank" rel="noopener" class="mycode_url">http://domain.example/</a>&amp;
```

Actual output:
```
<a href="http://domain.example/&amp" target="_blank" rel="noopener" class="mycode_url">http://domain.example/&amp;amp</a>;
```

dvz:
Resolves #4303

- merged the two expressions
- added `&` to excluded characters for path/query/fragment and added matching for `&amp;`, `&lt;`, `&gt;` (converted in `postParser::parse_html()`)
- removed prepending of whitespace to `$message` (presumably to match links at the beginning of the message), and added `^` to positive lookbehind
- changed numeric group references to named groups
- added comments

May conflict with #4314


inc/class_parser.php
```diff
@@ -1605,18 +1605,27 @@ function mycode_parse_video_disabled_callback($matches)
...
+		$message = preg_replace_callback(
+			"~
+				<a\\s[^>]*>.*?</a>|								# match and return existing links
+				(?<=^|[\s\(\)\[\>])								# character preceding the link
+				(?P<prefix>
+					(?:http|https|ftp|news|irc|ircs|irc6)://|	# scheme, or
+					(?:www|ftp)\.								# common subdomain
+				)
+				(?P<link>
+					[^\/\"\s\<\[\.]+\.*[\w]+					# host
```
lairdshaw: Parentheses are missing here around the pattern prior to the wildcard character, which should apply to the whole preceding match, not just the literal dot. The effect of this is that only two-part domains are currently matched.

lairdshaw:
I really like these changes. They make the code a lot more readable.

inc/class_parser.php
```diff
@@ -1605,18 +1605,27 @@ function mycode_parse_video_disabled_callback($matches)
...
 		// Don't create links within existing links (handled up-front in the callback function).
-		$message = preg_replace_callback("#<a\\s[^>]*>.*?</a>|([\s\(\)\[\>])(http|https|ftp|news|irc|ircs|irc6){1}(://)([^\/\"\s\<\[\.]+\.([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))(?![^<>]*?>)#ius", array($this, 'mycode_auto_url_callback'), $message);
-		$message = preg_replace_callback("#<a\\s[^>]*>.*?</a>|([\s\(\)\[\>])(www|ftp)(\.)(([^\/\"\s\<\[\.]+\.)*[\w]+(:[0-9]+)?(/([^\"\s<\[]|\[\])*)?([\w\/\)]))(?![^<>]*?>)#ius", array($this, 'mycode_auto_url_callback'), $message);
-		$message = my_substr($message, 1);
+		$message = preg_replace_callback(
+			"~
+				<a\\s[^>]*>.*?</a>|								# match and return existing links
+				(?<=^|[\s\(\)\[\>])								# character preceding the link
+				(?P<prefix>
+					(?:http|https|ftp|news|irc|ircs|irc6)://|	# scheme, or
```
euantorano, line 1618: I take it these comments are supposed to be lined up? if so, they aren't any longer based on the GitHub review UI. If they're not supposed to be lined up, can we collapse the whitespace a bit?

euantorano:
This looks good. As you guessed, there is a conflict though 😛 

A minor stylistic comment, but nothing major.

inc/class_parser.php
dvz: Yes, lined up - try appending `?ts=4` to GitHub's file preview URL.

Setting up a `.editorconfig` file like on the 1.9 branch should make the GitHub UI display these correctly: https://github.com/mybb/mybb/blob/6d6c8bfdd56fa8df2d1cd4ef68696805ab690850/.editorconfig#L17-L19

Sama34 (approved):
I did test, works.

PR #4325, opened 2021-03-20
Multiple email & Login method validation improvement

### LINKED ISSUE #4324: ACP Multiple email registration & Login method settings
- ACP Multiple email registration notice gets displayed even if there is no conflict.
- A query is always run based on the existing settings to check existence of multiple emails no matter if its required or not.

image

effone:
Resolves #4324


Starpaul20:
Works for me

PR #4326, opened 2021-03-20
Fixes #4319 - add hooks into admin/module/users/user.php

### LINKED ISSUE #4319: Inline user edit additional ACP hooks
Somehow the inline edit feature of the ACP users module doesn't use the datahandler.

If no data handler support is going to be added, it should then add additional hooks.

https://github.com/mybb/mybb/blob/7732e52795c7f6717e25c19bf16cb0a6b66ab288/admin/modules/user/users.php#L3008-L3020

https://github.com/mybb/mybb/blob/7732e52795c7f6717e25c19bf16cb0a6b66ab288/admin/modules/user/users.php#L2855-L2869

https://github.com/mybb/mybb/blob/7732e52795c7f6717e25c19bf16cb0a6b66ab288/admin/modules/user/users.php#L2497-L2503

Eldenroot:
Resolves #4319 

Thank you @lairdshaw for your php poetry mentoring :)

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -2496,6 +2496,8 @@ if($mybb->input['action'] == "inline_edit")
 				}
 			}
 
+		$plugins->run_hooks("admin_user_multiactivate", $to_update);
+
 			if(isset($to_update) && is_array($to_update))
 			{
 				$sql_array = implode(",", $to_update);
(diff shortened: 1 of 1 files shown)


Sama34:
Sorry I missed this on time, but these hooks run at the global scope, so passing variables is unnecessary.

PR #4333, opened 2021-03-22
`expcol` optimization

### LINKED ISSUE #4332: `expcol` issues
- [x] https://github.com/mybb/mybb/blob/feature/global.php#L1240 `display: show;`? Really?
- [x] We need to get rid of `_c` suffix. There is no usage of it. May be leftover from 1.6
- [x] The js functions are full of unwanted condition checks and redundant codes. Need to optimize.
- [x] Remove targeting table code (`<td>`) to support table-less themes. Also rectify adding unnecessary `tcat_collapse_collapsed` class to `.thead`.
- [x] Languagify hardcoded collapse alt / title texts ([-] / [+]);
- [x] Eager to add callback of collapse click to extend flexibility for theme developers.

effone:
Resolves #4332

### DIFF
--- a/global.php
+++ b/global.php
@@ -1235,10 +1235,7 @@ if(!empty($mybb->cookies['collapsed']))
 	$collapse = explode("|", $colcookie);
 	foreach($collapse as $val)
 	{
-		$ex = $val."_e";
-		$co = $val."_c";
-		$collapsed[$co] = "display: show;";
-		$collapsed[$ex] = "display: none;";
+		$collapsed[$val."_e"] = "display: none;";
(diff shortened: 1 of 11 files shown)


effone:
Left redundant classes that could have been combined to one. Responsible for adding bottom corner roundness only.
~~If `css3.css` is allowed to be modified I will add those changes as well.~~ (Won'tfix. This change may affect custom themes using the class for some other purpose, eg. thead opacity change on collapse)

Eldenroot:
Works fine, thx!

Sama34 (approved):

PR #4337, opened 2021-03-23
Add postParser output validation

### LINKED ISSUE #4305: Add HTML validation to postParser
Validate the `postParser` HTML output, log errors and, depending on policy, prevent output.
Add searchable error messages and Docs references.

Defense in depth solution to mitigate XSS vulnerabilities caused by broken output.

Related: #4293, #4303

dvz:
Calls `simplexml_load_string()` and captures LibXML errors.

The errors, depending on the validation policy, may display a warning and be logged by the error handler or, additionally, suppress output. The policy is currently set to `VALIDATION_REPORT_ONLY` to alert administrators but limit breaking functionality, and should be changed to `VALIDATION_REQUIRE` in subsequent versions.

The logged source and output is escaped to prevent serious malformation of the XML error log file.

Related new Docs: https://github.com/mybb/docs.mybb.com/pull/201/files

Resolves #4305

### DIFF
--- a/inc/class_error.php
+++ b/inc/class_error.php
@@ -147,9 +147,10 @@ class errorHandler {
 	 * @param string $message The error message
 	 * @param string $file The error file
 	 * @param integer $line The error line
+	 * @param boolean $allow_output Whether or not output is permitted
 	 * @return boolean True if parsing was a success, otherwise assume a error
 	 */
-	function error($type, $message, $file=null, $line=0)
+	function error($type, $message, $file=null, $line=0, $allow_output=true)
 	{
 		global $mybb;
 
@@ -202,38 +203,41 @@ class errorHandler {
 			$this->email_error($type, $message, $file, $line);
 		}
 
-		// SQL Error
-		if($type == MYBB_SQL)
+		if($allow_output === true)
 		{
-			$this->output_error($type, $message, $file, $line);
-		}
-		else
-		{
-			// Do we have a PHP error?
-			if(my_strpos(my_strtolower($this->error_types[$type]), 'warning') === false)
+			// SQL Error
+			if($type == MYBB_SQL)
 			{
 				$this->output_error($type, $message, $file, $line);
 			}
-			// PHP Error
 			else
 			{
-				if($mybb->settings['errortypemedium'] == "none" || $mybb->settings['errortypemedium'] == "error")
+				// Do we have a PHP error?
+				if(my_strpos(my_strtolower($this->error_types[$type]), 'warning') === false)
 				{
-					echo "<div class=\"php_warning\">MyBB Internal: One or more warnings occurred. Please contact your administrator for assistance.</div>";
+					$this->output_error($type, $message, $file, $line);
 				}
+				// PHP Error
 				else
 				{
-					global $templates;
-
-					$warning = "<strong>{$this->error_types[$type]}</strong> [$type] $message - Line: $line - File: $file PHP ".PHP_VERSION." (".PHP_OS.")<br />\n";
-					if(is_object($templates) && method_exists($templates, "get") && !defined("IN_ADMINCP"))
+					if($mybb->settings['errortypemedium'] == "none" || $mybb->settings['errortypemedium'] == "error")
 					{
-						$this->warnings .= $warning;
-						$this->warnings .= $this->generate_backtrace();
+						echo "<div class=\"php_warning\">MyBB Internal: One or more warnings occurred. Please contact your administrator for assistance.</div>";
 					}
 					else
 					{
-						echo "<div class=\"php_warning\">{$warning}".$this->generate_backtrace()."</div>";
+						global $templates;
+
+						$warning = "<strong>{$this->error_types[$type]}</strong> [$type] $message - Line: $line - File: $file PHP ".PHP_VERSION." (".PHP_OS.")<br />\n";
+						if(is_object($templates) && method_exists($templates, "get") && !defined("IN_ADMINCP"))
+						{
+							$this->warnings .= $warning;
+							$this->warnings .= $this->generate_backtrace();
+						}
+						else
+						{
+							echo "<div class=\"php_warning\">{$warning}".$this->generate_backtrace()."</div>";
+						}
 					}
 				}
 			}
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1977,6 +1977,7 @@ class postParser
 		);
 
 		libxml_use_internal_errors(true);
+		@libxml_disable_entity_loader(true);
 
 		simplexml_load_string('<root>'.$output.'</root>', 'SimpleXMLElement', 524288 /* LIBXML_PARSEHUGE */);
 
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -98,6 +98,19 @@ class postParser
 	public $clear_needed = false;
 
 	/**
+	 * Whether to validate the parser's HTML output. Validation errors will be logged/sent/displayed according to board settings.
+	 *
+	 * 0 - skip validation
+	 * 1 - validate if required PHP extensions (libxml, dom) are available
+	 * 2 - validate if required PHP extensions (libxml, dom) are available; block output on failure
+	 * 3 - always require validation; block output on failure
+	 *
+	 * @access public
+	 * @var int
+	 */
+	public $output_validation_policy = 1;
+
+	/**
 	 * Parses a message with the specified options.
 	 *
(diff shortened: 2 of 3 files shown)


inc/class_parser.php
```diff
@@ -1891,4 +1913,84 @@ function encode_url($url)
...
+	 */
+	function output_allowed($source, $output)
+	{
+		if($this->output_validation_policy === 0)
+		{
+			return true;
+		}
+		else
+		{
+			$output_valid = $this->validate_output($source, $output);
```
Sama34, line 1945: Should this be below the next `if` statement?

dvz: The function call also causes the errors to be logged; its result is not used in any check for policy `1`, but it is for higher values

awslabspl: > Should this be below the next `if` statement?

You are correct. @dvz should be corrected.

Sama34: No, I think @dvz explanation is clear.

Sama34:
@dvz do you have any practical case for testing this? (I suppose enabling HTML would make it easier for testing purposes.)

Also, how would admins change the validation policy (i.e: by a plugin)?

dvz:
Yes, validation errors may be occur after e.g. appending `<` to output, or currently in the _DVZ Mentions_ plugin (https://github.com/dvz/mybb-dvzMentions/blob/5d4fa3d6239a7414e164c57e0b1ec7a8f652acc5/inc/plugins/dvz_mentions/formatting.php#L41).

As long as any outstanding validation issues are not too common, and can be easily fixed (by MyBB, extension authors, forum administrators), it may be acceptable to require modifying the file directly to temporarily disable it (while the MyCodes, plugins, etc. are corrected). Adding a `$config` would make it easier if needed.

Potential documentation for the resulting errors: https://github.com/mybb/docs.mybb.com/pull/201/files

inc/class_parser.php
```diff
@@ -97,6 +97,19 @@ class postParser
...
+	 * Validation errors will be logged/sent/displayed according to board settings.
+	 *
+	 * 0 - skip validation
+	 * 1 - validate and log errors
+	 * 2 - validate and log errors; block output on failure
+	 *
+	 * @access public
+	 * @var int
+	 */
+	public $output_validation_policy = 1;
```
euantorano: Can we introduce some constants for these values?

inc/class_parser.php
```diff
@@ -1891,4 +1914,87 @@ function encode_url($url)
...
+		global $mybb, $error_handler;
+
+		$ignored_error_codes = array(
+			'XML_ERR_INVALID_CHAR' => 9,
+			'XML_ERR_UNDECLARED_ENTITY' => 26,
+			'XML_ERR_ATTRIBUTE_WITHOUT_VALUE' => 41,
+			'XML_ERR_TAG_NAME_MISMATCH' => 76, // the parser may output tags closed in different levels and siblings
+		);
+
+		libxml_use_internal_errors(true);
```
euantorano, line 1979: Do our system requirements already cover the requirement for `libxml`?

install/resources/mybb_theme.xml
```diff
@@ -13756,15 +13756,15 @@ if(use_xmlhttprequest == "1")
 				<td class="{$alt_bg}" align="center">{$issuedby}</td>
 	</tr>]]></template>
 		<template name="usercp_warnings_warning_post" version="1800"><![CDATA[<br /><small>{$lang->warning_for_post} <a href="{$warning['postlink']}">{$warning['post_subject']}</a></small>]]></template>
-		<template name="video_metacafe_embed" version="1807"><![CDATA[<iframe src="http://www.metacafe.com/embed/{$id}/" width="440" height="248" allowFullScreen frameborder=0></iframe>]]></template>
-		<template name="video_dailymotion_embed" version="1807"><![CDATA[<iframe frameborder="0" width="480" height="270" src="//www.dailymotion.com/embed/video/{$id}" allowfullscreen></iframe>]]></template>
+		<template name="video_metacafe_embed" version="1807"><![CDATA[<iframe src="http://www.metacafe.com/embed/{$id}/" width="440" height="248" allowfullscreen="true" frameborder=0></iframe>]]></template>
```
euantorano: Template version needs bumped.

install/resources/mybb_theme.xml
```diff
@@ -13756,15 +13756,15 @@ if(use_xmlhttprequest == "1")
...
+		<template name="video_dailymotion_embed" version="1807"><![CDATA[<iframe frameborder="0" width="480" height="270" src="//www.dailymotion.com/embed/video/{$id}" allowfullscreen="true"></iframe>]]></template>
```
euantorano: Template version needs bumped.

euantorano (changes requested):

inc/class_parser.php
dvz: It should be enabled by default, and may be satisfied by packages needed for SimpleXML

inc/class_parser.php
dvz: We can; will that be comprehensible for webmasters not familiar with PHP?

euantorano: Most webmasters not familiar with PHP are most likely not going to be looking at or modifying code without reading some level of instructions instructing them what to do I wouldn't think.

inc/class_error.php
```diff
@@ -202,38 +203,41 @@ function error($type, $message, $file=null, $line=0)
 			$this->email_error($type, $message, $file, $line);
 		}
 
-		// SQL Error
```
dvz, line 205: Wrapping inside the `if($allow_output === true)` block

dvz:
Modified error handler method to suppress output at https://github.com/mybb/mybb/blob/3f170f5fc2d117f1f97a3e632111c74065755d70/inc/class_error.php#L222

lairdshaw:
Given that this is the only remaining unmerged PR that Euan has indicated is a candidate for 1.8.27, I wanted to get in a reasonably timely review, so, I've both looked over its code as well as tested it. I think it's basically sound and useful, and should be merged. I do have a few comments/questions though:

1. When `postParser::output_validation_policy` is set to VALIDATION_REQUIRE, and if `$this->output_allowed()` returns false at the end of `postParser::parse_message()`, should there perhaps be _some_ sort of returned output - rather than, as at present, an empty string - to alert the end-user/admin as to _why_ there isn't any (non-error) output? Perhaps something simple like "<em>Output suppressed due to failed HTML validation. Please check your error log (if enabled).</em>"?

2. While the `$ignored_error_codes` are indeed ignored in determining whether or not to generate an error message, they are not (if present alongside other non-ignored error codes) actually omitted from that error message. Is this intentional? (Incidentally, I haven't carefully reviewed those ignored codes and why they are ignored; I have simply assumed that ignoring them is justified).

3. Given that errors are written to a plain-text error log file, is it perhaps less helpful in terms of the readability of that log file to run `$source` and `$output` through `htmlspecialchars_uni()` - as at present - than leaving those values raw and unencoded?

4. Should the hard-coded language string "Parser output validation failed.\n" perhaps be moved into a language file?

That's about all I've picked up for now. I do think this PR is worth merging.

dvz:
> should there perhaps be some sort of returned output - rather than, as at present, an empty string - to alert the end-user/admin as to why there isn't any (non-error) output?

It seemed like the cleanest way to suggest malfunction without interfering with user content. Prior to the error handler changes, a `MyBB Internal...` message appeared on the top of the page.

> [...] the `$ignored_error_codes` [...] are not [...] actually omitted from that error message. Is this intentional?

Yes, it's best to include all errors for clarity. E.g. problems that revolve around `XML_ERR_ATTRIBUTE_WITHOUT_VALUE` may cause further errors that are not on the ignore list.

> is it perhaps less helpful in terms of the readability of that log file to run `$source` and `$output` through `htmlspecialchars_uni()` - as at present - than leaving those values raw and unencoded?

Yes. It's escaped to:
- prevent breaking the XML file even further (it's not valid XML since MyBB doesn't create a root element, etc.), which may cause issues when processed automatically;
- prevent edge-case XSS problems, when the file is accessible publicly and processed as HTML in the browser;

we can choose to accept both of those risks to increase readability.

> Should the hard-coded language string "Parser output validation failed.\n" perhaps be moved into a language file?

IMO unnecessary - it's low-level information, and other strings (PHP warnings, tag names) are in English.

lairdshaw:
OK, to me this looks good to merge then, whether or not the escaping is removed.

Sama34:
I did test and seems to work for me.

PR #4339, opened 2021-03-25
patch-4138

dragonexpert:
Removes some PHP 8 Issues in #4138 that are global.  
inc/class_core.php mybb->parse_cookies
global.php $awaitingusers, $lockoutexpiry.

### DIFF
--- a/global.php
+++ b/global.php
@@ -91,7 +91,7 @@ $lang->load('global');
 $lang->load('messages');
 
 // Wipe lockout cookie if enough time has passed
-if(isset($mybb->cookies['lockoutexpiry'])
+if(isset($mybb->cookies['lockoutexpiry']))
 {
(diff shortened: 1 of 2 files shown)


dvz:
The `global.php` corrections may have been applied in this PR: https://github.com/mybb/mybb/pull/4095/files#diff-7b7d4d5231d83ea738098e31c83ba44004c829ccf7c74ac947ab02b054fe39ce

global.php
```diff
@@ -91,9 +91,12 @@
 $lang->load('messages');
 
 // Wipe lockout cookie if enough time has passed
-if(isset($mybb->cookies['lockoutexpiry']) && $mybb->cookies['lockoutexpiry'] < TIME_NOW)
```
effone, line 94: What exactly was wrong with this condition?

PR #4342, opened 2021-03-27
Fixes #4338 / #3754 - Relative/absolute path

### LINKED ISSUE #3754: Upload paths inaccuracy
Upload paths should support both forms, absolute and "relative" paths (if ./ is specified, you point the start of the path to MYBB_ROOT), as admin/modules/tools/system_health.php already does.

For example, on admin/modules/forum/attachments.php#L235, absolute paths would not be supported.

Or, the whole inc/functions_upload.php considers the admin directory one level up from MYBB_ROOT (what if I specify the admin directory two level ups?).

Not specifying MYBB_ROOT on upload paths (as function_upload does) could bring to unexpected behaviours (error due to the execution of tasks using cron, where $_SERVER['DOCUMENT_ROOT'] is not set):

&lt;error&gt;
    &lt;dateline&gt;1560304831&lt;/dateline&gt;
    &lt;script&gt;inc/functions_upload.php&lt;/script&gt;
    &lt;line&gt;160&lt;/line&gt;
    &lt;type&gt;2&lt;/type&gt;
    &lt;friendly_type&gt;Warning&lt;/friendly_type&gt;
    &lt;message&gt;opendir(./uploads/avatars): failed to open dir: No such file or directory&lt;/message&gt;
    &lt;back_trace&gt;#0  errorHandler-&gt;log_error(2, opendir(./uploads/avatars): failed to open dir: No such file or directory, inc/functions_upload.php, 160) called at [/***/inc/class_error.php:189]
#1  errorHandler-&gt;error(2, opendir(./uploads/avatars): failed to open dir: No such file or directory, inc/functions_upload.php
#2  opendir(./uploads/avatars) called at [/***/inc/functions_upload.php:160]
#3  remove_avatars(121577895) called at [/***/inc/datahandlers/user.php:1668]
#4  UserDataHandler-&gt;delete_content() called at [/***/inc/datahandlers/user.php:1536]
#5  UserDataHandler-&gt;delete_user(Array ([121577414] =&gt;

Original thread: [Upload paths inaccuracy](https://community.mybb.com/thread-223408.html)

### LINKED ISSUE #4338: Absolute uploadspath fails due to MYBB_ROOT prefix on mime check
To allow us to store the MyBB user uploaded attachments external to the webroot of the site we have set our `uploadspath` setting to an absolute directory e.g. `/foo/bar`

This all seems to work apart from the `finfo_open()`/`mime_content_type()` checks in [inc/functions_upload.php](https://github.com/mybb/mybb/blob/da7b32456e7b21c96f72c1093816c55cce249264/inc/functions_upload.php#L600) which prefixes the `uploadspath` path with `MYBB_ROOT`:

```php
$img_dimensions = @getimagesize($mybb->settings['uploadspath']."/".$filename);

$mime = "";
$file_path = $mybb->settings['uploadspath']."/".$filename;
if(function_exists("finfo_open"))
{
	$file_info = finfo_open(FILEINFO_MIME);
	list($mime, ) = explode(';', finfo_file($file_info, MYBB_ROOT.$file_path), 1);
	finfo_close($file_info);
}
else if(function_exists("mime_content_type"))
{
	$mime = mime_content_type($file_path);
}
```

Note how the `getimagesize()` **does not** prefix with `MYBB_ROOT` but then a few lines later in the call to `finfo_file()` it **does** prefix with it (and hence breaks for absolute directory paths).

Eldenroot:
Resolves #4338
Resolves #3754

Thank you @lairdshaw - another valuable lesson!

### DIFF
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -602,12 +602,12 @@ function upload_attachment($attachment, $update_attachment=false)
 		if(function_exists("finfo_open"))
 		{
 			$file_info = finfo_open(FILEINFO_MIME);
-			list($mime, ) = explode(';', finfo_file($file_info, MYBB_ROOT.$file_path), 1);
+			list($mime, ) = explode(';', finfo_file($file_info, $file_path), 1);
 			finfo_close($file_info);
 		}
 		else if(function_exists("mime_content_type"))
 		{
-			$mime = mime_content_type(MYBB_ROOT.$file_path);
+			$mime = mime_content_type($file_path);
 		}
 
 		if(!is_array($img_dimensions) || ($img_dimensions[2] != $img_type && !in_array($mime, $supported_mimes)))
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -61,14 +61,7 @@ function remove_attachment($pid, $posthash, $aid)
 
 	$db->delete_query("attachments", "aid='{$attachment['aid']}'");
 
-	if(defined('IN_ADMINCP'))
-	{
-	    $uploadpath = '../'.$mybb->settings['uploadspath'];
-	}
-	else
-	{
-	    $uploadpath = $mybb->settings['uploadspath'];
-	}
+	$uploadpath = $mybb->settings['uploadspath'];
 
 	// Check if this attachment is referenced in any other posts. If it isn't, then we are safe to delete the actual file.
 	$query = $db->simple_select("attachments", "COUNT(aid) as numreferences", "attachname='".$db->escape_string($attachment['attachname'])."'");
@@ -119,14 +112,7 @@ function remove_attachments($pid, $posthash="")
 		$query = $db->simple_select("attachments", "*", "pid='$pid'");
 	}
 
-	if(defined('IN_ADMINCP'))
-	{
-	    $uploadpath = '../'.$mybb->settings['uploadspath'];
-	}
-	else
-	{
-	    $uploadpath = $mybb->settings['uploadspath'];
-	}
+	$uploadpath = $mybb->settings['uploadspath'];
 
 	$num_attachments = 0;
 	while($attachment = $db->fetch_array($query))
--- a/inc/init.php
+++ b/inc/init.php
@@ -201,6 +201,12 @@ $settings['bbname_orig'] = $settings['bbname'];
 $settings['bbname'] = strip_tags($settings['bbname']);
 $settings['orig_bblanguage'] = $settings['bblanguage'];
 
+// Adjust a relative upload path to an absolute one
+if(my_substr($settings['uploadspath'], 0, 1) != "/")
+{
+	$settings['uploadspath'] = MYBB_ROOT.$settings['uploadspath'];
+}
+
 // Fix for people who for some specify a trailing slash on the board URL
 if(substr($settings['bburl'], -1) == "/")
 {
--- a/inc/init.php
+++ b/inc/init.php
@@ -202,7 +202,9 @@ $settings['bbname'] = strip_tags($settings['bbname']);
 $settings['orig_bblanguage'] = $settings['bblanguage'];
 
 // Adjust a relative upload path to an absolute one
-if(my_substr($settings['uploadspath'], 0, 1) != "/")
+$iswin = strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
+$char1 = my_substr($settings['uploadspath'], 0, 1);
+if($char1 != '/' && !($iswin && ($char1 == '\\' || preg_match('(^[a-zA-Z]:\\\\)', $settings['uploadspath']))))
 {
 	$settings['uploadspath'] = MYBB_ROOT.$settings['uploadspath'];
 }
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -409,7 +409,7 @@ max=9]]></optionscode>
 		</setting>
 		<setting name="uploadspath">
 			<title>Uploads Path</title>
-			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers). Please note that unless the set path starting with a dot ".", the path will be converted to absolute. Default value: "./uploads"]]></description>
+			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers). If the path starts with a dot "." it will be converted to an absolute path. Default value: "./uploads" (relative path).]]></description>
 			<disporder>10</disporder>
 			<optionscode><![CDATA[text]]></optionscode>
 			<settingvalue><![CDATA[./uploads]]></settingvalue>
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -409,7 +409,7 @@ max=9]]></optionscode>
 		</setting>
 		<setting name="uploadspath">
 			<title>Uploads Path</title>
-			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers).]]></description>
+			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers). Please note that unless the set path starting with a dot ".", the path will be converted to absolute. Default value: "./uploads"]]></description>
 			<disporder>10</disporder>
 			<optionscode><![CDATA[text]]></optionscode>
 			<settingvalue><![CDATA[./uploads]]></settingvalue>
@@ -417,7 +417,7 @@ max=9]]></optionscode>
 		</setting>
 		<setting name="useerrorhandling">
 			<title>Use Error Handling</title>
-			<description><![CDATA[If you do not wish to use the integrated error handling for MyBB, you may turn this option off. However, it is recommended that it stay on]]></description>
+			<description><![CDATA[If you do not wish to use the integrated error handling for MyBB, you may turn this option off. However, it is recommended that it stay on.]]></description>
 			<disporder>11</disporder>
 			<optionscode><![CDATA[onoff]]></optionscode>
 			<settingvalue><![CDATA[1]]></settingvalue>
--- a/install/resources/settings.xml
+++ b/install/resources/settings.xml
@@ -409,7 +409,7 @@ max=9]]></optionscode>
 		</setting>
 		<setting name="uploadspath">
 			<title>Uploads Path</title>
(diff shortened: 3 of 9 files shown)


Sama34:
What about updating the setting description to make it clear than unless starting with a dot `.` the path will be converted to absolute.

Eldenroot:
@Sama34 - I changed the description - I am not a native English speaker so maybe it would need some corrections/enhancements. Let me know :)

install/resources/settings.xml
```diff
@@ -409,15 +409,15 @@ max=9]]></optionscode>
 		</setting>
 		<setting name="uploadspath">
 			<title>Uploads Path</title>
-			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers).]]></description>
+			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers). Please note that unless the set path starting with a dot ".", the path will be converted to absolute. Default value: "./uploads"]]></description>
```
Sama34:
```suggestion
			<description><![CDATA[The path used for all board uploads. It <b>must be chmod 777</b> (on *nix servers). If the path starts with a dot "." it will be converted to an absolute path. Default value: "./uploads" (relative path).]]></description>
```

Sama34 (changes requested):

Eldenroot:
Thx @Sama34 !

install/resources/settings.xml
lairdshaw: Just a minor nitpick: technically, the "relative path" condition is not "starts with a dot" but "does not start with a forward slash".

Eldenroot:
Thank you guys for your valuable feedback, corrected in the last commit.

Sama34 (approved):

inc/functions_upload.php
```diff
@@ -119,14 +112,7 @@ function remove_attachments($pid, $posthash="")
...
 
-	if(defined('IN_ADMINCP'))
-	{
-	    $uploadpath = '../'.$mybb->settings['uploadspath'];
-	}
-	else
-	{
-	    $uploadpath = $mybb->settings['uploadspath'];
-	}
+	$uploadpath = $mybb->settings['uploadspath'];
```
Sama34, line 115: So what will happen if the user uses a relative path now? (I'm talking about the ACP here.) I will test this PR.

install/resources/upgrade35.php
```diff
@@ -164,12 +164,12 @@ function upgrade35_dbchanges4()
 	echo "<p>Adding index files to attachment directories...</p>";
 	flush();
 
-	$dir = @opendir('../'.$mybb->settings['uploadspath']);
+	$dir = @opendir($mybb->settings['uploadspath']);
```
Sama34, line 167: If the user is using a relative path what will happen here as well? Since the upgrade is ran from a sub folder.

inc/functions_upload.php
lairdshaw: Any relative path is converted to an absolute path just after settings are loaded in `inc/init.php`:

https://github.com/mybb/mybb/pull/4342/files#diff-09ba99a151ee25a01bd649b01fe94fab3b52d58c19f228a73a11e7b9050978a2R204-R209

install/resources/upgrade35.php
lairdshaw: Similarly as above, any relative path is converted to an absolute path just after settings are loaded, this time in `install/upgrade.php`:

https://github.com/mybb/mybb/pull/4342/files#diff-43f8477f93fbadcef16f90e9fbffd75dd83dc143d06fe9c7304ffb35e8c03794R121-R125

So, in all cases, we can rely on the path being absolute (or, at least, that's the intent - the changes haven't been carefully tested).

euantorano:
This seems good to me. I'm not sure that we have many people running MyBB on Windows, but it might be worth testing there as I've only tested on Linux and Mac.

inc/init.php
```diff
@@ -201,6 +201,12 @@
 $settings['bbname'] = strip_tags($settings['bbname']);
 $settings['orig_bblanguage'] = $settings['bblanguage'];
 
+// Adjust a relative upload path to an absolute one
+if(my_substr($settings['uploadspath'], 0, 1) != "/")
+{
+	$settings['uploadspath'] = MYBB_ROOT.$settings['uploadspath'];
+}
+
```
lairdshaw: @euantorano was right to question Windows compatibility, and @Eldenroot had mentioned Windows while we worked together on this PR, and it does need to be handled better. PHP on Windows allows for absolute paths that begin not with a forward slash but with either a backslash or a drive letter followed by colon followed by backslash. I suggest then that we replace the above code with this (tested on Windows):

```php
// Adjust a relative upload path to an absolute one
$iswin = strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
$char1 = my_substr($settings['uploadspath'], 0, 1);
if($char1 != '/' && !($iswin && ($char1 == '\\' || preg_match('(^[a-zA-Z]:\\\\)', $settings['uploadspath']))))
{
	$settings['uploadspath'] = MYBB_ROOT.$settings['uploadspath'];
}
```

install/upgrade.php
```diff
@@ -118,6 +118,12 @@
 $settings['bbname_orig'] = $settings['bbname'];
 $settings['bbname'] = strip_tags($settings['bbname']);
 
+// Adjust a relative upload path to an absolute one
+if(my_substr($settings['uploadspath'], 0, 1) != "/")
+{
+	$settings['uploadspath'] = MYBB_ROOT.$settings['uploadspath'];
+}
+
```
lairdshaw, line 126: As above, I suggest we replace this code with the code supplied above for better Windows compatibility.

effone: Why don't you just dump the controversy and go for `DIRECTORY_SEPARATOR`

lairdshaw: I don't understand. Can you explain in more detail?

effone: Its a constant of PHP which picks the right separator based on OS.

lairdshaw: Yep, yep, I know what it is, I just don't see how it helps here.

On Windows, an admin could use either a forward slash or a backslash when setting `uploadspath`. `DIRECTORY_SEPARATOR` isn't going to help us determine which was chosen by the admin, nor whether or not the admin chose to start an absolute path with a drive letter.

euantorano: Using `DIRECTORY_SEPARATOR` might not work perfectly either - path handling is hard 😢 

WordPress has a decent looking function that does what we're trying to do which might be worth having a glance at: https://developer.wordpress.org/reference/functions/path_is_absolute/

effone: We can run a preg_replace with `DIRECTORY_SEPARATOR` to whatever admin has set.
Can't we? Or I am missing something...

`$new_path = preg_replace('/[\\/\\\\]/', DIRECTORY_SEPARATOR, $path, -1);`

lairdshaw: > WordPress has a decent looking function that does what we're trying to do

Though it uses the same regex that I came up with independently, it looks flawed to me, in that the regex should only be tested on a Windows platform, since on a UNIX platform, C:\ is a valid name for a directory, and is not an absolute path.

Likewise for testing whether the first character is a backslash - that test should only be performed on a Windows platform.

Would encourage you guys to look at the code I supplied and point out any errors that you see. It seems serviceable to me.

And I still don't see how `DIRECTORY_SEPARATOR` helps in any meaningful way. So, yes, I think you're missing something.

effone: :(
Attempt to remove the absolute part from the path (if its there it will be removed, if its not there it will be as is) and force prepend. 
I guess I have lost the track. If its about OS and separator its definitely not about absolute / relative then.
Side note: IIRC we do something with avatar upload, worth a look.

I will jump out of it now, with my last thought:
`if(!in_array(my_substr($settings['uploadspath'], 0, 1)), ["/","\"])`

euantorano: > Would encourage you guys to look at the code I supplied and point out any errors that you see. It seems serviceable to me.

Will do, checking it in my email client wasn't ideal 😉

lairdshaw: > `if(!in_array(my_substr($settings['uploadspath'], 0, 1)), ["/","\"])`

The problems I see with that code are:

1. It doesn't recognise C:\ or D:\ or E:\ etc as the start of an absolute path on a Windows platform (which they are).
2. It mistakenly treats an opening backslash as indicating an absolute path on a UNIX platform.

lairdshaw: Thanks, @euantorano. The only improvement I can see is to recognise an opening file:// as indicating an absolute path - but are admins really going to enter an uploads path like that, and should we really support it?

effone: Better detect the OS then and do comparison.
again `if(DIRECTORY_SEPARATOR === "\")` its a Win

inc/init.php
euantorano: Note: `PHP_OS` reports the OS that PHP was __built for__, not the OS that PHP is __running on__. These are usually the same thing and is probably acceptable. other options include `php_uname('s')` or checking for one of the Windows-specific constants.

Otherwise, this looks good to me (though the regex is missing the delimiters). Only very minor point is the regex doesn't really need a capture group and could just be `preg_match('#^[a-zA-Z]:\\\\#')`

install/upgrade.php
lairdshaw: > Better detect the OS then and do comparison.

...which is exactly what the code I supplied *does*. I'm guessing you missed it? Here it is again:

https://github.com/mybb/mybb/pull/4342#discussion_r607987647

inc/init.php
lairdshaw: Yep, I think `PHP_OS` is acceptable for detecting Windows. If you've built PHP for Windows, you're not going to be running it on Linux/MacOS.

> though the regex is missing the delimiters

Globally enclosing parentheses are valid delimiters.

> the regex doesn't really need a capture group

It's not a capture group. Again, those parens are the delimiters. I prefer to use parentheses as delimiters because they don't impose any limits on the characters that can be used within the regex.

euantorano: > Yep, I think `PHP_OS` is acceptable for detecting Windows. If you've built PHP for Windows, you're not going to be running it on Linux/MacOS.
> 
> > though the regex is missing the delimiters
> 
> Globally enclosing parentheses are valid delimiters.
> 
> > the regex doesn't really need a capture group
> 
> It's not a capture group. Again, those parens are the delimiters. I prefer to use parentheses as delimiters because they don't impose any limits on the characters that can be used within the regex.

You learn something new every day! 🤣

Sama34 (approved):
With the changes I did it worked, tested on Windows.

Edit: Thanks for all of your contribution commenters ofc ;)

PR #2380, opened 2016-03-23
Fix #2375 custom mod tools permission checks missing

Destroy666x:
As the title says.

https://github.com/mybb/mybb/issues/2375

PR #4343, opened 2021-03-27
Fix #2375 custom mod tools permission checks missing

### LINKED ISSUE #2375: Custom mod tools group permission checks missing
A continuation of https://github.com/mybb/mybb/issues/1571
They're still missing from:
- moderation.php https://github.com/mybb/mybb/blob/fda43f3dc95421a50bf2f6eb0de0a92a10f1490b/moderation.php#L203
- search.php https://github.com/mybb/mybb/blob/9d82f2a12e0461e488a6c08be693d0ee0b558016/search.php#L684

Also, `LIKE '%,-1,%'` seems to be redundant, comparison should be enough because -1 is the only value.

Sama34:
Resolves #2375

#2380

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -97,17 +97,17 @@ if($mybb->user['uid'] == 0)
 		$forumsread = my_unserialize($mybb->cookies['mybb']['forumread']);
 	}
 
- 	if(is_array($forumsread) && empty($forumsread))
- 	{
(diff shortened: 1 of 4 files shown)


Eldenroot:
I applied to my board, but quite difficult to test correctly. Is there somebody else who can confirm it is working?

PR #4344, opened 2021-03-28
Fixes #3565 - lastpost_subject _ CSS fix

### LINKED ISSUE #3565: Enhancement: $lastpost_subject > 25

Eldenroot:
Resolves #3565

CSS solution... PHP solution was rejected #4288

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -448,6 +448,14 @@ table {
 	color: #999;
 }
 
+.lastpost_subject {
+	display: inline-block;
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	width: 200px;
+}
+
 .post.unapproved_post .post_author {
 	border-bottom-color: #ffb8be;
 }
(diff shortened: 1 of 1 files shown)


install/resources/mybb_theme.xml
```diff
@@ -448,6 +448,14 @@ table {
 	color: #999;
 }
 
+.lastpost_subject {
+	display: inline-block;
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	width: 200px;
```
Sama34, line 456: Can't the width be `100%` or removed? I know the default theme isn't mobile friendly but we won't want to make it worse.

Eldenroot: I dont think so... now it uses same width as td in table.

effone, line 456: You should not fix the size here. Table column has its own width specified.
Better use `display: block;` or a `div`.

Eldenroot: I have tried both and no success...

Sama34: Perhaps @justinsoltesz could give us some insight?

Eldenroot: @justinsoltesz @effone - any feedback how to solve this?

Eldenroot:
Bump

euantorano:
Any feedback @mybb/developers @mybb/sqa? We know that CSS changes don't get applied automatically on upgrade, but are we happy with the content of the changes? I'm okay with them personally.

Sama34 (approved):
Seems like it is not as easy to not use a set width, so I'm fine with it.

PR #4351, opened 2021-04-01
ShowThread Mode Toggle

### LINKED ISSUE #3791: Thread Mode Selector
Thread mode selector creates a dropdown to select mode. But...
- Why a selector is needed when there are only 2 options? Can be direct link based on set mode.
- The selector is driven by js and there is no fallback. The tool simply fails if the js is disabled.
- The selector trigger is bound to xmlhttp setting whereas it doesn't do any AJAX call.
- Unnecessary js functionality and js inclusion in template.

effone:
Resolves #3791

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -11508,7 +11508,7 @@ if(use_xmlhttprequest == "1")
 		<tr>
 			<td class="thead">
 				<div class="float_right">
-					<span class="smalltext"><strong><a href="javascript:void(0)" id="thread_modes">{$lang->thread_modes}</a>{$threadnoteslink}</strong></span>
+					<span class="smalltext"><strong><a href="showthread.php?mode={$thread_toggle}&amp;tid={$tid}&amp;pid={$pid}#pid{$pid}">{$lang->thread_toggle}</a>{$threadnoteslink}</strong></span>
 				</div>
 				<div>
 					<strong>{$thread['displayprefix']}{$thread['subject']}</strong>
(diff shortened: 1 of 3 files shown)


Ben-MyBB:
I still need to test this, but this setting also appears in the ACP when editing a user, I wonder if it should be changed there as well?

effone:
This PR addresses only the final output view.
All the decisions related to setting and user choice are made way prior to that.

Eldenroot:
Tested, works fine.

install/resources/mybb_theme.xml
```diff
@@ -11508,7 +11508,7 @@ if(use_xmlhttprequest == "1")
 		<tr>
 			<td class="thead">
 				<div class="float_right">
-					<span class="smalltext"><strong><a href="javascript:void(0)" id="thread_modes">{$lang->thread_modes}</a>{$threadnoteslink}</strong></span>
+					<span class="smalltext"><strong><a href="showthread.php?mode={$thread_toggle}&amp;tid={$tid}&amp;pid={$pid}#pid{$pid}">{$lang->thread_toggle}</a>{$threadnoteslink}</strong></span>
```
Sama34, line 11511: This should use the `get_thread_link()` function to build URLs, remember SEO is a core feature.

effone: `get_thread_link()` doesn't support custom fragments, like here we use `mode`

Sama34: Yes, we usually build the url in code, it has been suggested to fix this by introducing a global function to handle link building, which I agree with.

But to be frank, I'm unsure if that would be ideal for 1.8 at this point, and if not then this PR might as well just be left as is.

@euantorano @dvz

PR #4361, opened 2021-04-20
ACP CHMOD Check Fixes

### LINKED ISSUE #4360: ACP System Health CHMOD Check Issues
1. **Regression**: After merging PR https://github.com/mybb/mybb/pull/4342 some paths are displayed relative some are absolute.
image
The display effect should be neutralized to produce the results same as before the PR and after.

2. **Bug**: `$errors` was never reset before entering to CHMOD check. If `$errors` carries any value from previous code it may display wrong notice.

3. **Optimization**: Bloated repeated code chunks (around 150 lines) can be dried up and reduced down to just one third (around 50 lines).

4. The `CHMOD Status` heading is missing. The first heading is `File` whereas its listing directories also.

effone:
Resolves #4360 
Addresses #4342 Regression

### DIFF
--- a/admin/modules/tools/system_health.php
+++ b/admin/modules/tools/system_health.php
@@ -843,7 +843,7 @@ if(!$mybb->input['action'])
 
 	$errors = 0; // Reset errors
 	$table = new Table;
-	$table->construct_header($lang->file);
+	$table->construct_header($lang->checked_path);
 	$table->construct_header($lang->location);
 	$table->construct_header($lang->status_chmod, array('width' => 250));
 
--- a/admin/modules/tools/system_health.php
+++ b/admin/modules/tools/system_health.php
@@ -844,7 +844,8 @@ if(!$mybb->input['action'])
 	$errors = 0; // Reset errors
 	$table = new Table;
 	$table->construct_header($lang->file);
-	$table->construct_header($lang->location, array("colspan" => 2, 'width' => 250));
+	$table->construct_header($lang->location);
+	$table->construct_header($lang->status_chmod, array('width' => 250));
 
 	foreach ($writables as $langvar => $path)
 	{
--- a/admin/modules/tools/system_health.php
+++ b/admin/modules/tools/system_health.php
@@ -830,93 +830,41 @@ if(!$mybb->input['action'])
 		$table->output($lang->existing_db_backups);
 	}
 
-	if(is_writable(MYBB_ROOT.'inc/settings.php'))
+	$writable_directories = array(
+		'config_file' => 'inc/config.php',
+		'settings_file' => 'inc/settings.php',
+		'file_upload_dir' => ltrim($mybb->settings['uploadspath'], MYBB_ROOT . '.'),
+		'avatar_upload_dir' => ltrim($mybb->settings['avataruploadpath'], MYBB_ROOT . '.'),
+		'language_files' => 'inc/languages',
+		'backup_dir' => $config['admin_dir'] . '/backups',
+		'cache_dir' => 'cache',
+		'themes_dir' => 'cache/themes'
+	);
+
+	$errors = 0; // Reset errors
+	$table = new Table;
+	$table->construct_header($lang->file);
+	$table->construct_header($lang->location, array("colspan" => 2, 'width' => 250));
+
+	foreach ($writable_directories as $langvar => $path)
 	{
-		$message_settings = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_settings = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
+		$table->construct_cell("<strong>{$lang->{$langvar}}</strong>");
+		$table->construct_cell('./' . trim($path, '/'));
+
+		if (is_writable(MYBB_ROOT . $path))
+		{
+			$table->construct_cell("<span style=\"color: green;\">{$lang->writable}</span>");
+		}
+		else
+		{
+			$table->construct_cell("<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}");
+			++$errors;
+		}
+
+		$table->construct_row();
 	}
 
-	if(is_writable(MYBB_ROOT.'inc/config.php'))
-	{
-		$message_config = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_config = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	$uploadspath = $mybb->settings['uploadspath'];
-	if(is_writable($uploadspath))
-	{
-		$message_upload = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_upload = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	$avataruploadpath = $mybb->settings['avataruploadpath'];
-	if(my_substr($avataruploadpath, 0, 1) == '.')
-	{
-		$avataruploadpath = MYBB_ROOT . $mybb->settings['avataruploadpath'];
-	}
-	if(is_writable($avataruploadpath))
-	{
-		$message_avatar = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_avatar = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	if(is_writable(MYBB_ROOT.'inc/languages/'))
-	{
-		$message_language = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_language = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	if(is_writable(MYBB_ROOT.$config['admin_dir'].'/backups/'))
-	{
-		$message_backup = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_backup = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	if(is_writable(MYBB_ROOT.'/cache/'))
-	{
-		$message_cache = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_cache = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	if(is_writable(MYBB_ROOT.'/cache/themes/'))
-	{
-		$message_themes = "<span style=\"color: green;\">{$lang->writable}</span>";
-	}
-	else
-	{
-		$message_themes = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
-
-	if($errors)
+	if ($errors)
 	{
 		$page->output_error("<p><em>{$errors} {$lang->error_chmod}</span></strong> {$lang->chmod_info} <a href=\"https://docs.mybb.com/1.8/administration/security/file-permissions\" target=\"_blank\" rel=\"noopener\">MyBB Docs</a>.</em></p>");
 	}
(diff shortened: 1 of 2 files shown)


admin/modules/tools/system_health.php
```diff
@@ -830,93 +830,42 @@
...
-	}
-	else
-	{
-		$message_avatar = "<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}";
-		++$errors;
-	}
+	$writables = array(
+		'config_file' => 'inc/config.php',
+		'settings_file' => 'inc/settings.php',
+		'file_upload_dir' => ltrim($mybb->settings['uploadspath'], MYBB_ROOT . '.'),
```
lairdshaw, line 836: This assumes that the current value of the setting (as entered by the admin) is a relative path (which has been converted in `inc/init.php` to an absolute one). If the current setting value, prior to conversion, is an absolute path though, then stripping MYBB_ROOT from the beginning is the wrong action to take here, and will display an incorrect value.

effone: No, the trim affects to any kind provided, not only absolute paths. Notice, we are not only stripping the `MyBB_ROOT` part from an absolute path there, we are stripping the initial dot (`.`) from a relative path as well. Thats how its neutralizing regardless the type of the path provided.

The array ensures to have all ~~relative paths~~ (or better: may say splitted suffix to `MYBB_ROOT` of a path , not the actual path in effect), because we are prepending `MYBB_ROOT` to all values altogether while checking writability just after:

`if (is_writable(MYBB_ROOT . $path))`

So the array values are not path. They are partial strings of a path. I'll even not say relative paths, they are just text strings to make the path. I don't see any wrong action there.

lairdshaw: What I mean is that let's say in `Server and Optimization Options` I set `Uploads Path` to the absolute path `/path/to/mybb/uploads`. When I open the `System Health` page, shouldn't I expect to see the value of `File Uploads Directory` set to the exact same value that I'd entered in those settings, namely `/path/to/mybb/uploads`? But with this PR, I don't. Instead, I see the relative path `./uploads`. Isn't the fact that admins weren't always seeing here the same value that they entered in settings the whole reason you created this PR, in which case, shouldn't this PR ensure that I see `/path/to/mybb/uploads` here rather than `./uploads` in the scenario I've given?

effone: Again, those are not paths in the array. They are parts.
This action doesn't modify any path in system. It fetches the required part from a path. No matter what admin sets and what PR #4342 manipulates. It affects to all types of value provided.

If you are confused about the effect of the `ltrim` there, please see:

```php
echo ltrim('lairdshaw', 'airdl'); // shaw

echo ltrim('lairdshaw', 'drealbin'); // shaw
echo ltrim('bernardshaw', 'drealbin'); // shaw
```

effone: Sorry, may be I am not getting you right. Do you mean to say we should **display** all the absolute paths there?
The reason of my confusion is where exactly you have pointed. Thats not the visual part of the code, thats the working part. If you bother only about what admins see and what they should, you should have pointed here:

`$table->construct_cell('./' . trim($path, '/'));`

lairdshaw: I see I'd misunderstood how the second parameter of `ltrim()` worked, but now that I understand, my concerns are only magnified. e.g., if `$mybb->settings['uploadspath']` is set by the admin in the settings to `/path/to/some-other-dir/uploads`, then, if `MYBB_ROOT` is set to `/path/to/mybb`, then after the `ltrim` in this PR, the `uploadspath` setting will show up as the relative path ("part") as you want to call it, `./some-other-dir/uploads`, which _looks_ like it is relative to the MyBB root but isn't.

In any case, you _do_ seem not to be getting me right. Please just answer this simple question:

If in `Server and Optimization Options` the admin sets `Uploads Path` to an absolute path, should the admin then see an absolute path for this setting on the `System Health` page? Why or why not?

effone: Now, as I partially understand what you say, still there is a regression.
You want if admins sets an absolute path in `Server and Optimization Options` he should see an absolute path there. Right?
But what if he sets a relative path instead? After your PR, it still displays an absolute path for `Uploads Path` and thats the regression.

Now if I change that particular line to `'file_upload_dir' => ltrim($mybb->settings['uploadspath'], '.'),` then I believe both of our views will meet. Of course I need to manipulate the checking part also.

Is that OK with you?

lairdshaw: > Now, as I partially understand what you say, still there is a regression.
> You want if admins sets an absolute path in `Server and Optimization Options` he should see an absolute path there. Right?

Yes. Especially if that absolute path is outside of the MyBB root. e.g., MyBB is installed at `/path/to/mybb`, and the uploads path is set to `/some/wild/and/crazy/path/to/uploads`.

> But what if he sets a relative path instead? After your PR, it still displays an absolute path for `Uploads Path` and thats the regression.

Right. I understand that that's what your PR is intended to fix. And I'm happy to work with you to see how we can do that.

> Now if I change that particular line to `'file_upload_dir' => ltrim($mybb->settings['uploadspath'], '.'),` then I believe both of our views will meet.

I don't think so, because the later line (currently number 853):
```php
		$table->construct_cell('./' . trim($path, '/'));
```
still mistakenly ensures that an absolute path has its initial `/` converted to the apparently (but incorrectly) relative path `./`.

effone: > still mistakenly ensures that an absolute path has its initial / converted to the apparently (but incorrectly) relative path ./

As I said, **Of course I need to manipulate the checking part also.**

(edit: <snip>, wrong check)

Now that I understand that the upload path may or may not be depended to and part of `MYBB_ROOT` the same applies to avatar upload path as well. That may or may not be depended to `upload parh` or even `MYBB_ROOT`. I need to check whether it has made depended in core in any way.

lairdshaw: To try to avoid further extensive back-and-forth, I suggest two possible resolutions to your concerns:

1. We don't absolutise the uploads path early on in processing, but instead wrap it in a "path-absolutising" function everywhere it is used. That way, the System Health page by default has access to the original value.
2. We leave the early "absolutising" conversion of the uploads path as it is, but on the System Health page, we pull the original value out of the database for display and writeability checking.

lairdshaw: As for the avatar uploads path, as far as I understand, no changes have been made to it equivalent to those to the upload path in the recent PR #4342, but perhaps there should be.

effone: Alright, so I have rewritten the part, which I believe now OK and reflect correctly any presence of absolute path outside `MYBB_ROOT`. But yes, even if the admin sets an absolute path which is under `MYBB_ROOT` it will show up as relative, because it actually is.

Please let me know if the following is acceptable to you and I will push a commit:

```php

$writables = array(
	'config_file' => 'inc/config.php',
	'settings_file' => 'inc/settings.php',
	'file_upload_dir' => str_replace('/./', '/', $mybb->settings['uploadspath']),
	'avatar_upload_dir' => str_replace('/./', '/', $mybb->settings['avataruploadpath']),
	'language_files' => 'inc/languages',
	'backup_dir' => $config['admin_dir'] . '/backups',
	'cache_dir' => 'cache',
	'themes_dir' => 'cache/themes'
);

$errors = 0; // Reset errors
$table = new Table;
$table->construct_header($lang->checked_path);
$table->construct_header($lang->location);
$table->construct_header($lang->status_chmod, array('width' => 250));
foreach ($writables as $langvar => $path)
{
	$table->construct_cell("<strong>{$lang->{$langvar}}</strong>");
	$path = stripos($path, MYBB_ROOT) !== false ? $path : MYBB_ROOT . ltrim($path, './'); // TRIM IS REQUIRED BECAUSE AVATAR UPLOAD PATH IS STILL NOT ADDRESSED
	$table->construct_cell(str_replace(MYBB_ROOT, './', $path));
	if (is_writable($path))
	{
		$table->construct_cell("<span style=\"color: green;\">{$lang->writable}</span>");
	}
	else
	{
		$table->construct_cell("<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}");
		++$errors;
	}

	$table->construct_row();
}
```

~~And yes, avatar upload directory has to be addressed in the same way in your PR, I believe. Because its having the same issue.~~ (see comment below)

> Right. I understand that that's what your PR is intended to fix.

My PR is intended to fix 3 other issues apart from just this visual neat-pick and one of them is a **bug** otherwise I wouldn't have created the issue only for this.

effone: OK, after some brain storming I think we don't need to do the changes in avatar upload path, as the files are images and meant to be embedded in page: hence it has always to be relative / under `MYBB_ROOT`. Otherwise we need a proxy script in place to read the images, which is not present in current MyBB.

For avatars only what we can do is to explicitly declare in settings description that the path HAS TO BE relative.
Now I am forced to think how attachment image thumbnails are served if they are in absolute path out of `MyBB_ROOT`

lairdshaw: > Please let me know if the following is acceptable to you

I don't think it does the job it should do, no. To put my view in a nutshell: if it is to make useful changes, then this PR should ensure that the `System Health` page shows beside `File Uploads Directory` _exactly_ what the admin entered for the `Uploads Path` setting in `Server and Optimization Options`. Your new code does not achieve this, and makes some problematic transformations (confirmed by testing that in some scenarios, necessary parts of a path are removed).

effone: I would like to know:

1. Which necessary part is removed from the path?
2. What is the procedure of reverse-manipulation of the base settings already manipulated by PR #4342 and overwritten in the settings value itself? Where is the backup of the original value before manipulation? Why I need to run an extra query for the info I already had, but deliberately forgot?
3. What are the test case scenarios that has failed with this PR (new code, as above)?
4. Why we are at all considering some arbitrary absolute path out of domain root and in that case how we are managing the display of attachment image thumbnails in posts?

lairdshaw: > 1. Which necessary part is removed from the path?

It occurs in a scenario like this:

The admin sets `Uploads Path` in `Server and Optimization Settings` to:

`/some-dir/mybb/uploads`

when MYBB_ROOT is set to:

`/mybb`

In this scenario (though tested using different directories applicable on my dev machine), the new code you propose above for this PR will mistakenly display beside `File Uploads Directory` in `System Health`:

`/some-dir./uploads`

> 2. What is the procedure of reverse-manipulation of the base settings already manipulated by PR #4342 and overwritten in the settings value itself?

It is currently not possible to reliably reverse those changes, because it is impossible to know whether the setting was originally an absolute path or whether it was changed by PR #4342 from a relative path to an absolute one. Hence my suggestions above (I prefer the first): https://github.com/mybb/mybb/pull/4361#discussion_r617330863

> 3. What are the test case scenarios that has failed with this PR?

See number 1 above.

> 4. Why we are at all considering some arbitrary absolute path out of domain root

Because it's a valid value for the setting, and we should handle valid values for settings.

> and in that case how we are managing the display of attachment image thumbnails in posts?

I haven't confirmed this understanding, but my understanding is that neither full-sized nor thumbnail images are required to be web-accessible, only accessible to, and processed through, PHP via the web server, so that they can exist anywhere on the filesystem from/to which PHP (via the web server) can read and write, not necessarily under MYBB_ROOT. If I'm wrong about this, I hope you'll correct me.

effone: Attachment thumbnails are a very basic feature of MyBB and is there for ages. As per my understanding it was always considered that the upload attachments are under the domain root only.

If an admin deliberately inputs a settings like what you said, `/some-dir/mybb/uploads` then this is simply intentional and there are many settings like that in MyBB, infact anywhere - placing wrong settings simply fail the system. Thats the efficiency of Admin and thats why default values are there to hint. You can't ever make a setting input bullet-proof this way, no matter what you do. Speaking to the use case scenario you have pointed out that can be easily corrected in my above code and you also know how. In place of improving the effort your straight away rejection is stating that simply you are against this PR. 

I am leaving this PR here as a reference of the valuable discussion we already had for others.
You can close the issue / PR if you really feel its not a regression. Keeping your concept in consideration, if at all MyBB works this way then there is much more important core changes are required to support out-of-domain paths (starting with a proxy image dispatcher).

Sorry friend. I failed to conclude this on a common and agreed opinion.

lairdshaw: > As per my understanding it was always considered that the upload attachments are under the domain root only.

Well, in any case, deliberate or not, an uploads directory outside of the domain root works just fine (tested). This feature seems worth supporting given the extra security it provides given that the uploads directory can't then be accessed directly via the web.

> Speaking to the use case scenario you have pointed out that can be easily corrected in my above code and you also know how.

Correct. However, the original setting value (pre being altered by PR #4342) *can't* be determined, which is the real problem.

> you are against this PR.

I just don't understand the point of it. The only point to it that I can see would be to ensure that the _exact_ value of the `Uploads Path` setting entered by the admin in the settings is displayed in `System Health` given that that value is altered by PR #4342 - but it doesn't even achieve this.

So, what _is_ its point? To ensure that a relative path is displayed where possible - even if the admin entered an absolute path in the settings? If so, why?

> Sorry friend. I failed to conclude this on a common and agreed opinion.

That's OK. Disagreements happen, even to friends.

lairdshaw: If #4370 is accepted, then we could go with this (tested):
```php
$writables = array(
	'config_file' => 'inc/config.php',
	'settings_file' => 'inc/settings.php',
	'file_upload_dir' => $mybb->settings['uploadspath'],
	'avatar_upload_dir' => $mybb->settings['avataruploadpath'],
	'language_files' => 'inc/languages',
	'backup_dir' => $config['admin_dir'] . '/backups',
	'cache_dir' => 'cache',
	'themes_dir' => 'cache/themes'
);

$errors = 0; // Reset errors
$table = new Table;
$table->construct_header($lang->checked_path);
$table->construct_header($lang->location);
$table->construct_header($lang->status_chmod, array('width' => 250));
foreach ($writables as $langvar => $path)
{
	$table->construct_cell("<strong>{$lang->{$langvar}}</strong>");
	$table->construct_cell($path);
	if (is_writable(mk_path_abs($path)))
	{
		$table->construct_cell("<span style=\"color: green;\">{$lang->writable}</span>");
	}
	else
	{
		$table->construct_cell("<strong><span style=\"color: #C00\">{$lang->not_writable}</span></strong><br />{$lang->please_chmod_777}");
		++$errors;
	}

	$table->construct_row();
}
```
Your thoughts?

Sama34: Being that #4370 has been merged, what would be the resolution for this ?

PR #4364, opened 2021-04-25
Add field_exists check for new fields in upgrade52.php

### LINKED ISSUE #4297: Check table schema before altering it in upgrade52.php
The database changes in the `upgrade52.php` upgrade script should verify existing database table structure to avoid errors when the process is run more than once.

https://github.com/mybb/mybb/blob/35203a4145e379e58d4bad326fece8f19d189f8f/install/resources/upgrade52.php#L32-L44

dvz:
Resolves #4297


Sama34 (approved):
Seems fine to me.

PR #4379, opened 2021-05-29
Add validation for banned IPs

### LINKED ISSUE #4378: Missing input validation when banning IPs
When banning IP addresses or subnets in the ACP (Configuration -> Banning -> Banned IPs) there is no validation at all, if the entered address/subnet is valid at all.
One could enter `1.1.1.256`, `1.1.1.1/33`, `::1/129`, `1.1.1.1/-1` or any other data and MyBB will happily accept it.
I've noticed that there are some users that mistakenly entered invalid subnet masks and assumed that the IP addresses have been banned, because there was no check.

n-thumann:
Resolves #4378

This resolved the issue above by performing a check on entered IP addresses or subnets.
If the entered IP address is invalid, a error will be shown (that's the reason I've renamed one existing language string).

### DIFF
--- a/admin/modules/config/banning.php
+++ b/admin/modules/config/banning.php
@@ -35,6 +35,28 @@ if($mybb->input['action'] == "add" && $mybb->request_method == "post")
 		$errors[] = $lang->error_filter_already_banned;
 	}
 
+	if(!$errors && $mybb->input['type'] == 1)
+	{
+		$ip_address = $db->escape_string($mybb->input['filter']);
+		$subnet_mask = "0";
+		if(strpos($ip_address, "*") !== false)
+		{
+			$ip_address = str_replace("*", "0", $ip_address);
+		}
+		else if(strpos($ip_address, "/") !== false)
+		{
+			list($ip_address, $subnet_mask) = explode("/", $ip_address);
+		}
+
+		$is_valid_v4 = filter_var($ip_address, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) && $subnet_mask <= 32 && $subnet_mask >= 0;
+		$is_valid_v6 = filter_var($ip_address, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) && $subnet_mask <= 128 && $subnet_mask >= 0;
+
+		if (!$is_valid_v4 && !$is_valid_v6 || !ctype_digit($subnet_mask))
+		{
+			$errors[] = $lang->error_invalid_filter;
+		}
+	}
+
 	if(!$errors)
 	{
 		$new_filter = array(
@@ -100,7 +122,7 @@ if($mybb->input['action'] == "delete")
 	// Does the filter not exist?
 	if(!$filter['fid'])
 	{
-		flash_message($lang->error_invalid_filter, 'error');
+		flash_message($lang->error_filter_not_found, 'error');
 		admin_redirect("index.php?module=config-banning");
 	}
 
--- a/inc/languages/english/admin/config_banning.lang.php
+++ b/inc/languages/english/admin/config_banning.lang.php
@@ -34,7 +34,8 @@ $l['ip_address_desc'] = "Note: To ban a range of IP addresses use * (Ex: 127.0.0
 $l['ban_ip_address'] = "Ban IP Address";
 
 $l['error_missing_ban_input'] = "You did not enter a value to ban.";
-$l['error_invalid_filter'] = "The specified filter does not exist.";
+$l['error_invalid_filter'] = "The specified filter is not valid.";
+$l['error_filter_not_found'] = "The specified filter does not exist.";
 $l['error_filter_already_banned'] = "The filter you entered is already banned.";
 
 $l['success_ip_banned'] = "The IP address has been banned successfully.";


admin/modules/config/banning.php
```diff
@@ -35,6 +35,29 @@
 		$errors[] = $lang->error_filter_already_banned;
 	}
 
+	if($mybb->input['type'] == 1)
```
lairdshaw: Suggestion: wrap this conditional in an outer conditional `if(!$errors)`.

admin/modules/config/banning.php
```diff
@@ -35,6 +35,29 @@
...
+		if(strpos($ip_address, "*") !== false)
+		{
+			$ip_address = str_replace("*", "0", $ip_address);
+		}
+		else if(strpos($ip_address, "/") !== false)
+		{
+			list($ip_address, $subnet_mask) = explode("/", $ip_address);
+		}
+
+		$subnet_mask = intval($subnet_mask);
```
lairdshaw: This conversion obscures any errors that might be present in `$subnet_mask`; that is, the case in which the admin enters a value other than an integer for the subnet. Suggestion: replace the cast with an integer test, and append to `$errors` if the test fails. An appropriate test can be performed using the `ctype_digit()` function.

inc/languages/english/admin/config_banning.lang.php
```diff
@@ -34,7 +34,9 @@
 $l['ban_ip_address'] = "Ban IP Address";
 
 $l['error_missing_ban_input'] = "You did not enter a value to ban.";
-$l['error_invalid_filter'] = "The specified filter does not exist.";
+$l['error_invalid_filter'] = "The specified filter is not valid.";
+$l['error_filter_not_found'] = "The specified filter does not exist.";
+$l['error_filter_already_banned'] = "The filter you entered is already banned.";
```
lairdshaw, line 39: This line duplicates the line below. Suggestion: remove it.

lairdshaw (commented):
Generally, although I've picked a few minor issues, this looks like a nice addition - thanks for your contribution! Just a warning: it may not be merged prior to the (imminent) release of 1.8.27, in which case its merge will have to wait until the 1.9 rebase is completed.

dvz:
We can also validate it client-side using [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern); a proper option can be added to methods like https://github.com/mybb/mybb/blob/aa7691658f98e2f4186539312f15dccd23106d44/admin/inc/class_form.php#L100

admin/modules/config/banning.php
n-thumann: I added `!$errors` into the same condition, as I think it's a little bit more compact. Do you agree? :)

admin/modules/config/banning.php
n-thumann: Oh, good catch, thanks :)
Added the type check into the already existing validity check to assure that only one warning will be shown, if both IP address and subnet mask are invalid and to keep it compact.

inc/languages/english/admin/config_banning.lang.php
n-thumann: Whoops, done!

admin/modules/config/banning.php
lairdshaw: Yes, that works fine.

admin/modules/config/banning.php
lairdshaw: That works well too.

n-thumann:
> We can also validate it client-side using [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern);

Even though that would be even more user-friendly, I'll guess that that will result in a pretty clunky and difficult to maintain RegEx, especially with support for `*` and subnet mask 🤔

lairdshaw:
> Even though that would be even more user-friendly, I'll guess that that will result in a pretty clunky and difficult to maintain RegEx, especially with support for `*` and subnet mask thinking

Yep, that was my initial reaction, although, having said that, we *do* already have *this*[1] regex, which I think *might* fairly be described as "clunky and difficult to maintain", even given @dvz's admirable refactoring of it from its previous[2], even clunkier and harder to maintain state as two separately clunky regexes. So...

But yes, the regex under consideration *might* be a lot worse.

[1] https://github.com/mybb/mybb/blob/aa7691658f98e2f4186539312f15dccd23106d44/inc/class_parser.php#L1648-L1662

[2] https://github.com/mybb/mybb/blob/ee2c6a37055dbf76266adc06f46c0ab3a21a1971/inc/class_parser.php#L1524-L1525

lairdshaw:
By the way, @n-thumann, although I'm not an expert on Git etiquette, my understanding is that it's in general not a good idea to force-push changes to public repositories (including when they are the source of pull requests); better is to add the changes in a new commit. I hope that this comment doesn't come across as overly rude.

n-thumann:
> By the way, @n-thumann, although I'm not an expert on Git etiquette, my understanding is that it's in general not a good idea to force-push changes to public repositories (including when they are the source of pull requests); better is to add the changes in a new commit. I hope that this comment doesn't come across as overly rude.

Don't worry, it didn't :) I am used to fixup additional commits for a PR to keep the history clean, so that's always ready to merge without additional squashing needed by the maintainers, but I'll keep that in mind for future commits here.

Sama34:
I think adding a browser side validation should be considered with caution because we may find ourselves into scenarios where it does not report an invalid notice whereas the server-side may do so.

I think the possibility may not be worth it.

PR #4368, opened 2021-05-12
Inhibit drag-n-drop in the editor

lairdshaw:
Temporarily prevent drag-n-drop of images into the editor until PR #4359
is updated and included in 1.9.

Based on / inspired by @effone's code/approach in that PR.

PR #4380, opened 2021-06-03
Add event handler to strip [img] with data uri in SCEditor

### LINKED ISSUE #4078: Paste image into SCEditor gives data-uri instead of URL for [img] BBcode
I haven't upgraded my forum to 1.8.21 yet, thought it would be better to try &amp; test things in community forum to see if anything is wrong. Well..

Description: When you paste an image directly to editor, it creates huge URL (like 800k characters.) Forum can't show the image and trying to edit the message with 800k characters gives you "1406 - Data too long for column 'originaltext' at row 1" MySQL error.

Reproduction steps:
Right click on any image and click on "Copy image" (not link of image, but image itself)

Paste this image on message editor with Ctrl V

Notice that instead of your image you see that 800k characters between IMG codes.

Thread I tested it in Community Forums: https://community.mybb.com/thread-223398...pid1330487

Original thread: [Another image bug in editor](https://community.mybb.com/thread-223399.html)

dvz:
Resolves #4078

May result in `[img]data:...[/img]` fragments being deleted when quoting posts (including code fragments).

### DIFF
--- a/jscripts/post.js
+++ b/jscripts/post.js
@@ -41,6 +41,23 @@ var Post = {
 
 			Post.fileInput.parents().eq(1).hide();
 			Post.dropZone.parents().eq(1).show();
+
+			// prevent SCEditor from inserting [img] with data URI
+			var $message = document.querySelector('#message');
+
+			if ($message !== null) {
+				new MutationObserver(function () {
+					// run once #message is hidden by SCEditor, and the MyBBEditor instance becomes available
+
+					if (typeof MyBBEditor !== 'undefined' && MyBBEditor !== null) {
+						MyBBEditor.bind('valuechanged', function () {
+							MyBBEditor.val(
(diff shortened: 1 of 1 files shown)


Sama34:
Won't this conflict with possible third party plugins? Is this an alternative to #4368 ?

dvz:
> Won't this conflict with possible third party plugins?

It shouldn't if they don't reference or use the `[img]data:...[/img]` format.

> Is this an alternative to #4368 ?

Yes.

Sama34:
The strip of the desired data after pasting seems to work as expected. However, it seems the cursor position is returned back to the beginning of the editor every now and then when you click on the editor text area, after typing apparently. This only happens in the rich mode, the source mode seems to work just fine.

Sama34 (approved):

PR #4383, opened 2021-06-05
Fix "undefined" issues in ACP's Email Logs

dvz:
Resolves issues referenced in https://github.com/mybb/mybb/issues/4138#issuecomment-855217648


yuliu (approved):
It's OK to me and the test of the PR is good.

Another fix to `maillogs` might be to just remove `$log['find_from']` and `$log['find_to']` in the cell output from the two rows. Anyway the PR's method seems more bullet proof :).

PR #4385, opened 2021-06-10
Fix #4384 Soft deleted thread may be visible in User CP Home's subscr…

### LINKED ISSUE #4384: Soft deleted thread may be visible in User CP Home's subscriptions
Threads that were soft-deleted may be displayed in UCP's Home page in the _Thread Subscriptions with New Posts_ list.

The condition at https://github.com/mybb/mybb/blob/mybb_1826/usercp.php#L4300 includes a `t.visible != 0` check, but the code associated with soft-deleted threads is `-1`.

Conditions on the dedicated `usercp.php?action=subscriptions` page (https://github.com/mybb/mybb/blob/mybb_1826/usercp.php#L1565) include a `t.visible >= 0` check, but may still produce inconsistent behavior.

Reported in https://community.mybb.com/thread-232604.html

Sama34:
…iptions

Resolves #4384


usercp.php
```diff
@@ -4339,6 +4365,8 @@
 
 			if($subscriptions)
 			{
+				$latest_subscribed_threads = '';
+
```
dvz: It is already initialized a few lines below in `feature` (https://github.com/mybb/mybb/commit/ed09961345ac1ed5f8d34774f995b4c430389602)

dvz (changes requested):

dvz:
This changes logic around the `visible` state. Should soft deleted and unapproved threads be displayed in _Your Latest Threads_ and subscriptions?

dvz (approved):

Eldenroot:
> This changes logic around the `visible` state. Should soft deleted and unapproved threads be displayed in _Your Latest Threads_ and subscriptions?

I think yes they should... but these should have different color to know that they are unapproved/deleted. And these should be visible only to the author, not for visitors.

PR #4386, opened 2021-06-13
Fix discovered "undefined" issues

### LINKED ISSUE #4408: "Undefined array key" error when modifying forums on PHP 8
In class_datacache.php:

	if($perms[$gid])

should probably be:

	if(isset($perms[$gid]))

If there are no permissions defined for the forum, that check can trigger an invalid key access whenever permissions are being rebuilt.

### LINKED ISSUE #4409: Error suppression doesn't work on PHP 8
When the `@` error suppression operator is used in PHP 8, the value returned by `error_reporting()` to the error handler [has changed](https://www.php.net/manual/en/language.operators.errorcontrol.php).  This causes MyBB's custom error handler to deal with errors that should be suppressed.

In inc/class_error.php:

	if(error_reporting() == 0)

should be:

	if((error_reporting() & $type) == 0)

### LINKED ISSUE #4414: Display error in custom forum permission settings for guests in ACP
If you go to "Set custom permissions" for guests in ACP forum management, the "Miscellaneous" tab looks like this:

The error is caused by an typo in admin/modules/forum/management.php line 791.

if(!in_array($field['field'], $hidefields) &amp;&amp;...
should be
if(!in_array($field['Field'], $hidefields) &amp;&amp;...

Original thread: [Display error in custom forum permission settings for guests in ACP](https://community.mybb.com/thread-232970.html)

dvz:
Partially resolves .#4138


newreply.php
```diff
@@ -630,13 +630,13 @@
 
 				// Lets see if this post is on the same page as the one we're viewing or not
 				// if it isn't, redirect us
-				if($perpage > 0 && (($postcounter) % $perpage) == 0)
+				if($mybb->settings['postsperpage'] > 0)
 				{
-					$post_page = ($postcounter) / $mybb->settings['postsperpage'];
+					$post_page = ceil(($postcounter + 1) / $mybb->settings['postsperpage']);
```
lairdshaw, line 635: Shouldn't this `$postcounter + 1` expression rather be simply `$postcounter`?

polls.php
```diff
@@ -1150,7 +1158,7 @@
 	$votesarray = explode("||~|~||", $poll['votes']);
 	if(count($votesarray) > $poll['numoptions'])
 	{
-		$votesarray = array_slice(0, $poll['numoptions']);
+		$votesarray = array_slice($poll['numoptions'], 0, $poll['numoptions']);
```
lairdshaw: Shouldn't the first parameter to `array_slice()` here be `$votesarray` rather than `$poll['numoptions']`?

lairdshaw (commented):
Reviewed by: (1) performing a local lint test on all files via `php -l [filename]` and (2) careful, comprehensive visual inspection of the code to confirm that all changes are valid and don't seem to break anything.

Other than the two comments I've left re potential needs to alter the PR, everything looks good to me.

euantorano:
Regarding the lint, which version of PHP are you running locally? Just asking as we’ve been caught out with 5.2 incompatibilities before so it’s worth double checking 😀

lairdshaw:
Fair question. I'm currently running version 7.4.23. 8.x would be ideal of course, but it's not yet shipping with Fedora.

euantorano:
Ok, sounds good to me. Most hosts should be running 7.4 by now you would hope.

newreply.php
dvz: Would the new post be accounted for? The `$postcounter = $thread['replies'] + 1` initialization would only add the thread's opening post.

lairdshaw: Ah, yep, I missed that.

PR #4396, opened 2021-06-25
Fixes #4395

### LINKED ISSUE #4395: Extra "dots" in global.lang.php
Regression from this PR #4287

Eldenroot:
#4395

Resolves #4395

### DIFF
--- a/inc/languages/english/global.lang.php
+++ b/inc/languages/english/global.lang.php
@@ -333,8 +333,8 @@ $l['unapproved_attachments'] = "{1} unapproved attachments";
 $l['unread_report'] = "1 unread report";
 $l['unread_reports'] = "{1} unread reports";
 $l['groupleader_notice'] = "Group Leader Notice:";
-$l['pending_joinrequest'] = "1 pending group membership join request.";
-$l['pending_joinrequests'] = "{1} pending group membership join requests.";
+$l['pending_joinrequest'] = "1 pending group membership join request";
+$l['pending_joinrequests'] = "{1} pending group membership join requests";
 
 $l['search_user'] = "Search for a user";
 


Sama34:
Is there a community report on this ? The former issue adds no extra "dots" and I don't see them in the code.

lairdshaw:
> Is there a community report on this ? The former issue adds no extra "dots" and I don't see them in the code.

Without wishing to speak for @Eldenroot, I think he simply means that compared to the equivalent messages above those two - unapproved_thread, unapproved_threads, unapproved_post, unapproved_posts, unapproved_attachment, and unapproved_attachments - those two messages that were added in that prior issue were added so as to end in a full stop, but, for consistency, they should not end in a full stop, given that those equivalent messages don't. Hence, he removed the full stops. Seems legit to me.

Sama34:
The `mod_notice` language string adds the full stop or period for these messages you suggest :
https://github.com/mybb/mybb/blob/e462060b47fff996f2156cd2dc13046e0c419a29/inc/languages/english/global.lang.php#L326

However, the language strings related to this issue are rendered differently :
(Full stop or period is neither added in the `groupleader_notice` language string nor the `global_pending_joinrequests` template.)
https://github.com/mybb/mybb/blob/e462060b47fff996f2156cd2dc13046e0c419a29/inc/languages/english/global.lang.php#L335
https://github.com/mybb/mybb/blob/e462060b47fff996f2156cd2dc13046e0c419a29/install/resources/mybb_theme.xml#L4208

Below an attached preview from the community forums :
image

Thus, this PR is actually removing the full stop or period, which would make it inconsistent with other similar messages.

Otherwise, I can't see the _extra_ (2 or more) dots that the issue implies.

lairdshaw:
Thanks for the in-depth analysis. I take back that "seems legit". I now think we should close this PR and its associated issue.

Sama34:
@Eldenroot if you think we misunderstood don't hesitate into correcting us please.

Eldenroot:
OK guys, np :)

PR #4401, opened 2021-07-01
Fix #4400 [split] Viewing a user IP address in ACP doesn't work with …

### LINKED ISSUE #4400: Viewing a user IP address in ACP doesn't work with MyBB 1.8.27
Problem reported and confirmed with 1827.

Viewing a user IP address in ACP - Find Users, Options, Show IP addresses - Options - Show Users that have registered with this IP OR Show Users that have posted with this IP - returns all users.

Using PHP 7.3.28.  Workaround - use IP search in modcp.

Edit:  Should've started another thread.  If the mods could do the honours.  :blush: 
Subject - 1827 ACP IP search

Original thread: [[split] Viewing a user IP address in ACP doesn't work with MyBB 1.8.27](https://community.mybb.com/thread-232854.html)

Sama34:
…MyBB 1.8.27

Resolves #4400

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1890,9 +1890,8 @@ if($mybb->input['action'] == "ipaddresses")
 	{
 		$user['lastip'] = my_inet_ntop($db->unescape_binary($user['lastip']));
 		$popup = new PopupMenu("user_last", $lang->options);
-		$popup->add_item($lang->show_users_regged_with_ip,
-			"index.php?module=user-users&amp;action=search&amp;results=1&amp;conditions=".urlencode(my_serialize(array("regip" => $user['lastip']))));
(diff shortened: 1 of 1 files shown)


dvz:
Do we know what caused the issue?

Sama34:
https://github.com/mybb/mybb/commit/a484dee9802ecb525a7fee745195e2dfe774983c#diff-50b2bcac47c31533d9f4560598fcbcd89c0a06808eae2bed8e4abba66087e69e

Please note the following in the `build_users_view()` function :
https://github.com/mybb/mybb/blob/70469f5ad9fae24aeb88e6cdb760f5fca0bd7713/admin/modules/user/users.php#L3218-L3221

I decided on my approach instead of keeping the use of both `urlencode()` and `my_serialize()` to build the url.

PR #4412, opened 2021-07-09
Add opportunistic early setting loading using cache file

### LINKED ISSUE #4411: Cached settings not available during early errors
Currently, settings are loaded from the `inc/settings.php` file with a database fallback only after successfully establishing a database connection. This results in setting values (e.g. related to error handling configuration), in a potentially healthy cached file, not being available to the error handler when database connection problems occur.

dvz:
Resolves #4411

### DIFF
--- a/inc/init.php
+++ b/inc/init.php
@@ -109,6 +109,12 @@ if(empty($config['admin_dir']))
 // Load Settings
 $settings = array();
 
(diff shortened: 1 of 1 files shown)


lairdshaw (approved):
Visually inspected + tested. Seems to be working fine.

PR #4413, opened 2021-07-10
Fix #4410 - Unrendered font codes - plus add generic parser support for nestable callback MyCodes

### LINKED ISSUE #4410: Font codes showing in some posts
After update to 1.8.27 font codes are showing in some posts. 

Mentioned here: https://community.mybb.com/thread-232927.html

It can be seen on MyBB's forum here:  forum: https://community.mybb.com/post-1342828.html

I have found it on my forum in an old post too. 

Thank you.

Original thread: [Font codes showing in some posts](https://community.mybb.com/thread-232943.html)

lairdshaw:
Provide parser support for nestable font codes.

Resolves #4410

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -315,8 +315,8 @@ class postParser
 		global $cache, $lang, $mybb;
 		$this->mycode_cache = array();
 
-		$standard_mycode = $callback_mycode = $nestable_mycode = array();
-		$standard_count = $callback_count = $nestable_count = 0;
+		$standard_mycode = $callback_mycode = $nestable_mycode = $nestable_callback_mycode = array();
+		$standard_count = $callback_count = $nestable_count = $nestable_callback_count = 0;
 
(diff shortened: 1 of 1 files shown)


dvz:
Was the section moved to the other method because `postParser::$mycode_cache` doesn't support entries that are both `nestable` and `callback`? Would it be better to add this support?

lairdshaw:
> Was the section moved to the other method because `postParser::$mycode_cache` doesn't support entries that are both `nestable` and `callback`?

Yes, exactly.

> Would it be better to add this support?

Ideally, yes. I'll see what I can come up with.

Sama34:
I did test using the following post :
https://community.mybb.com/thread-226339-post-1342791.html#pid1342791

image

There seem to be some issues, but these seem to be related to the `[url]` MyCode, so as far as I can assure it seems your code fixes the `[font]` issue.

Sama34 (approved):

PR #4424, opened 2021-08-09
Adding forcedownload field in install

### LINKED ISSUE #4423: mybb_attachtypes.forcedownload not created on fresh install
Related to https://community.mybb.com/thread-233247.html : field mybb_attachtypes.forcedownload is not in install/resources/mysql_db_tables.php, just in install/resources/upgrade52.php.
So, in a fresh install, an error occurres when trying to edit attachment types

crazycatdevs:
Resolves #4423


dvz:
Note these will also need to be added to table & insert files for PostgreSQL and SQLite.

crazycatdevs:
Well, I did this change via web interface, not sure I can add changes in this PR but I'll try.

crazycatdevs:
Ok, this seems complete now. Let me know if I forgot a file

Sama34:
Default value for MySQL seems to be `0` but for the rest you used `1` (inserts too), I think this should be `0` by default.

crazycatdevs:
You're right. I'll correct that soon.
Don't you think that some attachement types (office documents) should be to 1 by default ?

Sama34:
Probably, but 1) browsers already prompt for some file types and 2) in 1.8 this would change behavior unexpectedly.

We could update some default data for 1.9

Sama34 (approved):

awslabspl:
This is producing an error on macOS Monterey Beta. Works fine on macOS BigSur

crazycatdevs:
> 
> 
> This is producing an error on macOS Monterey Beta. Works fine on macOS BigSur

Can you be more descriptive ? It's only changement in SQL queries, so it's OS independant, only the SQL engine/version could be affected.

PR #4427, opened 2021-08-11
Fix input type handling in ACP's Attachments

### LINKED ISSUE #4426: Delete and forum filter functions not working in ACP attachments module
Deleting attachments after a search and filtering the search of attachments by forum are not working anymore after #4207.

dvz:
Resolves #4426

### DIFF
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -712,7 +712,7 @@ if(!$mybb->input['action'])
 		$forum_cache = cache_forums();
 
 		// Searching for attachments in a specific forum, we need to fetch all child forums too
-		if($mybb->get_input('forum'))
+		if(!empty($mybb->input['forum']))
 		{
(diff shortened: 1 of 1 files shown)


Sama34 (approved):

demtor:
Please remember to include forum filter option fix here:
https://github.com/mybb/mybb/blob/ea30022ca1f4745b9c133b8ff131f3b78f374f52/admin/modules/forum/attachments.php#L707-L713

lairdshaw (approved):
Lightly tested. Seems to solve the problem. However, I have encountered a separate, unresolved problem: "Results per page" is by default set to zero. Is this just me, or do others encounter this problem?

dvz:
> I have encountered a separate, unresolved problem: "Results per page" is by default set to zero. Is this just me, or do others encounter this problem?

Created #4454

PR #4430, opened 2021-08-16
Fix #4428 Remove "MyBB Internal..." message output in error handler

### LINKED ISSUE #4428: Remove "MyBB Internal..." message output in error handler
Remove the output of the https://github.com/mybb/mybb/blob/ea30022ca1f4745b9c133b8ff131f3b78f374f52/inc/class_error.php#L225 message displayed at the top of the page when warnings occur.

This output is likely unexpected by administrators, and can break AJAX communication and cause further _headers already sent_ errors to be logged.
Numerous PHP Warnings are still expected when running on PHP 8 (#4138).

Sama34:
Resolves #4428


inc/class_error.php
```diff
@@ -203,7 +203,7 @@ function error($type, $message, $file=null, $line=0, $allow_output=true)
 			$this->email_error($type, $message, $file, $line);
 		}
 
-		if($allow_output === true)
+		if($allow_output === true && $mybb->settings['errortypemedium'] != "none")
```
dvz: This may prevent MyBB's error page from showing - instead we could change https://github.com/mybb/mybb/blob/2da5152718ad740bf4d2648829c856a5035a9a01/inc/class_error.php#L221 to
```php
elseif(in_array($mybb->settings['errortypemedium'], array('warning', 'both')))
```
so that all details are printed when _Error Type Medium_ is set either of these options.

dvz (changes requested):

inc/class_error.php
Sama34: image

Don't you think that is inconsistent with the description for the setting ?

dvz: Not if _errors_ imply _error details_ (the information being logged) - otherwise a blank page would be shown.

Sama34: Perhaps we could update the setting description from `Hide Errors and Warnings` to `Hide Errors and Warnings Details` then ?

dvz: Maybe
```diff
- The type of errors to show.
+ The type of errors to show with detailed information.
```

dvz:
A `$this->output_error()` for PHP Errors call seems to have been lost in https://github.com/mybb/mybb/pull/4430/commits/ac8e530b72c7179121fe04e04818abb04c4f413f, I opened #4450 with same changes

PR #4431, opened 2021-08-16
Fix #4420 registered users unable to change banned mail address

### LINKED ISSUE #4420: registered users unable to change banned mail address
When admin bans mail adresses that registered users are still currently using, and the Board Settings do not allow users to keep their already banned addresses, the forum forcibly redirects them to their User CP and demands a change of mail address.

https://github.com/mybb/mybb/blob/ea30022ca1f4745b9c133b8ff131f3b78f374f52/global.php#L1227-L1237

Unfortunately the same code prevents to actually finalize the email change, as the email change confirmation / activation mail (member.php action activate) is also forcibly redirected to the usercp.php mail change form.

https://github.com/mybb/mybb/blob/ea30022ca1f4745b9c133b8ff131f3b78f374f52/inc/languages/english/messages.lang.php#L369-L380

Since the activation link can not be followed the email change is never activated and users are stuck with their banned address.

It does not affect newly registered users still "Awaiting Activation" as their email change procedure seems to be different (old address is not retained).

Sama34:
Resolves #4420

### DIFF
--- a/global.php
+++ b/global.php
@@ -1226,7 +1226,10 @@ if($mybb->usergroup['canview'] != 1)
 // If they are, redirect them to change it
 if($mybb->user['uid'] && is_banned_email($mybb->user['email']) && $mybb->settings['emailkeep'] != 1)
 {
-	if(THIS_SCRIPT != 'usercp.php' || THIS_SCRIPT == 'usercp.php' && $mybb->get_input('action') != 'email' && $mybb->get_input('action') != 'do_email')
+	if(
+		!(THIS_SCRIPT == 'usercp.php' && in_array($mybb->get_input('action'), array('email', 'do_email'))) ||
+		!(THIS_SCRIPT == 'member.php' && $mybb->get_input('action') == 'activate')
+	)
 	{
 		redirect('usercp.php?action=email');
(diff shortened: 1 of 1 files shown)


global.php
```diff
@@ -1226,7 +1226,10 @@
 // If they are, redirect them to change it
 if($mybb->user['uid'] && is_banned_email($mybb->user['email']) && $mybb->settings['emailkeep'] != 1)
 {
-	if(THIS_SCRIPT != 'usercp.php' || THIS_SCRIPT == 'usercp.php' && $mybb->get_input('action') != 'email' && $mybb->get_input('action') != 'do_email')
+	if(
+		!(THIS_SCRIPT == 'usercp.php' && in_array($mybb->get_input('action'), array('email', 'do_email'))) ||
```
dvz:
```suggestion
		!(THIS_SCRIPT == 'usercp.php' && in_array($mybb->get_input('action'), array('email', 'do_email'))) &&
```

dvz (changes requested):
Currently causes a redirect loop

dvz (approved):

PR #4434, opened 2021-08-19
Fix input type for custom profile fields in ACP's Find Users

### LINKED ISSUE #4433: Custom profile field conditions ignored in ACP's Find Users
Regression from #4207

Values entered in _Where custom profile fields match_ are not taken into account when searching for users in _Find Users_ in the ACP.

Reported in https://github.com/mybb/mybb/commit/a484dee9802ecb525a7fee745195e2dfe774983c#r55121131

dvz:
Resolves #4433


lairdshaw (approved):
Lightly tested. Seems to fix the problem it was intended to fix.

PR #4435, opened 2021-08-19
Fixes #3634 Empty folders - message count of unread instead of inbox

### LINKED ISSUE #3634: Empty folders - message count of unread instead of inbox
https://community.mybb.com/thread-222444.html

It was introduced in v1820 - affected lines of code https://github.com/mybb/mybb/blob/feature/private.php#L1592-L1603

Eldenroot:
Empty folders - message count of unread instead of inbox

Thanks for fix - @doylecc

Resolves #3634

### DIFF
--- a/private.php
+++ b/private.php
@@ -1598,7 +1598,7 @@ if($mybb->input['action'] == "empty")
 	$plugins->run_hooks("private_empty_start");
 
 	$foldersexploded = explode("$%%$", $mybb->user['pmfolders']);
-	$folderlist = $unread = '';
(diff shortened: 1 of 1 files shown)


Sama34:
The variable should be removed from outside the loop then.

Eldenroot:
@Sama34 - please review again.

Thank you @lairdshaw for another lesson :)

Eldenroot:
@dvz and others, please could you check? Quite simple change but resolves annoying bug :)

PR #4438, opened 2021-08-28
Fix multiple thread moderation

### LINKED ISSUE #4437: Multiple thread moderation works only for one thread
Regression from PR #4170

When you try to approve, unapprove, restore or soft delete multiple threads (through the multiselect with checkboxes), only the first one in the list is moderated.

demtor:
Resolves #4437

### DIFF
--- a/inc/class_moderation.php
+++ b/inc/class_moderation.php
@@ -423,8 +423,8 @@ class Moderation
 			if($forum['usepostcounts'] != 0)
 			{
 				// On approving thread restore user post counts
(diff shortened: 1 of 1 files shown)


Sama34:
Seems about correc.

ghost:
> Seems about correc.

This is exactly what you said about #4170 but that clearly wasn’t the case, because it created this issue, so does it “seem” correct or did you test it for correctness I.e it fixing the issues that were introduced by the aforementioned PR.

PR #4440, opened 2021-08-30
Add separate callback function for `set_error_handler()`

### LINKED ISSUE #4439: Error handler's output may be incorrectly hidden
The `$allow_output` parameter of `errorHandler::error()` (added in 0328b9609770e97df180a9c483ad8760593c647d) may be misinterpreted when `array $errcontext`, compatible with PHP's [`set_error_handler()`](https://www.php.net/set_error_handler)'s callback signature, is passed in its place. This may result in triggered errors not being displayed.

dvz:
Resolves #4439

### DIFF
--- a/inc/class_error.php
+++ b/inc/class_error.php
@@ -137,11 +137,24 @@ class errorHandler {
 			$error_types = $error_types & ~$bit;
 		}
 		error_reporting($error_types);
-		set_error_handler(array(&$this, "error"), $error_types);
+		set_error_handler(array(&$this, "error_callback"), $error_types);
 	}
 
 	/**
-	 * Parses a error for processing.
+	 * Passes relevant arguments for error processing.
+	 *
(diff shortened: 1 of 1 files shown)


lairdshaw (commented):
Visually inspected + tested. Seems to work OK<s>, although I experienced an outcome that I couldn't explain, and the fact that I can't explain it is the only reason I'm not selecting "Approve", which I otherwise would:

This PR for some reason that I can't understand causes in at least one case an error/warning to be displayed where it wasn't prior to the PR - but since it seems that the error/warning *should* be displayed, this, whilst curious, does not seem to be problematic.</s>

Oh dear. I'm not sure how I missed that that's the entire point of the PR: to ensure that errors/warnings that were mistakenly not being displayed *are* displayed.

lairdshaw (approved):

PR #4443, opened 2021-08-31
Fix success deleted user on yourself

### LINKED ISSUE #4442: Fix success message when attempting to delete yourself
MyBB won't allow you to delete yourself however a success message is shown regardless, an error message should be shown correctly informing the user that their attempt has failed.

lewislarsen:
Resolves #4442

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1767,6 +1767,12 @@ if($mybb->input['action'] == "delete")
 		admin_redirect("index.php?module=user-users");
 	}
 
+	if($user['uid'] == $mybb->input['uid'])
+	{
+		flash_message($lang->error_delete_self, 'error');
+		admin_redirect("index.php?module=user-users");
+	}
+
 	// User clicked no
 	if($mybb->get_input('no'))
 	{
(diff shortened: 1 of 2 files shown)


admin/modules/user/users.php
```diff
@@ -1767,6 +1767,12 @@ function toggleBox(action)
 		admin_redirect("index.php?module=user-users");
 	}
 
+	if($mybb->user['uid'] == $mybb->input['uid'])
```
Sama34, line 1770:
```suggestion
	if($mybb->user['uid'] == $user['uid'])
```

inc/languages/english/admin/user_users.lang.php
```diff
@@ -91,6 +91,7 @@
 $l['error_remote_avatar_not_allowed'] = "Remote avatar URLs have been disabled by the forum administrator.";
 $l['error_invalid_user'] = "You have selected an invalid user.";
 $l['error_no_perms_super_admin'] = "You do not have permission to edit this user because you are not a super administrator.";
+$l['error_delete_self'] = "You cannot delete yourself.";
```
Sama34, line 94:
```suggestion
$l['error_delete_self'] = "You are not allowed to delete your own account while logged in.";
```

Sama34 (changes requested):

inc/languages/english/admin/user_users.lang.php
Xazin: It's kind of weird to write "while logged in", it assumes you could perform the action whilst not logged in, which is similarly impossible.

PR #4447, opened 2021-09-02
Fix `$value` check in `DefaultForm::generate_yes_no_radio()`

### LINKED ISSUE #4446: Value passed to `DefaultForm::generate_yes_no_radio()` can be evaluated differently on PHP 8
The condition `$value == "no"` at https://github.com/mybb/mybb/blob/8cb31f99c1fa19c4c209cb50d37b27592b37b45d/admin/inc/class_form.php#L757
can return different results on PHP 8 and previous versions, as a result of changes described on https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.string-number-comparision

This applies e.g. when `$value` is set to `0` (int).

dvz:
Resolves #4446

### DIFF
--- a/admin/inc/class_form.php
+++ b/admin/inc/class_form.php
@@ -754,7 +754,7 @@ class DefaultForm
 		global $lang;
 
 		// Checked status
-		if($value == "no" || $value === '0')
+		if($value === "no" || $value === '0' || $value === 0)
 		{
 			$no_checked = 1;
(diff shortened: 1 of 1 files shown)


lairdshaw (approved):
Visually inspected + tested locally. Seems to fix the problem and work as expected with no undesirable side-effects.

PR #4448, opened 2021-09-04
Fix `undefined` issues in `check_admin_permissions()`

### LINKED ISSUE #4444: Admin permission management broken in PHP 8
Issues include:

- [x] Numerous undefined array key errors for the administrator who's permissions have changed.
- [x] Updating permission values will wipe the previously applied permissions for the admin instead of adding to them.
- [x] #4446

PHP Version: 8.0.9
MyBB Version: 1.8.27

Have tested this across two different MyBB 1.8.27 installs and they were both affected equally.

dvz:
Resolves #4444

### DIFF
--- a/admin/inc/functions.php
+++ b/admin/inc/functions.php
@@ -359,7 +359,10 @@ function check_admin_permissions($action, $error = true)
 	{
 		$func = $action['module']."_admin_permissions";
 		$permissions = $func();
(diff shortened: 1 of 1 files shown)


lairdshaw (approved):
Visually inspected + lightly tested locally. Seems to resolve the problem and be working as expected.

PR #4455, opened 2021-09-18
Fix #4454: Faulty default per-page value in ACP Find Attachments

### LINKED ISSUE #4454: Default "Results per page" value in ACP's Find Attachments fails form validation
In ACP's _Find Attachments_ (_Forums & Posts → Attachments_), the default value of the _Results per page_ field is `0` (as a result of casting empty input value to integer), triggering client-side form validation.

Regression from #4095 (https://github.com/mybb/mybb/commit/1db7d2d5b38bf2d8f04c6b2fe23b32f1ed777731#diff-e1d6c9d5f9c1a7a6081fad1fea80cb6e060deee470f07fe35fb95ee5204a3391R943)

lairdshaw:
Resolves #4454

### DIFF
--- a/admin/modules/forum/attachments.php
+++ b/admin/modules/forum/attachments.php
@@ -42,6 +42,11 @@ $plugins->run_hooks("admin_forum_attachments_begin");
 $uploadspath_abs = mk_path_abs($mybb->settings['uploadspath']);
 
 $default_perpage = 20;
+$perpage = $mybb->get_input('perpage', MyBB::INPUT_INT);
+if(!$perpage)
+{
+	$perpage = $default_perpage;
+}
 
 if($mybb->input['action'] == "delete")
 {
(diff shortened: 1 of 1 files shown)


dvz:
`$mybb->get_input('perpage')` may have worked (looks like it's used that way in some files)

lairdshaw:
Yes, it would have worked, but the form field would then have been empty on first view. I thought it was more helpful for the admin to be able to see the default value of 20 there, rather than a blank field.

Sama34:
I think moving the `if(!$mybb->input['perpage'])` check right after the `admin_forum_attachments_delete` hook would have sufficed as well.

lairdshaw:
I think you're right, @Sama34, although I assume you mean right *before* that hook, and before the conditional that immediately precedes it (for the `delete` action).

That approach would require the least changes. I'm happy for you to submit an alternative PR using that approach which supersedes this one if you'd like to.

lairdshaw:
Sorry to kind of make that invitation redundant, @Sama34, but I figured it was appropriate to integrate your suggested change into this existing PR, but using the `$perpage` variable rather than overwriting the `$mybb->input['perpage']` variable, as my personal (aesthetic) sense is to prefer to avoid the latter (overwriting input variables or function parameters), although I also know it's common practice in MyBB.

Sama34 (approved):

PR #4461, opened 2021-10-02
Handle `postParser::validate_output()` return value with conditional

### LINKED ISSUE #4460: Parser output may be incorrectly suppressed when ignored errors occur
The [`postParser::validate_output()`](https://github.com/mybb/mybb/blob/mybb_1827/inc/class_parser.php#L1977) method's returned value is dependent on all parsing errors, rather than taking the ignored codes (`$ignored_error_codes`) into account, similar to the condition for logging the error.

This may result in over-aggressive suppression of output.

dvz:
Resolves #4460


yuliu:
Looks good.

PR #4465, opened 2021-10-08
Fix `undefined` issues

dvz:
Partially resolves .#4138

Addresses error reported in https://community.mybb.com/thread-233858.html


yuliu:
So far so good as I tested the PR.

PR #4467, opened 2021-10-09
Fix array key name for subforum viewers count in `build_forumbits()`

### LINKED ISSUE #4466: Subforums viewers count not accounted for on index page
The array key `viewing` at https://github.com/mybb/mybb/blob/mybb_1827/inc/functions_forumlist.php#L165-L168 does not exist; `viewers`, the key likely being referred to, is thus not added to the count, which results in views from subforums not being added to the parent's count (_x users browsing_ - _Show x viewing forum_ / `showforumviewing` setting).

dvz:
Resolves #4466


Sama34 (approved):

yuliu:
I'll need a bit more time to examine this PR and its resolving issue #4466. It seems to me there's something more hidden in that issue.

yuliu:
Comments added to the original issue.

yuliu (approved):

PR #4475, opened 2021-10-20
Fix #4471: Thread Subscription options do not persist when previewing…

### LINKED ISSUE #4471: Thread Subscription options do not persist when previewing a new thread
* Symptom: When selecting a "Thread Subscription" radio button other than the default (first) one on the `newthread.php` page, and then clicking "Preview Post", the selected radio button returns to the default.

* Diagnosis: Regression from 1.8's commit 2c4843c2fe2b743b6352c285af3ef481c9617c95 in the introduced function `get_subscription_method()`.

* Solution: In that function, move the following code immediately prior to the code above it commented as "Return user default if no thread id available, in case":
```
	if(isset($postoptions['subscriptionmethod']))
	{
		$method = trim($postoptions['subscriptionmethod']);
		return (in_array($method, $subscription_methods)) ? $method : $subscription_methods[0];
	}
```

Probably, this issue applies to the 1.8 branch too, although I haven't (yet) checked, so a 1.8 label might need to be added to the 1.9 label.

lairdshaw:
… a new thread

Resolves #4471

This fixes the issue for the 1.9 branch only. I haven't submitted a corresponding PR for 1.8 because we are nearing another release and I don't want to confuse things.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -3776,18 +3776,22 @@ function get_subscription_method($tid = 0, $postoptions = array())
 		$subscription_method = 0;
 	}
 
-	// Return user default if no thread id available, in case
-	if(!(int)$tid || (int)$tid <= 0)
-	{
-		return $subscription_methods[$subscription_method];
-	}
-
-	// If method not predefined set using data from database
+	// First check for a subscription method in $postoptions. Return it if found.
+	// e.g., When previewing a new thread, we want the user's choice
+	// of radio button to carry over.
 	if(isset($postoptions['subscriptionmethod']))
 	{
 		$method = trim($postoptions['subscriptionmethod']);
 		return (in_array($method, $subscription_methods)) ? $method : $subscription_methods[0];
 	}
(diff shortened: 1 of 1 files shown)


Eldenroot:
In my opinion we should bring these "easy and small" fixes to MyBB 1.8 branch as well... because we will get more bug fixes version than just 1.8.28 until MyBB 1.9 will be out :)

lairdshaw:
Maybe. In any case, we could go about it like this: we file all PRs for "easy and small" fixes against 1.9, and, if/when such a PR gets a passing review and is actually merged, somebody (hint hint!) files the corresponding PR against 1.8. Yes?

Eldenroot:
> Maybe. In any case, we could go about it like this: we file all PRs for "easy and small" fixes against 1.9, and, if/when such a PR gets a passing review and is actually merged, somebody (hint hint!) files the corresponding PR against 1.8. Yes?

Good idea, wait for merge and then if it is a small and easy fix -> backport it to MyBB 1.8.x branch. It will also close the issue :)

PR #4479, opened 2021-10-23
 Fixes #4480 - Threaded Mode Can't get back to old threads if sorted more recently

### LINKED ISSUE #4480: Forum view's sorting & filtering options missing when no threads displayed
From report here: https://community.mybb.com/thread-234049.html

This affects only MyBB 1.8.x branch, not MyBB 1.9.x branch.

Check is redundant, $threadcount cannot be negative.

Eldenroot:
From report here: https://community.mybb.com/thread-234049.html

Resolves #4480

### DIFF
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1529,7 +1529,7 @@ if($foruminfo['type'] != "c")
 	// Populate Forumsort
 	$forumsort = '';
 	
-	if($threadcount > 0)
+	if($threadcount >= 0)
 	{
 		eval("\$forumsort = \"".$templates->get("forumdisplay_forumsort")."\";");
 	}
--- a/forumdisplay.php
+++ b/forumdisplay.php
@@ -1528,12 +1528,8 @@ if($foruminfo['type'] != "c")
 
 	// Populate Forumsort
 	$forumsort = '';
-	
-	if($threadcount >= 0)
-	{
-		eval("\$forumsort = \"".$templates->get("forumdisplay_forumsort")."\";");
-	}
-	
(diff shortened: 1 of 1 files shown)


lairdshaw:
Elden, with the greatest of respect, does this bug apply to the 1.9 branch, and if so, could you please first file the PR on that branch? We are switching our focus to 1.9 development, and should not be filing PRs against 1.8 that could be first filed against 1.9.

Eldenroot:
I do not use MyBB 1.9 so I cannot test it.

I am not a good coder so I do not want to send untested PR to MyBB 1.9 branch. Anyway you can port it when this be valid, not?

forumdisplay.php
```diff
@@ -1529,7 +1529,7 @@
 	// Populate Forumsort
 	$forumsort = '';
 	
-	if($threadcount > 0)
+	if($threadcount >= 0)
```
Stefan-MyBB: Can `$threadcount` actually be negative? Otherwise the check doesn't make much sense.

lairdshaw:
Nevermind, Elden, I've just checked and the bug doesn't seem to be present in 1.9 anyhow. Stefan's review is on point though: it doesn't look as though `$threadcount` could ever be negative, so the conditional is redundant, and can be removed altogether.

forumdisplay.php
Eldenroot: You are right, thx!

Eldenroot:
@dvz - hi, can you add this to 1.8.30 milestone please? Small fix, no breaking changes and makes life easier :)

Eldenroot:
Thank you!

PR #4481, opened 2021-10-23
Updated composer dependencies

Starpaul20:
Brings all composer dependencies up to date. This requires the minimum PHP version to be bumped to ~~7.2.5~~ 7.3.

~~There is one deprecation notice when installing (phpunit/php-token-stream) however it doesn't seem to cause any issues.~~ Fixed.

### DIFF
--- a/inc/init.php
+++ b/inc/init.php
@@ -229,8 +229,10 @@ if(!defined("IN_INSTALL") && !defined("IN_UPGRADE") && $version['version_code'] 
 
 require_once __DIR__.'/src/bootstrap.php';
 
+use Illuminate\Support\Arr;
+
 MyBB\app('config')->set(
-	array_dot($config)
+	Arr::dot($config)
 );
 
 $error_handler->force_display_errors = false;
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,9 @@ matrix:
     - php: 8.0
     - php: 8.0
       env: setup=lowest
+    - php: 8.1
+    - php: 8.1
+      env: setup=lowest
     - php: nightly
     - php: nightly
       env: setup=lowest
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.2
-    - php: 7.2
-      env: setup=lowest
     - php: 7.3
     - php: 7.3
       env: setup=lowest
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,17 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
-    - php: 7.1
-      env: setup=lowest
     - php: 7.2
     - php: 7.2
       env: setup=lowest
     - php: 7.3
     - php: 7.3
+      env: setup=lowest
+    - php: 7.4
+    - php: 7.4
+      env: setup=lowest
+    - php: 8.0
+    - php: 8.0
       env: setup=lowest
     - php: nightly
     - php: nightly
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "issues": "https://github.com/mybb/mybb/issues"
     },
     "require": {
-        "php": "^7.2.5",
+        "php": ">=7.2.5",
         "illuminate/config": "7.*",
         "illuminate/container": "7.*",
         "illuminate/events": "7.*",
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "phpspec/phpspec": "^6.3",
         "phpunit/phpunit": "^8.5",
         "squizlabs/php_codesniffer": "^3.6",
-        "phpstan/phpstan": "0.12.99"
+        "phpstan/phpstan": "^1.3"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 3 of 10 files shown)


lairdshaw (commented):
That was quick!

Lightly tested and seems to be working well, other than that at the end of a test installation I got a message about `array_dot()` being undefined. I was able to fix the message by inserting in `install/index.php` the line:
```php
use Illuminate\Support\Arr;
```
just above the line:
```php
function install_done()
```
and then changing in that function the line:
```php
		array_dot($config)
```
to:
```php
		Arr::dot($config)
```
as I saw you'd done elsewhere.

lairdshaw (approved):
Seems to be working fine now.

Matslom:
Looks good, I've tested it and its working fine

cronlabspl (approved):
Working fine for me as well

inc/init.php
```diff
@@ -229,8 +229,10 @@
 
 require_once __DIR__.'/src/bootstrap.php';
 
+use Illuminate\Support\Arr;
+
 MyBB\app('config')->set(
-	array_dot($config)
+	Arr::dot($config)
```
yuliu, line 235: I don't understand here. If we're using package `Illuminate\Support` here, why isn't it required from the `composer.json`?

Starpaul20: Probably because it's included with other dependencies as a required package for installing.

Starpaul20:
I've updated the dependencies again and have tested them against PHP 8.1. No issues.

inc/init.php
yuliu: Hmm, IMO it should be required as we use the library explicitly. Although it might not be, it's still possible for this lib to be dropped from various `Illuminate` lib dependencies. I may be wrong.

yuliu:
I have some warnings & errors when `composer install` and `composer update` with the latest commit:

General `composer install` warning & error:
```
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
Your lock file does not contain a compatible set of packages. Please run composer update.
```

`composer install` on PHP 8.0:
```
  Problem 1
    - Root composer.json requires php ^7.2.5 but your php version (8.0.13) does not satisfy that requirement.
```

`composer install` on PHP 8.1:
```
  Problem 1
    - Root composer.json requires php ^7.2.5 but your php version (8.1.0) does not satisfy that requirement.
  Problem 2
    - phpspec/phpspec is locked to version 6.3.1 and an update of this package was not requested.
    - phpspec/phpspec 6.3.1 requires php ^7.2 || 8.0.* -> your php version (8.1.0) does not satisfy that requirement.
```

`composer update`:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php ^7.2.5 but your php version (8.0.13) does not satisfy that requirement.
```

I'm not sure if there's something wrong with my end.

Starpaul20:
That's weird, it works for me. I would double check to make sure you have all the correct files.

lairdshaw:
I had a similar problem (using PHP 8.0.14), off a fresh clone from GItHub, so I don't think it's a problem with correct files.

yuliu:
`>=` or `||` should be used. Check https://github.com/composer/composer/issues/9529 and https://getcomposer.org/doc/articles/versions.md#writing-version-constraints

yuliu:
With (not a suggestion)
```
"require": {
        "php": "^7.2.5||~8.0||~8.1",
...
```

`composer update` runs successfully on PHP 8.0. But the following errors on PHP 8.1:
>   Problem 1
>    - Root composer.json requires phpspec/phpspec ^6.3 -> satisfiable by phpspec/phpspec[6.3.0, 6.3.1, 6.3.x-dev].
>    - phpspec/phpspec[6.3.0, ..., 6.3.x-dev] require php ^7.2 || 8.0.* -> your php version (8.1.0) does not satisfy that requirement.

Starpaul20:
I've updated the version constraint to use a greater than or equal to comparison.

RE the phpspec issue: that seems to be caused by phpspec not yet supporting PHP 8.1, though it looks like it may come soon:  phpspec/phpspec#1406

yuliu:
Tested installing on PHP 7.4 and PHP 8.0 on Windows with no issues.

yuliu:
I don't think using only `>=` is a good idea which means the software will be automatically be announcing itself to be compatible with any future PHP versions.

dvz:
If it works on all versions, we can merge it to restore functionality and adjust requirements later, when upstream issues are resolved.

yuliu:
> If it works on all versions, we can merge it to restore functionality and adjust requirements later, when upstream issues are resolved.

No on PHP 8.1 (blocked by phpspec).

lairdshaw:
Thanks for the updates, @Starpaul20. Unfortunately, it seems that our blocker on PHP 8.1 is still with PHPSpec per the issue you linked to above, right? So, we unfortunately can't merge this PR just yet, even though it's worth keeping it as up to date as we can in the meantime, as you're doing. Does that match your understanding?

Starpaul20:
PHPSpec has been updated; also bumped PHP to 7.3, which also got rid of that php-token-stream deprecation notice.

lairdshaw:
Briefly tested on PHP 8.0.16 and seems to be working fine. Does anybody else want to test on (this and) other supported PHP versions, or otherwise have any objection to this PR finally being merged?

PR #4487, opened 2021-11-04
Fixes #4318 Banning system improvements

### LINKED ISSUE #4318: Banning system improvements
Currently the banning system is scattered around the code in different places, ideally a data handler for it should be implemented.

Currently, I could find the following places where hooks might be appropriate.

https://github.com/mybb/mybb/blob/7732e52795c7f6717e25c19bf16cb0a6b66ab288/inc/tasks/usercleanup.php#L55-L64

https://github.com/mybb/mybb/blob/7732e52795c7f6717e25c19bf16cb0a6b66ab288/admin/modules/user/users.php#L2551-L2560

https://github.com/mybb/mybb/blob/7732e52795c7f6717e25c19bf16cb0a6b66ab288/admin/modules/user/users.php#L2644-L2651

Starpaul20:
Resolves #4318 

This moves all things related to banning to the user datahandler.

- validate_ban - Checks if the user can be banned. Must be called for both banning and editing ban.
- insert_ban -  Handles all things related to banning a user.
- update_ban - Handles all things related to editing a ban.
- lift_ban - Handles all things related to lifting a ban.

All 4 functions have hooks for easy plugin use (what the original issue covered).

Also fixes a few things related to banning:
- Added a lift ban option to the Inline User Moderation. The function already existed but needed to be accessed via the ban user option.
- Split the ban user and edit ban pages in the Mod CP.
- Fixed a few issues related to picking ban groups when banning.


lairdshaw:
@Starpaul20,

Thank you for this impressive PR. I appreciate the opportunity to review it prior to its being (imminently, I expect) merged. I have reviewed it in two respects: lightly testing its functionality according to expected usage, and (somewhat more rigorously) visually inspecting the code. The testing succeeded very well, and the visual inspection too, with some more-or-less major caveats. I'll summarise them all here in this review comment, before annotating them in the code of the PR itself. Please feel free to respond to this comment, to the annotations in the code, or to both - as you see fit.

More serious:

Lines 3803 and 3885 of modcp.php: the lack of escaping of $mybb->input['uid'] in these queries looks like it potentially facilitates an SQL injection vulnerability.

Less serious:

Line 257 of admin/modules/user/banning.php: it appears (and I've tested to confirm) that 'usergroup' should not have been changed to 'gid' here.
Similarly, it *looks* as though in line 2758 of admin/modules/user/users.php, 'gid' should instead be 'usergroup', but I couldn't find a simple way to test this, so it remains "highly probable only".

Even less serious:

Lines 150 and 350 of admin/modules/user/banning.php: it appears as though these hooks should run before the calls to lift_ban()/insert_ban() respectively, but this is debatable.
Line 1999 of inc/datahandlers/user.php: it appears as though this hook should occur after the final update_query(), or before the insert_query() on line 1987, but again, this is debatable.

And even less serious:

Line 65 of inc/tasks/usercleanup.php: is this call to $cache->update_moderators() necessary, since that method is already called in the user data handler's lift_ban() method?
Line 3728 of modcp.php: $lifttime is not initialised to an array prior to this first reference to that variable. This does not appear to be an oversight on your part though, since the same is also the case on line 3917, which your PR appears not to touch.
Line 3754 of modcp.php: $banned is not initialised (to anything, let alone to an array, as it should be) prior to this - unless $errors evaluates to true.

OK, we're really lowering the bar for seriousness here:

In lines 1966, 1970, 2018, and 2022 of inc/datahandlers/user.php: "user" in these messages should probably instead be "ban". Also, perhaps these messages as a whole should be moved into the language file.

Starpaul20:
Fixed issues

lairdshaw:
Thanks. These changes work well, with one exception:

On line 3942 of `modcp.php`, `$banned` should not be set to an empty array, as it is already initialised earlier on (on line 3887 and potentially again on line 3904).

Also, further testing has revealed another minor issue - double-escaping of some values: elements of the array of data supplied to `$userhandler->set_data()` should not be escaped prior to supplying them to that function, because they are already escaped within the data handler prior to being inserted into / updated in the database via the handler's `insert_ban()` and `update_ban()` methods. Hopefully, the following list of lines from which escaping should be removed is complete, but I can't guarantee it:

`modcp.php`: 3667, 3668, 3669, 3836, 3837, 3838.
`admin/modules/user/banning.php`: 205, 206, 207, 333, 334, 335.
`admin/modules/user/user.php`: 2682, 2683, 2684, 2707, 2708, 2709.

After those fixes, we should be good to merge.

Starpaul20:
That should take care of it.

lairdshaw:
Yep, *almost*. I know this is probably annoying, but also in the removal of the initialisation of `$banned` on 3753 of `modcp.php`, which was the right thing to do and which I'd missed as necessary - good catch - that variable is now potentially again left uninitialised in some circumstances - namely, when `$errors` tests false on line 3713 above. How about we add an initialisation to an empty array somewhere above line 3713?

Starpaul20:
That should work.

lairdshaw:
Yes. It does. Unfortunately, I've picked up another need for correction:

In `modcp.php` on line 3952, the `usergroup` keys into `$banned` should instead be `gid` keys.

We'll get there! Hopefully, this is the last one.

Starpaul20:
Fixed! ;)

lairdshaw:
Thunderbirds are go. Merging...

PR #4491, opened 2021-11-07
Fix #4468 Collapse issues after rabase

### LINKED ISSUE #4468: Collapsible.init() should be expandables.init() in jscripts/general.js
This causes an error which can be seen in the console.

Probably, this was due to a mistake of mine during rebasing. I haven't investigated carefully as the fix is very straightforward.

Matslom:
Resolves #4468
Fixes collapses implemented in 1.9 theme.

There are some places in 1.8 theme that use collapses but they are not implemented in twig templates (search for eg. `$collapsedimg`). Idk if we will not use collapses in that places or they are waiting for implementation in templates. If not, we should clean up code from collapses.

### DIFF
--- a/global.php
+++ b/global.php
@@ -1078,26 +1078,17 @@ if($mybb->user['uid'] && is_banned_email($mybb->user['email']) && $mybb->setting
 
 // work out which items the user has collapsed
 $colcookie = '';
-if(!empty($mybb->cookies['collapsed']))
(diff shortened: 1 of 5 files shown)


lairdshaw:
Tested and seems to be working; a light review of the code checks out OK: LGTM and merging this PR.

PR #4493, opened 2021-11-08
Fixes #4492 - SCEditor returns unsupported mycode format...

### LINKED ISSUE #4492: SCEditor returns unsupported MyCode format for quotes with no username
Quoting a Guest, who didn't supply a username, may result in unparsed MyCode being displayed when using SCEditor.

When a `[quote="" ...]` MyCode tag with an empty author value and additional attributes is created in the [`parse_quoted_message()` function](https://github.com/mybb/mybb/blob/mybb_1829/inc/functions_posting.php#L176) and added to `<textarea>` elements, SCEditor may remove the  empty attribute code `=""` and output a `[quote ...]` tag, a format not supported by the MyBB parser.

Reported in:
- https://community.mybb.com/thread-214064.html
- https://community.mybb.com/thread-234164.html
- https://www.mybb.de/forum/thread-35184.html

Eldenroot:
Resolves #4492 

Fix by doylecc

### DIFF
--- a/inc/functions_posting.php
+++ b/inc/functions_posting.php
@@ -187,6 +187,13 @@ function parse_quoted_message(&$quoted_post, $remove_message_quotes=true)
 	{
 		$quoted_post['username'] = $quoted_post['userusername'];
 	}
+	else
+	{
+	if(empty($quoted_post['username']))
+	{
+		$quoted_post['username'] = htmlspecialchars_uni($lang->guest);
+	}
+}
 	// Clean up the message
 	$quoted_post['message'] = preg_replace(array(
 		'#(^|\r|\n)/me ([^\r\n<]*)#i',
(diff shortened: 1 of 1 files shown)


inc/functions_posting.php
```diff
@@ -187,6 +187,13 @@ function parse_quoted_message(&$quoted_post, $remove_message_quotes=true)
 	{
 		$quoted_post['username'] = $quoted_post['userusername'];
 	}
+	else
+	{
+	if(empty($quoted_post['username']))
```
dvz: Indentation here seems incorrect

dvz (changes requested):

dvz (approved):

PARADOXP (commented):
so what is the difference?

PR #4495, opened 2021-11-11
Fix #4489 broken attachments on 1.9

### LINKED ISSUE #4489: Attachments are not working on 1.9
There are multiple issues with attachments after rebase.
1. Broken attachment rendering in posts
```
<img src="attachment.php?aid=" class="attachment" alt="" title="Filename: 
Size: 
		">
```
2. Code that was introduced in 78d9d4ad833d6e10b89c81ac06d531925a160ab2 that need port to twig templates
3. Error codes are not show anywhere (eg. to big image size)

Matslom:
Resolves #4489

### DIFF
--- a/editpost.php
+++ b/editpost.php
@@ -215,20 +215,19 @@ if($mybb->settings['enableattachments'] == 1 && ($mybb->get_input('newattachment
 	{
 		if(isset($ret['success']))
 		{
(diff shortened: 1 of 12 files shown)


lairdshaw:
A fair bit of this PR seems to be reformatting and re-indenting, but hey, it seems to work "as advertised" given my light testing and light visual inspection of the code, and I'm grateful to @Matslom for providing it and putting the effort into getting it right, so I'm going to be bold enough, given the long delay, to merge it right now.

PR #4497, opened 2021-11-21
Fix #4166 Rework `show_fields_from()` for PostgreSQL and SQLite

### LINKED ISSUE #4166: DB show_fields_from() method's return varies in format for different db engines
The function show_fields_from($table) that shows the "show fields from" command for a specific table returns results in different formats for different db engines. It makes some particular uses of this function fail, for example, the merge system's column length calculation (https://github.com/mybb/merge-system/pull/260 ).

For mysql (now deprecated) and mysqli, the method directly uses SQL SHOW FIELDS FROM table_name to query, the results have columns (on MySQL 5.7) "Field", "Type", "Null", "Key", "Default", "Extra" and returns them.
mysql: https://github.com/mybb/mybb/blob/mybb_1824/inc/db_mysql.php#L1111
mysqli: https://github.com/mybb/mybb/blob/mybb_1824/inc/db_mysqli.php#L1105

For pqsql and sqlite, the method uses some different SQL queries to get an intermediate result and reformat it as the final results that have columns "Field", "Extra" and returns them. It's worth noting that the two methods have different results in the "Extra" field. For pgsql, the information of column types is totally forgotten.
pqsql: https://github.com/mybb/mybb/blob/mybb_1824/inc/db_pgsql.php#L1134
sqlite: https://github.com/mybb/mybb/blob/mybb_1824/inc/db_sqlite.php#L954

IMO we could use the skeleton of MySQL's SHOW FIELDS FROM's result to format results of pqsql and sqlite.

Original thread: [DB show_fields_from() method's return varies in format for different db engines](https://community.mybb.com/thread-229960.html)

yuliu:
Resolves #4166.
It's a work-in-progress PR but already committed changes can be reviewed. The remaining work may involve refine existing usings this function in MyBB.

In the PR, the function `show_fields_from()`'s returns for PostgreSQL and SQLite are made to have the same array keys as MySQL's, and with the following changes:
 - PostgreSQL's are fetched from `information_schema`
 - SQLite's are fetched using `PRAGMA table_info()`, primary key and auto_increment info are determined by `INTEGER PRIMARY KEY` type.
 - PostgreSQL's character varying /bit types will further have their length limit `n` appended to their types with parenthesises `(n)`.
 - PostgreSQL's numeric/decimal types will further have their precision `p` and scale `s` appended to their types with parenthesises `(p,s)`.

**Notes:**
 - The function doesn't return unified data types and default values across database engines. It just returns results from PostgreSQL's `information_schema` and SQLite's `PRAGMA table_info()` as  stated above. For example, for `VARCHAR` type, MySQL's may return `varchar` and PostgreSQL's may return `character varying`.
 - SQLite's returned data types are the same as how the `CREATE TABLE` statement defined, not the data type class or affinity explained [here](https://sqlite.org/datatype3.html).

### DIFF
--- a/inc/db_pgsql.php
+++ b/inc/db_pgsql.php
@@ -1146,7 +1146,7 @@ class DB_PgSQL implements DB_Base
 		$primary_key = $this->fetch_field($query, 'column_name');
 
 		$query = $this->write_query("
-			SELECT column_name, data_type, is_nullable, column_default, character_maximum_length
+			SELECT column_name, data_type, is_nullable, column_default, character_maximum_length, numeric_precision, numeric_precision_radix, numeric_scale
 			FROM information_schema.columns
 			WHERE table_name = '{$this->table_prefix}{$table}'
 		");
@@ -1164,9 +1164,15 @@ class DB_PgSQL implements DB_Base
 				$field['_extra'] = '';
 			}
 
+			// bit, character, text fields.
 			if(!is_null($field['character_maximum_length']))
 			{
 				$field['data_type'] .= '('.(int)$field['character_maximum_length'].')';
+			}
+			// numeric/decimal fields.
+			else if($field['numeric_precision_radix'] == 10)
+			{
+				$field['data_type'] .= '('.(int)$field['numeric_precision'].','.(int)$field['numeric_scale'].')';
 			}
 
 			$field_info[] = array(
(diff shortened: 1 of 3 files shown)


yuliu:
Any using this function in MyBB seems working with the PR.

The post handler uses it for determine post message length limitation which won't have any problem with the PR:
https://github.com/mybb/mybb/blob/89ba6fd39fca4ede328f69deca9aba7b57f2c5d1/inc/datahandlers/post.php#L263-L277

though it can be improved to be more PostgreSQL or SQLite, whose `TEXT` length limit is *unlimited**.

I'll set the PR to ready for review.

cronlabspl (approved):
LGTM

dvz (approved):

PR #4499, opened 2021-11-30
Fixes SQLite's `optmize_table()` by removing the function body

### LINKED ISSUE #4498: SQLite's `optimize_table()`
It looks like SQLite's optimization works on the database (schema) level but not the table level.

Currently, MyBB's using VACUUM is working on tables rather than databases/schemas, which is wrong: https://github.com/mybb/mybb/blob/mybb_1829/inc/db_sqlite.php#L923

And there's more. This function does fail when the VACUUM query tries to closeCursor() in https://github.com/mybb/mybb/blob/mybb_1829/inc/db_sqlite.php#L924 . 
Actually the error was from the VACUUM itself. We can read the error using PDO's errorInfo() and the error message states "cannot VACUUM - SQL statements in progress".

Then, there are two things we can get:
1. SQLite's VACUUM works on database (schema), not current MyBB's using on table.
2. As there would be any uncontrollable query [b]without  running before the DB class execute VACUUM on the database, there would (nearly) always that error preventing the db engine to optimize the database.

So I suggest to just remove the function body in SQLite and leave out SQLite's database optimization.

For future development, we could start thinking about SQLite's [`VACUUM [...] INTO ...`](https://www.sqlite.org/lang_vacuum.html#vacuum_with_an_into_clause) to optimize the database and generate a new db file and replace the old db file with it.

My old suggestion was: since the function is intended to run over tables, we may pass SQLite's implementation for it and add another optimize_db() function dedicated to SQLite for maintenance usage, such as Database Optimization and Database Backup. However, currently MyBB doesn't have database information such as database name stored in the DB object, it wouldn't be easy to implement such optimize_db() function. It'd be better to defer this part to 1.9.

Original thread: [SQLite's `optimize_table()`](https://community.mybb.com/thread-234290.html)

yuliu:
Resolves #4498

### DIFF
--- a/inc/db_sqlite.php
+++ b/inc/db_sqlite.php
@@ -920,8 +920,7 @@ class DB_SQLite implements DB_Base
 	 */
 	function optimize_table($table)
 	{
-		$query = $this->query("VACUUM ".$this->table_prefix.$table."");
-		$query->closeCursor();
+		// SQLite doesn't support table level optimization. TODO: implement an optimize_database() function using query `VACUUM $db_name`.
 	}
 
 	/**
--- a/inc/db_sqlite.php
+++ b/inc/db_sqlite.php
@@ -920,7 +920,8 @@ class DB_SQLite implements DB_Base
 	 */
 	function optimize_table($table)
 	{
-		// SQLite doesn't support table level optimization. TODO: implement an optimize_database() function using query `VACUUM $db_name`.
+		// SQLite doesn't support table level optimization.
+		// Using `VACUUM [main | $db_name]` may also be blocked by any opened query cursor, hence generating an error.
 	}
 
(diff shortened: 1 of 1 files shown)


inc/db_sqlite.php
```diff
@@ -920,8 +920,8 @@ function get_version()
 	 */
 	function optimize_table($table)
 	{
-		$query = $this->query("VACUUM ".$this->table_prefix.$table."");
-		$query->closeCursor();
+		// SQLite doesn't support table level optimization.
```
cronlabspl, line 923: Value of this change?

cronlabspl (commented):
Am unable to see real value of this change. @yuliu will you enlight me?

PR #4503, opened 2021-12-22
Fixes #4502 Sequence dropping in `drop_table()` of PostgreSQL and table prefix in `drop_table()` of PostgreSQL and SQLite

### LINKED ISSUE #4502: `drop_table()` in PostgreSQL and SQLite may result in errors
The primary part of DROP TABLE should work fine but the following DROP SEQUENCE part may not work:

https://github.com/mybb/mybb/blob/mybb_1...1286-L1293
https://github.com/mybb/mybb/blob/mybb_1...#L482-L487

As in the above code, there might be at least two issues:
1. The variable $table in the query doesn't have $table_prefix.
2. The query SQL to DROP SEQUENCE is using single quotes so the variables in it won't get expanded.
3. The query SQL to find any sequences that are still linked to the current table is wrong. The view that has sequence information is information_schema.sequences but not information_schema.constraint_column_usage which is used for constraints and keys.

About the sequence's add/drop:
 - In PostgreSQL 8.2 and later, CREATE SEQUENCE ... OWNED BY ... is allowed. Thus, when a table is dropped any sequences owned by it will also get dropped, created by either the serial data type or CREATE SEQUENCE ... OWNED BY ....
 - But in PostgreSQL 8.1 and previous versions, the OWNED BY clause hasn't been supported yet. ~~MyBB officially supports PostgreSQL &gt;= 8.1, so we have to maintain this code.~~

Original thread: [PostgreSQL's `drop_table()` may result in errors](https://community.mybb.com/thread-234455.html)

yuliu:
Resolves #4502.

**Note:**
1. `DROP SEQUENCE IF EXISTS` only exists in PostgreSQL 8.2 and higher. When MyBB ever bumps the required PostgreSQL version, this version compare should be changed or removed.
2. The `DROP SEQUENCE` part may be removed in future versions, eventually, when the `mybb_userfields` table related operations are fixed.

### DIFF
--- a/inc/db_pgsql.php
+++ b/inc/db_pgsql.php
@@ -1272,7 +1272,7 @@ class DB_PgSQL implements DB_Base
 		}
 
 		$table_prefix_bak = $this->table_prefix;
-		$this->table_prefix = $table_prefix;
+		$this->table_prefix = '';
 		$fields = array_column($this->show_fields_from($table_prefix.$table), 'Field');
 
 		if($hard == false)
--- a/inc/db_pgsql.php
+++ b/inc/db_pgsql.php
@@ -1271,9 +1271,13 @@ class DB_PgSQL implements DB_Base
 			$table_prefix = $this->table_prefix;
 		}
 
+		$table_prefix_bak = $this->table_prefix;
+		$this->table_prefix = $table_prefix;
+		$fields = array_column($this->show_fields_from($table_prefix.$table), 'Field');
+
 		if($hard == false)
 		{
-			if($this->table_exists($table))
+			if($this->table_exists($table_prefix.$table))
 			{
 				$this->write_query('DROP TABLE '.$table_prefix.$table);
 			}
@@ -1283,13 +1287,30 @@ class DB_PgSQL implements DB_Base
 			$this->write_query('DROP TABLE '.$table_prefix.$table);
 		}
 
-		$query = $this->query("SELECT column_name FROM information_schema.constraint_column_usage WHERE table_name = '{$table}' and constraint_name = '{$table}_pkey' LIMIT 1");
-		$field = $this->fetch_field($query, 'column_name');
+		$this->table_prefix = $table_prefix_bak;
 
-		// Do we not have a primary field?
-		if($field)
+		if(!empty($fields))
 		{
-			$this->write_query('DROP SEQUENCE {$table}_{$field}_id_seq');
+			foreach($fields as &$field)
+			{
+				$field = "{$table_prefix}{$table}_{$field}_seq";
+			}
+			unset($field);
+
+			if(version_compare($this->get_version(), '8.2.0', '>='))
+			{
+				$fields = implode(', ', $fields);
+				$this->write_query("DROP SEQUENCE IF EXISTS {$fields}");
+			}
+			else
+			{
+				$fields = "'".implode("', '", $fields)."'";
+				$query = $this->query("SELECT sequence_name as field FROM information_schema.sequences WHERE sequence_name in ({$fields}) AND sequence_schema = 'public'");
+				while($row = $this->fetch_array($query))
+				{
+					$this->write_query("DROP SEQUENCE {$row['field']}");
+				}
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


inc/db_pgsql.php
```diff
@@ -1271,9 +1271,13 @@ function drop_table($table, $hard=false, $table_prefix=true)
 			$table_prefix = $this->table_prefix;
 		}
 
+		$table_prefix_bak = $this->table_prefix;
+		$this->table_prefix = '';
```
Xazin, line 1275: What's the reason for doing this? I see on line 1290 that you assign it `$this->table_prefix = $table_prefix_bak` without accessing it at all inbetween line 1274 to 1290? You don't change `$table_prefix_bak` either after assigning `$this->table_prefix` to it.

You can remove this line, and then you can remove line 1290, at least it seems very redundant, maybe I am wrong, what's the reasoning?

inc/db_pgsql_pdo.php
```diff
@@ -470,20 +470,37 @@ public function drop_table($table, $hard = false, $table_prefix = true)
...
 			{
 				$this->write_query("DROP TABLE {$table_prefix}{$table}");
 			}
 		} else {
 			$this->write_query("DROP TABLE {$table_prefix}{$table}");
 		}
 
-		$query = $this->query("SELECT column_name FROM information_schema.constraint_column_usage WHERE table_name = '{$table}' and constraint_name = '{$table}_pkey' LIMIT 1");
-		$field = $this->fetch_field($query, 'column_name');
+		$this->table_prefix = $table_prefix_bak;
```
Xazin, line 486: I don't see the need to undo changes that haven't occurred?

inc/db_pgsql.php
yuliu: `table_exists()` uses `$this->table_prefix`, but yeah the code looks a bit redundant and confusing.

Xazin: It indeed does use `$this->table_prefix` but it does not change the value of it.

yuliu: The function `drop_table()` permits dropping any tables with any prefix. So you could write this part as either:
```php
$table_prefix_bak = $this->table_prefix;
$this->table_prefix = '';
// ...
$this->table_exists($table_prefix.$table);
// ...
$this->table_prefix = $table_prefix_bak;
```
or
```php
$table_prefix_bak = $this->table_prefix;
$this->table_prefix = $table_prefix;
// ...
$this->table_exists($table);
// ...
$this->table_prefix = $table_prefix_bak;
```

Xazin: What stops you from doing
```
$this->table_exists($table_prefix.$table);
```
Without any need to store `$this->table_prefix` in a temporary variable?

yuliu: Any non-empty `$table_prefix`.

yuliu: @Xazin that means:

 - `table_exists($table_prefix.$table)` will check the existence of table of name `$table_prefix.$this->table_prefix.$table`, so the first way sets `$this->table_prefix` to `''`.
 - `table_exists($table)` will check the table of name `$this->table_prefix.$table`, so the second way sets `$this->table_prefix` to `$table_prefix`.

and both of them save the original `$this->table_prefix` as `$table_prefix_bak` and restore it afterwards.

I don't know what confuses you exactly, or what is the way you'd like to do about this part?

Xazin: It is not that I am confused about the implementation, it is because I do not like this way of working around it. 

Optimally, make it so that you do not need to manipulate the table prefix so that it is very clear what is occurring in the code. But It is not up to me, so you will not have to make any changes.

I would advise though to leave comments there because what is happening does not give itself away easily just by reading the code. 

(Just my opinions - You do not have to mind it too much)

dvz: Other solutions that may be acceptable:
- add similar `bool $table_prefix` support to `DB_Base::table_exists()`
- remove `$table_prefix` support altogether (appears to be only used in `upgrade1.php`)

yuliu:
The latest commit fixes table prefix alteration in SQLite's `drop_table()`. I don't test this commit though.

PR #4514, opened 2022-01-16
Restore check for existing URLs in `postParser::mycode_auto_url_callback()`

dvz:
Restores a check for existing URLs returned by the calling function that are supposed to be skipped by the auto-link mechanism.

Likely a rebase regression.


lairdshaw:
Haven't looked into the origins of this but yes, it seems likely to be a rebase regression. Merging given that it's then pretty much a no-brainer.

PR #4532, opened 2022-04-04
Fix unable to  Simple search : unable to search some specific string …

BettyIsBoop:
Resolves 235334
 https://community.mybb.com/thread-235334.html


dvz:
Looks like this will need to be re-applied against the `feature` branch manually (`master` currently contains the old 1.6 series).

PR #4556, opened 2022-07-11
Lang Files - NewThread and NewReply Inconsistencies

### LINKED ISSUE #4553: Inconsistent Formatting & Semantics in Language Files
https://github.com/mybb/mybb/blob/f103d9adf6e938828433ac81dda0dfa6b7c4de26/inc/languages/english/newreply.lang.php#L22
https://github.com/mybb/mybb/blob/f103d9adf6e938828433ac81dda0dfa6b7c4de26/inc/languages/english/newthread.lang.php#L27
Colon should be inside `<strong>` and `<b>` tags to match with other lines in those files. ~~Super minor nit-picking, I know.~~

Additionally across lang files consider changing instances of `<b>` tags to `<strong>` (or vice versa). These tags have been made consistent within individual files, but there is a semantic difference between the two. Visually they are the same, but `<strong>` tags are emphasized *semantically* as important (ie: to screen readers) whereas `<b>` tags are solely visual. It's a small change code-wise but has an impact on accessibility. Which pieces of text should just be visually bold for sighted users? Which should be defined as semantically important? The same applies to the difference between `<i>` and `<em>` tags as well.

kjSagee:
Resolves #4553 

Fixed minor typo/inconsistent formatting.

Left bold/strong tags as-is for now because I don't have the time to do a thorough read-through of the lang files. See note in issue about semantics of `<b>` vs `<strong>`


Eldenroot:
Better to fix all in one PR...

PR #4574, opened 2022-08-26
PHP 8 -  `undefined` errors

RevertIT-alt:
Resolves `smilies.php`
Fix `@opendir` logic

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -23,6 +23,24 @@ Learn more on the MyBB development page:
 
 If you need any help with sending your code contributions, the [GitHub Help site](https://docs.github.com/en/github) is a good place to start. You can also get in touch with MyBB developers and the Community through our [support channels](https://mybb.com/support/).
 
+## Running MyBB
+
+MyBB requires a web server with PHP support and a database. [MyBB Documentation &mdash; Requirements &rarr;](https://docs.mybb.com/1.8/install/requirements/)
+
+If you want to use a manually-managed stack, copy code from the desired repository branch to your web server's HTTP-exposed directory.
+- Code for MyBB &le; 1.8 can be run as-is.
+- Code for MyBB &ge; 1.9 references external dependencies, installed using [Composer](https://getcomposer.org/) (`composer install`).
+
+To quickly setup a **self-contained development stack** with PHP + Xdebug + Composer auto-installation, web server, and database system(s), use [Docker](https://www.docker.com/products/docker-desktop/) Compose:
+
+```sh
+$ git clone https://github.com/mybb/mybb.git
+$ git clone https://github.com/mybb/deploy.git
+$ cd deploy
+$ docker compose up
+```
+[Using MyBB's Development Stack &rarr;](https://github.com/mybb/deploy/)
+
 ## MyBB on GitHub
 
 ### Branches
--- a/.github/README.md
+++ b/.github/README.md
@@ -21,7 +21,7 @@ Visit the [MyBB Community Forums](https://community.mybb.com) and join the [MyBB
 View [SECURITY.md](https://github.com/mybb/mybb/blob/feature/.github/SECURITY.md) for information on secure usage, incident recovery, and vulnerability reporting. To learn more about security research at MyBB, go to [**mybb.com/security**](https://mybb.com/get-involved/security/).
 
 ## Get Involved
-View [CONTRIBUTING.md](https://github.com/mybb/mybb/blob/feature/.github/CONTRIBUTING.md) for information on submitting changes, workflow, and best practices for working with the code repository. To learn more about contributing to the MyBB Project, go to [**mybb.com/get-involved**](https://mybb.com/get-involved/).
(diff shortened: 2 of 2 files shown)


dvz:
Can you attach/describe the errors that may appear before these changes?

RevertIT-alt:
Sure.

For example in input field enter "/images/smilies/" in "Mass Edit" under smilies in ACP.

You will get XX errors of "Undefined array key "/images/smilies/NAME_OF_SMILIE.png" - Line: 357"

This will fix this issue for smilies and prevent opendir before checking if directory actually exists. (For example someone deletes folder)

admin/modules/config/plugins.php
```diff
@@ -600,7 +600,8 @@
 function get_plugins_list()
 {
 	// Get a list of the plugin files which exist in the plugins directory
-	$dir = @opendir(MYBB_ROOT."inc/plugins/");
+	$dir = is_dir(MYBB_ROOT."inc/plugins/") && is_readable(MYBB_ROOT."inc/plugins/") ? opendir(MYBB_ROOT."inc/plugins/") : null;
```
dvz: It looks like there are a few places in the source where `opendir()` is called without additional checks nor the `@` operator. What kinds of errors occur with `@opendir()`?

admin/modules/config/post_icons.php
```diff
@@ -34,7 +36,7 @@
 			$errors[] = $lang->error_missing_path;
 		}
 
-		if(!$errors)
+		if(empty($errors))
```
dvz: These checks may not be needed as long as `$errors` was initialized

admin/modules/tools/tasks.php
```diff
@@ -87,21 +87,21 @@ function check_time_values($value, $min, $max, $return_type)
 			$errors[] = $lang->error_invalid_task_file;
 		}
 
-		$mybb->input['minute'] = check_time_values($mybb->input['minute'], 0, 59, 'string');
+		$mybb->input['minute'] = isset($mybb->input['minute']) ? check_time_values($mybb->input['minute'], 0, 59, 'string') : false;
```
dvz: Instead of a ternary operator, `$mybb->input` usage could be replaced with `$mybb->get_input()` that does the necessary checks and returns a default value:
https://github.com/mybb/mybb/blob/0bf8fdb90008c82e20cd6350ce74928e08a6bdda/inc/class_core.php#L464

admin/modules/tools/tasks.php
```diff
@@ -87,21 +87,21 @@ function check_time_values($value, $min, $max, $return_type)
...
 		}
 
-		$mybb->input['hour'] = check_time_values($mybb->input['hour'], 0, 59, 'string');
+		$mybb->input['hour'] = isset($mybb->input['hour']) ? check_time_values($mybb->input['hour'], 0, 59, 'string') : false;
 		if($mybb->input['hour'] === false)
 		{
 			$errors[] = $lang->error_invalid_hour;
 		}
 
 		if($mybb->input['day'] != "*" && $mybb->input['day'] != '')
```
dvz: If `day` is not set, this line may result in an error before the check below

admin/modules/tools/tasks.php
```diff
@@ -393,8 +393,8 @@ function check_time_values($value, $min, $max, $return_type)
 
 	$form_container = new FormContainer($lang->edit_task);
 	echo $form->generate_hidden_field("tid", $task['tid']);
-	$form_container->output_row($lang->title." <em>*</em>", "", $form->generate_text_box('title', $task_data['title'], array('id' => 'title')), 'title');
-	$form_container->output_row($lang->short_description." <em>*</em>", "", $form->generate_text_box('description', $task_data['description'], array('id' => 'description')), 'description');
+	$form_container->output_row($lang->title." <em>*</em>", "", $form->generate_text_box('title', $task_data['title'] ?? null, array('id' => 'title')), 'title');
+	$form_container->output_row($lang->short_description." <em>*</em>", "", $form->generate_text_box('description', $task_data['description'] ?? null, array('id' => 'description')), 'description');
```
dvz: This operator was [added in PHP 7.0](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op), so we'd need to use something else (e.g. assign all expected `$task_data` values individually using `$mybb->get_input()` instead of `$task_data = $mybb->input;`)

(one of the reasons why the lint test is failing)

admin/modules/config/post_icons.php
```diff
@@ -114,7 +116,8 @@
 			}
 
 			$path = $mybb->input['pathfolder'];
-			$dir = @opendir(MYBB_ROOT.$path);
+			$dir = !empty(MYBB_ROOT.$path) && is_dir(MYBB_ROOT.$path) && is_readable(MYBB_ROOT.$path) ? opendir(MYBB_ROOT.$path) : null;
```
dvz: (`empty()` doesn't seem to support non-variable expressions on PHP <= 5.5 - one of the reasons why the lint test is failing)

PR #4596, opened 2022-10-15
Fix #3793: Caching fallback breaks when external cache system has problems

### LINKED ISSUE #3793: Caching fallback breaks when external cache system has problems

lairdshaw:
Resolves #3793.

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -2392,295 +2392,33 @@ function my_set_array_cookie($name, $id, $value, $expires="")
 	$mybb->cookies['mybb'][$name] = $newcookie;
 }
 
-/*
- * Arbitrary limits for _safe_unserialize()
- */
-define('MAX_SERIALIZED_INPUT_LENGTH', 10240);
-define('MAX_SERIALIZED_ARRAY_LENGTH', 256);
-define('MAX_SERIALIZED_ARRAY_DEPTH', 5);
-
 /**
- * Credits go to https://github.com/piwik
- * Safe unserialize() replacement
- * - accepts a strict subset of PHP's native my_serialized representation
- * - does not unserialize objects
+ * Simply wrap around PHP's default unserialize() function, disallowing the
+ * unserialisation of objects.
  *
- * @param string $str
- * @return mixed
- * @throw Exception if $str is malformed or contains unsupported types (e.g., resources, objects)
- */
-function _safe_unserialize($str)
-{
-	if(strlen($str) > MAX_SERIALIZED_INPUT_LENGTH)
-	{
-		// input exceeds MAX_SERIALIZED_INPUT_LENGTH
-		return false;
-	}
-
-	if(empty($str) || !is_string($str))
-	{
-		return false;
-	}
-
-	$stack = $list = $expected = array();
-
-	/*
-	 * states:
-	 *   0 - initial state, expecting a single value or array
-	 *   1 - terminal state
-	 *   2 - in array, expecting end of array or a key
-	 *   3 - in array, expecting value or another array
-	 */
-	$state = 0;
-	while($state != 1)
-	{
-		$type = isset($str[0]) ? $str[0] : '';
-
-		if($type == '}')
-		{
-			$str = substr($str, 1);
-		}
-		else if($type == 'N' && $str[1] == ';')
-		{
-			$value = null;
-			$str = substr($str, 2);
-		}
-		else if($type == 'b' && preg_match('/^b:([01]);/', $str, $matches))
-		{
-			$value = $matches[1] == '1' ? true : false;
-			$str = substr($str, 4);
-		}
-		else if($type == 'i' && preg_match('/^i:(-?[0-9]+);(.*)/s', $str, $matches))
-		{
-			$value = (int)$matches[1];
-			$str = $matches[2];
-		}
-		else if($type == 'd' && preg_match('/^d:(-?[0-9]+\.?[0-9]*(E[+-][0-9]+)?);(.*)/s', $str, $matches))
-		{
-			$value = (float)$matches[1];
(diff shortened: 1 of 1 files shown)


lairdshaw:
Superseded by #4608.

Issue #3689, opened 2019-06-04
1.9 Theme System


dvz:
MyBB 1.8 themes' information, contained in XML files, are expected to be split into the filesystem directly (templates, stylesheets, and additional assets) and a `package.json` manifest file with properties and metadata.

- What **theme properties** should the manifest file include?
- What **metadata** should the manifest file include?
- How should **theme installation & modification** work (Composer, CSS pre-processors)?
- How should **MyBB compatibility updates** work?
- How should **stylesheet information** be stored?
- Hoe should the new system handle **inheritance**? How should that information be stored?
- What system could be implemented for applying **theme updates**?
- What level of support should be left for **1.8 templates & stylesheet** to maintain 1.8 extension compatibility? How?

euantorano:
We have also had some discussion about managing JavaScript scripts related to themes. The idea was that we'd have a JS manager similar to the stylesheet manager with some of the following features:

- JS scripts could have dependencies (eg: `a.js` depends on `jQuery.js` so we ensure `jQuery.js` is loaded)
- JS scripts could be combined and minified, then cached
- JS scripts would have cache busting when using combined and minified versions

euantorano:
Regarding theme installation, I would love to package themes as Composer packages so that they can be installed that way as well as via the standard upload procedure (including simply uploading a ZIP file to the ACP I hope).

Composer's JSON schema already has a [`type`](https://getcomposer.org/doc/04-schema.md#type) property. I imagine us having our own defined types (`mybb-theme`, `mybb-plugin`, `mybb-translation`).

Composer's schema also has an [`extra`](https://getcomposer.org/doc/04-schema.md#extra) feels that allows other arbitrary entries. Given that it contains other fields we'd want to show for themes (author, name, license, URL, etc. I've considered using the standard `composer.json` as the actual theme manifest. We'd define some standard `extra` fields that we understand, and otherwise use the standard fields.

WildcardSearch:
As to theme info, I would start by drawing a direct comparison from what we are doing now. As it stands, the inheritance system that is currently in place works fairly well.

The main issue with the current system that I see in practice is that upgrades require quite a bit of knowledge from admins because of how updates are handled. If there was a way to improve this process, that would be a welcome change.

I'm currently in the process of mimicking the existing inheritance pattern for a plugin that needs a similar system. I honestly haven't found any ways to improve the current setup, but I am still looking into it.

[...]

Issue #552, opened 2014-05-27
Quick Reply not working if captcha is present


JN-Jones:
If the captcha (default or recaptcha) is present in the quick reply form (guests can post) they can't submit the post, only preview it.

PR #4601, opened 2022-11-23
Develop the 1.9 theme and plugin system

lairdshaw:
This is a WIP but already with many features and commits.

See the individual commits for the implemented features.

I'm happy though (and hope) for it to be merged ASAP by anybody who thinks it's worth doing that.

The two major features lacking at time of initial PR are:

1. The equivalent of 1.8's find_replace_templatesets() for plugins (on which I've made a decent start but on which I've also not yet reached the point of committing let alone pushing anything).
2.  Tracking and resolution of conflicts in updated resources, especially templates and stylesheets - the equivalent of 1.8's "Find Updated Templates".

Related issue: #3689


admin/modules/config/plugins.php
```diff
@@ -415,22 +544,52 @@
...
+
+	if($do_upgrade || $action == "activate")
+	{
+		if(!$do_upgrade)
+		{
+			$message = $lang->success_plugin_activated;
+		} else {
+			// Plugin is compatible with this version?
+			$plugininfo = read_json_file(MYBB_ROOT."inc/plugins/{$codename}/plugin.json");
+			if (empty($plugininfo['compatibility']) || $plugins->is_compatible($plugininfo['compatibility']) == false) {
```
Sama34, line 582: I think this check is done above (#552) for upgrades

admin/modules/config/plugins.php
```diff
@@ -517,36 +680,53 @@
 
 	$plugins->run_hooks("admin_config_plugins_plugin_list");
 
+	$form = new Form('index.php?module=config-plugins&amp;action=upload', 'post', '', 1);
```
Sama34, line 683: Was a permission for uploading plugins added ?

It probably should and a separated page for this should be added.

admin/modules/config/plugins.php
```diff
@@ -517,36 +680,53 @@
...
-			if(!function_exists($infofunc))
+			foreach($plugins_list as $plugin_file)
 			{
-				continue;
-			}
-
-			$plugininfo = $infofunc();
-			$plugininfo['codename'] = $codename;
+				$codename = str_replace('.php', '', $plugin_file);
+				$plugininfo = read_json_file(MYBB_ROOT."inc/plugins/{$codename}/plugin.json", $err_msg, /*$show_errs = */false);
```
Sama34, line 703: I thought only template updates were going to be the major requirement for 1.8 plugins to work on 1.9 but it seems that is no longer the case ?

Are required changes documented somewhere or somehow ?

admin/modules/style/themes.php
```diff
@@ -2047,188 +1910,184 @@ function checkAction(id)
...
-				$css_to_insert .= "{$field}: {$value};\n";
-			}
-		}
+	if($mybb->request_method == "post")
+	{
+		$err_msg = false;
+		if (!is_mutable_theme($codename, $mybb->settings['themelet_dev_mode'])) {
+			$err_msg = $lang->error_immutable_theme;
+		} else {
+			// TODO: Abstract this code given that it is duplicated for saving in "advanced" mode below
```
Sama34, line 1969: Maybe abstract before merging ?

admin/modules/style/themes.php
```diff
@@ -2312,86 +2173,72 @@ function checkAction(id)
...
-	$query = $db->simple_select("themestylesheets", "*", "name='".$db->escape_string($mybb->input['file'])."' AND tid IN ({$parent_list})", array('order_by' => 'tid', 'order_dir' => 'desc', 'limit' => 1));
-	$stylesheet = $db->fetch_array($query);
-
-	// Does the theme not exist?
-	if($db->num_rows($query) == 0)
-	{
-		flash_message($lang->error_invalid_stylesheet, 'error');
-		admin_redirect("index.php?module=style-themes");
-	}
+	// Possible TODO: Implement support on this same page for editing of any SCSS module
```
Sama34, line 2180: We should document these TODOs for later.

global.php
```diff
@@ -123,7 +122,7 @@
 
 	$user = array(
 		'uid' => $mybb->user['uid'],
-		'style' => $mybb->get_input('theme', MyBB::INPUT_INT),
+		'style' => $mybb->input['theme'],
```
Sama34, line 125: Is this already sanitized ? I see you removed some sanitizing here.

inc/class_language.php
```diff
@@ -134,14 +134,32 @@ function set_language($language="", $area="user")
...
+	/**
+	 * A wrapper around load() with plugin-friendly argument order.
+	 *
+	 * @param string  $section       The section name.
+	 * @param string  $plugin_code   The codename of the plugin if we are loading one of a plugin's sections.
+	 *                               If not supplied, it is assumed to be identical to $section.
+	 * @param boolean $forceuserarea Should use the user area even if in admin? For example for datahandlers
+	 * @param boolean $supress_error supress the error if the file doesn't exist?
+	 */
+	function loadp($section, $plugin_code='', $forceuserarea=false, $supress_error=false)
```
Sama34, line 146: Is this really necessary ? It is probably the method names that bugs me the most.

inc/functions.php
```diff
@@ -9091,3 +9238,387 @@ function mk_path_abs($path, $base = MYBB_ROOT)
...
+	foreach ($moved as $from => $to) {
+		if ($del_source) rename($to, $from);
+	}
+
+	// Earlier return possible
+	return false;
+}
+
+// Copied from https://stackoverflow.com/a/15111679 with edits for coding style
+// TODO: Try to reconcile this with the preexisting similar function my_rmdir_recursive() above,
```
Sama34, line 9409: Maybe do this before merging so that core function doesn't lose meaning

inc/functions_themes.php
```diff
@@ -0,0 +1,1042 @@
...
+ * License: http://www.mybb.com/about/license
+ *
+ */
+
+use ScssPhp\ScssPhp\Compiler;
+
+/**
+ * Gets the hierarchy of themelets present in the filesystem, including their properties.
+ *
+ * @return array The themelet hierarchy where the first index is type ('themes' or 'plugins').
```
Sama34, line 17: Wouldn't be better to separate functions/methods/classes between themes and plugins ? Even if the happen to interreference commonly.

For instance, the plugins class already exists.

inc/functions_themes.php
```diff
@@ -0,0 +1,1042 @@
...
+ *                                          CSS (when false).
+ *                    `inheritance_chain` => an array of themelets in ascendant order of inheritance,
+ *                                           terminating in the themelet to which the resource
+ *                                           resolves. Each entry in the array is an array indexed
+ *                                           by:
+ *                      `codename`  => string, the codename of the theme/plugin.
+ *                      `is_plugin` => boolean, indicating whether or not the codename refers to a
+ *                                              plugin (otherwise, it resolves to a theme).
+ */
+define('RTR_RETURN_PATH'       , 1);
```
Sama34, line 720: Probably should move these constants to the top.

inc/functions_user.php
```diff
@@ -255,9 +255,17 @@ function verify_user_password($user, $password)
 		/** @var Illuminate\Contracts\Hashing\Hasher $hashDriver */
 		$hashDriver = \MyBB\app('hash')->driver($parameters['user']['password_algorithm']);
 
-		return $hashDriver->check($parameters['password'], $parameters['user']['password'], [
+		$result = $hashDriver->check($parameters['password'], $parameters['user']['password'], [
 			'salt' => $user['salt'],
 		]);
+
+		if (!$result && user_password_needs_rehash($user)) {
```
Sama34, line 262: What is the relation of this with themes/templates ? I presume you are fixing a non-related issue ?

inc/plugins/hello/hello.php
```diff
@@ -326,13 +159,13 @@ function hello_activate()
 	require_once MYBB_ROOT.'inc/adminfunctions_templates.php';
 
 	// Edit the index template and add our variable to above {$forums}
-	find_replace_templatesets('index', '#'.preg_quote('{$forums}').'#', "{\$hello}\n{\$forums}");
+// 	find_replace_templatesets('index', '#'.preg_quote('{$forums}').'#', "{\$hello}\n{\$forums}");
```
Sama34, line 162: Shouldn't this be deleted ?

inc/themes/core.default/frontend/styles/LICENCE
```diff
@@ -0,0 +1,31 @@
+[This licence applies to the *.scss files in this directory and its subdirectories]
+
+BSD 3-Clause License
```
Sama34, line 3: Why do we use a different license here ?

install/common.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+function clone_and_archive_default_theme()
```
Sama34, line 3: Why do we want to keep a copy of the default theme ?

install/index.php
```diff
@@ -1767,10 +1772,29 @@ function insert_templates()
...
+
+	echo $lang->theme_step_imported;
+	$output->print_footer('configuration');
+
+	return;
+
+
+
+
+	/** TODO: Migrate as necessary to 1.9 any of the below 1.8 code (now being
```
Sama34, line 1790: This is probable a cleanup to do before merging.

install/resources/settings.xml
```diff
@@ -303,10 +303,18 @@ min=0]]></optionscode>
 			<settingvalue><![CDATA[0]]></settingvalue>
 			<isdefault>1</isdefault>
 		</setting>
+		<setting name="themelet_dev_mode">
+			<title>Development Mode (Themelets)</title>
```
Sama34, line 307: Hiding this to the file level has been discussed ? We probably don't want production users to mess with the development features that easily.

resource.php
```diff
@@ -0,0 +1,58 @@
...
+define('IN_MYBB', 1);
+
+define('NO_ONLINE', 1);
+
+define('THIS_SCRIPT', 'resource.php');
+
+// We'd prefer not to load this file, and instead load inc/init.php alone, because it is more
+// lightweight, but we need it for its resolution of the current theme for the current user, which
+// it sets $theme to, and which is used within resolve_themelet_resource().
+require_once './global.php';
```
Sama34, line 20: I'm thinking on the 1.8 context here, so might be wrong, but this might be too heavy to load for each resource, as far as I understand the below code and this file's purpose.

dvz:
# Miscellaneous Changes
- [x] ~~Old template system removals~~
  merged via #4603
- [ ] `undefined` PHP 8 fixes (i.a. `global.php`)
- [ ] password handling
- [ ] 1.9 update file for DB structure changes
- [ ] _Hello World_ plugin
  #4614
- [ ] `admin/modules/*/module_meta.php` changes
- [ ] `inc/plugins/[codename]/languages/*` support
- [ ] install refactoring (`my_is_dir_writable()`)
- [ ] invisible character changes 

# Theme System Implementation
This is an implementation of the [MyBB 1.9 Theme System specification](https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md).

## Code Changes Overview
- helper functions (JSON read/write, copying directories, zip)
- Theme System internals
  - file data interpretation (equivalent to DB queries)
  - file & Resource path resolution
  - inheritance resolution
  - intermediate rendering functions (e.g. `\MyBB\template()`)
  - Twig, scssphp execution
  - core callback in compiled Twig templates (PHP code) to modify local Template variables
  - custom Twig functions (e.g. `asset()`, `asset_url()`)
- asset processing
  - external SCSS moved to the repository, and now processed by the Theme System
  - checking append conditions
  - adding assets via `{{ asset() }}` Twig function for centralized injection to page HTML
  - cache maintenance
  - asset routing
- ACP management
  - Package management
  - Package uploading
  - Theme Properties editing
  - Resource CRUD operations
  - user/developer feedback & tools

## Specification Deviations
The code generally follows the linked abstract specification; the following exceptions are known:

### ~~Major~~
- [x] ~~**_Processing (Compatible)_ Archiving Method**~~.
  One of the considered _Themelet Versioning Methods_, fulfilling the desire for [data retention](https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md#version-data-retention).

  Removed recently in favor of the Method selected in the document, preserving common code and triggers.

### Intermediate
- [ ] **Database Logic Removed**
  The code makes a full switch from using the database to using files for Theme-related matters.

  This means that:
  - the relevant tables (`themes`, `themestylesheets`, `templatesets`, `templategroups`, `templates`) are dropped
  - the in-database references to themes (`forums.style`, `users.style`) no longer use a foreign key (`themes.tid`), but a string identifying the theme directory (codename or equivalent)
  - the old logic for fetching theme data from the database (i.a. in `global.php`) is removed
  - the old logic for managing in-database resources is removed in the ACP's _Themes_ module
  - runtime data, including database row keys, attached to themes (e.g. `usergroups.gid` for permissions) is saved in files in source Theme Packages

  Meanwhile, the specification lays out preserving the barebones database structures & core logic in order to:
  - maintain the separation of source code and runtime data (then it may be possible to e.g. apply restrictive file permissions)
  - have a single pivot point between the DB & filesystem realms  (i.e. a `themes` row with a `tid` key)
  - continue increasing the referential integrity in the database (adding explicit foreign keys, propagating changes/deletions, rejecting invalid data)
  - for a period of time, offer compatibility with Plugins inserting & using their 1.8 Templates & Stylesheets (and notify administrators which Themes have deprecated Resources attached)

  The specification's approach would mean that a Theme can be "installed" (instantiated) from one of the Packages, which themselves remain unchanged, and storing MyBB installation-specific data in the database.
  This PR's code creates sub-Packages instead, where runtime data is stored & updated in files, to avoid modifying other source Packages that should be preserved in their original state (like the Core Theme).

### Minor
  - [ ] the data directory is not named `data/`, but `storage/` (_of what?_) (see also: https://github.com/mybb/meta/blob/main/architecture/directory-structure.md)
  - [ ] archive Themelet paths don't seem to follow [ABNF](https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md#abnf) `package-name` (e.g. `myplugin`), but `resource-namespace` (e.g. `ext.myplugin`) instead - at https://github.com/mybb/mybb/blob/e1f73aff771bd26425ee9c95cebf4f8afac7dbe7/inc/functions_themes.php#L301
  - [ ] asset inclusion/scheduling Twig function not named `asset()` (assuming we're comfortable to settle on that one), but `attach()` instead
  - [ ] the code state still indicates that running upgrade scripts will be required (removing _upgrade not needed_)

    (the _Cold Duplicate_ Method can trigger it independently on version bump; the trigger/hint call itself -  `clone_and_archive_default_theme()` may be useful in practice to deal with all intensive tasks right away, rather than after exiting the upgrade script)

(this is only an assessment of deviations with global impact, and may itself be incomplete)

## Completeness of the Theme System
Tracking in the [MyBB Extension System project](https://github.com/orgs/mybb/projects/3).

## Potential Refactoring
- Internal:
  Structure:
  - `MyBB\Extensions` class (Package-level operations; repository pattern)
  - `MyBB\Themelets` / `MyBB\ThemeSystem` class (Theme-related operations; repository pattern: mass querying, CRUD Resource functions)
  - `MyBB\Theme` class (can be instantiated, e.g. for Inheritance Chain resolution)
  - type constants, used across the codebase when possible
    - `MyBB\Extensions::TYPE_{THEME|PLUGIN}`
    - `MyBB\Theme::TYPE_{CORE|ORIGINAL|BOARD}`

  Logic:
  - centralized generation, parsing, and validation of identifiers:
    - Package names
    - namespace names
    - Resource paths
    - filesystem paths
  - centralized propagation of Resource changes accepted in CRUD/hint callback functions

- ACP:
  - move self-contained logic to classes/functions, telling the application _what_ to do instead of _how_
  - DRY
    - aim for the module to be an UI overlay for internal CRUD functions
    - merge common Templates & Stylesheets code where possible

Code quality for new code (usually `inc/src/`):
- [PSR-12](https://www.php-fig.org/psr/psr-12/) for PHP
- typing & documentation: rely on language features in PHP 7.4 & modern JavaScript; for unsupported/complex types & annotations: PHPDoc, [Psalm](https://psalm.dev/), [JSDoc](https://jsdoc.app/)
- [strict typing](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict)

admin/modules/config/plugins.php
dvz: We should:
- advertise the ACP's _Plugins_ module as extremely dangerous regarding permissions altogether
- reject Theme Packages with `.php` files
- add a config flag to disable uploads of any Packages through the ACP (i.e. impossible to modify with any permission level)

and IMO:
- disallow ACP uploading Plugin Packages - which contain `.php` files - by default (similarly, we might allow it with a config flag; IMO still not worth the risk)

otherwise the upload feature would serve vulnerability escalation to RCE on a silver platter.

admin/modules/config/plugins.php
dvz: Not sure about this implementation, but the plan was to allow plugins to function mostly unaffected (only modifications of existing templates would break).

We want to introduce `.json` manifest files for both Themes and Plugins - in 1.9, it would be required for Themes, but only optional for Plugins (as an alternative to `[codename]_info()`), preserving compatibility in this area.

install/common.php
dvz: [Version Data Retention](https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md#version-data-retention).

Before a different-than-planned approach was removed from this PR, it was necessary to call this, but now it should be opportunistic (to avoid copying on the first forum visit after update, or via the task system), and running the installer would not be necessary unless we actually need to run the release-specific migrations.

install/resources/settings.xml
dvz: This relates to a mode for Extension authors, allowing to bypass Resource caches, edit Packages otherwise protected from accidental writes, etc. In the end it shouldn't impact internal stability or security.

Indeed, right now it may be confusing, and names appearing in the UI (like _Development Mode_) will need to be revisited, to make sure they're clear and approachable (without sacrificing accuracy).
I've added _Reconcile developer/debug modes_ in the UX section of https://github.com/orgs/mybb/projects/3.

resource.php
dvz: This is only expected in "Development Mode". Its primary purpose is to allow skipping asset cache and proxy files from internal Packages instead (`inc/`), when they cannot be accessed directly via HTTP (a [best practice](https://docs.mybb.com/1.8/administration/security/protection/#restrict-web-access-to-internal-files) we want to [double down on](https://github.com/mybb/meta/blob/main/architecture/directory-structure.md)).

admin/modules/config/plugins.php
Sama34: I agree with all your points but the first one which I quite don't understand.

Also, won't disabling PHP files on plugin packages kill the feature altogether ? What plugins don't use PHP files ?

install/common.php
Sama34: So the default theme here is mutable ? I thought the theme being archived here was the inmutable version of the default theme.

admin/modules/config/plugins.php
dvz: RE the 1st point: as Plugins can arbitrarily change the application's logic, it may make sense to warn administrators on the permission selection page, and in security guides, that they shouldn't grant access to the _Plugins_ section of the ACP to anyone else unless necessary.

I've corrected the comment to indicate that only the new feature of in-ACP HTTP uploads is problematic (as opposed to FTP or similar level access).

resource.php
Sama34: > This is only expected in "Development Mode".

Oh, ok, so only disabling otherwise seems to be missing I think.

install/common.php
dvz: The kind of _mutation_ that would be relevant here is the official changes occurring between different versions of the Core Theme.

The tools for applying upstream changes in child themes would benefit from a version-to-version diff presented to the user, demonstrating the changes that have to be somehow mirrored in the child theme (when, for whatever reason, it cannot happen automatically).

Core and Original Themes will indeed be protected from accidental changes, and the UI will guide users away from it unless they're in an "expert" "Development Mode".

resource.php
dvz: Do you mean reverting to normal URLs?

This procedure should be handled automatically - normal paths would be used as arguments in custom Twig functions ([`asset_url()`/`asset()`](https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md#asset-management), which may be named differently in this PR), which will adjust the URLs to either point to the exposed cache directory (during normal operation), or `resource.php` when the special mode is on.

Sama34:
> * runtime data, including database row keys, attached to themes (e.g. `usergroups.gid` for permissions) is saved in files in source Theme Packages

I'm unsure about this, seems like some runtime data like group permissions would be better off if stored in the DB as well.

resource.php
Sama34: I meant to disable the `resource.php` file when the special mode is off, at least in the file itself I couldn't notice a code that blocks it. It might be on the `global.php` file or `init.php` so if so ignore my comment.

da-anda:
sorry if this is the wrong way to bring this up, but since you are (thankfully) refactoring the templating and plugin system, may I kindly suggest that you will support composer to install plugins? As of now, it is a huge PITA to manage forum installations along with the plugins via GIT, have proper deployment methods and what not. And since you are migrating to a json file for plugin metadata anyways and that plugins now have their own directory, using a `composer.json` for the plugin metadata etc would be awesome and ease up plugin updates.

In addition, may I suggest to add a hook to the language loader class which forum "owners" could use to hook into via a custom plugin and adjust some language labels of either the core or other plugins? Right now you have to patch files which is a nightmare to maintain when you try to keep everything updated with security patches and what not.

Thanks.

PS: future wishlist :)
- while breaking existing plugins anyways, language files should NOT be provided as PHP files but rather YAML, JSON or XLIFF (if you would like to use a translation standard)
- use something like the symfony service container for autoloading/class registration
- replace the hooks with a proper event system (like symfony events) which allows for priorities

admin/index.php
```diff
@@ -827,5 +827,9 @@
 
 $plugins->run_hooks("admin_load");
 
+if (!empty($page->plugin)) {
+	$modules_dir = MYBB_ROOT.'inc/plugins/'.$page->plugin.'/admin/modules';
+}
+
 require $modules_dir."/".$run_module."/".$action_file;
```
da-anda, line 834: would it make sense to check first if the file actually exists?

admin/modules/config/plugins.php
```diff
@@ -372,30 +374,153 @@
 	$page->output_footer();
 }
 
-// Activates or deactivates a specific plugin
-if($mybb->input['action'] == "activate" || $mybb->input['action'] == "deactivate")
+$is_upload = ($action == 'upload');
```
da-anda, line 377: in this day and age, shouldn't more strict comparissons be used? ( `===` instead of `==`)

admin/modules/config/plugins.php
```diff
@@ -415,22 +544,52 @@
...
+	}
+
+	if($do_upgrade || $action == "activate")
+	{
+		if(!$do_upgrade)
+		{
+			$message = $lang->success_plugin_activated;
+		} else {
+			// Plugin is compatible with this version?
+			$plugininfo = read_json_file(MYBB_ROOT."inc/plugins/{$codename}/plugin.json");
```
da-anda, line 581: why do you read the json file twice in case of an upgrade? See line 549. It might make sense to add a new method `getPluginInfo(string $codename): ?array` which will perform the parsing of the json file and then return a cached version of this in case of subsequent calls.

admin/modules/config/plugins.php
```diff
@@ -517,36 +680,53 @@
...
+	$table->construct_row();
+	$table->output($lang->upload_plugin);
+	$form->end();
+
 	if(!empty($plugins_list))
 	{
 		$a_plugins = $i_plugins = array();
 
-		foreach($plugins_list as $plugin_file)
+		if(!empty($plugins_list))
```
da-anda, line 698: unless github is messing up the diff view, this condition is superfluous as it's literally done 4 lines above.

admin/modules/config/plugins.php
da-anda: I honestly would prefer if you would force the same structure for everything, including a certain folder structure inside the ZIP files. This a) would finally allow to script plugin installations (or ideally use composer for it) and b) simplify the whole plugin system and make things more maintainable. So if you already break the plugin system, take the chance and do it properly once instead of over and over again over the next couple releases.

Personally I would also implement a `PluginInterface` (along with either a trait or an abstract class for convenience) that defines `initialize`, `enable`, `disable`, `install`, `uninstall` etc methods and force the plugins to wrap their methods into a proper, ideally namespaced class. It's not really much more work for plugin authors to convert the currently global functions into a class, and on MyBB side you could get rid of all the `if(function_exists())` stuff and simply call it.

This class could in the end also hold the metadata and what not.

admin/modules/config/plugins.php
```diff
@@ -517,36 +680,53 @@
...
+				}
 
-			if(isset($active_plugins[$codename]))
-			{
-				// This is an active plugin
-				$plugininfo['is_active'] = 1;
+				if(isset($active_plugins[$codename]))
+				{
+					// This is an active plugin
+					$plugininfo['is_active'] = 1;
```
da-anda, line 720: abstracting the pluginInfo into a class with dedicated setters and strict return values would IMO be a better approach over an array with unknown indices if you are rewriting this part anyways atm (code completion in IDEs anyone?).

admin/modules/style/themes.php
```diff
@@ -303,162 +396,282 @@
...
+			$infofile2 = '';
+			$infofile = "$unzipped_path/theme.json";
+			if (!($themeinfo = read_json_file($infofile))) {
+				$errors[] = $lang->sprintf($lang->error_bad_unzipped_json_file, htmlspecialchars_uni($infofile));
+			} else if (empty($themeinfo['codename'])) {
+				$errors[] = $lang->sprintf($lang->error_missing_theme_file_property, 'codename', htmlspecialchars_uni($infofile));
+			} else if ($themeinfo['codename'] != $codename) {
+				$errors[] = $lang->sprintf($lang->error_codename_mismatch, htmlspecialchars_uni($themeinfo['codename']), htmlspecialchars_uni($codename));
+			} else if (file_exists(MYBB_ROOT."inc/themes/{$codename}")) {
+				$infofile2 = MYBB_ROOT."inc/themes/{$codename}/theme.json";
```
da-anda, line 524: is it in general a good idea to have themes inside the `inc` folder which usually is protected from public access? Or are the theme assets at a different location? If so, I would suggest to keep them in one location and rather force their directory structure to have a `public` folder with assets that must be accessible from outside while the rest can be protected via .htaccess or nginx rules.

admin/modules/style/themes.php
```diff
@@ -303,162 +396,282 @@
...
+			switch ($type) {
+				case 0:
+					if ($_FILES['local_file']['error'] == 4) {
+						// UPLOAD_ERR_NO_FILE
+						$errors[] = $lang->error_uploadfailed_php4;
+					} else {
+						// Find out if there was an error with the uploaded file
+						if ($_FILES['local_file']['error'] != 0) {
+							$errors[] = $lang->error_uploadfailed.$lang->error_uploadfailed_detail;
+							switch($_FILES['local_file']['error']) {
```
da-anda, line 423: couldn't this entire block be simplified to
```
$errorNo = (int)$_FILES['local_file']['error'];
$errors[] = $lang->{'error_uploadfailed_php' . $errorNo} ?? $lang->sprintf($lang->error_uploadfailed_phpx, $errorNo);
```

da-anda (commented):
in general I would split up some of these huge functions into multiple ones

dvz:
Our plans are generally in line with your suggestions, @da-anda.

Eventually, we want to adjust how MyBB stores data to improve clarity, safety, and deployment & management. These changes would include storage formats and [locations](https://github.com/mybb/meta/blob/main/architecture/directory-structure.md) (i.a. for metadata and language phrases).

Accordingly, extension packages would be treated internal code, with front-end assets copied to exposed directories as needed (with SCSS transpilation, etc.).

1.9 focuses on the theme system, but we'll want to bring plugin metadata to the same format, and make it possible to install all extensions via Composer/CLI/API.

Better support of manual, and theme-supplied language phrases - that would be loaded automatically - should make it in at some point.

MyBB 1.9 introduces Laravel's Container in new code.

Issue #4602, opened 2022-12-08
undefined variable $usersbrowsing in eval()'d code (92) in ./showthread.php(1617).


tool-buddy:
Issue reproduced under MyBB 1.8.32 and PHP 8.1 and 8.0. Issue does not happen under PHP 7.4
The following warnings appear at the top of the page when loading  thread page.

![image](https://user-images.githubusercontent.com/23341378/206450161-3e5f4125-2874-401b-bb75-10f0b5fdff2b.png)

yuliu:
There are 3 PHP "undefined" warnings:

1. undefined array key "allowautourl" in `./inc/class_parser.php`(584).
2. undefined property `MyLanguage::$thread_modes` in eval()'d code (46) in `./showthread.php`(1617).
3. undefined variable `$usersbrowsing` in eval()'d code (92) in `./showthread.php`(1617).

I can't reproduce the 1st & 2nd ones:
 - In the 1st one warning, `allowautourl` is a setting key for "Automatically Parse URLs Into Links", which is introduced in 1.8.27. If you experience this warning, your board may be malfunctioning.
 - The missing `thread_modes` language entry key in the 2nd warning is not found in current code base. It seems the language entry is added and/or used by a plugin or a custom theme/template.

The 3rd one will be fixed shortly.

Issue #4593, opened 2022-10-05
Fix PHP 8 compatibility issues since version 1.8.31


dvz:

yuliu:
Warning [2] Undefined array key "doticon" - Line: 4687 - File: \usercp.php PHP 8.1.11 (WINNT)

https://github.com/mybb/mybb/blob/0cd318136a10b029bb5c8a8f6dddf39d87519797/usercp.php#L4687

Before using it, there was:

https://github.com/mybb/mybb/blob/0cd318136a10b029bb5c8a8f6dddf39d87519797/usercp.php#L4588-L4596

.. where `$thread['doticon']` may not be set due to `$mybb->settings['dotfolders']` (Use 'dot' Icons) configured to `0`.

yuliu:
Warning [2] Undefined array key "XXX" - Line: 4160 - File: \admin\modules\user\users.php PHP 8.1.11 (WINNT)

https://github.com/mybb/mybb/blob/0cd318136a10b029bb5c8a8f6dddf39d87519797/admin/modules/user/users.php#L4160

For a specific profile filed for a specific user, especially the `checkbox` type. If the user hasn't selected one checkbox item, the key `$val` in the array variable `$selected_options[$val]` won't exist.

yuliu:
Warning [2] Undefined array key XXXXX - Line: 2341 - File: \private.php PHP 8.1.11 (WINNT)

Warning [2] Trying to access array offset on value of type null - Line: 2342 - File: \private.php PHP 8.1.11 (WINNT)

Warning [2] Undefined array key "usergroup" - Line: 2343 - File: \private.php PHP 8.1.11 (WINNT)
/ Undefined array key "displaygroup"

https://github.com/mybb/mybb/blob/0cd318136a10b029bb5c8a8f6dddf39d87519797/private.php#L2338-L2349

If a user was deleted and the user had been a recipient in a private message sent by another user, when the message sender accesses their **Sent Items** box, these warnings will appear.

That means if the user was deleted, it won't be added to the array `$cached_users`.

https://github.com/mybb/mybb/blob/0cd318136a10b029bb5c8a8f6dddf39d87519797/private.php#L2253-L2262

A fix may take care of two types of messages, those sent to a single user or those sent to multiple recipients.

It's worth noting that when the sender opens a sent message, for multi-recipient messages, deleted users won't display in the recipient list, whereas a "N/A" will display in the "Sent Items" listing popup.

yuliu:
[...]

Issue #4530, opened 2022-04-01
Mark forum read doesn't work with php 8.0


stryderunknown:
I just did a test upgrade to php 8.0 (while running 1.8.30) and while a majority of things seemed to work, the Mark Forums Read (from the bottom of the main index) didn't.  Reverting back to php 7.4 and it works again, so there is something buried in the code that isn't compatible.  I thought I would mention it so it can be found before 8.0 gets full support.

dvz:
Are any [errors logged](https://docs.mybb.com/1.8/faq/errors/#enabling-error-logs) when that happens?

stryderunknown:
There are plenty of errors logged for 8.0, although none of them specifically state "Mark thread Read".  
A number of undefined variables shown below. 

< error>
< dateline>1648765705</dateline>
< script>global.php(978) : eval()'d code</script>
< line>20</line>
< type>2</type>
< friendly_type>Warning</friendly_type>
  	< message>Undefined variable $menu_help</message>
  	< back_trace>#0  errorHandler->error() called at [/inc/class_error.php:153]
  #1  errorHandler->error_callback() called at [/global.php(978) : eval()'d code:20]
  #2  eval() called at [/global.php:978]
  #3  require_once() called at [/showthread.php:28]
  < /back_trace>
  < /error>

 
 < error>
 	<dateline>1648765705</dateline>
 	<script>global.php(978) : eval()'d code</script>
 	<line>39</line>
 	<type>2</type>
 	<friendly_type>Warning</friendly_type>
 	<message>Undefined variable $unreadreports</message>
 	<back_trace>#0  errorHandler->error() called at [/inc/class_error.php:153]
 #1  errorHandler->error_callback() called at [/global.php(978) : eval()'d code:39]
 #2  eval() called at [/global.php:978]
 #3  require_once() called at [/showthread.php:28]
 < /back_trace>
 < /error>
 
 < error>
 	< dateline>1648765705</dateline>
 	< script>global.php</script>
 	< line>1087</line>
 	< type>2</type>
 	< friendly_type>Warning</friendly_type>
 	< message>Undefined array key "fid4"</message>
 	< back_trace>#0  errorHandler->error() called at [/inc/class_error.php:153]
 #1  errorHandler->error_callback() called at [/global.php:1087]
 #2  require_once() called at [/showthread.php:28]
 < /back_trace>
 < /error>
 
 < error>
 	<dateline>1648765705</dateline>
[...]

PR #4613, opened 2022-12-17
Fixes PHP 8 compatibility issues.

yuliu:
1. The issue for tracking: #4593.
1. The message of each commit explains what it fixes (the PHP 8 compatibility issue).
1. The PR includes fixes for issues reported in #4602.
1. The PR includes fixes for some issues reported in https://github.com/mybb/mybb/issues/4530#issuecomment-1089224347 . This one may be closed when the PR is merged, since other issues reported in it can't be reproduced.
1. The PR doesn't fix all remaining PHP 8 compatibility issues. That means there are still more.
1. Some more commits will be pushed later, before marking the PR as ready.


yuliu:
In commit https://github.com/mybb/mybb/pull/4613/commits/6b2fa22764219045a831bf2faaf7ac78bac9a878

 - Move post(s):
    - Undefined variable `$return_url` in eval()'d code (L27) in `moderation.php` (L2669).

Note: not sure if the variable `$return_url` used in template `moderation_inline_moveposts` is truly needed.

yuliu:
Remarks in commit https://github.com/mybb/mybb/pull/4613/commits/64e8476a8ecee61ecc053f342252746106e686f9

 - `$pm['saveasdraft']` and `$pm['options']['XXX']` are added following how a pm is prepared in [`send_pm()`](https://github.com/mybb/mybb/blob/c3c474a5b70f9fb4bb368b5b58c623b1c9e2fc16/inc/functions.php#L8974):
https://github.com/mybb/mybb/blob/c3c474a5b70f9fb4bb368b5b58c623b1c9e2fc16/inc/functions.php#L9079-L9085
 - `$pm['saveasdraft'] = 0` is optional because the default for this option is `0`.
 - `$pm['options']['signature']` is set to `0`, given that the pms are to inform moderators/admins of a report.

Eldenroot:
Will be merged into 1.8.33?

yuliu:
Remarks for commit https://github.com/mybb/mybb/pull/4613/commits/128f8ea6e7c9cb8b3dc8a8975076714d995465d9

 - This is only a workaround for eliminating PHP warnings caused by the real issue reported in #4630.

yuliu:
Remarks for commit https://github.com/mybb/mybb/pull/4613/commits/b153eb7861c904c94ea4a992215216b1fab15599

 - The fixed typo here (`canonlyviewownthread` -> `canonlyviewownthreads`) doesn't fully resolve the PHP warning issue:
   ```
   Warning [2] Undefined array key "canonlyviewownthreads" - Line: 1228 - File: inc\datahandlers\post.php PHP 8.1.13
   ```

 - The real issue behind this typo fix is about getting forum permissions:
    - `forum_permissions()` which calls
    - `fetch_forum_permissions()` that may return a set of group permissions where relevant permissions are missing --- those only exist in forum permissions (stored in `mybb_forumpermissions`) but not usergroup permissions (in `mybb_usergroups`), e.g., `canonlyviewownthreads`, `canonlyreplyownthreads`, etc.

yuliu:
Since the code base is already in v1.8.33, while the PR is based on v1.8.32, it'll cost much more time to pinpoint recent reported issues.

More PHP 8 compatibility issues will be addressed in other PRs, and this one is marked as ready for review.

yuliu:
More issues relating to the video mycode parsing have been reported in https://community.mybb.com/thread-237617.html . I'll push some commits that'll cover these issues later, to this PR.

kgraefe:
Not sure if that's the right place, but I found some more in the admin panel of MyBB 1.8.33 that are not yet fixed, as far as I can tell from scanning through the PR:
mybb

This only happens when a view is selected that does not match any user.

Edit: You probably guessed it, but FTR: the error happens on PHP 8.0, 8.1 and 8.2. It does not occur on PHP 7.4.

yuliu:
Remarks for commit https://github.com/mybb/mybb/pull/4613/commits/d0af36303c83730f5e51830d76fd4ef9cbf444ce

 - The commit should have fixed all issues related to the parser reported in https://community.mybb.com/thread-237617.html .
 - The following line was removed as the variable `$title` isn't used in the `video_metacafe_embed` template or anywhere else in the parser class. The removal may affect plugins that rely on the variable.

    https://github.com/mybb/mybb/blob/c3c474a5b70f9fb4bb368b5b58c623b1c9e2fc16/inc/class_parser.php#L1531

 - Currently, the parser can't correctly deal with URLs with any number of continuous `/` in its `path` part. [[REF]](https://stackoverflow.com/questions/10161177/url-with-multiple-forward-slashes-does-it-break-anything)

PR #4625, opened 2022-12-19
Fixes #4621 A replacement word can be equivalent to its word filter v…

### LINKED ISSUE #4621: A replacement word can be equivalent to its word filter via edit
Not a critical bug, but when we create a word filter and its replacement, both are checked to don't be equivalent, but when we use the edit function, they're not checked and can be equivalent

Original thread: [A replacement word can be equivalent to its word filter via edit](https://community.mybb.com/thread-237228.html)

Starpaul20:
…ia edit

Resolves #4621

Several filter checks were missing so I added them all.

### DIFF
--- a/admin/modules/config/badwords.php
+++ b/admin/modules/config/badwords.php
@@ -207,6 +207,43 @@ if($mybb->input['action'] == "edit")
 
 		if(!$errors)
 		{
+			$query = $db->simple_select("badwords", "bid", "badword = '".$db->escape_string($mybb->input['badword'])."' AND bid != '".$badword['bid']."'");
+
+			if($db->num_rows($query))
+			{
+				$errors[] = $lang->error_bad_word_filtered;
+			}
+		}
+
+		$badword_check = trim($mybb->input['badword']);
+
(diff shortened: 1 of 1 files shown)


admin/modules/config/badwords.php
```diff
@@ -205,6 +205,43 @@
...
+				$errors[] = $lang->error_bad_word_filtered;
+			}
+		}
+
+		$badword_check = trim($mybb->input['badword']);
+
+		if($mybb->get_input('regex', MyBB::INPUT_INT))
+		{
+			// Check validity of defined regular expression
+			if((@preg_match('#'.$badword_check.'#is', null) === false))
```
dvz: Would it make sense to use an empty string (`''`) instead of `null`? The `@` avoids the following error (PHP >= 8.1), but it may cause issues in the future.

```
Deprecated: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated
```

https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.null-not-nullable-internal

PR #4626, opened 2022-12-19
Fixes #4624 Custom User Title Maximum Length (Not Working As Intended)

### LINKED ISSUE #4624: Custom User Title Maximum Length (Not Working As Intended)
Template: usercp_profile_customtitle

Variable: {&#36;mybb-&gt;settings['customtitlemaxlength']}

Whole Piece: &lt;input type=&quot;text&quot; class=&quot;textbox&quot; name=&quot;usertitle&quot; size=&quot;25&quot; maxlength=&quot;{$mybb-&gt;settings['customtitlemaxlength']}&quot; value=&quot;{$newtitle}&quot; /&gt;

In Admin CP -&gt; Configuration -&gt; Settings -&gt; Profile Options the setting "Custom User Title Maximum Length" is described as "Maximum length a user can enter for the custom user title. Set to 0 to disable the limitation."

However, inputting 0 does not disable the limitation. It disables the ability to type anything at all because it makes the maxlength for input 0.

Original thread: [Custom User Title Maximum Length (Not Working As Intended)](https://community.mybb.com/thread-236967.html)

Starpaul20:
Resolves #4624

### DIFF
--- a/install/resources/mybb_theme.xml
+++ b/install/resources/mybb_theme.xml
@@ -6779,7 +6779,7 @@ if(use_xmlhttprequest == "1")
 </tr>]]></template>
 		<template name="modcp_banuser_lift" version="1801"><![CDATA[<div class="float_right"><a href="modcp.php?action=liftban&amp;uid={$banned['uid']}&amp;my_post_key={$mybb->post_code}">{$lang->lift_ban}</a></div>]]></template>
 		<template name="modcp_banuser_liftlist" version="1800"><![CDATA[<option value="{$time}"{$selected}>{$title}{$thattime}</option>]]></template>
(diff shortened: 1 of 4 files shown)


yuliu:
ModCP > Profile Editor (`/modcp.php?action=editprofile&uid=`) also has this input field that may need a fix.

dvz:
Will it not error out for certain values due to DB field limits?
https://github.com/mybb/mybb/blob/c234fe2ff0e8ecac242d6239185529be8c703dd6/install/resources/mysql_db_tables.php#L1068
If that's the case, perhaps the limit should be mirrored as `max="250"` in the setting (removing the _0 to disable_ bit).

Starpaul20:
Good catch, I've reverted my previous fixes and changed the setting code instead.

PR #4660, opened 2023-03-02
PHP 8 Warning for get more smilies

Eldenroot:
https://community.mybb.com/thread-237763.html

Resolves #ISSUE_ID


dvz:
Looks like the variable is not needed, and can be removed from the `misc_smilies_popup` template instead.

Eldenroot:
@dvz - better?

schauan:
with class="modal" the close - "x" is in background of the list and I see only a little bit of "x". 
With class="modal_{$pid}" it is in foreground.  

See https://community.mybb.com/thread-237763-post-1396822.html#pid1396822

PR #4662, opened 2023-03-02
PHP 8 Warnings - bunch of fixes

Eldenroot:
https://community.mybb.com/thread-237773.html

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -515,7 +515,7 @@ class PostDataHandler extends DataHandler
 		$permissions = user_permissions($uid);
 
 		// Fetch the forum this post is being made in
-		if(!$post['fid'])
+		if(empty['fid'])
 		{
 			$query = $db->simple_select('posts', 'fid', "pid = '{$post['pid']}'");
 			$post['fid'] = $db->fetch_field($query, 'fid');
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -515,7 +515,7 @@ class PostDataHandler extends DataHandler
 		$permissions = user_permissions($uid);
 
 		// Fetch the forum this post is being made in
-		if(empty['fid'])
+		if(empty(['fid']))
 		{
 			$query = $db->simple_select('posts', 'fid', "pid = '{$post['pid']}'");
 			$post['fid'] = $db->fetch_field($query, 'fid');
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -515,7 +515,7 @@ class PostDataHandler extends DataHandler
 		$permissions = user_permissions($uid);
 
 		// Fetch the forum this post is being made in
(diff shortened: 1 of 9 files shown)


inc/functions_forumlist.php
```diff
@@ -487,7 +487,7 @@ function get_forum_lightbulb($forum, $lastpost, $locked=0)
 	global $mybb, $lang, $db, $unread_forums;
 
 	// This forum is a redirect, so override the folder icon with the "offlink" icon.
-	if($forum['linkto'] != '')
+	if(isset($forum['linkto']) && $forum['linkto'] != '')
```
Stefan-MyBB: You could merge both conditions with !empty.

printthread.php
```diff
@@ -20,6 +20,7 @@
 
 // Load global language phrases
 $lang->load("printthread");
+$lang->load("archive");
```
Stefan-MyBB: It makes little sense to load that language file here.

Stefan-MyBB (changes requested):

printthread.php
Eldenroot: Are you sure?

Eldenroot:
Part is fixed. Thank you!

Eldenroot:
Is there anybody who can review this and help me a little bit with this one? Thank you!

printthread.php
Stefan-MyBB: In any case, I would not expect this language file to be used anywhere outside the archive.

inc/datahandlers/post.php
```diff
@@ -515,7 +515,7 @@ function verify_image_count()
 		$permissions = user_permissions($uid);
 
 		// Fetch the forum this post is being made in
-		if(!$post['fid'])
+		if(empty['fid'])
```
dvz: A missing `(` makes the syntax checks fail

printthread.php
dvz: Worth reviewing if both `Pages:` phrases should be included like this:
https://github.com/mybb/mybb/blob/af60e3a9e03b7e4be0d4d6a848945d026c233b0b/install/resources/mybb_theme.xml#L9894

inc/datahandlers/post.php
Eldenroot: Hmm, checked both and still invalid... reverted back.

printthread.php
Eldenroot: I will revert this part for this PR.

inc/datahandlers/post.php
dvz: The variable name was also lost, so in the end we'll probably want: `if(empty($post['fid']))`

Eldenroot: Thank you, this did the trick!

printthread.php
Eldenroot: Reverted, a new ticket  should be opened.

PR #4668, opened 2023-03-02
PHP 8 Search results display with invalid sid

Eldenroot:
https://community.mybb.com/thread-237717.html

### DIFF
--- a/search.php
+++ b/search.php
@@ -60,7 +60,7 @@ if($mybb->input['action'] == "results")
 	$query = $db->simple_select("searchlog", "*", "sid='$sid'");
 	$search = $db->fetch_array($query);
 
-	if(!$search['sid'])
+	if(isset($search['sid']) && !$search['sid'])
 	{
 		error($lang->error_invalidsearch);
 	}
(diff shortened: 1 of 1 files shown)


search.php
```diff
@@ -60,7 +60,7 @@
 	$query = $db->simple_select("searchlog", "*", "sid='$sid'");
 	$search = $db->fetch_array($query);
 
-	if(!$search['sid'])
+	if(isset($search['sid']) && !$search['sid'])
```
Stefan-MyBB: This looks wrong to me. I suggest `if(empty($search['sid']))`.

search.php
```diff
@@ -152,7 +152,7 @@
 
 	// Work out if we have terms to highlight
 	$highlight = "";
-	if($search['keywords'])
+	if(isset($search['keywords']))
```
Stefan-MyBB: `!empty` could be used here.

Stefan-MyBB (changes requested):

Eldenroot:
@Stefan-MyBB - thank you, it should be better now...

PR #4669, opened 2023-03-02
PHP 8 Full text search warnings

Eldenroot:
https://community.mybb.com/thread-237723.html

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -1709,7 +1709,7 @@ function perform_search_mysql_ft($search)
 
 	// Searching a specific thread?
 	$tidsql = '';
-	if(isset($search['tid']) && $search['tid'])
+	if(!empty($search['tid']))
 	{
 		$tidsql = " AND t.tid='".(int)$search['tid']."'";
 	}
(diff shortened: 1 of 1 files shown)


inc/functions_search.php
```diff
@@ -1709,7 +1709,7 @@ function perform_search_mysql_ft($search)
 
 	// Searching a specific thread?
 	$tidsql = '';
-	if($search['tid'])
+	if(isset($search['tid']) && $search['tid'])
```
Stefan-MyBB: I suggest using `if(!empty($search['tid']))` here.

inc/functions_search.php
```diff
@@ -1727,7 +1727,7 @@ function perform_search_mysql_ft($search)
 	if($search['postthread'] == 1)
 	{
 		// No need to search subjects when looking for results within a specific thread
-		if(!$search['tid'])
+		if(isset($search['tid']) && !$search['tid'])
```
Stefan-MyBB: That looks wrong to me. I suggest using `!empty`.

Stefan-MyBB (changes requested):

Eldenroot:
@Stefan-MyBB - thank you, it should be fixed now

inc/functions_search.php
```diff
@@ -1727,7 +1727,7 @@ function perform_search_mysql_ft($search)
 	if($search['postthread'] == 1)
 	{
 		// No need to search subjects when looking for results within a specific thread
-		if(!$search['tid'])
+		if(!empty($search['tid']))
```
dvz: The logic got modified - `empty()` alone (rather than `!empty()`) would be appropriate in place of a `!` negation

dvz (changes requested):

PR #4670, opened 2023-03-02
PHP 8 Warnings in utf8mb4 conversion process

Eldenroot:
https://community.mybb.com/thread-237738.html

### DIFF
--- a/admin/modules/tools/system_health.php
+++ b/admin/modules/tools/system_health.php
@@ -238,7 +238,7 @@ if($mybb->input['action'] == "utf8_conversion")
 		$table->construct_cell($lang->please_wait);
 		$table->construct_row();
 
-		$table->output($converting_table." {$mybb->input['table']}");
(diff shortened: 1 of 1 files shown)


admin/modules/tools/system_health.php
```diff
@@ -357,7 +357,7 @@
 		}
 
 		// Any fulltext indexes to recreate?
-		if(is_array($fulltext_to_create))
+		if(!empty($fulltext_to_create))
```
Stefan-MyBB, line 360: This is fine. But the variable needs to be initzialized around line 281.

yuliu:
PR #4613 has a fix for this in commit https://github.com/mybb/mybb/pull/4613/commits/7c32e61044d98a8fe15d8ce6b94018a25c001371 .

Eldenroot:
OK, so I will close this one.

PR #4681, opened 2023-04-04
PHP 8 - Fix various datahandler `event.php` and `calendar.php` errors

RevertIT-alt:
There could be more issues, but those are current ones that I manage to find and fix.

### DIFF
--- a/inc/datahandlers/event.php
+++ b/inc/datahandlers/event.php
@@ -70,7 +70,7 @@ class EventDataHandler extends DataHandler
 	function verify_name()
 	{
 		$name = &$this->data['name'];
-		$name = trim($name);
+		$name = trim_blank_chrs($name);
 		if(!$name)
 		{
 			$this->set_error("missing_name");
--- a/inc/datahandlers/event.php
+++ b/inc/datahandlers/event.php
@@ -87,7 +87,7 @@ class EventDataHandler extends DataHandler
 	function verify_description()
 	{
 		$description = &$this->data['description'];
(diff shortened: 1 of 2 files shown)


inc/datahandlers/event.php
```diff
@@ -577,18 +577,17 @@ function update_event()
...
 		if(isset($event['endtime']))
 		{
 			$this->event_update_data['endtime'] = (int)$event['endtime'];
-			$this->event_update_data['usingtime'] = (int)$event['usingtime'];
+			$this->event_update_data['usingtime'] = isset($event['usingtime']) ? (int) $event['usingtime'] : 0;
 		}
 		else
 		{
 			$this->event_update_data['endtime'] = 0;
-			$this->event_update_data['usingtime'] = 0;
```
dvz, line 591: Is it supposed to not be set anymore?

RevertIT: Yeah,  `L580` and `L586`  are throwing errors, and `L591` is redundant because errors are already thrown before we come to set a default value.

dvz: Looks like it may no longer be set/corrected when neither `starttime` nor `endtime` are set. It may be better to leave it to preserve the intended logic.

RevertIT: Submitted a correction, this now should cover all the instances of `endtime/usingtime`

Eldenroot:
@dvz is this finished?

PR #4684, opened 2023-04-04
PHP 8 - Fix `announcements` errors

RevertIT-alt:
1. Fix empty `title` and `message`
2. Fix the potentially fatal error by incorrect data type entered in the form.

### DIFF
--- a/admin/modules/forum/announcements.php
+++ b/admin/modules/forum/announcements.php
@@ -589,11 +589,11 @@ if($mybb->input['action'] == "edit")
 		}
 
 		// Gather start and end date data
-		$startday = $mybb->input['starttime_day'];
+		$startday = $mybb->get_input('starttime_day', MyBB::INPUT_INT);
 		$start_time = $mybb->input['starttime_time'];
-		$startmonth = $mybb->input['starttime_month'];
+		$startmonth = $mybb->get_input('starttime_month', MyBB::INPUT_INT);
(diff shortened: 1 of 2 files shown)


admin/modules/forum/announcements.php
```diff
@@ -589,23 +589,23 @@ function checkAction(id)
...
+		$startdateyear = $mybb->get_input('starttime_year', MyBB::INPUT_INT);
 
 		if($mybb->input['endtime_type'] == 1)
 		{
 			// Set time
 			$endtime_checked[1] = 'checked="checked"';
 			$endtime_checked[2] = '';
 
-			$endday = $mybb->input['endtime_day'];
+			$endday = $endmonth = $mybb->get_input('endtime_day', MyBB::INPUT_INT);
```
dvz: Looks like `$endmonth = ` shouldn't be there

RevertIT: I can't remember now why I did that. But looks like I did it in some sort of hacky way to prevent an undefined variable or something. Or maybe it was a mistype from my side.

Eldenroot:
Is this still valid?

admin/modules/forum/announcements.php
Sama34:
```suggestion
			$endday = $mybb->get_input('endtime_day', MyBB::INPUT_INT);
```

If there is any undefined variable warning then we can fix that later.

PR #4687, opened 2023-04-10
PHP 8 - Fix admin language module

RevertIT:
- Initialize `$no_write` and prevent `Undefined variable $no_write - Line: 172 - File: admin\modules\config\languages.php`
- Replace `$config` with `$mybb->config` and set correct value `admin_dir` to prevent `Undefined array key "admindir" - Line: 1034 - File: \admin\modules\config\languages.php`


dvz:
It may be safe to remove these config references, as that value refers to the renamed `admin/` directory, as opposed to the language directories like `inc/languages/english/admin/`.

The directory seems to be stripped anyway, and prepended as `admin/` at https://github.com/mybb/mybb/blob/af60e3a9e03b7e4be0d4d6a848945d026c233b0b/admin/modules/config/languages.php#L445-L456

RevertIT:
That's correct.  Removed redundant config var.

PR #4690, opened 2023-04-19
Replace remaining `unserialize()` calls with `my_unserialize()`

### LINKED ISSUE #4617: Replace first-party usage of PHP's `unserialize()` with `my_unserialize()`
Use the custom function for consistency and security (https://github.com/mybb/mybb/issues/3793#issuecomment-1332349456).

References:
- _Warning_ for https://www.php.net/unserialize

dvz:
Resolves #4617

### DIFF
--- a/inc/class_datacache.php
+++ b/inc/class_datacache.php
@@ -134,7 +134,7 @@ class datacache
 			$query = $db->simple_select("datacache", "title,cache");
 			while($data = $db->fetch_array($query))
 			{
-				$this->cache[$data['title']] = unserialize($data['cache']);
+				$this->cache[$data['title']] = my_unserialize($data['cache']);
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


inc/class_datacache.php
```diff
@@ -134,7 +134,7 @@ function cache()
 			$query = $db->simple_select("datacache", "title,cache");
 			while($data = $db->fetch_array($query))
 			{
-				$this->cache[$data['title']] = unserialize($data['cache']);
+				$this->cache[$data['title']] = my_unserialize($data['cache']);
```
yuliu, line 137: This call would very probably impact performance of large forums with many categories/forums, and so does the following call `$data = my_unserialize($cache_data['cache']);` in line 220 when the `forums` cache is read.

dvz: We can go back to `unserialize()` for the problematic calls (or all replacements in this PR), and set `allowed_classes` to `false`.

yuliu: Yes. Contents of most cache entries could be large strings, and that'd potentially cause performance issue.

PR #4694, opened 2023-04-29
Update functions.php

ghost:
The original code had a `warning message` that the third parameter of `preg_replace` and `str_replace` functions were being passed null values, which is deprecated in PHP. The third parameter in both of these functions is the subject or message that needs to be manipulated. Passing null as the message would trigger a warning in PHP.

To fix this issue, a check was added to ensure that `$message` is not null before calling `preg_replace` and `str_replace` functions. This check ensures that these functions are called only when `$message` is a non-null string.

In this modified code, I have used the ternary operator `?` to check if `$message` is a non-null string. If `$message` is a non-null string, we perform the corresponding replacement operation on it. Otherwise, we return `$message` as it is, without any manipulations.

### DIFF
--- a/admin/modules/config/settings.php
+++ b/admin/modules/config/settings.php
@@ -1704,6 +1704,7 @@ if($mybb->input['action'] == "change")
 					$multivalue = explode(',', $setting['value']);
 				}
 
+				$option_list = array();
 				for($i = 0; $i < $typecount; $i++)
 				{
 					$optionsexp = explode("=", $type[$i]);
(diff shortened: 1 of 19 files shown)


Stefan-MyBB:
Do you have an example when null is passed to htmlspecialchars_uni?

RevertIT:
This happens only with PHP 8.1 >=. But yes, it should be updated to be future-proof. 

There are multiple null passings via some popular plugins.

Btw, this PR can be further improved, if the message is null return false.

ghost:
> if the message is null return false.

Will update the PR with a new improvement.

> Do you have an example when null is passed to htmlspecialchars_uni?

No, i saw the error immediatly fixed it.

RevertIT:
I would do something like
```php
if (!is_string($message)) {
    return false; // or return '';
}
```
because it checks null assignments as well and remove redundant ternary operators from the regex lines below. 

I'm not sure if this is going to be implemented as it is only getting triggered under very specific circumstances and is mostly due to faulty plugins.

PR #4700, opened 2023-05-15
PHP 8 - `admin/groups` module

RevertIT:
You can thank me later. 

There were over 30+ undefined errors alone if god forbid we had errors while editing the group.


Sama34:
I might be missing something but it seems some of these changes to the `get_input()` are unnecessary. Unnecessary in the sense that there shouldn't be warnings by reviewing the code.

PR #4701, opened 2023-05-15
PHP 8 - `admin/mod_tools`  module

RevertIT:


admin/modules/config/mod_tools.php
```diff
@@ -2220,8 +2220,8 @@ function checkAction(id)
 	$form_container->output_row($lang->approve_unapprove_posts." <em>*</em>", '', $form->generate_select_box('approveposts', $approve_unapprove, $mybb->input['approveposts'], array('id' => 'approveposts')), 'approveposts');
 	$form_container->end();
 
-	$selectoptions = "<option value=\"-1\"{$do_not_split_checked}>{$lang->do_not_split}</option>\n";
-	$selectoptions .= "<option value=\"-2\"{$split_same_checked} style=\"border-bottom: 1px solid #000;\">{$lang->split_to_same_forum}</option>\n";
+	$selectoptions = "<option value=\"-1\"" . (isset($do_not_split_checked) ? $do_not_split_checked : '') . ">{$lang->do_not_split}</option>\n";
+	$selectoptions .= "<option value=\"-2\"" . (isset($split_same_checked) ? $split_same_checked : '') . " style=\"border-bottom: 1px solid #000;\">{$lang->split_to_same_forum}</option>\n";
```
Stefan-MyBB: Initializing these variables probably (not inside the if condition) might make more sense.

RevertIT: It can be defined above https://github.com/mybb/mybb/blob/c6ea09bf0e9017777179dc13d25e11da45e74561/admin/modules/config/mod_tools.php#L1430
but not really needed tbh

admin/modules/config/mod_tools.php
```diff
@@ -2220,8 +2220,8 @@ function checkAction(id)
 	$form_container->output_row($lang->approve_unapprove_posts." <em>*</em>", '', $form->generate_select_box('approveposts', $approve_unapprove, $mybb->input['approveposts'], array('id' => 'approveposts')), 'approveposts');
 	$form_container->end();
```
Sama34:
```suggestion
if(!isset($do_not_split_checked))
{
	$do_not_split_checked = '';
}

if(!isset($split_same_checked))
{
	$split_same_checked = '';
}

```

admin/modules/config/mod_tools.php
```diff
@@ -2220,8 +2220,8 @@ function checkAction(id)
...
 
-	$selectoptions = "<option value=\"-1\"{$do_not_split_checked}>{$lang->do_not_split}</option>\n";
-	$selectoptions .= "<option value=\"-2\"{$split_same_checked} style=\"border-bottom: 1px solid #000;\">{$lang->split_to_same_forum}</option>\n";
+	$selectoptions = "<option value=\"-1\"" . (isset($do_not_split_checked) ? $do_not_split_checked : '') . ">{$lang->do_not_split}</option>\n";
```
Sama34:
```suggestion
	$selectoptions = "<option value=\"-1\"{$do_not_split_checked}>{$lang->do_not_split}</option>\n";
```

admin/modules/config/mod_tools.php
Sama34:
```suggestion
	$selectoptions .= "<option value=\"-2\"{$split_same_checked} style=\"border-bottom: 1px solid #000;\">{$lang->split_to_same_forum}</option>\n";
```

Sama34 (changes requested):
No need to do it before the if statement.

admin/modules/config/mod_tools.php
Sama34: No needed but doing it that way will decrease the ugliness of our code base :P

RevertIT: I mean it's more a matter of taste. I still consider the PR you submitted the "ugly" hacky way of doing what I did with extra lines. Why would you initialize those variables in GET when those are already initialized in POST requests, wouldn't it be better to add them before both POST and GET, and reduce unnecessary duplication?

Like I said matter of taste. I'll submit a PR correction which I consider to be the correct way, but you could still disagree and we would get the same output regardless.

Sama34: My code suggestion might indeed be incorrect, consider it a "normal" suggestion more than a "best" approach suggestion. I went back into the code but not as back as necessary, probably.

Sama34: I think the major concern here is that in-line HTML is disliked and we have made some progress into removing it altogether from the front-end, and will eventually be reworked on the future for the admin side.

Taking that into consideration we would like to avoid adding code that might delay or put load into the work necessary for such achievement.

It would be appreciated if you take into consideration this explanation when contributing to in-line HTML within the script, but you are free to not to.

Sama34 (approved):

PR #4702, opened 2023-05-17
Use PHP's own `unserialize()` for performance reasons in chosen locations

dvz:
Addresses problems raised in https://github.com/mybb/mybb/pull/4690#discussion_r1194906185

Related: #3793


yuliu (approved):
Looks good and locally tested fine.

PR #4707, opened 2023-05-27
Fix #2349 better replace for or and and keywords

### LINKED ISSUE #2349: Searching for "orange" will find "angel" in non-fulltext search
Non-Fulltext-Search in MyBB ignores "and" and "or" of the first entered keyword of the search.
If I search for "orange", MyBB will actually search for "ange". I can't search for "andi", because Mybb will complain that my search keyword is only one character long.

The reason is a bug in clean_keywords function: 

https://github.com/mybb/mybb/blob/featur...h.php#L243
https://github.com/mybb/mybb/blob/featur...h.php#L248

At these lines, "and" and "or" will be removed from the search string, if they are at the beginning of the first keyword.

I suggest removing the two "if"-statements completely, as they are broken in many ways. Not only will they currently unintentionally remove the first characters of a matching keyword, they also don't protect against "and" and "or" keywords at the first position of a search query (the intention of these lines) if fixed to not cut off longer keywords:

"or test" would become "test" (correct)
"and test" would become "test" (correct)
"or or test" would become "or test" (wrong)
"and and test" would become "and test" (wrong)
"or and test" would become "test" (correct)
"and or test" woud become "or test" (wrong)
and so on ...

Original thread: [Searching for "orange" will find "angel" in non-fulltext search](http://community.mybb.com/thread-189418.html)

BettyIsBoop:
Resolves #2349

### DIFF
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -282,9 +282,9 @@ function clean_keywords($keywords)
 	$keywords = str_replace("*", "%", $keywords);
 	$keywords = preg_replace("#\s+#s", " ", $keywords);
 	$keywords = str_replace('\\"', '"', $keywords);
-
+	// trim for blank characters
+	$keywords = trim_blank_chrs($keywords);
 	// Search for "and" or "or" and remove if it's at the beginning
-	$keywords = trim($keywords);
 	$keywords = preg_replace('/^((and|or)(\s|$))+/', '', $keywords);
 
 	if(!$keywords)
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -285,13 +285,13 @@ function clean_keywords($keywords)
 
 	// Search for "and" or "or" and remove if it's at the beginning
 	$keywords = trim($keywords);
-	if(my_strpos($keywords, "or") === 0)
+	if(my_strpos($keywords, "or ") === 0)
 	{
 		$keywords = substr_replace($keywords, "", 0, 2);
 		$keywords = " ".$keywords;
 	}
 
-	if(my_strpos($keywords, "and") === 0)
+	if(my_strpos($keywords, "and ") === 0)
 	{
 		$keywords = substr_replace($keywords, "", 0, 3);
 		$keywords = " ".$keywords;
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -285,17 +285,7 @@ function clean_keywords($keywords)
 
 	// Search for "and" or "or" and remove if it's at the beginning
 	$keywords = trim($keywords);
-	if(my_strpos($keywords, "or ") === 0)
-	{
-		$keywords = substr_replace($keywords, "", 0, 2);
-		$keywords = " ".$keywords;
-	}
-
-	if(my_strpos($keywords, "and ") === 0)
-	{
-		$keywords = substr_replace($keywords, "", 0, 3);
-		$keywords = " ".$keywords;
-	}
+	$keywords = preg_replace('/^((and|or)(\s|$))+/', '', $keywords);
 
 	if(!$keywords)
 	{
--- a/inc/functions_search.php
+++ b/inc/functions_search.php
@@ -283,7 +283,7 @@ function clean_keywords($keywords)
 	$keywords = preg_replace("#\s+#s", " ", $keywords);
 	$keywords = str_replace('\\"', '"', $keywords);
 	// trim for blank characters
-	$keywords = trim_blank_chrs($keywords);
+	$keywords = trim($keywords);
 	// Search for "and" or "or" and remove if it's at the beginning
 	$keywords = preg_replace('/^((and|or)(\s|$))+/', '', $keywords);
 


lairdshaw:
This fix fails in three scenarios:
1. One or more spaces precede the opening "and" or "or" (it fails to remove the "and" or "or").
2. The string starts with more than one "and" or "or" (it only removes the first).
3. The string consists in a single "and" or "or" (it fails to remove that "and" or "or").

The solution provided in this comment works in all of those scenarios [edit: for the first, only when the `$keywords` argument is replaced by `trim($keywords)`], and should, I think, be preferred:
https://github.com/mybb/mybb/issues/2349#issuecomment-189898399

RevertIT:
it should use `trim_blank_chrs()` to match all varieties of blank characters. eg: "invisible blank name" characters.

BettyIsBoop:
> One or more spaces precede the opening "and" or "or" (it fails to remove the "and" or "or").

Same with current one

> The string starts with more than one "and" or "or" (it only removes the first).

Same with current one

> The string consists in a single "and" or "or" (it fails to remove that "and" or "or").

OK

Personally i use on my forum because current solution broke more times. I check with regexp.

Simple issue open on 2016 … still exist 7 years after, Let Me Laugh …

BettyIsBoop:
> This fix fails in three scenarios:
> 
>     1. One or more spaces precede the opening "and" or "or" (it fails to remove the "and" or "or").

[Like before : trim remove space](https://github.com/mybb/mybb/pull/4707/files#diff-f22f376bfbc597df3fd386d21431706a9c94f158a9cc32763bb7612a780f5ba9R287)


>     2. The string starts with more than one "and" or "or" (it only removes the first).

Like current

> 
>     3. The string consists in a single "and" or "or" (it fails to remove that "and" or "or").

Fixed


> The solution provided in this comment works in all of those scenarios [edit: for the first, only when the `$keywords` argument is replaced by `trim($keywords)`], and should, I think, be preferred: [#2349 (comment)](https://github.com/mybb/mybb/issues/2349#issuecomment-189898399)

Done

lairdshaw:
> > This fix fails in three scenarios:
> > ```
> > 1. One or more spaces precede the opening "and" or "or" (it fails to remove the "and" or "or").
> > ```
> 
> [Like before : trim remove space](https://github.com/mybb/mybb/pull/4707/files#diff-f22f376bfbc597df3fd386d21431706a9c94f158a9cc32763bb7612a780f5ba9R287)

You're right - I'd missed that. Arguably, as @RevertIT suggests, that `trim()` should be converted to `trim_blank_chrs()`, but that's probably not necessary for this PR.

> > The solution provided in this comment works in all of those scenarios [edit: for the first, only when the `$keywords` argument is replaced by `trim($keywords)`], and should, I think, be preferred: [#2349 (comment)](https://github.com/mybb/mybb/issues/2349#issuecomment-189898399)
> 
> Done

Untested but LGTM.

Sama34 (approved):
This seems to work when searching for "andy", it finds the following thread titles :

> Tempore technologia paternoster tandy
> Andyravida technologia tristique urna nullandy
> Computandytor quis viverra ipsum venenatis

When searching for "ORAS", it finds the following thread titles :

> ORAS, Pokemon nunc risus aliquam risus
> Eget continentes pellentesque cinque ORAS
> Andyravida technoloRASgia tristique urna nullandy

Before the patch the same search resulted in the "minimum length" error.

We appreciate your contribution !

RevertIT:
@Sama34 Try searching with quotes `' " `and special characters to see if it will break syntax, can't really test it myself nowadays

Shnoulle:
> You're right - I'd missed that. Arguably, as @RevertIT suggests, that trim() should be converted to trim_blank_chrs(), but that's probably not necessary for this PR.

It's for user entered data when search, a simple trim for spaced added by copy/paste : i really don't think need more.

RevertIT:
> It's for user entered data when search, a simple trim for spaced added by copy/paste : i really don't think need more.

Well, it's actually needed, trim doesn't trim `invisible` characters as shown below:
https://3v4l.org/ObKdY#v8.2.6

Output:
```php
Trim blank character
string(19) "   f   "


Trim blank character via chrs function 
string(1) "f"


Trim normal space 
string(1) "f"
```

Shnoulle:
> > It's for user entered data when search, a simple trim for spaced added by copy/paste : i really don't think need more.
> 
> Well, it's actually needed, trim doesn't trim `invisible` characters as shown below: https://3v4l.org/ObKdY#v8.2.6
> 

1. **It's another issue**
2. You think mybb user use u+XXXX when do a search ? And if by mistake : they enter an invisible character : they don't find what the want … ,not a real issue.


This fix one issue, and not another one.

Sama34:
> @Sama34 Try searching with quotes `' " `and special characters to see if it will break syntax, can't really test it myself nowadays

Searching for the following yields the same (positive) results as my previous post :
```
"andy"
"oras"

```

Searching for the following yields a _Sorry, but no results were returned ..._ error result:
```
'andy"
'oras"
'andy'
'oras'
"andy'
"oras'
"&andy%"
"&oras%"
/oras
```
I stopped testing after all these terms. My environment is:
> MyBB 1.8.33
> PHP 8.2.6
> SQL MySQL (PDO) 8.0.33-0ubuntu0.22.04.2

BettyIsBoop:
> > @Sama34 Try searching with quotes `' " `and special characters to see if it will break syntax, can't really test it myself nowadays
> 
> Searching for the following yields the same (positive) results as my previous post :
> 
> I stopped tested after all these terms. I'm testing on: MyBB 1.8.33 PHP 8.2.6 SQL MySQL (PDO) 8.0.33-0ubuntu0.22.04.2

I really think it's totally unrelated to this issue …

Did you compare with current ? Dis you compare with "toty" or 'toty" etc ?

RevertIT:
> You think mybb user use u+XXXX when do a search ? And if by mistake : they enter an invisible character : they don't find what the want … ,not a real issue.

Doesn't change the fact it's better to implement now than forget down the road if the native function exits.

lairdshaw:
> Doesn't change the fact it's better to implement now than forget down the road if the native function exits.

Yeah, given that it's got a benefit (handling more cases of whitespace), the question seems to be "Is there any reason _not_ to?" The only potentially legit one I can think of is "It's less performant", but given that this isn't performance-critical code, and the performance impact is anyway very small, that doesn't seem to be an _actually_ legit reason. So, I'm in favour of the change, but also fine with it being achieved via a separate issue+PR. My 2c.

Eldenroot:
> > You think mybb user use u+XXXX when do a search ? And if by mistake : they enter an invisible character : they don't find what the want … ,not a real issue.
> 
> Doesn't change the fact it's better to implement now than forget down the road if the native function exits.

I agree. @BettyIsBoop please could you edit your PR to cover all? Thx!

I hope this will be merged into 1.8.35 :)

dvz:
Can none of the listed entities be part of a valid search phrase (perhaps in other languages/scripts)?

If there are no known issues, we can keep using the inbuilt functions - if there's a problem, these can be replaced across the application later.

BettyIsBoop:
> > > You think mybb user use u+XXXX when do a search ? And if by mistake : they enter an invisible character : they don't find what the want … ,not a real issue.
> > 
> > 
> > Doesn't change the fact it's better to implement now than forget down the road if the native function exits.
> 
> I agree. @BettyIsBoop please could you edit your PR to cover all? Thx!
> 
> I hope this will be merged into 1.8.35 :)

Done

Sama34:
I can't recall what the status of this is. It did look fine to me back then, but I'm unsure what the final conclusion was.

If somebody could confirm so we can either merge or close this, please.

Regards.

lairdshaw:
I'm not aware of any objections unless @dvz's [comment above](https://github.com/mybb/mybb/pull/4707#issuecomment-1583610478) indicates one (I'm not really sure what he's saying). Otherwise, nothing's changed my own endorsement of this PR. Does anybody else object?

dvz:
> the question seems to be "Is there any reason not to?"

[`trim_blank_chrs()`](https://github.com/mybb/mybb/blob/a4f0aa37effa9dbabf1ca88386ea8b3edf4abfd4/inc/functions.php#L8727C1-L8727C17) removes characters blacklisted for IDN (to prevent spoofing), while the search encompasses regular content - legitimate characters may be excluded unnecessarily.

lairdshaw:
> legitimate characters may be excluded unnecessarily.

OK, I understand your original comment now. Fair enough. It seems to be a reason not to use `trim_blank_chrs()` (although I note incidentally that in the MyBB codebase, `trim_blank_chrs()` is used for more than just sanitising IDNs).

BettyIsBoop:
> > legitimate characters may be excluded unnecessarily.
> 
> OK, I understand your original comment now. Fair enough. It seems to be a reason not to use `trim_blank_chars()` (although I note incidentally that in the MyBB codebase, `trim_blank_chars()` is used for more than just sanitising IDNs).

**Seriously** ?

https://github.com/mybb/mybb/pull/4707#issuecomment-1566287998

2 years after ?

lairdshaw:
> **Seriously** ?
> 
> [#4707 (comment)](https://github.com/mybb/mybb/pull/4707#issuecomment-1566287998)
> 
> 2 years after ?

Hmm? I wrote "arguably" in that comment. @dvz made an argument. It seems persuasive enough. Anyhow, it's a pretty minor issue and I'm not invested in the outcome.

Sama34:
If I understand correctly, practically removing support for characters blacklisted for IDN seems like a dealbreaker to me.

I confess I may not be understanding the topic to the extent some of you may.

Maybe, a refactor of the search feature, perhaps using some external library or dependency, would be better ? I doubt we are the first out there to try to implement a reasonably secure and working search feature.

If so is feasible, maybe this would be better handled in 1.9 or 1.10 and lesser feature breaking changes may be applied to 1.8

lairdshaw:
> If I understand correctly, practically removing support for characters blacklisted for IDN seems like a dealbreaker to me.

Just to be sure I'm parsing that sentence and its logic correctly, are saying that you think we should use `trim()` and *not* `trim_blank_chrs()` here?

Sama34:
What I'm saying is that we should avoid any change that breaks support for special characters.

I don't want to pretend to know better because I haven't researched for a solution myself and I have limited myself mostly to just testing the PR.

To be frank, I don't understand why we don't:
1. Break all search phrases by a space,
2. Check that the first phrase, or any phrase for that matter, is a valid search term,
3. Glue them back together

Fixes are not required to be one line edits; if the fix requires more refactoring, then so be it.

If this doesn't make sense then, well, that is what I meant with not knowing better.

To be honest, I'm a bit confused about this issue and its discussion.

lairdshaw:
> What I'm saying is that we should avoid any change that breaks support for special characters.

OK, that aligns with how I interpreted you, as saying that we should use `trim()` and not `trim_blank_chars()` here, which is @dvz's position, and which I don't have a problem with. Does anybody else still want us to switch to using `trim_blank_chrs()` here?

lairdshaw:
> To be frank, I don't understand why we don't:
> 
>     1. Break all search phrases by a space,
> 
>     2. Check that the first phrase, or any phrase for that matter, is a valid search term,
> 
>     3. Glue them back together

~Would you agree that that's probably beyond the scope of this PR, which is fixing the specific issue #2349?~

~Beyond that, I haven't looked at the search code to assess whether there exist any problems that this approach would fix. Are you aware of any?~

Those comments were a bit hasty. Your suggested approach *would* fix this specific issue. I just wonder whether, given that we've got a working solution, it's productive to change approaches.

BettyIsBoop:
> What I'm saying is that we should avoid any change that breaks support for special characters.


I just want to allow user to find orange when they search orange …  ([issue exit since 2014](https://community.mybb.com/thread-189418.html) …)

I don't want to fix all search issue.

See the original issue : https://github.com/mybb/mybb/issues/2349

inc/functions_search.php
```diff
@@ -282,20 +282,10 @@ function clean_keywords($keywords)
 	$keywords = str_replace("*", "%", $keywords);
 	$keywords = preg_replace("#\s+#s", " ", $keywords);
 	$keywords = str_replace('\\"', '"', $keywords);
-
+	// trim for blank characters
+	$keywords = trim_blank_chrs($keywords);
```
Sama34:
```suggestion
	$keywords = trim($keywords);
```

Sama34 (dismissed):

Sama34:
A new issue may be opened for the `trim` vs `trim_blank_characters` issue if necessary.

PR #4716, opened 2023-06-03
Fixes #4715 - Undefined array key username

### LINKED ISSUE #4715: Undefined array key username
The below reference code can lead to undefined array key username error.

https://github.com/mybb/mybb/blob/bc03b9be63cf2f1c31b672f16d64552e4f0736d7/member.php#L1724

The root cause is that the code is run through irrespective of the result of the check in line 1721.

I would suggest the following fix:

```
$input_username = '';

if(!empty($mybb->input['username']))
{
	$input_username = htmlspecialchars_uni($mybb->input['username']);
}
```

Eldenroot:
Resolves #4715 

Fix proposed by @drecomx

### DIFF
--- a/member.php
+++ b/member.php
@@ -1721,7 +1721,12 @@ if($mybb->input['action'] == "resetpassword")
 		{
 			$input_username = '';
 		}
-		$input_username = htmlspecialchars_uni($mybb->input['username']);
+		$input_username = '';
+
+		if(!empty($mybb->input['username']))
+		{
+			$input_username = htmlspecialchars_uni($mybb->input['username']);
+		}
 
 		eval("\$activate = \"".$templates->get("member_resetpassword")."\";");
 		output_page($activate);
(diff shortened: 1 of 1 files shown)


drecomx:
These lines are obsolete from my point of view:
		if(!isset($mybb->input['username']))
		{
			$input_username = '';
		}

as my suggested fix initialises $input_username to an empty string and populates it with a value if provided in input.

Sama34 (approved):

Sama34:
Oh, yes, the lines above should be removed as they would become redundant with the provided fix.

Sama34 (changes requested):
Lines 1720-1723 are redundant now.

Stefan-MyBB:
Why not this?
```php
$input_username = htmlspecialchars_uni($mybb->get_input('username'));
```

Sama34 (approved):

PR #4730, opened 2023-06-16
fix error in php8 when set_time_limit is disabled

### LINKED ISSUE #4729: @set_time_limit with php8
If the hoster has disabled the set_time_limit function the error supression with the @ operator does no longer work with php 8 [https://www.php.net/manual/en/language.operators.errorcontrol.php](url)
Thus the installation ist not possible (Line 11 in install/index.php). Other files are also affected.

Possible Solution?
if(strpos(ini_get('disable_functions'),'set_time_limit')===false) 
    @set_time_limit(0);

opt-natter:
@operator changed in php8
a check if the function is avaible is necessary before using it

Resolves #4729

### DIFF
--- a/admin/modules/tools/backupdb.php
+++ b/admin/modules/tools/backupdb.php
@@ -147,7 +147,8 @@ if($mybb->input['action'] == "backup")
 			admin_redirect("index.php?module=tools-backupdb&action=backup");
 		}
 
-		@set_time_limit(0);
+		if(strpos(ini_get('disable_functions'),'set_time_limit')===false)
+			@set_time_limit(0);
 
 		// create an array with table prefix appended for checks, as full table names are accepted
 		$binary_fields_prefixed = array();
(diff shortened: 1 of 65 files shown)


chack1172:
You can use function_exists() instead of checking the ini setting. In PHP 8 function_exists return false for disabled functions.

Sama34:
An internal function for this would be better IMO

Sama34:
@dvz what do you think about this ?

dvz:
I'd probably still use a `$seconds` argument and pass `0`/`false`/`null` for clarity.

We can also delete it from the upgrade files and call it in a single place, e.g. before the upgrade functions are called:
https://github.com/mybb/mybb/blob/3da3513d3db68408df9d7ad912f3ee0e2d1b09d8/install/upgrade.php#L434

A similar approach was used in the 1.9 installer: https://github.com/mybb/mybb/blob/976321e8f9474540925d0d6d09e0cfdc0d79848d/inc/src/Maintenance/Process/Model/UpgradeModel.php#L202-L204

PR #4753, opened 2023-09-11
Fix #4752: Remove the deprecated "All rights reserved" phrase

### LINKED ISSUE #4752: Remove the deprecated "All rights reserved" phrase from all copyright notices
```
"The requirement to add a notice became obsolete and essentially
deprecated on August 23, 2000, as every country that was a member of
the Buenos Aires Convention (which is the only copyright treaty
requiring this notice to be used) is also a member of the Berne
Convention which requires protection be granted without any formality
of notice of copyright."
```

taken from [debian-legal mailing list](https://lists.debian.org/debian-legal/2007/06/msg00252.html)

The next reply of the email says,

```
> 5) Is it *advisable* to use such a terminology in a copyright statement 
> in a file licensed under the GPL?

No.
```

Since all copyright notices in mybb are after 2002, I suggest removing this phrase from all copyright notices.

delgh1:
The phrase "All rights reserved" is not required to be granted protections, and it basically has no legal effects now. Using this phrase in a file or a project licensed under GPL/LGPL is not advisable.

See also: https://lists.debian.org/debian-legal/2007/06/msg00252.html
Resolves #4752

### DIFF
--- a/admin/inc/class_form.php
+++ b/admin/inc/class_form.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * MyBB 1.8
- * Copyright 2014 MyBB Group, All Rights Reserved
+ * Copyright 2014 MyBB Group
  *
  * Website: http://www.mybb.com
  * License: http://www.mybb.com/about/license
--- a/admin/inc/class_page.php
+++ b/admin/inc/class_page.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * MyBB 1.8
- * Copyright 2014 MyBB Group, All Rights Reserved
+ * Copyright 2014 MyBB Group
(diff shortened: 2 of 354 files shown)


Sama34:
As it has no effect I won't consider merging this change to the feature branch myself.

Ben-MyBB:
Can this be targeted to 1.9 branch or shall we close this?

Sama34:
A new PR will be required if we want this for 1.9. Suck PR would probably be one of the last ones to merge so there is no rush for it.

We appreciate your suggestion @delgh1, and we encourage you or anyone to send a PR once 1.9 is near release.

Regards.

PR #4761, opened 2023-09-24
Misc.php warning

### LINKED ISSUE #4760: Misc.php warning
https://community.mybb.com/thread-239244.html

Eldenroot:
Resolves #4760


misc.php
```diff
@@ -179,7 +179,7 @@
 		$query = $db->simple_select("searchlog", "*", "uid='{$mybb->user['uid']}' AND dateline > '$timecut'", array('order_by' => "dateline", 'order_dir' => "DESC"));
 		$last_search = $db->fetch_array($query);
 		// Users last search was within the flood time, show the error
-		if($last_search['sid'])
+		if(!empty($last_search['sid']))
```
Stefan-MyBB: I suggest to use `if($last_search)` instead. It's shorter and more readable. $last_search is null when there is no result.

Eldenroot: Thank you!

dvz:
Should this still be a draft?

PR #4765, opened 2023-10-06
Use standard delimiter for mail

### LINKED ISSUE #4734: eMail issues with PHP mail()...
Because I did not receive any response on the MyBB Community website for 4 days, I am posting the bug here!

MyBB Version 1.8.34
PHP Version 8.0.29
SQL Engine MySQLi 8.0.33

This is a plain english default installation without any plugins.

I was able to find why no email was able to be sent from anywhere in the MyBB forum.

Inside the file "/inc/mailhandlers/php.php" there is 2 lines causing the problem. The headers lines are not separated with CRLF (\r\n), and because the mailhandler delimiter is changed in the code blocking also the mass mail.

PHP ask for a CRLF (\r\n) separator between headers lines...
This is the section where the headers and delimiter are wrongly changed in the file "/inc/mailhandlers/php.php" :

```
if($this->sendmail)
{
// to be commented $this->headers = str_replace("\r\n", "\n", $this->headers);
$this->message = str_replace("\r\n", "\n", $this->message);
// to be commented $this->delimiter = "\n";
}

```

Without the commented lines, this is causing the PHP mail() silent error!

dvz:
Resolves #4734

### DIFF
--- a/inc/mailhandlers/php.php
+++ b/inc/mailhandlers/php.php
@@ -36,6 +36,7 @@ class PhpMail extends MailHandler
 		global $lang, $mybb;
 
 		// For some reason sendmail/qmail doesn't like \r\n
+		/*
 		$this->sendmail = @ini_get('sendmail_path');
 		if($this->sendmail)
 		{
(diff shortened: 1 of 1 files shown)


Sama34:
Why would we keep it by commenting it out ?

dvz:
I'm not sure if it would cause problems on setups where this was actually needed. 

The problem may be addressed in server configuration, but regular users may be unable to modify it, so the bit of code would still be there to uncomment in those cases.

Then, if it turns out to be a common issue we can create a setting or a [config flag](https://docs.mybb.com/1.8/administration/configuration-file/), or if no problems are reported, remove it completely after some time.

PR #4768, opened 2023-10-17
Update member.php - php 8.x warning (COPA)

### LINKED ISSUE #4767: member.php warning 1.8.36 (COPA)
https://community.mybb.com/thread-239282.html

Eldenroot:
Resolves #4767

### DIFF
--- a/member.php
+++ b/member.php
@@ -391,7 +391,7 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 			$captcha->invalidate_captcha();
 		}
 
-		if($mybb->settings['regtype'] != "randompass" && !empty($mybb->cookies['coppauser']))
+		if($mybb->settings['regtype'] != "randompass" && empty($mybb->cookies['coppauser']))
 		{
 			// Log them in
 			my_setcookie("mybbuser", $user_info['uid']."_".$user_info['loginkey'], null, true, "lax");
--- a/member.php
+++ b/member.php
@@ -391,13 +391,13 @@ if($mybb->input['action'] == "do_register" && $mybb->request_method == "post")
 			$captcha->invalidate_captcha();
 		}
 
-		if($mybb->settings['regtype'] != "randompass" && !isset($mybb->cookies['coppauser']))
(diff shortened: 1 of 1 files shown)


member.php
```diff
@@ -743,6 +743,11 @@
 				my_setcookie("coppauser", 1, -0);
 				$under_thirteen = true;
 			}
+			else
+			{
+				my_setcookie("coppauser", 0, -0);
```
Stefan-MyBB, line 750: I haven't tested this but I think it's incorrect.
The code only checks the existence of that specific cookie but not its value.
https://github.com/mybb/mybb/blob/bc8e0c6bcf068b9787bb75f415710660c2371e60/member.php#L394-L400
So it basically sets it to true in both cases.

dvz: We'll probably want to:
- default `$under_thirteen` to `false` outside the `// Is COPPA checking enabled?` block, and
- add `!empty(...)` to `$mybb->cookies['coppauser']` where it causes `undefined` issues

instead.

Sama34: As per @dvz suggestion, the following line :
```PHP
if($mybb->settings['coppa'] != "disabled" && ($mybb->cookies['coppauser'] == 1 || $under_thirteen))
```

Should be :
```PHP
if($mybb->settings['coppa'] != "disabled" && (!empty($mybb->cookies['coppauser']) || $under_thirteen))
```

Finally, after line :
```PHP
my_setcookie("coppadob", "{$mybb->input['bday1']}-{$mybb->input['bday2']}-{$birthday_year}", -1);
```

Add :
```PHP
$under_thirteen = false;
```

This is the simplest solution, avoiding any possible issues like the one suggested by @Stefan-MyBB 

The 2 changes from `isset(...)` to `empty(...)` on lines 394 and 400 seem to be fine.

Eldenroot: Thank you, should be fixed right now!

Eldenroot:
Should be fixed. Thank you all for help!

member.php
dvz: It's not immediately clear that `$under_thirteen` is always set - how about setting it to `false` by default at around line 720 (instead of this area, outside of all the conditionals)?

Eldenroot:
Thx

PR #4771, opened 2023-10-26
moderation.php: set default move forum in dropdown to current forum

### LINKED ISSUE #4772: Default to current forum in "Move / Copy Thread" dropdown

dsyalex:
While this may sound like a silly default since it's the one place the thread will not need to be moved to, if you have a large number of forums and a common workflow is to move the thread to a sub-forum, this prevents a lot of scrolling through the dropdown to find the right forum.

### DIFF
--- a/moderation.php
+++ b/moderation.php
@@ -1020,7 +1020,7 @@ switch($mybb->input['action'])
 
 		$plugins->run_hooks("moderation_move");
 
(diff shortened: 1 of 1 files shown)


Stefan-MyBB:
Just as a note: If you move a lot of threads to specific sub-forums, you should set up custom moderation tools to do that. That's what I did to save some time.

dsyalex:
That's neat, thanks Stefan.

PR #4774, opened 2023-11-26
Fixes #4773 - Improve 2FA QRCode presentation

### LINKED ISSUE #4773: Improve 2FA QRCode presentation
https://community.mybb.com/thread-239405.html

Eldenroot:
Resolves #4773


Sama34:
Would it be reasonable to add a language string for this ?

Eldenroot:
Is it needed?

Sama34:
I would think so, for non Latin-alphabet based languages the most.

Eldenroot:
I am still not sure, there is almost nothing to translate...

PR #4784, opened 2024-01-22
Fixes #4783 - Invalid AdminCP login warning 1.8.37 PHP 8.2

### LINKED ISSUE #4783: Invalid AdminCP login warning 1.8.37 PHP 8.2
admin/index.php
If a user attempts to login to AdminCP without permissions, a warning is generated.
&#36;login_message and error condition is not defined.

&lt;error&gt;
	&lt;dateline&gt;1705314831&lt;/dateline&gt;
	&lt;datetime&gt;2024-01-15 10:33:51 UTC -0700&lt;/datetime&gt;
	&lt;script&gt;admin/index.php&lt;/script&gt;
	&lt;line&gt;619&lt;/line&gt;
	&lt;type&gt;2&lt;/type&gt;
	&lt;friendly_type&gt;Warning&lt;/friendly_type&gt;
	&lt;message&gt;Undefined variable $login_message&lt;/message&gt;
	&lt;back_trace&gt;#0  errorHandler-&gt;error() called at [/inc/class_error.php:153]
#1  errorHandler-&gt;error_callback() called at [/admin/index.php:619]
&lt;/back_trace&gt;
&lt;/error&gt;
Find beginning line 538	my_unsetcookie('adminsid');
} and replace with 	my_unsetcookie('adminsid');
	if($mybb-&gt;get_input('do') == 'login')
	{
		$login_message = $lang-&gt;error_mybb_not_admin_account;
	}
}
Then add language string in inc/languages/english/admin/global.lang.php after line 54.
$l['error_mybb_not_admin_account'] = &quot;Username is not an administrator.&quot;;
Thank you Laird for helping me track this one down.

Original thread: [Invalid AdminCP login warning 1.8.37 PHP 8.2](https://community.mybb.com/thread-240083.html)

Eldenroot:
Resolves #4783

### DIFF
--- a/inc/languages/english/admin/global.lang.php
+++ b/inc/languages/english/admin/global.lang.php
@@ -52,7 +52,7 @@ $l['error_admin_session_expired'] = "Your administration session has expired.";
 $l['error_invalid_ip'] = "Your IP address is not valid for this session.";
 $l['error_mybb_admin_lockedout'] = "This account has been locked out.";
 $l['error_mybb_admin_lockedout_message'] = "Your account is currently locked out after failing to login {1} times. You have been sent an email with instructions on how to unlock your account.";
-$l['error_mybb_not_admin_account'] = "Username is not an administrator.";
+$l['error_mybb_not_admin_account'] = "You do not have permission to access the administration control panel.";
 
 $l['error_invalid_username'] = "The username you entered is invalid.";
 $l['error_invalid_uid'] = "The user id you entered is invalid.";
--- a/inc/languages/english/admin/global.lang.php
+++ b/inc/languages/english/admin/global.lang.php
@@ -52,6 +52,7 @@ $l['error_admin_session_expired'] = "Your administration session has expired.";
 $l['error_invalid_ip'] = "Your IP address is not valid for this session.";
 $l['error_mybb_admin_lockedout'] = "This account has been locked out.";
 $l['error_mybb_admin_lockedout_message'] = "Your account is currently locked out after failing to login {1} times. You have been sent an email with instructions on how to unlock your account.";
+$l['error_mybb_not_admin_account'] = "Username is not an administrator.";
 
 $l['error_invalid_username'] = "The username you entered is invalid.";
 $l['error_invalid_uid'] = "The user id you entered is invalid.";
--- a/admin/index.php
+++ b/admin/index.php
@@ -536,6 +536,10 @@ if(empty($mybb->usergroup['cancp']) && !$is_super_admin || !$mybb->user['uid'])
 	$db->delete_query("adminsessions", "uid = '{$uid}'");
 	unset($mybb->user);
 	my_unsetcookie('adminsid');
+	if($mybb->get_input('do') == 'login')
+	{
+		$login_message = $lang->error_mybb_not_admin_account;
+	}
 }
 
 if(!empty($mybb->user['uid']))


inc/languages/english/admin/global.lang.php
```diff
@@ -52,6 +52,7 @@
 $l['error_invalid_ip'] = "Your IP address is not valid for this session.";
 $l['error_mybb_admin_lockedout'] = "This account has been locked out.";
 $l['error_mybb_admin_lockedout_message'] = "Your account is currently locked out after failing to login {1} times. You have been sent an email with instructions on how to unlock your account.";
+$l['error_mybb_not_admin_account'] = "Username is not an administrator.";
```
Stefan-MyBB: This error message sounds strange and does not match the other messages. IMHO a more generic "You do not have the permission..." would be better.

lairdshaw: I think a generic error doesn't supply enough information. How about something like: "The account with which you attempted to log in does not have administrative privileges."?

effone: A login attempt of a non admin should always throw a 404. Why we reveal the route of admin login is correct?

dvz: Note that front-end and ACP sessions are separate, so a login form seems unavoidable.

If we generalize 
https://github.com/mybb/mybb/blob/d1b1d45994e5b190e49973b8d658a89bc4e6fd3f/inc/languages/english/admin/global.lang.php#L177
the new phrase could read `You do not have permission to access the administration control panel.`

lairdshaw (approved):

Sama34:
I think @effone comment is on point, we shouldn't be explicit about the permission, instead we should trow an invalid user message so it is vague.

It seems conflicting as well because it seems this error could also be presented if there is no uid :
https://github.com/mybb/mybb/pull/4784/files#diff-59a8ada2c7b98dfd72d4b79144d2327149f0a18901f0fd3b31bd7cf1f4760f89R529

So how come there is no uid and the user has no permission at the same time ?

lairdshaw:
> I think @effone comment is on point, we shouldn't be explicit about the permission, instead we should trow an invalid user message so it is vague.

I'm not sure what this achieves: how does it obscure the ACP's entry point? Couldn't a fake/honeypot ACP also show the same error message, making it impossible to distinguish the real one from the fake one?

> It seems conflicting as well because it seems this error could also be presented if there is no uid : https://github.com/mybb/mybb/pull/4784/files#diff-59a8ada2c7b98dfd72d4b79144d2327149f0a18901f0fd3b31bd7cf1f4760f89R529
> 
> So how come there is no uid and the user has no permission at the same time ?

There _is_ a uid: it's the first half of the conditional that tests true in this scenario. The login was successful (much earlier, on [line 245](https://github.com/mybb/mybb/blob/d1b1d45994e5b190e49973b8d658a89bc4e6fd3f/admin/index.php#L245)). The first half of the conditional though determines that the user, despite successfully logging in, doesn't have admin privs. Inside the `if` block, we then simply test whether we arrived here _from_ a login attempt, in which case we know to show the message (without the test, we would show the message on first loading `admin/index.php`, prior to a login attempt).

The alternative is to perform this setting of `$login_message` to `$lang->error_mybb_not_admin_account` after [line 247](https://github.com/mybb/mybb/blob/d1b1d45994e5b190e49973b8d658a89bc4e6fd3f/admin/index.php#L247), either unconditionally (confusing) or after performing the same assignments as on [lines 513-527](https://github.com/mybb/mybb/blob/d1b1d45994e5b190e49973b8d658a89bc4e6fd3f/admin/index.php#L513-L527) then performing the same first half of the test as on line 529 (redundant).

Because it was either confusing or redundant, I chose this way instead (yes, ultimately, this code came from me, via HLFadmin then Eldenroot).

If there's a better way, then let's take it.

PR #4785, opened 2024-01-22
Fixes #4775 - Part 1/2 - Avatar warnings when editing user profile in…

### LINKED ISSUE #4775: Fix PHP 8 compatibility issues since version 1.8.37
- [x] https://community.mybb.com/thread-239883.html

   > Warning [2] Trying to access array offset on value of type bool - Line: 2730 - File: inc/functions.php PHP 8.1.26 (Linux)

  #4776

Eldenroot:
… admincp

#4775 - part 1/2

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -632,7 +632,7 @@ if($mybb->input['action'] == "edit")
 				}
 			}
 			// Are we setting a new avatar from a URL?
-			else if($mybb->input['avatar_url'] && $mybb->input['avatar_url'] != $user['avatar'])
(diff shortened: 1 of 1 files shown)


admin/modules/user/users.php
```diff
@@ -632,7 +632,7 @@
 				}
 			}
 			// Are we setting a new avatar from a URL?
-			else if($mybb->input['avatar_url'] && $mybb->input['avatar_url'] != $user['avatar'])
+			else if(isset($mybb->input['avatar_url']) && $mybb->input['avatar_url'] != $user['avatar'])
```
dvz: Will this not trigger the code when an empty input is sent? If so, `!empty(...)` or `$mybb->get_input(...)` may be better

PR #4798, opened 2024-04-12
Update `replyto` in merge/split moderation actions

### LINKED ISSUE #4789: Awaiting Moderation thread count issue
When moved posts are set to 'unapproved', they count as a thread awaiting moderation, not a post, as the 'reply-to' flag is set to 0. Maybe the code global.php is a little too streamlined?!

```
	$unapproved_posts = $unapproved_threads = 0;
	$query = $db->simple_select("posts", "replyto", "visible = 0");
	while($unapproved = $db->fetch_array($query))
	{
		if($unapproved["replyto"] == 0){
			$unapproved_threads++;
		} else {
			$unapproved_posts++;
		}
	}
```

dvz:
Resolves #4789

### DIFF
--- a/inc/class_moderation.php
+++ b/inc/class_moderation.php
@@ -1595,6 +1595,9 @@ class Moderation
 			update_first_post($thread['tid']);
 		}
 
+		// Attach moved posts to the first post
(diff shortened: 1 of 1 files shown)


Sama34:
A method would be nice on the future for this.

Issue #4630, opened 2023-01-02
PrivateMessages formerly sent by any now deleted users will be recognized as sent by "MyBB Engine"


yuliu:
## Background:

 - PrivateMessages formerly sent by any now deleted users will have `0` for field `fromid` in the database table `mybb_privatemessages` records.

 - The function [`send_pm()` in `inc/functions.php`](https://github.com/mybb/mybb/blob/mybb_1832/inc/functions.php#L8966-L8974) allows `0`, `-1` or any positive UID as the input for `$fromid`:

    https://github.com/mybb/mybb/blob/c3c474a5b70f9fb4bb368b5b58c623b1c9e2fc16/inc/functions.php#L8966-L8974

 - In the above function, `-1` for `$fromid` is usually used in sending a notification PM when:
    - The PM is sent **as/from an anonymous user** for notifying the target user of being warned. [REF](https://github.com/mybb/mybb/blob/mybb_1832/warnings.php#L105-L126)
    - The PM is for notifying a new post replied to a subscribed thread. [REF](https://github.com/mybb/mybb/blob/mybb_1832/inc/datahandlers/post.php#L1309-L1319)

 - However, `-1`, meant to be "MyBB Engine", for the PM sender's UID (`$fromid`) will never be saved because:
    - `$fromid`'s `-1` will be changed to `0`. [REF](https://github.com/mybb/mybb/blob/mybb_1832/inc/functions.php#L9057-L9060)
    - , and the database table `mybb_privatemessages`'s corresponding column `fromid` is defined `unsigned` for MySQL. [REF](https://github.com/mybb/mybb/blob/mybb_1832/install/resources/mysql_db_tables.php#L564)

 - Then, the desired username "MyBB Engine", meant for the above PMs whose `UID=0`, will also be shown for the PMs that sent by any deleted users, which may confuse end-users.

## Possible fix would include:

 - Changes in MySQL database definitions and new upgrade scripts.
 - Various changes in `class PMDataHandler`.
 - Functions related to pm (e.g., `send_pm()`) and code that uses these functions.
 - Changes for PM box interface and related language entries.

yuliu:
This issue also causes some PHP warnings in PHP 8+:

 - Undefined array key "uid" and "username" in `verify_sender()` in `inc/datahandlers/pm.php` ([L149-L150](https://github.com/mybb/mybb/blob/mybb_1832/inc/datahandlers/pm.php#L149-L150)).
[...]

PR #4800, opened 2024-04-15
Fix `undefined` issues in `PMDataHandler` for system messages

dvz:
Resolves .https://github.com/mybb/mybb/issues/4775#issuecomment-1906701491

Related: #4630

### DIFF
--- a/inc/datahandlers/pm.php
+++ b/inc/datahandlers/pm.php
@@ -129,6 +129,16 @@ class PMDataHandler extends DataHandler
 			return true;
 		}
 
+		if($pm['fromid'] <= 0)
+		{
+			$pm['sender'] = array(
+				"uid" => 0,
+				"username" => ''
+			);
+
+			return true;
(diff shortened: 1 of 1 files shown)


inc/datahandlers/pm.php
```diff
@@ -129,6 +129,16 @@ function verify_sender()
 			return true;
 		}
 
+		if($pm['fromid'] <= 0)
+		{
+			$pm['sender'] = array(
+				"uid" => 0,
```
Sama34, line 135: Shouldn't this be `$pm['fromid']` in case the value is `-1` ?

dvz: It looks like it would end up being `0` (#4630), but if there are conditions for `-1` it may need adjustment (conditions added in this PR simply check for both)

Sama34: Of course..

PR #4813, opened 2024-05-21
Fix #4812.

### LINKED ISSUE #4812: PHP 8 Warning: Undefined array key "quick_login"
There’s an unchecked array index access in https://github.com/mybb/mybb/blob/d6a96019025de9149014e06b1df252e6122e5630/inc/datahandlers/login.php#L112

Happens after 3 failed login attempts. (At least in [1.8.37](https://github.com/mybb/mybb/releases/tag/mybb_1837). ~~Might have been circumvented on a higher level in 1.8.38.~~)

> Warning [2] Undefined array key "quick_login" - Line: 112 - File: inc/datahandlers/login.php PHP 8.2.19 (Linux)
> File Line Function
> /inc/class_error.php 153 errorHandler->error
> /inc/datahandlers/login.php 112 errorHandler->error_callback
> /inc/datahandlers/login.php 77 LoginDataHandler->verify_captcha
> /inc/datahandlers/login.php 266 LoginDataHandler->verify_attempts
> /member.php 1766 LoginDataHandler->validate_login

smnscp:
Resolves #4812

### DIFF
--- a/inc/datahandlers/login.php
+++ b/inc/datahandlers/login.php
@@ -109,7 +109,7 @@ class LoginDataHandler extends DataHandler
 				return true;
 			}
 		}
-		else if(isset($mybb->input['quick_login']) && $mybb->input['quick_login'] == 1 && $mybb->input['quick_password'] && $mybb->input['quick_username'])
+		else if(!empty($mybb->input['quick_login']) && !empty($mybb->input['quick_password']) && !empty($mybb->input['quick_username']))
 		{
 			$this->set_error('regimagerequired');
 			return false;
--- a/inc/datahandlers/login.php
+++ b/inc/datahandlers/login.php
@@ -109,7 +109,7 @@ class LoginDataHandler extends DataHandler
 				return true;
 			}
 		}
-		else if($mybb->input['quick_login'] == 1 && $mybb->input['quick_password'] && $mybb->input['quick_username'])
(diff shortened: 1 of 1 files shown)


inc/datahandlers/login.php
```diff
@@ -109,7 +109,7 @@ function verify_captcha()
 				return true;
 			}
 		}
-		else if($mybb->input['quick_login'] == 1 && $mybb->input['quick_password'] && $mybb->input['quick_username'])
+		else if(isset($mybb->input['quick_login']) && $mybb->input['quick_login'] == 1 && $mybb->input['quick_password'] && $mybb->input['quick_username'])
```
Stefan-MyBB: You could probably use `!empty($mybb->input['quick_login'])` or $mybb->get_input() here.

Sama34: Try `$mybb->get_input('quick_login', \MyBB::INPUT_INT) === 1` instead.

smnscp: Thanks for your suggestions! Thinking it over, we only care about the truthiness of `quick_login`, not its numeric value. So, I prefer the `!empty` approach:

```suggestion
		else if(!empty($mybb->input['quick_login']) && $mybb->input['quick_password'] && $mybb->input['quick_username'])
```

Apart from that, for consistency’s sake we should check the other array keys, too, rather than assume that `quick_password` and `quick_username` must be set if `quick_login` is:

```suggestion
		else if(!empty($mybb->input['quick_login']) && !empty($mybb->input['quick_password']) && !empty($mybb->input['quick_username']))
```

Issue #3695, opened 2019-06-08
"Additional Contact Information" should be reworked.


send-your-nudes:
The problem is that some of the fields are old and need upgrading

1. Google Hangouts will be discontinued around 2020.
2. Yahoo Messenger is discontinued
3. ICQ is an old school along with Skype.

Now how to improve it:
Like kawaii or someone mentioned new standards are coming.
Maybe include fields such as:
Instagram, Facebook, Discord, Steam, Whatsapp, Viber, or anything you see as appropriate.

I think this shouldn' take much of the time.

Eldenroot:
All suggestion and bug reports should be placed on MyBB community board, not here.

@euantorano please close this one.

euantorano:
The plan is that there won’t be any built in contact fields, there will instead be some pre-set custom profile fields for contact fields. I believe this might already be implemented in 1.9.

euantorano:
@eldenroot we’re not too bothered about forcing people to use the forums anymore, so long as people don’t start using GitHub to post support questions that just confuse things.

Sick1G:
This is related to 1.8, if so, can we have it removed then?

PS: For me, posting enhancements looks way clearer here somehow.

effone:
I have seen discussions before that additional contact informations will be removed altogether to make use with custom profile fields; however I'd prefer an additional contact module which will be validated through Oauth before adding the field to profile. This data can further be used for core social login feature.

Sick1G:
Would that be for 1.8?

effone:
Ofcourse its possible even for 1.8 but I doubt such huge feature will be allowed in 1.8 branch.
..... or, will be accepted at all.

euantorano:
I’d definitely prefer a big change like that for 1.9 rather than 1.8

PR #4818, opened 2024-05-26
Remove ICQ

### LINKED ISSUE #4816: Remove ICQ field
ICQ will be shut down on June 26.
Source: https://icq.com/desktop/de#windows

effone:
Resolves #4816


Sama34:
Maybe we would be better off holding this for 1.9 ? It might be possible administrators would dislike losing this data in a minor update, it seems an abrupt change considering it can be simply disabled by default instead for 1.8

Stefan-MyBB:
It makes little sense to keep the data as it has simply become useless. The field is not removed because ICQ has lost importance, but because it has been shut down for good.

dvz:
It could be that some translations/boards repurpose the field for more relevant/local equivalents.

Stefan-MyBB:
AIM and Yahoo have been removed in a minor updates: #3199 #3695

Sama34:
This seems to affect the merge system. Next time, I think we should stick to keeping fields in the DB and drop them later in major versions.

PR #4824, opened 2024-05-29
Fixes #4811 Group promotions ignore the user display group when promo…

### LINKED ISSUE #4811: Group promotions ignore the user display group when promoting users
**Describe the bug**
When removing a user primary or secondary group which was set as their display group, the task process completely ignores this and leaves the display group with no modification, even if the user is no longer a member of such group.

**To Reproduce**
1. Make a user assign their primary group as their display group. Or make a user assign any of their secondary groups as their display group.
2. Use the promotion system to promote a user to a state where such primary or secondary group is lost.
4. A user will now have their display group intact, even if the user is no longer a member of such group.

**Expected behavior**
Once lost membership to any group, their display group should be updated accordingly if necessary.

**Additional context**
It might be argued this to be an expected behavior. However, I think it is not.

Eldenroot:
…ting users

Group promotions ignore the user display group when promoting users

Fixes #4811

### DIFF
--- a/inc/tasks/promotions.php
+++ b/inc/tasks/promotions.php
@@ -167,7 +167,7 @@ function task_promotions($task)
 			$plugins->run_hooks('task_promotions', $args);
 		}
 
-		$query2 = $db->simple_select("users", "uid,{$usergroup_select}", $sql_where);
+		$query2 = $db->simple_select("users", "uid,usergroup,additionalgroups,displaygroup,{$usergroup_select}", $sql_where);
 
 		$uids = array();
 		while($user = $db->fetch_array($query2))
@@ -203,7 +203,15 @@ function task_promotions($task)
 			}
 
 			$uids[] = $user['uid'];
+			if((int)$user['displaygroup'] === (int)$user['usergroup'] || in_array((int)$user['displaygroup'], array_map('intval', explode(',', $user['additionalgroups']))))
+			{
+				if(!isset($updateDisplayGroups))
+				{
+					$updateDisplayGroups = [];
+				}
 
+				$updateDisplayGroups[(int)$user['uid']] = (int)$user['uid'];
+			}
 
 			if($usergroup_select == "additionalgroups")
 			{
@@ -212,7 +220,12 @@ function task_promotions($task)
 					// Did the user already have the additional usergroup?
 					array_pop($log_inserts);
 					array_pop($uids);
-				}
+				} elseif(isset($updateDisplayGroups[(int)$user['uid']]))
+				{
+					$db->update_query('users', array('displaygroup' => 0), "uid='{$user['uid']}'");
+
+					unset($updateDisplayGroups[(int)$user['uid']]);
+                }
 			}
 
 			if((count($uids) % 20) == 0)
@@ -220,6 +233,15 @@ function task_promotions($task)
 				if($usergroup_select == "usergroup")
 				{
 					$db->update_query("users", array('usergroup' => $promotion['newusergroup']), "uid IN(".implode(",", $uids).")");
+
+					if(!empty($updateDisplayGroups))
+                    {
+						$userIDs = implode(',', $updateDisplayGroups);
+
+						$db->update_query('users', array('displaygroup' => 0), "uid IN ({$userIDs})");
+
+						$updateDisplayGroups = [];
+					}
 				}
 
 				if(!empty($log_inserts))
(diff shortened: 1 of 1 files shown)


inc/tasks/promotions.php
```diff
@@ -212,14 +220,28 @@ function task_promotions($task)
 					// Did the user already have the additional usergroup?
 					array_pop($log_inserts);
 					array_pop($uids);
-				}
+				} elseif(isset($updateDisplayGroups[(int)$user['uid']]))
```
effone: Code style: new line before `elseif`

inc/tasks/promotions.php
```diff
@@ -203,7 +203,15 @@ function task_promotions($task)
 			}
 
 			$uids[] = $user['uid'];
+			if((int)$user['displaygroup'] === (int)$user['usergroup'] || in_array((int)$user['displaygroup'], array_map('intval', explode(',', $user['additionalgroups']))))
+			{
+				if(!isset($updateDisplayGroups))
+				{
+					$updateDisplayGroups = [];
```
effone: Use `$updateDisplayGroups = array();`

inc/tasks/promotions.php
```diff
@@ -212,14 +220,28 @@ function task_promotions($task)
...
 				{
 					$db->update_query("users", array('usergroup' => $promotion['newusergroup']), "uid IN(".implode(",", $uids).")");
+
+					if(!empty($updateDisplayGroups))
+                    {
+						$userIDs = implode(',', $updateDisplayGroups);
+
+						$db->update_query('users', array('displaygroup' => 0), "uid IN ({$userIDs})");
+
+						$updateDisplayGroups = [];
```
effone: Use `$updateDisplayGroups = array();`

inc/tasks/promotions.php
```diff
@@ -237,6 +259,15 @@ function task_promotions($task)
...
 			{
 				$db->update_query("users", array('usergroup' => $promotion['newusergroup']), "uid IN(".implode(",", $uids).")");
+
+				if(!empty($updateDisplayGroups))
+				{
+					$userIDs = implode(',', $updateDisplayGroups);
+
+					$db->update_query('users', array('displaygroup' => 0), "uid IN ({$userIDs})");
+
+					$updateDisplayGroups = [];
```
effone: Use `$updateDisplayGroups = array();`

effone: Why this is required to reset?

Sama34: > Why this is required to reset?

Because `format_name` does not check for the actual membership of the user to the `displaygroup`. So users might be shown as _belonging_ to any group regardless of that being the case.

I think this is a bug, not a feature.

My suggestions should be tested, tho, as I did not test them myself.

inc/tasks/promotions.php
Eldenroot: Thx!

inc/tasks/promotions.php
Eldenroot: Thx!

inc/tasks/promotions.php
Eldenroot: Thx!

inc/tasks/promotions.php
```diff
@@ -167,7 +167,7 @@ function task_promotions($task)
 			$plugins->run_hooks('task_promotions', $args);
 		}
 
-		$query2 = $db->simple_select("users", "uid,{$usergroup_select}", $sql_where);
+		$query2 = $db->simple_select("users", "uid,usergroup,additionalgroups,displaygroup,{$usergroup_select}", $sql_where);
```
Sama34, line 170: `$usergroup_select` might seem redundant now, but it is passed by references in the plugin hook above, so the best would be to keep for now.

inc/tasks/promotions.php
```diff
@@ -203,7 +203,15 @@ function task_promotions($task)
 			}
 
 			$uids[] = $user['uid'];
+			if((int)$user['displaygroup'] === (int)$user['usergroup'] || in_array((int)$user['displaygroup'], array_map('intval', explode(',', $user['additionalgroups']))))
```
Sama34, line 206: From reading the code, it seems the conditional here should be inverted. It seems to me the code is resetting the display group of users even though they are members of such groups.

Maybe further testing is advised.

Sama34 (changes requested):

Sama34:
https://github.com/mybb/mybb/pull/4824#discussion_r1673382803

inc/tasks/promotions.php
Eldenroot: Not sure about this one... @effone ?

Sama34:
I think there is something fundamentally wrong with this suggested change and my suggested fix too.

Maybe a new review should be done to properly understand the issue and provide a better or proper solution to it.

PR #4825, opened 2024-05-29
Bypass Post Approval for Mods

### LINKED ISSUE #4623: Post edited by Admin requires approval again
I don't know whether MyBB script is like this or have I to modify something.

Steps:
1. I kept moderation tick for the post of a member on new post and also after editing.
2. Member created a thread so it requires approval.
3. Admin approve it via post approval moderation tool from the thread.
4. So, the thread is approved.
5. Now, Admin edit that member's thread then also that thread goes for approval. 

Point 5 is weird to me. 
As per my thinking, the post edited by the admin MUST not go for approval again.
Is anything need to be modified in permission to remove again required approval after editing or it is not possible to do in MyBB script.?

Hope you understand.
Thank you.

Original thread: [Post edited by Admin requires approval again](https://community.mybb.com/thread-234548.html)

effone:
Resolves #4623

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -1884,9 +1884,10 @@ class PostDataHandler extends DataHandler
 
 		// Decide on the visibility of this post.
 		$ismod = is_moderator($post['fid'], "", $uid);
-
+		$editor_ismod = is_moderator($post['fid'], "", $mybb->user['uid']);
+		
 		// Keep visibility for unapproved and deleted posts
-		if($existing_post['visible'] == 0)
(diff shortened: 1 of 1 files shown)


inc/datahandlers/post.php
```diff
@@ -1884,17 +1884,18 @@ function update_post()
 
 		// Decide on the visibility of this post.
 		$ismod = is_moderator($post['fid'], "", $uid);
-
+		$editor_ismod = is_moderator($post['fid'], "", $mybb->user['uid']);
```
Sama34, line 1887:
```suggestion
		$editor_ismod = is_moderator($post['fid']);
```

PR #4845, opened 2024-07-22
[PHP8 warnings] assign default value for $icon in editpost

### LINKED ISSUE #4844: PHP 8 Warning in editpost.php: undefined variable $icon
`$icon` is undefined when viewing a post preview (editpost.php)

**Analysis**
`$icon` is only defined if `$mybb->input['previewpost']` is **not** set.
https://github.com/mybb/mybb/blob/e12fc4a60781e733c16db81eb02bffa52570a4f6/editpost.php#L607-L610
`$icon` is only accessed if `$mybb->input['previewpost'] **is** set.
https://github.com/mybb/mybb/blob/e12fc4a60781e733c16db81eb02bffa52570a4f6/editpost.php#L822-L858

MyBB version: 1.8.38
PHP version: PHP 8.3.9

kikelkik:
Resolves #4844 

The Analysis in #4844 shows that `$icon` is always not defined when needed and vice-versa.

Maybe, instead of/additionally to providing a default value for the variable, the problem should be fixed directly? 
It seems to me that `if(!isset($mybb->input['previewpost']))` should be `if(!empty($mybb->input['previewpost']))` instead..?
But you never know, if `$icon` is used in any included template...

### DIFF
--- a/editpost.php
+++ b/editpost.php
@@ -231,7 +231,7 @@ if($mybb->settings['enableattachments'] == 1 && ($mybb->get_input('newattachment
 			$usage = $db->fetch_array($query);
 			$ret['usage'] = get_friendly_size($usage['ausage']);
 		}
-		
+
 		header("Content-type: application/json; charset={$lang->settings['charset']}");
 		echo json_encode($ret);
 		exit();
@@ -603,6 +603,7 @@ if(!$mybb->input['action'] || $mybb->input['action'] == "editpost")
 	$plugins->run_hooks("editpost_action_start");
 
 	$preview = '';
+	$icon = '';
 	if(!isset($mybb->input['previewpost']))
 	{
 		$icon = $post['icon'];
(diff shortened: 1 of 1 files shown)


Sama34:
The value of `mybb->input['previewpost']` is unimportant, so `isset()` is preferable. We just interchangeably use it around the code.

It seems to me that `{$icon}` isn't used in any default template related to the edit post page, so I think you could just replace `$postinfo['icon'] = $icon;` with `$postinfo['icon'] = $post['icon'];` and remove the first code block you shared in the issue report.

kikelkik:
Sounds good - I changed the code accordingly

kikelkik:
@Sama34 can we merge this PR or is there anything else to do here?

Sama34:
I was waiting for feedback, but yes, it seems fine to me.

Thank yo for your contribution !

PR #4867, opened 2024-12-23
Warning [2] Undefined array key 1 - Line: 801 - File: forumdisplay.ph…

Sama34:
Also:
`Warning [2] Undefined variable $codeinsert - Line: 3881 - File: inc/functions.php PHP 8.3.6 (Linux)`
`Warning [2] Undefined array key "processed" - Line: 2732 - File: admin/modules/user/users.php PHP 8.3.6 (Linux)`

### DIFF
--- a/inc/class_core.php
+++ b/inc/class_core.php
@@ -283,7 +283,7 @@ class MyBB {
 		}
 		else
 		{
-			$this->request_method = "";
+			$this->request_method = "get";
 		}
 
 		// If we've got register globals on, then kill them too
--- a/inc/class_core.php
+++ b/inc/class_core.php
@@ -277,13 +277,17 @@ class MyBB {
 		$this->parse_incoming($_GET);
 		$this->parse_incoming($_POST);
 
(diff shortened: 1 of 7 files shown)


inc/class_core.php
```diff
@@ -277,14 +277,18 @@ function __construct()
...
 			$this->request_method = "post";
 		}
-		else if($_SERVER['REQUEST_METHOD'] == "GET")
+		else if(isset($_SERVER['REQUEST_METHOD']) && $_SERVER['REQUEST_METHOD'] == "GET")
 		{
 			$this->request_method = "get";
 		}
+		else
+		{
+			$this->request_method = "";
```
Stefan-MyBB: Wouldn't it make sense to default to "get"? MyBB doesn't support other request methods anyway.

Sama34: Probably, I just went with the logic around it.

dvz: There may be some plugins with checks like `$mybb->request_method == "get"` - we may want to keep the original logic, with a similar fix.

PR #4871, opened 2025-01-06
Fixes #4868 - PHP 8.4 mybb error when logging in

### LINKED ISSUE #4868: PHP 8.4 mybb error when logging in
There seems to be an "Authorization code mismatch. Are you accessing this function correctly? Please go back and try again." error when logging in php8.4

This seems to happen the very first forum end login after an installation.

Original thread: [PHP 8.4 mybb error when logging in](https://community.mybb.com/thread-242648.html)

Eldenroot:
Resolves #4868

### DIFF
--- a/inc/class_session.php
+++ b/inc/class_session.php
@@ -65,7 +65,14 @@ class session
 		// Attempt to find a session id in the cookies.
 		if(isset($mybb->cookies['sid']) && !defined('IN_UPGRADE'))
 		{
-			$sid = $db->escape_string($mybb->cookies['sid']);
+			$cookies = explode('; ', $_SERVER['HTTP_COOKIE']);
+			$mycookie = [];
+			foreach($cookies as $cookie)
+			{
+				$kv = explode('=', $cookie);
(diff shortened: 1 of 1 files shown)


inc/class_session.php
```diff
@@ -65,7 +65,14 @@ function init()
 		// Attempt to find a session id in the cookies.
 		if(isset($mybb->cookies['sid']) && !defined('IN_UPGRADE'))
 		{
-			$sid = $db->escape_string($mybb->cookies['sid']);
+			$cookies = explode('; ', $_SERVER['HTTP_COOKIE']);
```
Stefan-MyBB, line 68: This looks like a weird workaround to me. If there is an issue with `$mybb->cookies` (likely affecting other features) that should be fixed instead.

inc/class_session.php
```diff
@@ -65,7 +65,14 @@ function init()
...
+			$mycookie = [];
```
Stefan-MyBB, line 69: This syntax is not compatible with older PHP versions.

PR #4893, opened 2025-04-15
fix: uninitialized errors array in ACP user module

### LINKED ISSUE #4835: ACP error in absence settings (/admin/modules/user/users.php)
Editing the absence settings in the ACP
If the returndate is selected in the past, there will be an error 500 with the following:

Fatal error: Uncaught Error: [] operator not supported for strings in .../admin/modules/user/users.php:824 Stack trace: #0 .../admin/index.php(834): require() #1 {main} thrown in .../admin/modules/user/users.php on line 824

For quick fix add the following to line 816 of /admin/modules/user/users.php
$errors = array();

Original thread: [ACP error in absence settings (/admin/modules/user/users.php)](https://community.mybb.com/thread-241186.html)

arcalys:
- Initalized `errors` array.

Resolves #4835

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -813,6 +813,8 @@ if($mybb->input['action'] == "edit")
 				}
 			}
 
+			$errors = array();
+
 			if(!empty($extra_user_updates['moderateposts']) && !empty($extra_user_updates['suspendposting']))
 			{
 				$errors[] = $lang->suspendmoderate_error;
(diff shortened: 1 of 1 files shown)


dvz:
Could that overwrite previously assigned `$errors`?

Perhaps
https://github.com/mybb/mybb/blob/0bdad6909064d314983de1dd65c5c5425dfcb905/admin/modules/user/users.php#L592
should be changed instead.

arcalys:
> Could that overwrite previously assigned `$errors`?
> 
> Perhaps
> 
> https://github.com/mybb/mybb/blob/0bdad6909064d314983de1dd65c5c5425dfcb905/admin/modules/user/users.php#L592
> 
> should be changed instead.

You're right, updated the PR

PR #4894, opened 2025-04-16
fix: support IP addresses in mycode_auto_url

### LINKED ISSUE #3813: mybb doesn't parse IP address urls in threads

### LINKED ISSUE #4727: Parser assumes that characters have been encoded in a post when HTML is enabled
The following line assumes that characters have been encoded :
https://github.com/mybb/mybb/blob/35c62a1d607fff2332a7229ef06702ac3fac0a88/inc/class_parser.php#L1693

Which is untrue when HTML is enabled.

Reference :
https://community.mybb.com/thread-237311-post-1382420.html#pid1382420

arcalys:
### Changed

- Modified the expression for `mycode_auto_url` to support additional IPv4/IPv6 parsing:
  - Added `[?#][^\"\s<\[&]*` as an alternative for path/query/fragment parsing.
  - Introduced a negative lookbehind (`(?<![.,;:'\"?!])`) to account for the removal of `[\w\/\)]` (trailing punctuation).
- Updated host expression:
  - Explicitly allowed dashes in domain labels.
  - Explicitly excluded closing bracket from valid characters.
- Split the path/query/fragment expression for better readability.

### Fixed
- Fixed handling of query/fragment without a preceding / (e.g., example.com?query, example.com#fragment).
- Fixed IPv4 address-only URLs that end with a single digit (e.g., 1.2.3.4).
- Fixed IPv6 addresses.

Resolves #3813 
Resolves #4727

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1695,19 +1695,23 @@ class postParser
 		// Don't create links within existing links (handled up-front in the callback function).
 		$message = preg_replace_callback(
 			"~
-				<a\\s[^>]*>.*?</a>|								# match and return existing links
-				(?<=^|[\s\(\)\[\>])								# character preceding the link
+				<a\\s[^>]*>.*?</a>|													# match and return existing links
+				(?<=^|[\s\(\)\[\>])													# character preceding the link
 				(?P<prefix>
-					(?:http|https|ftp|news|irc|ircs|irc6)://|	# scheme, or
-					(?:www|ftp)\.								# common subdomain
+					(?:http|https|ftp|news|irc|ircs|irc6)://|						# scheme, or
+					(?:www|ftp)\.													# common subdomain
 				)
 				(?P<link>
-					(?:[^\/\"\s\<\[\.]+\.)*[\w]+				# host
-					(?::[0-9]+)?								# port
-					(?:/(?:[^\"\s<\[&]|\[\]|&(?:amp|lt|gt);)*)?	# path, query, fragment; exclude unencoded characters
-					[\w\/\)]
+					(?:
+						\[[0-9a-fA-F:]+(?:%[0-9a-zA-Z._-]+)?\]|						# IPv6 address with optional zone
+						(?:(?:\d{1,3}\.){3}\d{1,3})|								# IPv4 address
+						(?:[^\/\"\s<\[\]\.]+\.)*[\w-]+								# domain name
+					)
+					(?::[0-9]+)?													# optional port number
+					(?:/(?:[^\"\s<\[&]|\[\]|&(?:amp|lt|gt);)*|[?#][^\"\s<\[&]*)?	# optional path, query or fragment (excluding unencoded characters)
 				)
-				(?![^<>]*?>)									# not followed by unopened > (within HTML tags)
+				(?<![.,?!])															# exclude trailing punctuation
+				(?![^<>]*?>)														# not followed by unopened > (within HTML tags)
 			~iusx",
 			array($this, 'mycode_auto_url_callback'),
 			$message
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1710,7 +1710,7 @@ class postParser
 					(?::[0-9]+)?													# optional port number
 					(?:/(?:[^\"\s<\[&]|\[\]|&(?:amp|lt|gt);)*|[?#][^\"\s<\[&]*)?	# optional path, query or fragment (excluding unencoded characters)
 				)
-				(?<![.,?!])															# exclude trailing punctuation
+				(?<![.,;:`'\"?!])													# exclude trailing punctuation
 				(?![^<>]*?>)														# not followed by unopened > (within HTML tags)
 			~iusx",
 			array($this, 'mycode_auto_url_callback'),
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1695,25 +1695,25 @@ class postParser
 		// Don't create links within existing links (handled up-front in the callback function).
 		$message = preg_replace_callback(
 			"~
-				<a\\s[^>]*>.*?</a>|													# match and return existing links
-				(?<=^|[\s\(\)\[\>])													# character preceding the link
+				<a\\s[^>]*>.*?</a>|									# match and return existing links
+				(?<=^|[\s\(\)\[\>])									# character preceding the link
 				(?P<prefix>
-					(?:http|https|ftp|news|irc|ircs|irc6)://|						# scheme, or
-					(?:www|ftp)\.													# common subdomain
+					(?:http|https|ftp|news|irc|ircs|irc6)://|		# scheme, or
+					(?:www|ftp)\.									# common subdomain
 				)
 				(?P<link>
 					(?:
-						\[[0-9a-fA-F:]+(?:%[0-9a-zA-Z._-]+)?\]|						# IPv6 address with optional zone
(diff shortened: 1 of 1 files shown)


arcalys:
Tested with the following values:
```
http://[::1]
http://[2001:4860:4860::8888]
http://[2001:db8::2]:8080/status
http://[fe80::1%eth0]
http://[fe80::1%eth0]:8080/path/to/file
http://1.1.1.1:3000/resource
http://1.2.3.4:8000
http://1.2.3.4?query=val
http://10.10.10.10:9000/test?debug=true
http://127.0.0.1
http://127.0.0.1:3000/dashboard
http://127.0.0.1:8080/path/to/resource#fragment
http://192.168.100.200:8080/index.html
http://example.co.uk
http://example.com
http://exämple.com 
http://example.com:443?debug=true 
http://example.com:80
http://example.com:8080/test
http://example.com/abc/xyz#frag
http://example.com/test?param=foo 
http://internal.service.local
http://localhost
http://localhost:12345/some/path?foo=bar#anchor
http://localhost:8080#frag
http://localhost?debug=true
http://localhost/abc/xyz#anchor
http://printer.local
http://xn--d1acpjx3f.xn--p1ai 
https://[::1]:443/
https://[::1]#section
https://[2001:db8::1] 
https://[2001:db8::1]?query=test
https://[2001:db8::1]/path/to/page#section
https://[2001:db8::2]:443/test/path
https://[2001:db8:abcd:0012::0]/abc/def
https://[2606:4700:4700::1111]:443/
https://[fd00::1234]:8443/
https://[fe80::1%eth0] 
https://[fe80::1%eth0]#header
https://[fe80::1%lo0]:9090/ 
https://1.2.3.4
https://192.168.1.100/api/v1/
https://example.com:443?param=value
https://example.com:443/
https://example.com?param=value
https://example.com/path/to/file
https://example.com/path/to/file?query=test
https://example.com/path#fragment
https://example.com#fragment
https://localhost:8080/
https://localhost?query=value
https://localhost#anchor
https://my-site.internal:10443/dashboard
https://some.long.subdomain.example.co.uk:5000/query?foo=bar
https://some.sub.domain.example.com
https://somehost:8443
```

dvz:
Atomic changes for reference:
- _host_ / _domain name_:
  - `\<` changed to `<`
  - added `\]`
  - added `-` in last character class
- _path, query, fragment_:
  - added `[?#][^\"\s<\[&]*` alternative
- `[\w\/\)]` removed (introduced via https://github.com/mybb/mybb/commit/3a4afb63df6dec01f9f5c790f14c6516ea5d27f4)
- added `(?<![.,?!])` negative lookbehind

Fixes:
- IPv6 addresses
- IPv4 address-only URLs ending with a single digit (`1.2.3.4`)
- query/fragment without preceding `/` (`example.com?query`, `example.com#fragment`)
---

1. As `[\w\/\)]` was removed, this seems to result in unlisted trailing characters being ultimately included in links (e.g. `:`, `;`, `` ` ``). It may be enough to add these (and others, if any).

   Similarly, we can consider adding these to `mycode_auto_url_callback`.

2. We should make the _path, query, fragment_ expression more readable at this point, e.g.:
   ```regex
   (?:/[^\"\s<\[&?#]*)?				# optional path
   (?:\?(?:[^\"\s<\[?#]|&(?:amp|lt|gt);)*)?	# optional query
   (?:\#(?:[^\"\s<\[]|&(?:amp|lt|gt);)*)?		# optional fragment
   ```
   (also: non-URL-encoded `&` disallowed in path, and allowed in fragment)

3. If you can, can you amend the commit to remove unrelated whitespace changes?

   The current _common subdomain_ comment may be more accurate, as only one is expected in the processed URL.

---

Related: The `dev-1.9` branch has a [`ParserTest`](https://github.com/mybb/mybb/blob/c7539bea0f3670511f1c68a10706c95b5cbe3baf/tests/Unit/ParserTest.php) - once this change is propagated, we can add an appropriate test for links and punctuation.

arcalys:
Hi @dvz 

- Removed the unrelated white space changes, sorry, editor habits 
- Restored the original comment for subdomain
- Added extra trailing punctuation (`;`,`:`, `'`, `"` and `)

I'll split the path/fragment part tomorrow

arcalys:
Split the path/query/fragment expression. Do you want the extra punctuation added to `mycode_auto_url_callback`?  
Also updated the PR description to match everything you pointed out.

dvz:
We can probably leave the callback unless anything comes up that's worth addressing (otherwise `[url]` can handle special cases).

Some circumstances to fix:
```
[http://example.com?query]
http://example.com/search.php?forums[]=123
```

by:
- extending the negated character classes with the closing `]`
- adding back the `[` `]` sequence alternative for _query_ (previously left out)

so something like:
```regex
(?:/[^\"\s<\[\]&?#]*)?				# optional path
(?:\?(?:[^\"\s<\[\]?#]|\[\]|&(?:amp|lt|gt);)*)?	# optional query
(?:\#(?:[^\"\s<\[\]]|&(?:amp|lt|gt);)*)?	# optional fragment
```

The indentation for comments can now be shortened (GitHub may render these as misaligned sometimes). There are some accidental spaces in between, and tabs after comments.

inc/class_parser.php
```diff
@@ -1695,19 +1695,25 @@ function mycode_auto_url($message)
...
-					[\w\/\)]
+					(?:
+						\[[0-9a-fA-F:]+(?:%[0-9a-zA-Z._-]+)?\]|		# IPv6 address with optional zone
+						(?:(?:\d{1,3}\.){3}\d{1,3})|				# IPv4 address
+						(?:[^\/\"\s<\[\]\.]+\.)*[\w-]+				# domain name
+					)
+					(?::[0-9]+)?									# optional port number
+					(?:/[^\"\s<\[\]&?#]*)?							# optional path
+					(?:\?(?:[^\"\s<\[\]?#]|\[\]|&(?:amp|lt|gt);)*)?	# optional query
+					(?:\#(?:[^\"\s<\[\]]|&(?:	|lt|gt);)*)?		# optional fragment
```
dvz: Looks like `amp` got replaced with a tab here

dvz (changes requested):

inc/class_parser.php
arcalys: Fixed, sorry about that.

dvz:
It looks like we'd have a regression into a previously fixed issue: #4303
```
http://example.com/?query&amp;
```

We may be able to avoid it by allowing certain entities to be included with the `;`:
```regex
(?:
    (?<=&amp;)|(?<=&lt;)|(?<=&gt;)|	# allow trailing entities
    (?<![.,;:`'\"?!])			# exclude other trailing punctuation
)
```

The special entity capturing could be removed as well - letting encoded entities to be captured by the general expression:
```diff
- (?:\?(?:[^\"\s<\[\]?#]|\[\]|&(?:amp|lt|gt);)*)?
+ (?:\?(?:[^\"\s<\[\]?#]|\[\]                )*)?
```
```diff
- (?:\#(?:[^\"\s<\[\]]|&(?:amp|lt|gt);)*)?
+ (?:\#   [^\"\s<\[\]]                 *)?
```

arcalys:
Updated:
- `http://domain.example/&` links to `http://domain.example`
- `http://example.com/?query&amp;` links to `http://example.com?query&amp;`

PR #4896, opened 2025-04-19
Add a post_session_load hook to mirror the pre_ one

Gargaj:


dvz:
Can you share about the use cases for the new hook?

Gargaj:
The obvious one would be a custom login system where you get to evaluate if the user already has a MyBB session and if not, perhaps use some other means to log them in (e.g. via some another type of session).

Sama34:
Coincidently, I recently reviewed the possibility for a plugin to hook in there, but the existing `pre_session_load` did seem lacking.

I agree some hook should be added, the suggested position seems appropriate for me too but could be different.

@Gargaj is the login data handler hooks not sufficient for your requirement ?

Gargaj:
> @Gargaj is the login data handler hooks not sufficient for your requirement ?

No, because I want to abolish the (forum) login entirely and want the session data to come from somewhere else.

PR #4899, opened 2025-04-20
fix(installer): duplicate port in suggested board URL

arcalys:
### Fixed

- Duplicate non standard port when port is already present in `$_SERVER['HTTP_HOST']`.

### Example

image


dvz:
Thanks; can we test for `:` alone? We can also use `str_contains()`.

The 3 conditions be can now split into separate lines, as well.

arcalys:
> Thanks; can we test for `:` alone? We can also use `str_contains()`.
> 
> The 3 conditions be can now split into separate lines, as well.

Updated

PR #4900, opened 2025-04-20
fix: SQL error when accessing private inbox

arcalys:
When trying to access the private messages inbox, got the following error:
```
SQL Error
    0 - ERROR: invalid input syntax for type smallint: "Array" LINE 1: ...l FROM mybb_privatemessages WHERE uid='1' AND folder='Array' ^
Query
    SELECT COUNT(*) AS total FROM mybb_privatemessages WHERE uid='1' AND folder='Array'
```

### Fixed

- Used `fid` instead of `folder`.

### DIFF
--- a/private.php
+++ b/private.php
@@ -2029,7 +2029,7 @@ if(!$mybb->input['action'])
 	$private['orderarrow'][$sortby] = true;
 
 	// Do Multi Pages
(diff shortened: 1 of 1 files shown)


dvz:
It looks like there are ~7 more references to `$folder` expecting it to be an ID in the file.

PR #4902, opened 2025-04-21
fix(admin): incorrect success message when attempting to delete own account

### LINKED ISSUE #4442: Fix success message when attempting to delete yourself

arcalys:
### Fixed

- Displayed an error message instead.

Based on https://github.com/mybb/mybb/pull/4443
Resolves #4442

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1806,6 +1806,12 @@ if($mybb->input['action'] == "delete")
 		admin_redirect("index.php?module=user-users");
 	}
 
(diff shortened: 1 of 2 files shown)


arcalys:
This behavior also exists in 1.8.x. Should this be directed to the feature branch @dvz?

dvz:
It's low priority, so we can keep it as is.

PR #4903, opened 2025-04-22
refactor(acp): use local fonts in theme

### LINKED ISSUE #4838: Use local fonts in ACP
Replace references to remote assets with local ones in Admin CP:
- https://github.com/mybb/mybb/blob/f591907c17af916291dce8824a688edafe36399b/admin/styles/default/main.css#L6
- https://github.com/mybb/mybb/blob/f591907c17af916291dce8824a688edafe36399b/admin/styles/default/login.css#L6

Some fonts may be available in https://github.com/mybb/mybb/tree/dev-1.9/jscripts/fonts

arcalys:
### Changed

- Used locally available fonts (Asap, Open Sans) and moved their definition to a dedicated `fonts.css`.

Fixes #4838

### DIFF
--- a/admin/styles/default/fonts.css
+++ b/admin/styles/default/fonts.css
@@ -1,0 +1,20 @@
+@font-face {
+    font-family: 'Asap';
+    src: url('/jscripts/fonts/Asap/Asap-VariableFont_wght.ttf') format('truetype');
+    font-weight: 400 600;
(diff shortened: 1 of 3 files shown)


dvz:
We may need to use relative paths instead to keep sub-directory installations working  (`example.com/forum/admin/`).

PR #4911, opened 2025-04-24
fix: use new_not_null and new_default_value in 1.8.39 upgrade script

arcalys:
### Fixed

- Incorrect `modify_column` calls (`varchar(120) NOT NULL default ''` one liner is invalid in PostgreSQL).


dvz:
Introduced in #4864

PR #4919, opened 2025-05-04
fix: quotes in font mycode

### LINKED ISSUE #2458: Font BBCODE
Hello,
From Class_parser.php:
....
`$nestable_mycode['font']['regex'] = "#\[font=([a-z0-9 ,\-_'\"]+)\](.*?)\[/font\]#si";
$nestable_mycode['font']['replacement'] = "<span style=\"font-family: $1;\">$2</span>";`
...

this BBCODE : [font=arial"] anything[/font]  produce invalid tag:

`<span style="font-family: arial";">anything</span>`

this solve the problem:
`$nestable_mycode['font']['regex'] = "#\[font=([a-z0-9 ,\-_]+)\](.*?)\[/font\]#si";`
the best way is limit font choice: $defined_user_fonts
`$nestable_mycode['font']['regex'] = "#\[font=(arial|tahoma)\](.*?)\[/font\]#si";`

arcalys:
### Fixed

- Remove leading/trailing quotes and spaces from matched font name to prevent invalid CSS generation as seen in the related issue.

Resolves #2458

### DIFF
--- a/inc/class_parser.php
+++ b/inc/class_parser.php
@@ -1220,12 +1220,30 @@ class postParser
 	*/
 	function mycode_parse_font_callback($matches)
 	{
-		// Replace any occurrence(s) of double quotes in fonts with single quotes.
-		// A back-fix for double-quote-containing MyBB font tags in existing
-		// posts prior to the client-side aspect of this fix for the
-		// browser-independent SCEditor bug of issue #4182.
-		$fonts = str_replace('"', "'", $matches[2]);
-
+		preg_match_all("/
+			(?:                                 
+				\"(?:[^\"]*)\"|	# double-quoted font                    
+				'(?:[^']*)'|	# single-quoted font 
+				[^,]+			# unquoted font name (anything but comma)
+			)
+		/x", $matches[2], $matchedFonts);
+		
+		$cleanFonts = array();
+		
+		foreach ($matchedFonts[0] as $font)
+		{
+			// Remove leading/trailing quotes and spaces
+			$font = trim($font, "\"' ");
+	
+			// If font name contains a space or comma, wrap it in single quotes
+			if (strpos($font, ' ') !== false || strpos($font, ',') !== false) {
+				$font = "'{$font}'";
+			}
+	
+			$cleanFonts[] = $font;
+		}
+	
+		$fonts = implode(', ', $cleanFonts);
 		return "<span style=\"font-family: {$fonts};\" class=\"mycode_font\">{$matches[3]}</span>";
 	}
 


inc/class_parser.php
```diff
@@ -1220,11 +1220,9 @@ function mycode_parse_url($url, $name="")
...
 	function mycode_parse_font_callback($matches)
 	{
-		// Replace any occurrence(s) of double quotes in fonts with single quotes.
-		// A back-fix for double-quote-containing MyBB font tags in existing
-		// posts prior to the client-side aspect of this fix for the
-		// browser-independent SCEditor bug of issue #4182.
-		$fonts = str_replace('"', "'", $matches[2]);
+		// Clean font name by removing quotes and extra spaces for safe CSS output.
+		$fonts = trim($matches[2], "\"' ");
+		$fonts = str_replace(array("'", '"'), '', $fonts);
```
lairdshaw: It seems to me that removing all internal quotes may cause problems. Per the [MDN font-family article](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family): "String values must be quoted".

inc/class_parser.php
```diff
@@ -1220,12 +1220,23 @@ function mycode_parse_url($url, $name="")
 	*/
 	function mycode_parse_font_callback($matches)
 	{
-		// Replace any occurrence(s) of double quotes in fonts with single quotes.
-		// A back-fix for double-quote-containing MyBB font tags in existing
-		// posts prior to the client-side aspect of this fix for the
-		// browser-independent SCEditor bug of issue #4182.
-		$fonts = str_replace('"', "'", $matches[2]);
-
+		$fonts = explode(',', $matches[2]);
```
lairdshaw: This seems to assume that commas do not (cannot) occur within quoted font families. Given my reading of that same MDN article, this assumption is false.

inc/class_parser.php
arcalys: Hi,
You're right. I’ve updated the callback function to properly loop through the font list and wrap font names with spaces in quotes (which wasn’t happening before). Let me know if you'd prefer it handled differently.

lairdshaw: Belatedly (I didn't see your response until after my below critique): re your fix, it seems justifiable and even preferable to auto-quote any font-family containing spaces as that fix now does, however, it seems to me that commas in already-quoted font-families should be ignored when exploding the fonts into that list.

arcalys: Updated the function to use a regex that preserves commas inside quoted strings.

Tested with the following:
```
[font="Arial, Helvetica, sans-serif"]Arial[/font]
[font='Times New Roman, serif']Times New Roman with comma[/font]
[font=Verdana]Verdana[/font]
[font="Comic Sans MS, cursive, sans-serif"]Comic Sans MS[/font]
[font="Special Font"]Special Font[/font]
```
Parsed versions:
```html
<span style="font-family: Arial, Helvetica, sans-serif;" class="mycode_font">Arial</span><br>
<span style="font-family: 'Times New Roman, serif';" class="mycode_font">Times New Roman with comma</span><br>
<span style="font-family: Verdana;" class="mycode_font">Verdana</span><br>
<span style="font-family: 'Comic Sans MS', cursive, sans-serif;" class="mycode_font">Comic Sans MS</span><br>
<span style="font-family: 'Special Font';" class="mycode_font">Special Font</span>
```

lairdshaw: LGTM.

PR #4922, opened 2025-05-07
fix(parser): code blocks newlines and tabs/spaces

### LINKED ISSUE #2849: [code] Tags don't preserve whitespace in 1.8.12
In addition, if a custom css rule is used to set `white-space: pre`, every newline is doubled. (This is due to the nl2br function which should be edited to not run on <code> tags).

arcalys:
### Added

- Added `white-space` and `tab-size` properties to `.codeblock code`.

### Changed

- Updated the logic around `nl2br` usage in `parse_message` to avoid using it for code blocks.

### Removed

- Replacements for tabs and spaces in `mycode_parse_code`.

---
Tested with spaces, tabs, mixed tabs and spaces, after upgrading the theme.
Resolves #2849


Sama34:
Thank you for fixing this annoying bug.

Sama34:
Seems like cherry-picking won't cut it for this, as the differences between 1.8 and 1.9 make it so that a review is necessary.

@arcalys could you push a PR for 1.9?

dvz:
Does it rely on CSS to function properly? Note that the theme's CSS file is not updated automatically during upgrade. If so, it could be changed just in 1.9.

arcalys:
> Does it rely on CSS to function properly? Note that the theme's CSS file is not updated automatically during upgrade. If so, it could be changed just in 1.9.

Yeah it does rely on CSS

Sama34:
I did post a link to this page in the internal thread for the next release to keep a track of it.

PR #4929, opened 2025-05-10
enhancement: add composer config.autoloader-suffix

### LINKED ISSUE #2960: Set config.autoloader-suffix for Composer
Set the [`autoloader-suffix`](https://getcomposer.org/doc/06-config.md#prepend-autoloader) Composer config value in `composer.json` to prevent random values being generated on each run. Allows easier package comparison and build reproducibility.

arcalys:
### Added

- `autoloader-suffix` to composer configuration.

Resolves #2960

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "config": {
+        "autoloader-suffix": "MyBB",
         "vendor-dir": "inc/vendor",
         "bin-dir": "inc/vendor/bin",
         "optimize-autoloader": true,


dvz:
Are there downsides to simply using `MyBB`?

[Drupal included](https://www.drupal.org/project/drupal/issues/2176265) a number, but it looks like they needed to avoid conflicts with hardcoded autoloader files.

arcalys:
The only scenario I can think of where this could be a problem is in automated tests where multiple isolated instances might be booted within the same PHP process. I don't know what you have planned on that side, but it can probably be set to MyBB and revisited later if necessary.

dvz:
Sounds good, let's use `MyBB` then.

arcalys:
> Sounds good, let's use `MyBB` then.

Updated

PR #4933, opened 2025-05-19
Replace `utf8_decode` Fixes #4807

### LINKED ISSUE #4807: `utf8_*` functions deprecated since PHP 8.2
snap

6 existences in code base.
3 in xmlhttp
1 in functions
2 in string (3rd party library - diff)

Sama34:
Partially resolves #4807

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,8 @@
     },
     "require": {
         "php": "^8.2",
-        "ext-pdo": "*",
         "ext-json": "*",
+        "ext-pdo": "*",
         "illuminate/config": "^10.0",
         "illuminate/container": "^10.0",
         "illuminate/events": "^10.0",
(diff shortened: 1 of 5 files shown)


arcalys:
Perhaps we should:
- Check if `mb_convert_encoding()` is available.
- If not, check if `iconv()` is available.
- If neither is available, trigger an error.

Sama34:
It seems `\Symfony\Polyfill\Mbstring\mb_convert_encoding()` should be available. Should we use that instead ? @dvz

I think we should use that instead.

dvz:
It may make it easier (then we can simply call `mb_*` functions without handling the `iconv` fallback everywhere).

In that case, we'll be able to remove/simplify the `mbstring` & `iconv` conditions throughout the codebase.

Note that the simple polyfill still requires `iconv` (so the _mbstring or iconv_ requirement wouldn't change).

PR #4941, opened 2025-05-24
fix: disallow HTTP access to inc/

### LINKED ISSUE #4676: Disallow HTTP access to `inc/`
Extend default server configuration directives to deny access to files under `inc/`.

Related:
- https://github.com/mybb/meta/blob/main/architecture/directory-structure.md

arcalys:
### Added

- `inc/` location rule to Nginx and Apache configuration examples.

Resolves #4676

### DIFF
--- a/htaccess-nginx.txt
+++ b/htaccess-nginx.txt
@@ -9,6 +9,10 @@ location /admin/backups {
     deny all;
 }
 
+location ~ /inc/ {
+    internal;
+}
+
 rewrite ^/forum-([0-9]+).html$ /forumdisplay.php?fid=$1 last;
 rewrite ^/forum-([0-9]+)-page-([0-9]+).html$ /forumdisplay.php?fid=$1&page=$2 last;
 rewrite ^/thread-([0-9]+).html$ /showthread.php?tid=$1 last;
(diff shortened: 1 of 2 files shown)


Sama34:
To be honest, I only tested on nginx, but the htaccess change seems fine.

dvz:
https://httpd.apache.org/docs/2.4/mod/core.html#location

Does it work as expected?

`admin/backups/` has a `.htaccess` included (https://github.com/mybb/mybb/blob/dev-1.9/admin/backups/.htaccess)

Sama34:
You are correct, nothing seems to work if attempting this from within the root `.htaccess` file outside the `/inc/` directory.

@arcalys could you simply copy the `/admin/backups/.htaccess` file inside the `/inc/` directory ? That just works.

The modifications to the `.htaccess` file in this PR simply breaks the forum, my bad.

arcalys:
I will do that @Sama34 
Sorry I was hoping for someone else to test on Apache and did not think it would be merged right away, am only using Nginx

https://github.com/mybb/mybb/pull/4986

Sama34:
To be frank, Apache sucks (specially because I don't use it 🙃) I had to test XAMPP to test this.

It was ultimately my fault for finger crossing that it was functional without testing.

PR #4943, opened 2025-05-24
fix(theme): thread rating visuals

arcalys:
### Added

- Added extra styling for thread rating to fix the visuals.

### Example

Before:
image

After:
image

### DIFF
--- a/inc/themes/core.base/frontend/styles/pages/_thread.scss
+++ b/inc/themes/core.base/frontend/styles/pages/_thread.scss
@@ -114,3 +114,77 @@
         }
     }
 }
+
+.inline_rating {
+    display: flex;
+    align-items: center;
+    margin: 10px 0;
+}
+
+.star_rating {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    display: flex;
+    align-items: center;
+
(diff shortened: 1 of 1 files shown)


Sama34:
I recall ratings are leaving, we are dropping that feature.

arcalys:
> I recall ratings are leaving, we are dropping that feature.

I’ve seen the issue, but they’re still there for now. I’m not sure if they’ll actually be dropped in 1.9 or if the removal will be postponed.

dvz:
https://github.com/mybb/mybb/issues/3498
https://community.mybb.com/thread-220554.html

We could comment out the threads bits, as was done in `forumdisplay.twig` to clean it up - before a wider PR to remove the rating-related code that may also point out what should be included in a potential plugin (@kawaii seemed interested).

arcalys:
> #3498 https://community.mybb.com/thread-220554.html
> 
> We could comment out the threads bits, as was done in `forumdisplay.twig` to clean it up - before a wider PR to remove the rating-related code that may also point out what should be included in a potential plugin (@kawaii seemed interested).

I will comment it out then. Closing this.

PR #4944, opened 2025-05-24
feat: add groups legend management

arcalys:
This is an attempt to integrate a groups legend into the "Who's online" block on the index page. It’s certainly not perfect and may need to be updated or improved, but for now, it’s working well for me.

### Added

- Added a new setting (defaults to 1) to control whether the groups legend is displayed.
- Added a new toggle in the groups management page to decide if a group should appear in the legend. Defaults to 1 for Administrators, Super Moderators, and Moderators.
- Added a new element in the Twig index view to display the groups legend above the members list, if enabled.

### Example

Index page:
image

Setting in the ACP > Forum Home Options:
image

Group management:
image

Feedback and suggestions are welcome.

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -926,6 +926,7 @@ if($mybb->input['action'] == "edit")
 				"canbereported" => $mybb->get_input('canbereported', MyBB::INPUT_INT),
 				"canbeinvisible" => $mybb->get_input('canbeinvisible', MyBB::INPUT_INT),
 				"canchangewebsite" => $mybb->get_input('canchangewebsite', MyBB::INPUT_INT),
+				"showinlegend" => $mybb->get_input('showinlegend', MyBB::INPUT_INT),
 				"showforumteam" => $mybb->get_input('showforumteam', MyBB::INPUT_INT),
 				"usereputationsystem" => $mybb->get_input('usereputationsystem', MyBB::INPUT_INT),
 				"cangivereputations" => $mybb->get_input('cangivereputations', MyBB::INPUT_INT),
@@ -1072,6 +1073,7 @@ if($mybb->input['action'] == "edit")
 
 	$general_options = array();
 	$general_options[] = $form->generate_check_box("showmemberlist", 1, $lang->member_list, array("checked" => $mybb->get_input('showmemberlist', MyBB::INPUT_INT)));
+	$general_options[] = $form->generate_check_box("showinlegend", 1, $lang->show_in_legend, array("checked" => $mybb->get_input('showinlegend', MyBB::INPUT_INT)));	
 	if($usergroup['gid'] != "1" && $usergroup['gid'] != "5")
 	{
 		$general_options[] = $form->generate_check_box("showforumteam", 1, $lang->forum_team, array("checked" => $mybb->get_input('showforumteam', MyBB::INPUT_INT)));
@@ -1387,6 +1389,8 @@ if($mybb->input['action'] == "disporder" && $mybb->request_method == "post")
 			$db->update_query('usergroups', $sql_array, "gid = '{$gid}'");
 		}
 	}
+	
+	$cache->update_usergroups();
 
 	// Log admin action
 	log_admin_action();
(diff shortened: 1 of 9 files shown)


dvz:
Would it be better to use the main `usergroups` datacache instead? This would eliminate the hassle of maintaining the derivative data.

We can use `format_name()`, or [Twig's `replace()`](https://twig.symfony.com/doc/3.x/filters/replace.html) (then without `|raw`) to deal with `{username}` dynamically.

The lang string was added after `forum_team` - we can similarly make the code more intuitive by adding the new lines for input/columns/defaults close to the equivalent _Show Team_ lines (adding everything before _team_ may even make sense, so that _General Options_ have increasing specificity); and the setting after `showwol`, since it's displayed under it (`<disporder>` will need correction).

arcalys:
> Would it be better to use the main `usergroups` datacache instead? This would eliminate the hassle of maintaining the derivative data.
> 
> We can use `format_name()`, or [Twig's `replace()`](https://twig.symfony.com/doc/3.x/filters/replace.html) (then without `|raw`) to deal with `{username}` dynamically.
> 
> The lang string was added after `forum_team` - we can similarly make the code more intuitive by adding the new lines for input/columns/defaults close to the equivalent _Show Team_ lines (adding everything before _team_ may even make sense, so that _General Options_ have increasing specificity); and the setting after `showwol`, since it's displayed under it (`<disporder>` will need correction).

Thanks a lot for the feedback!

I'm not sure why I didn't think of using the existing groups cache earlier — I've now updated everything to rely on that instead.

I also moved the `showinlegend` toggle to appear before `showforumteam`, and switched to using `format_name()` for rendering the group display. The general setting has been relocated next to `showwol`.

admin/modules/user/groups.php
```diff
@@ -1070,6 +1071,10 @@
 
 	$general_options = array();
 	$general_options[] = $form->generate_check_box("showmemberlist", 1, $lang->member_list, array("checked" => $mybb->input['showmemberlist']));
+	if($mybb->settings['showgroupslegend'] != 0)
+	{
+		$general_options[] = $form->generate_check_box("showinlegend", 1, $lang->show_in_legend, array("checked" => $mybb->input['showinlegend']));
+	}
```
dvz: Would the missing input result in the option set to `0` for the group when the setting is disabled?

It may be fine to show the option regardless.

arcalys: Correct, it would reset to 0 for a group that was previously shown in the legend, which would be awkward when reactivating the global setting. Removed the conditional and now display the option regardless of the setting's status.

Sama34: You may use a peeker

arcalys: Is it necessary, or should the input be visible at all times? It doesn't bother me that it's always displayed, especially since I know disabling the global setting won't have any effect anyway.

Sama34: Always showing the input seems congruent to other behaviour, so no peeker is fine I think.

arcalys: Ok then!

dvz:
Should the legend be displayed below the list of users instead?

arcalys:
Your choice, I don't have any preference.

dvz:
It seems customary, so let's move it below. Otherwise LGTM.

arcalys:
Updated:
image

And rebased the PR to keep it clean.

arcalys:
Migration if PR is accepted:
```php
#pgsql
$db->add_column("usergroups", "showinlegend", "smallint NOT NULL default '0'");

# sqlite and mysql
$db->add_column("usergroups", "showinlegend", "tinyint(1) NOT NULL default '0'");
```

inc/themes/core.base/frontend/templates/index/index.twig
```diff
@@ -32,6 +32,14 @@
                         {% if user.invisible == 1 and (mybb.usergroup.canviewwolinvis == 1 or user.uid == mybb.user.uid) %}*{% endif %}
                     {% endfor %}
                 </p>
+                {% if mybb.settings.showgroupslegend %}
```
Sama34:
```suggestion
                {% if mybb.settings.showgroupslegend and groups %}
```

Otherwise, some HTML will be added even if there are no groups to display.

Why do we need a global setting for this again ? I think the global setting to be unnecessary now.

arcalys: If you want to disable the legend entirely and have many groups, it's probably easier to just toggle the global setting.

PR #4952, opened 2025-06-08
chore: deprecate thread ratings

### LINKED ISSUE #3498: Deprecate Thread Ratings in 1.9

arcalys:
### Removed

- Removed everything related to thread ratings except:
  - Database table and fields.
  - Legacy theme structure in `mybb_theme.xml`.
  - The `merge_thread_ratings` and `acp_recount_thread_ratings` functions (deprecated instead). 

### Changed

- Renamed `posting_rating` to `posting` in language keys and forum management.

Let me know if you'd prefer everything to be commented instead of removed.

Resolves #3498


Sama34:
I think usages of `mybb.settings.allowthreadratings`,`foruminfo.allowtratings`, etc should be removed from code too.

[ratethread.php](https://github.com/mybb/mybb/blob/dev-1.9/ratethread.php) should be deprecated somehow, maybe empty its contents or just deleted, etc.

arcalys:
The suggestion only mentioned templates, but I can remove everything related to thread rating—except the database objects—if you prefer. It will still be easy for anyone who wants to move it to a plugin to trace everything that existed through the PR.

arcalys:
Removed the javascript files and the dedicated PHP file @Sama34.
Should I remove the rating logic in the code as well (forumdisplay, showthread,...)? If not, I'll need to revert the last commit and keep the setting.

Sama34:
I think everything else except the database objects should be removed or deprecated.

Functions (if any) should be deprecated instead of removed, in case some plugin use them. I think there is a php8 attribute for deprecated content.

arcalys:
> I think everything else except the database objects should be removed or deprecated.
> 
> Functions (if any) should be deprecated instead of removed, in case some plugin use them. I think there is a php8 attribute for deprecated content.

Updated the PR, removing everything except the existing functions (outside of `ratethread.php`) that have been marked as deprecated.

PR #4955, opened 2025-06-10
feat: add upgrade script for 1.9.0

### LINKED ISSUE #4592: Remove unused `mail_parameters` setting
Remove the _Mail Settings → Additional Parameters for PHP's mail()_ ([`mail_parameters`](https://github.com/mybb/mybb/blob/mybb_1830/install/resources/settings.xml#L2331-L2338)) setting, which became no-op with [MyBB 1.8.31](https://mybb.com/versions/1.8.31/) when the value was moved to the [Configuration File](https://docs.mybb.com/1.8/faq/mail/#additional-parameters-for-phps-mail), but was left to preserve existing data that administrators could copy over.

### LINKED ISSUE #4831: Add 1.9.0 upgrade script
Create an upgrade script for MyBB 1.8-to-1.9 migration, accounting for changes made so far.

The file can be named `upgrade100.php`, and corrected before stable release (#4830).

Existing drafts:
- https://github.com/mybb/mybb/blob/fc06044714f29aa37fa4d263de33776294b3a1b4/install/resources/upgrade56.php#L34-L44 (needs i.a. restructured upgrade script format, #4599)

arcalys:
### Added

- 1.9 upgrade script base with existing migrations (except for themes/templates related tables).

Resolves #4592
Resolves #4831

### DIFF
--- a/inc/upgrades/upgrade100.php
+++ b/inc/upgrades/upgrade100.php
@@ -134,7 +134,7 @@ function upgrade100_dbchanges()
     // Update moved threads
     $db->query("
         UPDATE ".TABLE_PREFIX."threads
-        SET closed = '2', moved = ".$moved_tid_substring."
+        SET closed = '0', moved = ".$moved_tid_substring."
         WHERE closed LIKE 'moved|%' AND (moved IS NULL OR moved = 0);
     ");
 
--- a/inc/upgrades/upgrade100.php
+++ b/inc/upgrades/upgrade100.php
@@ -36,7 +36,7 @@ function upgrade100_dbchanges()
     $db->modify_column("users", "password", "varchar(500)", "set", "''");
     $db->modify_column("users", "style", "varchar(30)", "set", "''");
 
-    // Add userfields columns 
+    // Add userfields columns
     foreach (["fid4", "fid5", "fid6"] as $fid) {
         if (!$db->field_exists($fid, "userfields")) {
             $db->add_column("userfields", $fid, "text NOT NULL");
@@ -72,6 +72,7 @@ function upgrade100_dbchanges()
                     type varchar(50) NOT NULL default ''
                 );");
             }
+            $moved_tid_substring = "SUBSTRING(closed FROM 7)::INTEGER";
             break;
 
         case 'sqlite':
@@ -98,6 +99,7 @@ function upgrade100_dbchanges()
                     type varchar(50) NOT NULL default ''
                 );");
             }
+            $moved_tid_substring = "SUBSTR(closed, 7)";
             break;
 
         default: // MySQL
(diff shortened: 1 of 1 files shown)


inc/upgrades/upgrade100.php
```diff
@@ -0,0 +1,85 @@
...
+ * MyBB 1.9
+ * Copyright 2014 MyBB Group, All Rights Reserved
+ *
+ * Website: http://www.mybb.com
+ * License: http://www.mybb.com/about/license
+ *
+ */
+
+/**
+ * Upgrade Script: 1.9.0
```
dvz: This would be the previous version - we can temporarily use `1.8.x`, like with the upgrade number.

https://github.com/mybb/mybb/tree/dev-1.9/inc/upgrades#source-versions

inc/upgrades/upgrade100.php
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+$upgrade_detail = array(
+    'revert_all_templates' => 0,
+    'revert_all_themes' => 0,
+    'revert_all_settings' => 0
+);
+
+@set_time_limit(0);
```
dvz: Not needed - executed on a higher level at https://github.com/mybb/mybb/blob/33b979a117d6a2fc355315ae2ce2184c682912b2/inc/src/Maintenance/Process/Model/UpgradeModel.php#L203

dvz (changes requested):
We'll need conditionals to check if the changes are necessary for queries that may otherwise fail. This will allow the upgrade to be run numerous times without side effects.

dvz:
Looks like the new upgrade script runs without issues in GUI & CLI, so this can be tested normally.

inc/upgrades/upgrade100.php
```diff
@@ -0,0 +1,133 @@
...
+                $db->write_query("CREATE TABLE ".TABLE_PREFIX."securitylog (
+                    uid int NOT NULL default '0',
+                    ipaddress bytea NOT NULL default '',
+                    dateline int NOT NULL default '0',
+                    type varchar(50) NOT NULL default ''
+                );");
+            }
+            break;
+
+        case 'sqlite3':
```
dvz: This should be simply `sqlite`

(looks like historical MyBB versions may have included a number)

inc/upgrades/upgrade100.php
```diff
@@ -0,0 +1,133 @@
...
+            if (!$db->field_exists("contact", "profilefields")) {
+                $db->add_column("profilefields", "contact", "tinyint(1) NOT NULL default '0' AFTER disporder");
```
dvz: Wouldn't `AFTER` cause problems on SQLite as well?

dvz (changes requested):

inc/upgrades/upgrade100.php
arcalys: Updated

inc/upgrades/upgrade100.php
arcalys: Indeed, saw some `AFTER` clauses in previous scripts and did not check the docs.  Removed for SQLite

inc/upgrades/upgrade100.php
```diff
@@ -125,9 +127,17 @@ function upgrade100_dbchanges()
...
                 ) ENGINE=MyISAM;");
             }
+            $moved_tid_substring = "CAST(SUBSTRING(closed, 7) AS SIGNED)";
             break;
     }
 
+    // Update moved threads
+    $db->query("
+        UPDATE ".TABLE_PREFIX."threads
+        SET closed = '2', moved = ".$moved_tid_substring."
```
dvz: Apparently `2` doesn't have any meaning (#4966); the previous value seems to be discarded, so `closed` can be set to `0`

arcalys: Updated

dvz (approved):

PR #4962, opened 2025-06-28
Correct mistaken comparisons of $thread['closed'] with zero

### LINKED ISSUE #4961: Delete button sometimes mistakenly not shown in postbit on PHP 8
This is because in PHP 8, an empty string no longer compares equal to zero[1], but the database field `mybb_threads`.`closed` is typically set to an empty string (and is only set to zero when reopening a closed thread), meaning that the `$thread['closed'] == 0` condition here evaluates false on PHP 8 when we want it to evaluate true:

https://github.com/mybb/mybb/blob/b8cc332a27e145c33effaccec90e23c103ae5193/inc/functions_post.php#L632

[1] https://www.php.net/manual/en/migration80.incompatible.php

lairdshaw:
In PHP 8, an empty string no longer compares equal to zero, and these comparisons mistakenly assume that it (still) does.

Resolves #4961


lairdshaw:
Closing - I meant to file this against the feature branch, not dev-1.9. Will refile. Feel free to delete.

Sama34:
> Closing - I meant to file this against the feature branch, not dev-1.9. Will refile. Feel free to delete.

Welcome to the new primary repository :D (this happens to me too... t.t)

Issue #4959, opened 2025-06-21
Implemente the proper error status code for error, no permission pages.


Sama34:
Currently, 1.8 outputs 200 HTTP status codes for error pages.

Ideally, 404 status code should be used for all error pages, and 403 for no permission pages.

There may be additional status codes reasonable for specific features.

I think this may be better off configured in the `config.php` file.

Ideally, plugins should be able to hijack whatever implementation is applied.

See: https://community.mybb.com/thread-243653-post-1401815.html#pid1401815

arcalys:
Do we need this to be configurable in config.php? Most likely we’re just going to use 403 for explicitly forbidden actions and 404 for everything else?

Ben-MyBB:
Another https://github.com/mybb/mybb/issues/4352

Sama34:
> Do we need this to be configurable in config.php? Most likely we’re just going to use 403 for explicitly forbidden actions and 404 for everything else?

I think my suggestion to allow some configuration was based on the limitation of the `error` hook:
https://github.com/mybb/mybb/blob/ce3ded05a5b88404a56355cf1584cceb9486928f/inc/functions.php#L902C23-L902C32

I think we should instead update this hook to pass on all arguments of the function. That should be left for another issue.

Regards

PR #4967, opened 2025-07-01
Refactor/return 403 status code for error_no_permission

### LINKED ISSUE #2864: Return correct HTTP status codes
The default MyBB forum/thread/user not found/does not exist pages return HTTP status code 200 (success) currently. I think that PHP's http_response_code function should be used and make those pages return the correct HTTP 404 status code.

Could this be extended to make the 'no permissions' error page return a status code of 403 (forbidden) or 401 (unauthorised)?

Original thread: [Return correct HTTP status codes](https://community.mybb.com/thread-212936.html)

arcalys:
### Added

- Added a `status_code` parameter to the `error` function, defaulting to 200, to avoid breaking existing behavior immediately.
- Added parameters and return type hints for the `error` function.

### Fixed

- Corrected the parameters description for the `error` function.

Part of #4959 
Resolves #2864

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -884,10 +884,12 @@ function get_child_list($fid)
 /**
  * Produce a friendly error message page
  *
- * @param string $error The error message to be shown
+ * @param string $error_message The error message to be shown
  * @param string $title The title of the message shown in the title of the page and the error table
+ * @param string $error_page Optional additional error page
+ * @param int $status_code HTTP status code to send with the error page
  */
-function error($error_message = "", $title = "", $error_page = "")
+function error($error_message = "", $title = "", $error_page = "", $status_code = 200)
 {
 	global $db, $lang, $mybb, $plugins;
 
@@ -901,6 +903,8 @@ function error($error_message = "", $title = "", $error_page = "")
 		echo json_encode(array("errors" => array($error_message)));
 		exit;
 	}
+	
+	http_response_code($status_code);
 
 	$timenow = my_date('relative', TIME_NOW);
 	reset_breadcrumb();
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -1027,7 +1027,7 @@ function error_no_permission()
 		'redirect_url' => $redirect_url
 	]);
 
-	error($error_message, $lang->error_nopermission, $error_page);
+	error($error_message, $lang->error_nopermission, $error_page, 403);
 }
 
 /**
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -889,7 +889,7 @@ function get_child_list($fid)
  * @param string $error_page Optional additional error page
  * @param int $status_code HTTP status code to send with the error page
  */
-function error($error_message = "", $title = "", $error_page = "", $status_code = 200)
+function error(string $error_message = "", string $title = "", string $error_page = "", int $status_code = 200): never
 {
 	global $db, $lang, $mybb, $plugins;
 


inc/functions.php
```diff
@@ -884,10 +884,12 @@ function get_child_list($fid)
...
  * Produce a friendly error message page
  *
- * @param string $error The error message to be shown
+ * @param string $error_message The error message to be shown
  * @param string $title The title of the message shown in the title of the page and the error table
+ * @param string $error_page Optional additional error page
+ * @param int $status_code HTTP status code to send with the error page
  */
-function error($error_message = "", $title = "", $error_page = "")
+function error($error_message = "", $title = "", $error_page = "", $status_code = 200)
```
Sama34: Regardless of use implementing wide type hints, I think we should start using them for new or updated code like this line.

Also, `never` as the return type declaration, although I understand if you decide not to as you are not adding a new function here.

inc/functions.php
```diff
@@ -901,6 +903,8 @@ function error($error_message = "", $title = "", $error_page = "")
 		echo json_encode(array("errors" => array($error_message)));
 		exit;
 	}
+	
+	http_response_code($status_code);
```
Sama34, line 907: Isn't `200` implied ? I'm unsure if the best practice is to always send a response code, so forgive my ignorance if so.

editpost.php
```diff
@@ -43,7 +43,7 @@
 
 if(!$post || ($post['visible'] == -1 && $mybb->input['action'] != "restorepost"))
 {
-	error($lang->error_invalidpost);
+	error(error_message: $lang->error_invalidpost, status_code: 404);
```
Sama34: The first named parameter can be skipped. I have no preference on this matter, so this is more a comment to bring up and decide on a pattern for future code so we follow the same.

@dvz

Sama34 (commented):
I think the error_no_permission throwing a `403` status code could easily be implemented in this PR as it should be only be 1 line edit.

I suppose all other `error()` usages can be updated in a different PR, as these may be numerous.

inc/functions.php
arcalys: Added the types and return type

inc/functions.php
arcalys: It is indeed implied. I chose to set it explicitly here in case another part of the code might have modified the status code beforehand.

arcalys:
> I think the error_no_permission throwing a `403` status code could easily be implemented in this PR as it should be only be 1 line edit.
> 
> I suppose all other `error()` usages can be updated in a different PR, as these may be numerous.

I kept the new parameter and the 403 status for `error_no_permission`, dropping the other commit (will be done through an other PR).

inc/functions.php
Sama34: I suppose plugins can update their usage of error() if so, then.

inc/functions.php
```diff
@@ -884,10 +884,12 @@ function get_child_list($fid)
...
  * Produce a friendly error message page
  *
- * @param string $error The error message to be shown
+ * @param string $error_message The error message to be shown
  * @param string $title The title of the message shown in the title of the page and the error table
+ * @param string $error_page Optional additional error page
+ * @param int $status_code HTTP status code to send with the error page
  */
-function error($error_message = "", $title = "", $error_page = "")
+function error(string $error_message = "", string $title = "", string $error_page = "", int $status_code = 200): never
```
Sama34, line 892: Just thinking now, maybe we should add constants instead of "magic" numbers ? I'm unsure if it is worth or best practice for this case, as the status codes seem to be universally standardized. @dvz

Sama34:
> dropping the other commit (will be done through an other PR).

Oh, I didn't mean to suggest dropping the `error()` changes in this PR, I apologize if that felt as rude or demanding.

But if you already removed them, I have no issue either.

Thank you for all your contributions !

arcalys:
> > dropping the other commit (will be done through an other PR).
> 
> Oh, I didn't mean to suggest dropping the `error()` changes in this PR, I apologize if that felt as rude or demanding.
> 
> But if you already removed them, I have no issue either.
> 
> Thank you for all your contributions !

No worries, it wasn’t rude at all. You were right — it makes more sense to have a separate PR to update all the other `error()` calls at once.

inc/functions.php
Sama34: I forgot to send this message last time ...

editpost.php
dvz: It could be skipped when obvious (expected 1st argument & passed value or variable name - mutually reinforcing). E.g. for `error()`, when using both `message` and `title` with `error()`, specifying names locally would be helpful.

inc/functions.php
dvz: We might eventually, possibly from a library (e.g. https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/HttpFoundation/Response.php).

It may be fine using named parameters with integers for now, though.

Sama34: Oh, if we are already requiring that, we should just go ahead and use it already.

It would also motivate developers to start using the same for new code for both the core and plugins.

arcalys: I'd prefer to avoid integrating a new dependency in this PR and move forward with this change. Maybe we can open an issue to track the potential new dependency and any related refactoring if needed?

Sama34: I'm a bit confused because these constants seem to be available already in both my local machine and test server.

But maybe my understanding of _dependency_ is a bit off so we maybe we should move on and go back to it later, yes.

Sama34:
image

PR #4973, opened 2025-07-03
Missing checks for `set_time_limit()` availability

### LINKED ISSUE #4729: @set_time_limit with php8

Sama34:
From https://github.com/mybb/mybb/pull/4730

Resolves #4729


Sama34:
@dvz do you know why the PHPUnit Tests is failing ?

dvz:
Looks like the `ParserTest` needs corrections for #4919.

Is `disable_functions` necessary? https://php.watch/versions/8.0/disable_functions-redeclare

It may be fine to reuse the same code where needed:
```php
if (function_exists('set_time_limit')) {
    set_time_limit(0);
}
```

arcalys:
> Looks like the `ParserTest` needs corrections for #4919.
> 

Done through https://github.com/mybb/mybb/pull/4977

Sama34:
> Looks like the `ParserTest` needs corrections for #4919.
> 
> Is `disable_functions` necessary? https://php.watch/versions/8.0/disable_functions-redeclare
> 
> It may be fine to reuse the same code where needed:
> 
> ```
> if (function_exists('set_time_limit')) {
>     set_time_limit(0);
> }
> ```

I kept `disable_functions` to keep the function as in the 1.8 PR, but I'm fine with removing it, so I will do so.

Do you think the unit test file in this PR is superfluous ? I'm willing to drop it because it seems unnecessary to me.

Edit: if you mean to drop the `my_set_time_limit` I may agree, it seems superfluous. I just attempted to migrate the code from 1.8, but I'm fine with just dropping this function altogether for 1.9

dvz:
The test does seem redundant.

Did we have more reports for #4729? It seemed like we could have left the calls, which were removed in 1.9. It primarily introduces a function that may be removed later.

Sama34:
OK, the latest code should suffice.

I think there is no need to revert the change in 1.8 but let me know if you disagree.

Ben-MyBB:
@arcalys Please could you review?

arcalys (approved):
Might want to rename the PR and/or rebase.

PR #4976, opened 2025-07-03
fix(captcha): cloudflare turnstile twig template and class

### LINKED ISSUE #4659: Add support for Cloudflare Turnstile for captcha
Please add Turnstile to selection of captcha

arcalys:
### Fixed

- Incorrect class for the turnstile html element.
- Incorrect template call for turnstile in `captcha::build_cfturnstile()`.
- Incorrect type for `captcha::type`, which is supposed to be an integer but receives `$mybb->settings['captchaimage']` which is a string. This makes the `$this->type === self::CFTURNSTILE` comparison invalid.

Resolves #4659

### DIFF
--- a/inc/class_captcha.php
+++ b/inc/class_captcha.php
@@ -106,7 +106,7 @@ class captcha
 	{
 		global $mybb, $plugins;
 
-		$this->type = $mybb->settings['captchaimage'];
+		$this->type = (int) $mybb->settings['captchaimage'];
 
 		$args = array(
 			'this' => &$this,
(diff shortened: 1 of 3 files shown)


Sama34:
You rock.

I think there is a conflict because your repository may be outdated, I will merge anyway.

arcalys:
> You rock.
> 
> I think there is a conflict because your repository may be outdated, I will merge anyway.

That was due to https://github.com/mybb/mybb/pull/4977
I did not have time to rebase 

Thanks for merging!

Sama34:
> > You rock.
> > I think there is a conflict because your repository may be outdated, I will merge anyway.
> 
> That was due to #4977 I did not have time to rebase
> 
> Thanks for merging!

I think most of the time you can simply sync your repository branch from the GitHub UI.

But if there is no "real" conflict, I think it should be no problem to merge ultimately.

PR #4980, opened 2025-07-07
Update <b> & <i> tags for accessibility

### LINKED ISSUE #4553: Inconsistent Formatting & Semantics in Language Files

Sama34:
Resolves #4553


dvz:
Limit it to 1.9, perhaps?

Sama34:
Maybe, but I think providing changes in favour of accessibility should be a priority and the EOL support for 1.8 is what, probably 2027 (guess) ?

I'm open for rejecting, but let's wait for additional feedback.

BKPepe (approved):

arcalys:
What should we do with this?

dvz:
We can merge if you like; a mirror PR for 1.9 would be good to get possible conflicts out of the way (and see if any similar corrections are needed in 1.9 code).

arcalys:
> We can merge if you like; a mirror PR for 1.9 would be good to get possible conflicts out of the way (and see if any similar corrections are needed in 1.9 code).

I'll prepare the mirror PR

PR #4985, opened 2025-07-08
Fixes #4979 Shorten parameter names for error()

### LINKED ISSUE #4979: Shorten parameter names for `error()`
Change parameter names of the `error()` function:
- from `$error_message` to `$message`
- from `$error_page` to `$page`

The `$title` parameter has no such prefix.

for easier usage with named arguments.

## Impact
[Named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments) (PHP 8) are not used in MyBB ≤ 1.8, but change may affect usage in plugins.

Starpaul20:
Resolves #4979

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -886,10 +886,10 @@ function get_child_list($fid)
  *
  * @param string $message The error message to be shown
  * @param string $title The title of the message shown in the title of the page and the error table
- * @param string $page Optional additional error page
+ * @param string $extra_content Optional additional error page content
  * @param int $status_code HTTP status code to send with the error page
  */
-function error(string $message = "", string $title = "", string $page = "", int $status_code = 200): never
+function error(string $message = "", string $title = "", string $extra_content = "", int $status_code = 200): never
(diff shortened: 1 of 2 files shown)


inc/functions.php
```diff
@@ -884,23 +884,23 @@ function get_child_list($fid)
 /**
  * Produce a friendly error message page
  *
- * @param string $error_message The error message to be shown
+ * @param string $message The error message to be shown
  * @param string $title The title of the message shown in the title of the page and the error table
- * @param string $error_page Optional additional error page
+ * @param string $page Optional additional error page
```
```suggestion
 * @param string $page Optional additional error page content
```

Sama34:
Just suggested a small addition to the comment section because it is lacking.

Everything seems fine to me.

Edit: Actually never mind, reviewed again and `$page` should be something like `$extra_content` because it is simply confusing. That variable name doesn't describe what it is at all.

Sama34 (approved):
LGTM

PR #4986, opened 2025-07-09
fix: incorrect htaccess configuration for inc/ folder

arcalys:
### Fixed

- Removed incorrect `inc` location configuration in main htaccess example and instead duplicated `/admin/backups/.htaccess` to `/inc/.htaccess`.


Sama34:
I did test the in my local machine and this file just works, using XAMPP for macOS.

If you could double-check and merge @dvz thanks !

PR #5009, opened 2025-08-11
Update default smilies to SVG format

### LINKED ISSUE #4999: Update smilies to SVG format
https://community.mybb.com/thread-243754.html

https://github.com/Paul-Verbitsky/MyBB-Vector-Smilies

It seems images in the `/images/smilies` directory would need to be replaced.
https://github.com/mybb/mybb/tree/dev-1.9/images/smilies

The `/inc/seeds/db_inserts.php` file would need to be updated for new installations.
https://github.com/mybb/mybb/blob/dev-1.9/inc/seeds/db_inserts.php

And probably some upgrade process should be added to the upgrade script to update smilies when they match the 1.8 counterpart (to avoid updating modified smilies).

arcalys:
### Changed

- Replaced smilies with the new SVG versions provided by @Paul-Verbitsky.

Also, I’m not entirely sure what behaviour we want for the smilies upgrade, so the upgrade script might need to be updated or reworked. If you have time, I’d appreciate it if you could test it out.

Resolves #4999


dvz:
It could perform some checks to prevent duplicates when inserting new ones (we don't seem to have `UNIQUE(find)` nor `UNIQUE(image)`).

If it makes things easier, it could load data for all/new entries from the seeds file instead of having a similar array in the upgrade file.

`disporder` may need taking care of, especially if the default `0` will result in the new extra ones being shown at the start of the list.

Using a separate function may help keep the file clean.

inc/upgrades/upgrade100.php
```diff
@@ -53,6 +53,14 @@ function upgrade100_dbchanges()
             if (!$db->field_exists("moved", "threads")) {
                 $db->add_column("threads", "moved", "int NOT NULL default '0'");
             }
+
```
Sama34, line 60: This and several blocks below seem unrelated to this PR, is this intentional ?

Probably the branch should just be updated/synched.

lairdshaw:
> This and several blocks below seem unrelated to this PR, is this intentional ?

Yep, see @arcalys's note up top:

> ⚠️ This is currently based on my other [pull request](https://github.com/mybb/mybb/pull/5004) since I needed the fix in the upgrade script.

inc/upgrades/upgrade100.php
lairdshaw: > This and several blocks below seem unrelated to this PR, is this intentional ?

Yep, see @arcalys's comment up top:

> ⚠️ This is currently based on my other [pull request](https://github.com/mybb/mybb/pull/5004) since I needed the fix in the upgrade script.

arcalys: Yeah I needed the fix from my other pull request in order to test the upgrade script

arcalys:
> It could perform some checks to prevent duplicates when inserting new ones (we don't seem to have `UNIQUE(find)` nor `UNIQUE(image)`).
> 
> If it makes things easier, it could load data for all/new entries from the seeds file instead of having a similar array in the upgrade file.
> 
> `disporder` may need taking care of, especially if the default `0` will result in the new extra ones being shown at the start of the list.
> 
> Using a separate function may help keep the file clean.

Moved the logic to a dedicated function, handled disporder and prevented duplicates.

For now I kept the array in the file, but let me know if this needs to be removed.

arcalys:
Updated the PR to include the support for adding SVG smilies in the ACP when using "Add multiple smilies".

PR #5013, opened 2025-08-14
Fix template and styling for message edition

arcalys:
### Changed

- Updated the postbit template and related stylesheet to properly display the `edited by` message and reason.

### Example

image


dvz:
Did we want to use some tooltip for those? @justinsoltesz

PR #5021, opened 2025-08-19
Disable home icon in footer when home settings are not set

### LINKED ISSUE #5020: If homepage name is blank a home icon appears in footer with no text showing a style inconsistency
Until the homepage name setting is updated in the ACP, there is a styling inconsistency. We should include a default value similiar to the board name field.

arcalys:
- Added extra condition to `footer.twig` to avoid displaying home icon when related settings (homename and homeurl) are not set.

Resolves #5020

### DIFF
--- a/inc/themes/core.base/frontend/templates/partials/footer.twig
+++ b/inc/themes/core.base/frontend/templates/partials/footer.twig
@@ -71,12 +71,14 @@
                                 </a>
                             </li>
                         {% endif %}
-                        <li class="footer-menu__item">
-                            <a href="{{ mybb.settings.homeurl }}" class="footer-menu__link">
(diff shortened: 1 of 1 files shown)


inc/themes/core.base/frontend/templates/partials/footer.twig
```diff
@@ -71,12 +71,14 @@
                                 </a>
                             </li>
                         {% endif %}
+                        {% if mybb.settings.homeurl and mybb.settings.homename %}
                         <li class="footer-menu__item">
```
dvz: we could indent the content to mirror preceding conditions

arcalys: Updated :)

PR #5039, opened 2025-09-08
Log successful admin CP logins to security log

### LINKED ISSUE #4087: log admin cp log ins & prevent some log pruning
mybb really should log admin cp log ins as well as prevent some stuff from being able to be pruned from the adminstrator's logs.

arcalys:
### Added

- Security log entry for successful admin CP logins.

Resolves #4087

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -264,6 +264,7 @@ elseif($mybb->input['do'] == "login")
 		}
 
 		$plugins->run_hooks("admin_login_success");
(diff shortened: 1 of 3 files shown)


dvz:
Shorten to `login_admincp` instead? Other entries seem to imply success, too.

arcalys:
> Shorten to `login_admincp` instead? Other entries seem to imply success, too.

Updated

PR #5040, opened 2025-09-08
Normalize line endings (LF only)

### LINKED ISSUE #2157: Wrong line endings in php files (Development Standard)
I noticed that some of the files of the project don't follow the http://docs.mybb.com/1.8/development/standards/

MyBB Development Standards Wrote:New Lines

All new lines should be denoted by the UNIX line feed (\n, also known as "LF") and not include the Windows carriage return (\r, also known as "CR"). CR and CRLF are not permitted; LF is the only permitted line ending.

The full list of the 1.8.5 release with wrong line endings (crlf instead of lf only):

admin\styles\default\style.php
inc\3rdparty\2fa\GoogleAuthenticator.php
inc\db_sqlite.php
install\resources\upgrade25.php
member.php
usercp2.php

It would be nice to see these, maybe with 1.8.6, corrected to follow the UNIX standard of lf only like the rest of the files. I know that such a change isn't easy to do (especially for the last two files), because it may result in difficulties for merges from other developers/forks in some cases.

Original thread: [Wrong line endings in php files (Development Standard)](http://community.mybb.com/thread-173279.html)

arcalys:
### Changed

Normalized line endings for a few remaining files:
```
./admin/styles/default/style.php
./admin/styles/default/login.css
./jscripts/jeditable/jeditable.min.js
./inc/3rdparty/diff/Diff/Op/index.html
./inc/3rdparty/diff/Diff/ThreeWay/Op/index.html
./inc/3rdparty/diff/Diff/ThreeWay/index.html
./inc/3rdparty/2fa/GoogleAuthenticator.php
```

Resolves #2157


dvz:
Looks like that will be all files affected.

We can defer collective white-space changes (so also: removing trailing whitespace at end of lines & including `LF` at end of files - part of [PSR-12](https://www.php-fig.org/psr/psr-12/)) in old code to ease comparing & porting changes until 1.8 EOL (or its announcement, when we limit it to backporting), and then commit them at once.

PR #5047, opened 2025-09-09
Remove HTML tags in buddy list error messages

arcalys:
### Fixed

- Removed HTML in buddy list error messages (no buddies, no online buddies,...). 
image


dvz:
It would be better to remove the HTML altogether (related: #756), and avoid `|raw`.

arcalys:
> It would be better to remove the HTML altogether (related: #756), and avoid `|raw`.

Updated the PR, removed the `<em>` tags from the strings.

PR #5055, opened 2025-09-16
Add usercount to usergroups cache and pagination to ACP groups module

### LINKED ISSUE #4436: Add Pagination and Cache to User-Groups module
The Admin user-groups module [here](https://github.com/mybb/mybb/blob/feature/admin/modules/user/groups.php) does not have any pagination for the listing of groups, or any caching on the fetching of the total “# of users”.

This lack of caching causes significant performance issues on page load for large boards, while pagination is just something it should get, like other pages have recently thanks to the latest version. 

`$query = $db->query("
				SELECT g.gid, COUNT(u.uid) AS users
				FROM ".TABLE_PREFIX."users u
				LEFT JOIN ".TABLE_PREFIX."usergroups g ON (CONCAT(',', u.additionalgroups, ',') LIKE CONCAT('%,', g.gid, ',%'))
				WHERE g.gid != '0' AND g.gid is not NULL GROUP BY g.gid
			"); `

Testing this on a local board with about 10-15 groups and 700,000 users (all in the same group besides the default admin account) yielded a consistent 10-15 second page load, this however would significantly increase to amounts I cannot imagine on a board that is in use with more moving parts.

arcalys:
### Added
- Pagination to the Admin CP User-Groups module.

### Changed
- Updated the `update_usergroups()` cache function to include `usercount` for each group.
- Replaced the user count SQL queries in the Admin CP User-Groups module with a lookup from the cache.

Resolves #4436

### DIFF
--- a/admin/modules/user/groups.php
+++ b/admin/modules/user/groups.php
@@ -1403,6 +1403,25 @@ if($mybb->input['action'] == "disporder" && $mybb->request_method == "post")
 
 if(!$mybb->input['action'])
 {
+	$query = $db->simple_select("usergroups", "COUNT(gid) as usergroups");
+	$total_rows = $db->fetch_field($query, "usergroups");
+	$pagenum = $mybb->get_input('page', MyBB::INPUT_INT);
+	if($pagenum)
+	{
+		$start = ($pagenum-1) * 20;
+		$pages = ceil($total_rows / 20);
+		if($pagenum > $pages)
+		{
+			$start = 0;
+			$pagenum = 1;
+		}
+	}
+	else
+	{
+		$start = 0;
+		$pagenum = 1;
+	}
+
 	$plugins->run_hooks("admin_user_groups_start");
 
 	if($mybb->request_method == "post")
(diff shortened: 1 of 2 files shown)


dvz:
Looks like the cached count would be at high risk of becoming stale - it would need updating in i.a. `join_usergroup()`, `leave_usergroup()`, and account state events. It may be better to contain the problem to the ACP module.

The usergroup memberships would be one of the main candidates for DB normalization - created #5122.

arcalys:
> Looks like the cached count would be at high risk of becoming stale - it would need updating in i.a. `join_usergroup()`, `leave_usergroup()`, and account state events. It may be better to contain the problem to the ACP module.
> 
> The usergroup memberships would be one of the main candidates for DB normalization - created #5122.

Should we close this, or would you like me to rebase and keep only the pagination commit?

Sama34:
Hi,

Rebasing would suffice, as querying in the ACP shouldn't be a problem right now.

Thank you for your contribution !

Regards.

dvz:
For pagination alone, it may come down to whether the overall experience is better - dividing a long list may mean building the data more than once to find a group over several pages (depending on implementation).

Issue #5018, opened 2025-08-17
No users found when searching for a user to private message


Ben-MyBB:
On 1.9, no results are returned when searching for a username in the private message function.

arcalys:
Issue in `xmlhttp.php` due to `$theme` not being set and [this line](https://github.com/mybb/mybb/blob/dev-1.9/xmlhttp.php#L123) causing an issue. Affects all similar search inputs.

Ben-MyBB:
Was this resolved in #5085 @arcalys ?

arcalys:
> Was this resolved in [#5085](https://github.com/mybb/mybb/pull/5085) [@arcalys](https://github.com/arcalys) ?

Yes it was. The user search field worked when I tested it after that PR was merged

Ben-MyBB:
Fixed in #5085

Issue #4954, opened 2025-06-08
Active Themes management


dvz:
Add options for managing active themes. Builds upon #4953

UI:
- [x] Admin CP's _Themes_ page (listing, default theme)
- [x] Admin CP's theme assignment (forums, users, spiders)
- [x] front-end (User CP, footer theme switch)

Related:
- [Diagram: Types of Controls in the View Domain](https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md#controls)

PR #5085, opened 2025-10-01
Add Theme instance internals

### LINKED ISSUE #4953: Active Themes internals
Add handling of active themes, associating Theme Packages with entries in the `mybb_themes` table.

Replaces hardcoded logic like: 
https://github.com/mybb/mybb/blob/c10caef61e62e05e8f42687014afb81712e9d84b/global.php#L472-L473

Internals:
- [x] active theme class/logic
- [x] runtime logic with Theme Package association

Related:
- https://github.com/mybb/meta/blob/main/architecture/mybb-1.9-theme-system.md#theme-runtime-information

dvz:
Resolves #4953

Introduces a basic OOP for entries in the `mybb_themes` table, which in MyBB 1.9 will link database references (`mybb_themes.package`) to Theme Packages.

Theme Packages (represented as `Extensions\Theme\Theme`), have a one-to-many relationship with the entries (represented as `Database\Models\Theme`), containing instance-specific settings and permissions.

Applies preliminary changes in schemas, the 1.9 upgrade, runtime code, and inserts an entry for the default _Base_ theme.

### DIFF
--- a/global.php
+++ b/global.php
@@ -114,8 +114,10 @@ if(function_exists('mb_internal_encoding') && !empty($lang->settings['charset'])
 	@mb_internal_encoding($lang->settings['charset']);
 }
 
-// Select the board theme to use.
-$loadstyle = '';
+// Select the Theme model to use.
+$repository = \MyBB\app(\MyBB\Database\Repositories\ThemeRepository::class);
+$tid = null;
+$theme_model = null;
 $load_from_forum = $load_from_user = 0;
 $style = array();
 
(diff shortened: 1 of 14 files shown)


arcalys:
This also resolves #5018

arcalys:
Tested your PR branch—everything works fine except for the admin panel (Themes), which now throws an error due to the missing pid field (replaced here with package). I assume this is intended to be addressed in #4954 rather than in this PR?

PR #5091, opened 2025-10-02
Fixes #5090 - Uploaded files and dirs not deleted from CDN static path

### LINKED ISSUE #5090: Uploaded files and dirs not deleted from CDN static path
https://community.mybb.com/thread-237760.html

Eldenroot:
Resolves #5090

### DIFF
--- a/inc/functions_upload.php
+++ b/inc/functions_upload.php
@@ -840,79 +840,104 @@ function add_attachments($pid, $forumpermissions, $attachwhere, $action=false)
 }
 
 /**
- * Delete an uploaded file both from the relative path and the CDN path if a CDN is in use.
+ * Delete an uploaded file both from the provided absolute path
+ * as well as from the associated CDN path if a CDN is in use.
  *
- * @param string $path The relative path to the uploaded file.
+ * @param string $abs_path The absolute path to the uploaded file.
+ *                         Must resolve to a directory under MYBB_ROOT.
  *
  * @return bool Whether the file was deleted successfully.
  */
-function delete_uploaded_file($path = '')
+function delete_uploaded_file($abs_path)
 {
 	global $mybb, $plugins;
 
 	$deleted = false;
 
-	$deleted = @unlink($path);
-
-	$cdn_base_path = rtrim($mybb->settings['cdnpath'], '/');
-	$path = ltrim($path, '/');
-	$cdn_path = realpath($cdn_base_path . '/' . $path);
-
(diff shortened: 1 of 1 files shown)


inc/functions_upload.php
```diff
@@ -840,79 +840,104 @@ function add_attachments($pid, $forumpermissions, $attachwhere, $action=false)
...
-	);
+		$rel_path = substr($realpath, strlen(MYBB_ROOT));
+
+		if (!empty($mybb->settings['usecdn']))
+		{
+			$cdn_base_path = rtrim($mybb->settings['cdnpath'], '/');
+			if (!empty($cdn_base_path))
+			{
+				$cdn_path = realpath("{$cdn_base_path}/{$rel_path}");
+				$deleted = $deleted && @unlink($cdn_path);
```
Sama34, line 874: I think we should stop using the `@` operant like this because behaviour changed in php8+

Better to avoid in any new code or remove when updating specific lines, etc.

This may require a bit more of checks around the code.

inc/functions_upload.php
```diff
@@ -840,79 +840,104 @@ function add_attachments($pid, $forumpermissions, $attachwhere, $action=false)
...
+			}
+		}
+
+		$hook_params = array(
+			'path' => &$rel_path,
```
Sama34, line 879: What is the implication for plugins here ? @lairdshaw 

Would it be such that delaying this change for 1.9 is to be considered ?

inc/functions_upload.php
lairdshaw: Fair enough. I guess we'd call `is_writable()` on the parent directory first to confirm that the file can be deleted.

inc/functions_upload.php
lairdshaw: As far as I can see, the only implication is that if plugins have already been adapted to the original change from relative paths to absolute paths being supplied to this function - which (I thought was what) broke deletion of CDN files - then they will have to revert their adaptation. Otherwise, the supplied arguments will be identical.

(Caveat: I haven't looked into this code since first sharing it, and, as I said at the time, I didn't test it much even then.)

inc/functions_upload.php
Sama34: @Eldenroot will you update your code ?

Sama34: The suggestion is to avoid using the `@` operator and instead use `is_writable()` where applicable

PR #5095, opened 2025-10-09
Hide remaining canratethreads elements from the ACP

### LINKED ISSUE #3498: Deprecate Thread Ratings in 1.9

arcalys:
### Changed

- Forum management: added an extra deprecation message for the `canratethreads` permission and included it in the `hidefields` array to prevent displaying a checkbox without text (since the corresponding language keys have already been removed).
- Groups management: removed the `canratethread` permission checkbox (same issue, no text).

Fixes #3498


Ben-MyBB:
#3498

PR #5098, opened 2025-10-09
Fix incorrect language key used for submit in ACP help documents

### LINKED ISSUE #5110: Fix incorrect language key used for submit in ACP help documents
Fix incorrect language key used for submit in ACP help documents

arcalys:
### Fixed

- Incorrect language key used for the submit button in the ACP help documents management.

Resolves #5110


Ben-MyBB:
#5110

PR #5102, opened 2025-10-15
Use tooltip for posts edit information in postbit

### LINKED ISSUE #5112: Use tooltip for posts edit information in postbit
Use tooltip for posts edit information in postbit

arcalys:
### Changed

- Updated the postbit template to use a tooltip for post edit information (as originally mentioned in [#5013](https://github.com/mybb/mybb/pull/5013)). I’m not a CSS expert, so this might still need some tweaks or improvements.

### Example
Before:
image

After:
image

Resolves #5112

### DIFF
--- a/inc/themes/core.base/frontend/templates/postbit/postbit.twig
+++ b/inc/themes/core.base/frontend/templates/postbit/postbit.twig
@@ -95,9 +95,27 @@
             <input type="checkbox" class="checkbox post__inline-mod" name="inlinemod_{{ post.pid }}" id="inlinemod_{{ post.pid }}" value="1" {% if post.inlinechecked %}checked="checked"{% endif %}/>
         {% endif %}
 
+        {# Post edit information #}
         {% if post.editedmsg %}
-            <a href="#" class="post__edit-log" title="" id="edited_by_{{ post.pid }}">{{ include('partials/icon.twig', {icon: 'pencil-alt', class: ''}, with_context = false) }}</a>
-            <span class="post__edit-reason">({{ post.editnote|raw }} {{ post.editedprofilelink|raw }}.{% if post.editreason %} <em>{{ lang.postbit_editreason }}: {{ post.editreason }}</em>{% endif %})</span>
+            <span class="post__edit-log">
+                <button popovertarget="editlog-{{ post.pid }}" popovertargetaction="toggle" class="post__edit-log-button" aria-label="{{ lang.postbit_editreason }}" style="anchor-name: editlog-button-{{ post.pid }};">
+                    {{ include('partials/icon.twig', { icon: 'pencil-alt', class: '' }, with_context = false) }}
+                </button>
+                <div
+                    id="editlog-{{ post.pid }}"
+                    popover="auto"
+                    class="post__edit-log-popover"
+                    style="position-anchor: editlog-button-{{ post.pid }};"
+                    role="dialog"
+                >
+                    {{ post.editnote|raw }} {{ post.editedprofilelink|raw }}
+                    {% if post.editreason %}
+                        <div class="edit-reason">
+                            {{ lang.postbit_editreason }}: <em>{{ post.editreason }}</em>
+                        </div>
+                    {% endif %}
+                </div>
+            </span>
         {% endif %}
 
         {# IP address #}
--- a/inc/themes/core.base/frontend/styles/pages/_posts.scss
+++ b/inc/themes/core.base/frontend/styles/pages/_posts.scss
@@ -56,20 +56,62 @@ $post-spacing: 25px;
     &__edit-log {
         margin-#{$left}: 4px;
         font-size: 0.8em;
+        display: inline-block;
+        color: $alt-font-color-2;
 
-        &:link,
-        &:visited {
-            color: $alt-font-color-2;
+        &-button {
+            background: none;
+            border: none;
+            color: inherit;
+            cursor: pointer;
+            font-size: 1em;
+            padding: 0;
+
+            &:hover,
+            &:focus {
+                color: $link-hover;
+            }
         }
 
-        &:hover {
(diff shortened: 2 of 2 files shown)


Ben-MyBB:
#5112

Eldenroot:
Hmm, I am not a fan of this change.

arcalys:
> Hmm, I am not a fan of this change.

Could you elaborate a bit on what specifically you don’t like about the change?
In my first PR, I fixed the edit reason not appearing, and dvz mentioned that they might have considered using a tooltip for that.

Sama34:
I thought there was a new native HTML tag to use for tooltips, but I suppose I mixed that up (with dropdowns?).

I personally like the idea and your concept, but I'm more concerned about the accessibility downsides.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role#accessibility_concerns

From reading the CSS, it seems to me this is an on hover only tooltip. How does it behave in touch capable devices is also to be considered.

We could probably argue that the icon is an indicator of an edit itself, but what about the edit reason, couldn't that be considered _important_ information that should be easily accessible ?

We may consider summary as well:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/summary

Open for feedback, of course, specially from people with more knowledge on accessibility concerns.

Regards.

arcalys:
> I thought there was a new native HTML tag to use for tooltips, but I suppose I mixed that up (with dropdowns?).
> 
> I personally like the idea and your concept, but I'm more concerned about the accessibility downsides. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tooltip_role#accessibility_concerns
> 
> From reading the CSS, it seems to me this is an on hover only tooltip. How does it behave in touch capable devices is also to be considered.
> 
> We could probably argue that the icon is an indicator of an edit itself, but what about the edit reason, couldn't that be considered _important_ information that should be easily accessible ?
> 
> We may consider summary as well: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/summary
> 
> Open for feedback, of course, specially from people with more knowledge on accessibility concerns.
> 
> Regards.

As I mentioned in the PR description, I’m not a CSS or HTML/accessibility expert, so I’m open to improvements.
I tested adding `tabindex`, `aria-describedby`, `role="tooltip"`, and `&:focus .tooltiptext` in the CSS: keyboard navigation now works, and screen readers can read the tooltip correctly (tested with orca on my end).

I’m not sure if this is sufficient, or if it’s the best approach, but in my opinion the current way the edit reason is displayed doesn’t fit well into the postbit.

inc/themes/core.base/frontend/templates/postbit/postbit.twig
```diff
@@ -95,9 +95,19 @@
...
         {% if post.editedmsg %}
-            <a href="#" class="post__edit-log" title="" id="edited_by_{{ post.pid }}">{{ include('partials/icon.twig', {icon: 'pencil-alt', class: ''}, with_context = false) }}</a>
-            <span class="post__edit-reason">({{ post.editnote|raw }} {{ post.editedprofilelink|raw }}.{% if post.editreason %} <em>{{ lang.postbit_editreason }}: {{ post.editreason }}</em>{% endif %})</span>
+            <span class="post__edit-log tooltip">
+                {{ include('partials/icon.twig', {icon: 'pencil-alt', class: ''}, with_context = false) }}
+                <span class="tooltiptext">
+                    {{ post.editnote|raw }} {{ post.editedprofilelink|raw }}
+                    {% if post.editreason %}
+                        <div class="edit-reason">
+                            {{ lang.postbit_editreason }}: <em>{{ post.editreason|raw }}</em>
```
dvz: Does `post.editreason` need to be `raw`?

dvz:
> I thought there was a new native HTML tag to use for tooltips, but I suppose I mixed that up (with dropdowns?).

Perhaps:
- https://developer.mozilla.org/en-US/docs/Web/API/Popover_API
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning

With any implementation, the content should remain interactive.

I imagine the icon may become a button used by features/plugins that show detailed edit history.

inc/themes/core.base/frontend/templates/postbit/postbit.twig
arcalys: No, it does not need to use raw; that was a mistake I made.

arcalys:
Updated the PR to use a popover button with anchor positioning. Please note that anchor positioning is not enabled in all browsers by default (for example, I had to manually enable it in Firefox). 

Display on Chrome:
image

Display on Firefox:
image

As mentioned earlier in the PR, I’m not a CSS or HTML expert, so please let me know if anything can be improved.

dvz:
Changing it to a `<button>` right away sounds good.

If we want to use the new features now, some fallback may be necessary:
- the MDN and [related demo](https://mdn.github.io/dom-examples/popover-api/popover-hint/) shows the menu nested inside `<button>` to use `position: absolute`
- [`@supports`](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports) could be useful to apply either set conditionally: https://developer.chrome.com/docs/css-ui/anchor-positioning-api

Sama34:
image

image

Firefox on macOS Sequoia & Android 12

I think that if modern browsers won't support this out of the box, then we should go with the second best option available.

Regards.

arcalys:
> 
> Firefox on macOS Sequoia & Android 12
> 
> I think that if modern browsers won't support this out of the box, then we should go with the second best option available.
> 
> Regards.

Yeah that's what I said above, in Firefox I had to enable it (`about:config` > `layout.css.anchor-positioning.enabled`).

Ben-MyBB:
@justinsoltesz

PR #5104, opened 2025-10-16
Show reply button on posts in threaded mode

### LINKED ISSUE #5114: Show reply button on posts in threaded mode
Enabled the commented reply button in threaded mode and updated the condition.

arcalys:
### Changed

- Enabled the commented reply button in threaded mode and updated the condition.

Resolves #5114


Ben-MyBB:
#5114

PR #5107, opened 2025-10-18
Use get_thread_link for pagination in thread listing

### LINKED ISSUE #5116: Use get_thread_link for pagination in thread listing
Use get_thread_link for pagination in thread listing

arcalys:
### Changed

- Used `get_thread_link()` for pagination in thread listings (pagination was previously non-functional).

Resolves #5116


Ben-MyBB:
#5116

PR #5117, opened 2025-10-22
Introduce Twig UserExtension to style usernames in templates

arcalys:
Introduces a new Twig extension (`UserExtension`) to provide user-related functions and filters in templates.

As a first feature, it adds the `format_name` filter, which formats and styles usernames according to their user/group settings. This complements existing helpers like `get_profile_link()` allowing templates to display usernames safely without using `|raw`.

The filter accepts either a full user array or a user ID, making it flexible for different template contexts (and to avoid looking up the user when not necessary).

As an example, missing username styling in thread and private message quick replies has been updated to use this new filter.

@dvz I’m not sure if this is exactly what you had in mind or the approach you wanted to take, but I thought I’d give it a try. I’m happy to adjust or replace this if it’s not the way you’d prefer to tackle this.

### DIFF
--- a/inc/src/Twig/Extensions/UserExtension.php
+++ b/inc/src/Twig/Extensions/UserExtension.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace MyBB\Twig\Extensions;
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+
+class UserExtension extends AbstractExtension
+{
+    public function getFilters(): array
+    {
+        return [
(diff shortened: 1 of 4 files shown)


dvz:
Use same name as the PHP function `format_name()`?

arcalys:
> Use same name as the PHP function `format_name()`?

That'd make sense indeed, I'll update the PR.

arcalys:
Rebased and updated the comments to match `format` instead of `style`.

PR #5125, opened 2025-10-26
Fix 4139 Update hardcoded image file formats

### LINKED ISSUE #3847: Implement support for WebP
The [WebP image format](https://en.wikipedia.org/wiki/WebP) is combines all the advantages of JPG, PNG, and GIF:

* Good compression ratio like JPEG (a PNG-like lossless compression is also supported)
* Progressive opacity like PNG 
* Animations like GIF

File size is usually significantly smaller than the other formats. WebP is supported natively by all recent browsers, and supported in PHP image functions since PHP 7.1.

But it cannot be used in a MyBB forum because it is not among the supported image formats.

### LINKED ISSUE #4139: Hardcoded picture extensions limits the creation of thumbnails

Sama34:
This attempts to fix the following issues:

Resolves #3847
Resolves #4139

Notes:

- `.jpe` file format was removed.

Feedback is welcome.

### DIFF
--- a/admin/modules/config/post_icons.php
+++ b/admin/modules/config/post_icons.php
@@ -141,7 +141,24 @@ if($mybb->input['action'] == "add_multiple")
 					if($file != ".." && $file != ".")
 					{
 						$ext = get_extension($file);
-						if($ext == "gif" || $ext == "jpg" || $ext == "jpeg" || $ext == "png" || $ext == "bmp")
+						if(in_array($ext, array(
+							'apng',
+							'png',
+							'avif',
+							'gif',
+							'jpg',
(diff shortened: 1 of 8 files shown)


Sama34:
This will be dropped for 1.8 and instead fixed in 1.9 see #5314

Issue #4918, opened 2025-05-04
Discrepancy between unsigned integer values (MySQL <> PostgreSQL)


arcalys:
In the current database schema, columns defined as unsigned in MySQL do not have corresponding constraints in PostgreSQL, creating a discrepancy between the two databases.

This leads to negative values being allowed and stored in PostgreSQL when proper validation is not in place beforehand.

Example with `disporder` in `mybb_usergroups`:
```sql
# MySQL
disporder smallint(6) unsigned NOT NULL,

# PostgreSQL
disporder smallint NOT NULL default '0',
```

![Image](https://github.com/user-attachments/assets/d9b26d9f-375d-4863-8e44-5ffb287e5e91)

While proper validation should be in place, we should also add CHECK constraints (e.g.: `disporder smallint NOT NULL DEFAULT 0 CHECK (disporder >= 0)`) in PostgreSQL for the unsigned equivalents to ensure data consistency.

arcalys:
@Sama34 @dvz 
For this case, should I add `CHECK` constraints in PostgreSQL for all columns that were `UNSIGNED INTEGER` in the MySQL schema (except those converted to `SERIAL`), or would you prefer using a domain type (a custom data type with built-in constraints)?

Example:

```sql
create domain uint as integer check (value >= 0);
```
```php
$tables[] = "CREATE TABLE mybb_adminsessions (
	sid varchar(32) NOT NULL default '',
	uid int NOT NULL default '0',
	loginkey varchar(50) NOT NULL default '',
	ip bytea NOT NULL default '\x'::bytea,
	dateline uint NOT NULL default '0',
	lastactive uint NOT NULL default '0',
	data text NOT NULL default '',
	useragent varchar(200) NOT NULL default '',
	authenticated smallint NOT NULL default '0'
);";
```
versus:
```php
$tables[] = "CREATE TABLE mybb_adminsessions (
	sid varchar(32) NOT NULL default '',
	uid int NOT NULL default '0',
	loginkey varchar(50) NOT NULL default '',
	ip bytea NOT NULL default '\x'::bytea,
	dateline int NOT NULL default '0' CHECK (dateline >= 0),
	lastactive int NOT NULL default '0' CHECK (lastactive >= 0),
	data text NOT NULL default '',
	useragent varchar(200) NOT NULL default '',
	authenticated smallint NOT NULL default '0'
);";
````

dvz:
[...]

PR #5126, opened 2025-10-28
Add check constraints for disporder columns

arcalys:
### Added

- Added `CHECK` constraints for disporder columns in SQLite and PostgreSQL.

Part of #4918

### DIFF
--- a/inc/schemas/sqlite_db_tables.php
+++ b/inc/schemas/sqlite_db_tables.php
@@ -164,7 +164,7 @@ $tables[] = "CREATE INDEX mybb_buddyrequests_touid ON mybb_buddyrequests (touid)
 $tables[] = "CREATE TABLE mybb_calendars (
 	cid INTEGER PRIMARY KEY,
 	name varchar(100) NOT NULL default '',
-	disporder smallint NOT NULL default '0',
+	disporder smallint NOT NULL default '0' CHECK (disporder >= 0),
 	startofweek tinyint(1) NOT NULL default '0',
 	showbirthdays tinyint(1) NOT NULL default '0',
 	eventlimit smallint(3) NOT NULL default '0',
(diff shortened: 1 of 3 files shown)


Sama34:
From the doc for both MySQL and MariaDB, it seems they support `CHECK` constraints. Why would we be leaving these out then ?

Of course, it may be left for a different PR, so this is really just a question nor a request to implement in the same PR.

Regards.

arcalys:
> From the doc for both MySQL and MariaDB, it seems they support `CHECK` constraints. Why would we be leaving these out then ?
> 
> Of course, it may be left for a different PR, so this is really just a question nor a request to implement in the same PR.
> 
> Regards.

That’s already implicitly covered on the MySQL side — the equivalent disporder fields are declared as UNSIGNED INT, which enforces non-negative values at the type level. The CHECK constraints were therefore only needed for PostgreSQL and SQLite, which don’t have unsigned integer types.

I'm not opposed to adding the equivalent CHECK constraints to MySQL, but that would be redundant and would require converting the unsigned integers to signed integers first.

arcalys:
I still need to add the upgrade script, by the way. One question arises: what should I do regarding SQLite? SQLite does not support `ALTER TABLE … ADD CONSTRAINT`.

inc/schemas/sqlite_db_tables.php
```diff
@@ -339,7 +339,7 @@
 	document TEXT NOT NULL,
 	usetranslation tinyint(1) NOT NULL default '0',
 	enabled tinyint(1) NOT NULL default '0',
-	disporder smallint NOT NULL default '0'
+	disporder smallint NOT NULL default '0' CHECK (disporder >= 0
```
dvz: Looks like missing `)` here and below

dvz:
We can probably skip SQLite in the upgrade, then.

It may also be worth resetting potentially invalid values before schema changes (for all engines).

inc/schemas/sqlite_db_tables.php
arcalys: Fixed

arcalys:
> We can probably skip SQLite in the upgrade, then.
> 
> It may also be worth resetting potentially invalid values before schema changes (for all engines).

I will add the reset to the upgrade script along with the constraints.

Sama34:
> I'm not opposed to adding the equivalent CHECK constraints to MySQL, but that would be redundant and would require converting the unsigned integers to signed integers first.

Never mind, of course, if it is redundant then there is no need to add.

Regards.

arcalys:
Updated the upgrade script, adding the check constraints if they do not exist, and resetting the column value if < 0. Tested with multiple upgrades on my end, but an additional test would be welcome.

Issue #4793, opened 2024-03-31
get_profile_link escapes & in profile URL due to htmlspecialchars, breaks URL


AniLeo:
Hello,

On MyBB 1.8.37 (and ever since I remember), the profile URL in the contact form is escaped in a way where it breaks the URL.

For instance:
Expected: https://forums.rpcs3.net/member.php?action=profile&uid=1
Actual: https://forums.rpcs3.net/member.php?action=profile&amp;uid=1

The `&` is escaped to `&amp;`, breaking the URL.

Culprit code is:
https://github.com/mybb/mybb/blob/9a701b248308c499641d40ca09b5eb3bda76f9c6/inc/functions.php#L6405

https://github.com/mybb/mybb/blob/9a701b248308c499641d40ca09b5eb3bda76f9c6/contact.php#L219

To reproduce, you need to be logged into a forum account, and use the contact.php form to send a message. Then check the received email and see the user profile URL in the email.

Let me know if you need further details,
Ani

PR #5129, opened 2025-11-04
Remove |raw usage for links, usernames and dates in showthread template

arcalys:
### Removed

- Computed links and values server-side for threads, posts, and users.

### Changed

- Replaced all `|raw` filter usage with appropriate Twig filters depending on the scenario:
  - `my_date` for dates
  - `format_name` for usernames
  - `get_profile_link` for user profile links
  - `get_thread_link` for thread links

This is a test with the showthread template only to check if this is the way you guys would like to go.

### DIFF
--- a/inc/themes/core.base/frontend/templates/showthread/showthread.twig
+++ b/inc/themes/core.base/frontend/templates/showthread/showthread.twig
@@ -244,13 +244,27 @@
                     {% for post in threadedbits %}
                         {% if post.bitactive %}
                             <div class="threaded-mode__post threaded-mode__post--active" style="margin-left: {{ post.indentsize }}px;">
-                                <span class="threaded-mode__date">{{ post.postdate|raw }}</span>
-                                <span class="threaded-mode__author">{{ lang.by }} {{ post.profilelink|raw }}</span>
+                                <span class="threaded-mode__date">{{ post.postdate }}</span>
+                                <span class="threaded-mode__author">
+                                    {{ lang.by }}
+                                    {% if post.uid > 0 %}
+                                        <a href="{{ get_profile_link(post.uid) }}">{{ post.uid|format_name }}</a>
+                                    {% else %}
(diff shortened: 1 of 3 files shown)


inc/themes/core.base/frontend/templates/showthread/showthread.twig
```diff
@@ -244,13 +244,27 @@
...
+                                    {% else %}
+                                        {{ post.username|default(lang.guest) }}
+                                    {% endif %}
+                                </span>
                             </div>
                         {% else %}
                             <div class="threaded-mode__post" style="margin-left: {{ post.indentsize }}px;">
-                                <span class="threaded-mode__date"><a href="showthread.php?tid={{ thread.tid }}&amp;pid={{ post.pid }}&amp;mode=threaded">{{ post.postdate|raw }}</a></span>
-                                <span class="threaded-mode__author">{{ lang.by }} {{ post.profilelink|raw }}</span>
+                                <span class="threaded-mode__date"><a href="showthread.php?tid={{ thread.tid }}&amp;pid={{ post.pid }}&amp;mode=threaded">{{ post.postdate }}</a></span>
```
Sama34, line 259: Shouldn't this use `get_post_link` (with a new parameter similar to `get_thread_link`) ?

dvz:
We could investigate whether the `get_*_link()` functions could use `urlencode()` instead of `htmlspecialchars_uni()` (related: #4793).

Sama34:
Maybe we should fix that first before merging this.

arcalys:
> Maybe we should fix that first before merging this.

I'll submit a second PR and rebase this afterwards.

arcalys:
I’ll address the functions first (move to `urlencode`); we can revisit this later if needed.

Sama34:
Will you reopen this or open a new one ? I ask because this and #5131 don't seem to reference any specific issue.

PR #5131, opened 2025-11-06
Add mode parameter to get_post_link

arcalys:
### Added
- Added rewrite rules to support the new `mode` parameter (`linear`, `threaded`) in pretty URLs.

### Changed
- Replaced manually constructed post URL in the `showthread` template with a call to `get_post_link()` as an example.
- Refactored `get_post_link()`:
  - Added parameter and return types.
  - Applied `urlencode()` to individual parameters and used `strtr()` for placeholder substitution.
- Updated the related `get_post_link` Twig function to match the new signature and behavior.

This follows a discussion in https://github.com/mybb/mybb/pull/5129.  
I have not updated the other `get_*_link` functions yet.  
If this approach is accepted, I will submit a separate PR to update the remaining link-generation functions accordingly.


Sama34:
I'm unsure about this one.

- `get_post_link()` API is different from `get_thread_link()`
- I think `get_post_link()` should add a `action` parameter just like `get_thread_link()`
- Maybe `get_post_link()` should be a wrapper for `get_thread_link()` ? Thus adding `$pid` as a parameter to `get_thread_link()`
- I see no reason not to add `$mode` to `get_thread_link()` in this PR.

@dvz what you think ?

Regards.

PR #5143, opened 2025-11-16
Show default avatar for guest and installer accounts

### LINKED ISSUE #5142: MyBB Installer Avatar
For consistency, the thread which is generated following a new install of 1.9 does not show the standard avatar for "MyBB Installer".

arcalys:
### Changed

- Moved the `$post['showavatar']` block outside the `$post['userusername']` check to ensure avatars are processed for `uid=0`.

Resolves #5142


arcalys:
Can you double-check this, @Ben-MyBB?
Tested with `showavatars` both enabled and disabled on my end; behaviour appears correct.

PR #5147, opened 2025-11-20
DarkMode Toggle ReWork

joogiebear:
Made a slight change to the dark mode toggle method. 
Prevent flashing
Friendly Redirection page fix


joogiebear:
Forum Discussion:

https://community.mybb.com/thread-244396.html

Ben-MyBB:
@justinsoltesz

justinsoltesz:
There should be a third option to tie it to your system light/dark setting.

I also don't think it belongs in the header bar. I could see it in the user bar instead, though I think the best place for it is probably the footer bar. That's where the quick theme and language selectors have always been. (Other forum softwares place the light/dark/system option there as well)

joogiebear:
Pushed light/dark/system button to footer.

Ben-MyBB:
@justinsoltesz thoughts here? :)

PR #4376, opened 2021-05-22
Ability to suspend private messaging ability

Xazin:
https://community.mybb.com/thread-195240.html

Based on the above-planned feature, the ability to suspend a user's private messaging abilities. If a user is suspended using private messaging, the user can neither send nor receive new private messages.

Options to moderate suspension of PM added both AdminCP and ModCP similar to suspend posts and moderate posts.

_Note: Should also be mentioned, this restricts access to 'private.php', similar to if a usergroup didn't have access to Private Messaging._

PR #5153, opened 2025-11-22
Add the ability to suspend a user's private messaging

### LINKED ISSUE #5134: Ability to suspend private messaging ability
From  #4376 

This feature was rejected for the 1.8 branch, but would be nice to have in the 1.9 release.

arcalys:
Based on https://github.com/mybb/mybb/pull/4376.

### Added

- Added the ability to suspend a user's private messaging. Based on Xazin's work, the main difference is that users can still access their inbox and message folders, but they cannot compose new messages or reply to existing ones.

### Fixed

- Fixed an issue where moderator options in the ACP were not saved because of an incorrect conditional.

Resolves #5134

### DIFF
--- a/inc/languages/english/admin/user_users.lang.php
+++ b/inc/languages/english/admin/user_users.lang.php
@@ -257,6 +257,9 @@ $l['suspend_for'] = "Suspend for:";
 $l['suspended_perm'] = "<p><small>Suspended permanently.<br />Enter a new time below to change or untick this option to remove this suspension.</small></p>";
 $l['suspend_length'] = "<p><small>Remaining Suspension: <span style=\"color: {2};\">{1}</span>.<br />Enter a new time below to change or untick this option to remove this suspension.</small></p>";
 
+$l['suspend_pm'] = "Suspend private messaging";
+$l['suspend_pm_info'] = "Suspend {1} from sending and receiving new private messages.";
+
 $l['suspendsignature_error'] = "You selected to suspend this user's signature, but didn't enter a valid time period. Please enter a valid time to continue or untick the option to cancel.";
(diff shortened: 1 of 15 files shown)


arcalys:
Keeping this as a draft until the questions in the issue are addressed.

inc/languages/english/admin/user_users.lang.php
```diff
@@ -257,6 +257,9 @@
 $l['suspended_perm'] = "<p><small>Suspended permanently.<br />Enter a new time below to change or untick this option to remove this suspension.</small></p>";
 $l['suspend_length'] = "<p><small>Remaining Suspension: <span style=\"color: {2};\">{1}</span>.<br />Enter a new time below to change or untick this option to remove this suspension.</small></p>";
 
+$l['suspend_pm'] = "Suspend private messaging";
+$l['suspend_pm_info'] = "Suspend {1} from sending new private messages.";
```
dvz: This explicit mention of _sending_ only would need to be changed as per https://github.com/mybb/mybb/issues/5134#issuecomment-3566929455

arcalys: Forgot this one, updated!

private.php
```diff
@@ -554,6 +554,17 @@
 	{
 		error_no_permission();
 	}
+	
+	if($mybb->user['suspendpm'] === 1)
```
Sama34: Will this exact check be valid always ? I don't see the logic for it, so think maybe not.

private.php
```diff
@@ -667,6 +678,17 @@
 		error_no_permission();
 	}
 
+	if($mybb->user['suspendpm'] === 1)
```
Sama34: Same.

Sama34 (changes requested):
In `messenger.twig` you should update `{% if mybb.usergroup.cansendpms %}` to `{% if mybb.usergroup.cansendpms and mybb.user.suspendpm == 0 %}`.

PR #5155, opened 2025-11-22
Fix intentionally bad requests as array submits causing PHP fatal errors

### LINKED ISSUE #5150: Intentionally bad requests as array submits causing PHP fatal errors
https://community.mybb.com/thread-244560-post-1404955.html#pid1404955

Sama34:
Resolves #5150 (probably requires a new issue for the `attachments.php` refactor.)

Also refactors the `attachments.php` logic to load `/inc/init.php` instead of `global.php`.

Feedback is welcome.


dvz:
Dropping `$nosession['avatar']` seems OK; we could keep `$nosession` recognition for plugins.

`NO_ONLINE` for thumbnails could be moved from `global.php` in 1.8, if no compatibility issues are expected.

There are a few areas with non-standard initialization (`attachment.php` seems conceptually similar to `xmlhttp.php`); there will be some View/Theme-related changes for initialization, and we can leave a refactoring for later if there's no particular need now.

The ACP change can be dropped here (we can handle it when forward-porting 1.8 commits).

Ben-MyBB:
What about https://github.com/mybb/mybb/pull/5151 ?

Sama34:
I was expecting ACP related changes to disappear once the repository was synched to upstream, so I'm a bit confused with the reason it remains.

I will check it, of course; ACP changes are not intended part of this PR.

About the `attachment.php` refactoring, what do you mean with _later_ ? You mean for 1.9 or for 1.10 ?

I would suggest to just refactor this for 1.19, but surely we can do an independent issue for it.

@Ben-MyBB that would be the suggested solution for 1.8 as to keep refactoring for 1.9 mostly

Regards.

dvz:
The big hunk copies part of `global.php` - I wonder if it's worth the code duplication (https://github.com/mybb/mybb/pull/5155#discussion_r2553359525).

There are 3 similar, theme-related code sections across `global.php`, `xmlhttp.php`, and this `attachment.php`:
- establishing the Theme DB entry (`Database\Models\Theme`) to use ([example](https://github.com/mybb/mybb/blob/a55243ad859f0eb5b4ed8285c89df1fd76c04e25/xmlhttp.php#L69-L119))
- overriding the default Theme Package (`Extensions\Theme`) and View ([example](https://github.com/mybb/mybb/blob/a55243ad859f0eb5b4ed8285c89df1fd76c04e25/xmlhttp.php#L121-L137))
- building the legacy `$theme` array ([example](https://github.com/mybb/mybb/blob/a55243ad859f0eb5b4ed8285c89df1fd76c04e25/xmlhttp.php#L140-L202))

The 1st has more logic in `global.php`, but I'll be looking to put the 3rd into a function, which would reduce duplication here.


> I was expecting ACP related changes to disappear once the repository was synched to upstream, so I'm a bit confused with the reason it remains.

They're not there yet (changes from 1.8.x are usually cherry-picked to `dev-1.9` after a release).

Sama34:
I will probably merge as is (duplication of `global.php` logic) and leave that as a separated issue unless there is explicit disagreement.

Though I will probably redo the PR if rebasing fails (meanwhile on hold for feedback).

dvz:
Does it duplicate it for performance only now?

Sama34:
Seems like you are correct (or now I understand better).

Yes, the change seems to be unnecessary for the issue reported. I will update/redo accordingly and make a separate issue.

Issue #5122, opened 2025-10-25
Normalize usergroup membership data


dvz:
Use a pivot table for usergroup memberships, linking `mybb_users` and `mybb_usergroups` records.

Currently, user group information is stored in `mybb_users` table fields:
- `usergroup`
- `displaygroup`
- `additionalgroups` in comma-separated format (CSV)

Existing values may be left for compatibility and caching.

Sama34:
https://community.mybb.com/thread-205925.html

PR #5161, opened 2025-11-23
Exclude banned users from newest-member stats calculation

### LINKED ISSUE #5156: Banned User should not be in ( welcome list)
If we bann new user (spam) it should not be shown on board statistics as welcome new memeber { Username (banned)}

arcalys:
### Changed

- Updated the query retrieving the last registered member in `update_stats` to exclude the banned user group.

Resolves #5156

### DIFF
--- a/inc/functions.php
+++ b/inc/functions.php
@@ -2846,9 +2846,44 @@ function update_stats($changes = array(), $force = false)
 	// Fetch latest user if the user count is changing
 	if(array_key_exists('numusers', $changes))
 	{
-		$query = $db->simple_select("users", "uid, username", "", array('order_by' => 'regdate', 'order_dir' => 'DESC', 'limit' => 1));
+		$banned_groups = [];
+		$conditions = [];
+
+		$query = $db->simple_select("usergroups", "gid", "isbannedgroup = 1");
+		while($group = $db->fetch_array($query))
+		{
+			$banned_groups[] = (int)$group['gid'];
+		}
+
+		if(!empty($banned_groups))
+		{
+			$conditions[] = "u.usergroup NOT IN (" . implode(',', $banned_groups) . ")";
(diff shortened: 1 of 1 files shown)


Ben-MyBB:
What about preventing a banned person's username being shown as the most recent member?

arcalys:
The most recent member is taken from the stats cache. So updating this and refreshing the cache seems to be enough.

arcalys:
Do we want this included in 1.8 as well @Ben-MyBB ?

Ben-MyBB:
I would just leave it for 1.9 personallyOn 29 Nov 2025, at 14:24, Axel Fontana ***@***.***> wrote:﻿arcalys left a comment (mybb/mybb#5161)
Do we want this included in 1.8 as well @Ben-MyBB ?

Sama34:
The issue with the banned group is that custom groups can be marked as banned groups :
image

~~I think this would also apply for other PRs.~~

Regards.

arcalys:
> The issue with the banned group is that custom groups can be marked as banned groups : image
> 
> ~I think this would also apply for other PRs.~
> 
> Regards.

Will check how we can handle that, it's a bit more complicated with all groups (additionalgroups,...).

Sama34:
Just do the regular ugly `CONCAT` stuff, and we will fix the ugliness eventually on #5122

arcalys:
@Ben-MyBB @Sama34 
Updated to go through all banned groups. Tested with:
- Default banned group membership (primary, additional)
- Newly created banned group (primary, additional).

inc/functions.php
Sama34, line 2779:
```suggestion
```

Sama34: I was going to commit before merging but clicked too quick hehe

PR #5166, opened 2025-11-26
Add the ability to hide time spent online

### LINKED ISSUE #5158: Hide members time online via option
By default there is no option to show/hide the time online of a member.
It's possible to entirely use ghost mode (being invisible) only.

This additional option can be added with less effort in a future release and be achieved by a usergroup option and/or a member setting.
For example:

ACP > Groups > Edit User Group:
Can view timeonline [Yes|No]
($mybb->usergroup['canviewtol'])

User CP > Edit Options > Privacy:
"Hide my total time online" [yes|No]
($memprofile['hidetimeonline'])

Perhaps make this a general option in forum configuration (enable/disable at all).

arcalys:
### Added

- Added the ability to hide time spent online based on a new user option (`showtimespentonline`) that defaults to true.

Resolves #5158

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -563,6 +563,7 @@ if($mybb->input['action'] == "edit")
 			"hideemail" => $mybb->get_input('hideemail'),
 			"subscriptionmethod" => $mybb->get_input('subscriptionmethod'),
 			"invisible" => $mybb->get_input('invisible'),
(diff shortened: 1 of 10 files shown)


Sama34:
Can we implement snake case for new database table and column names ? It is already hard to read the code, so I think this is important from a documentation point of view.

Regards.

Issue #4416, opened 2021-07-17
Enhancement: $subject_length > 85


AniLeo:
https://github.com/mybb/mybb/blob/feature/inc/datahandlers/post.php#L226

This value is the maximum length a thread subject can have.

It should ideally be configurable in ACP, even if there's an upper limit, allowing values up to 100 or 120. I need to have this limit at 100 as 85 is not enough for my use case, thus needing to change it on code. 

Since this is stored on database `mybb_threads` -> `subject` as a varchar(120), there's seemingly no issue in increasing this limit up to 120 right now.

Eldenroot:
Invalid, rejected... duplication, I made a PR, they reverted it---wait for 1.9

AniLeo:
What PR are you referring to? I can only find #4344 and #4288, which are not about the same thing

arcalys:
Should we simply adapt these checks to use 120 characters as well, @dvz @Sama34?  
This also applies to private messages, which have a 120-character limit in the database.

dvz:
Possibly; the `RE:` insertion logic may affect this.

arcalys:
We could adjust the truncation logic to something like this:

```php
$max_subject_length = 120;
$ellipsis = '...';
$reply_prefix = 'RE: ';

if (my_strlen($thread['reply_subject']) > $max_subject_length - my_strlen($reply_prefix)) {
    $thread['reply_subject'] = my_substr(
        $thread['reply_subject'],
        0,
        $max_subject_length - my_strlen($reply_prefix) - my_strlen($ellipsis)
    ) . $ellipsis;
}
```

Both the ellipsis and the reply prefix could even be registered as language keys.

Sama34:
I think there are 3 issues pointed in this discussion:

- Maximum character limit is hardcoded, a setting should be added which value should be between 1 and 120.
- Truncated subject logic should be moved to CSS and not PHP logic.
- Adding `RE: ` to post subjects makes no sense, was this based on how emails work ?

I think we should limit this issue to fix the first point listed above and leave the other 2 to its own issues and discuss them there instead.

Regards.

Eldenroot:
My pro with thé CSS solution was rejected by mybb team... I think it would be enough, clean solution.

Sama34:
Could you link to it ? I see no issue with using CSS for 1.9

dvz:
IMO we don't need a setting, and can support the database column maximum. The values would need checking on insert; for displaying, 1.9 uses CSS.

[...]

PR #5172, opened 2025-11-28
Increase max length for PM subjects to match database column

arcalys:
### Changed

- Moved the subject length to a property and increased value to 120 characters to match the database column (`VARCHAR(120)`).

Related to #4416


Sama34:
I think we should start implementing typed properties for new or updated code.

PR #5174, opened 2025-11-30
Display errors for empty subject/messages in PM preview

### LINKED ISSUE #4115: Preview empty new message doesn't throw errors
Hello,

Steps:

Go to /private.php?action=send

Click Preview button

You do not see errors, that the message is empty, as it happens after clicking save as a draft or preview an empty post on forums

Original thread: [Preview empty new message doesn't throw errors](https://community.mybb.com/thread-228514.html)

arcalys:
### Changed

- Added display of errors for empty subject or message when previewing private messages.
- Removed `$options = $mybb->get_input('options', MyBB::INPUT_ARRAY);` from the second preview block because it's already set a few lines before.

### Example

image

Resolves #4115

### DIFF
--- a/private.php
+++ b/private.php
@@ -736,7 +736,6 @@ if($mybb->input['action'] == "send")
 	// Preview
 	if(!empty($mybb->input['preview']))
 	{
-		$options = $mybb->get_input('options', MyBB::INPUT_ARRAY);
 		$query = $db->query("
 			SELECT u.username AS userusername, u.*, f.*
 			FROM ".TABLE_PREFIX."users u
@@ -751,6 +750,8 @@ if($mybb->input['action'] == "send")
 		$post['message'] = $mybb->get_input('message');
 		$post['subject'] = htmlspecialchars_uni($mybb->get_input('subject'));
 		$post['icon'] = $mybb->get_input('icon', MyBB::INPUT_INT);
+		$post['to'] = $to;
+		$post['bcc'] = $bcc;
 		if(!isset($options['disablesmilies']))
 		{
 			$options['disablesmilies'] = 0;
@@ -766,6 +767,8 @@ if($mybb->input['action'] == "send")
 		{
 			$post['includesig'] = 1;
 		}
+
+		$post['options'] = $options;
 
 		// Merge usergroup data from the cache
 		$data_key = array(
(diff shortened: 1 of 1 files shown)


Sama34:
Why not just `validate_pm()`'d ? To be frank, the logic seems a bit confusing and inconsistent with how posts are handled.

arcalys:
> Why not just `validate_pm()`'d ? To be frank, the logic seems a bit confusing and inconsistent with how posts are handled.

I originally did not use `validate_pm()` because the PM data used for the preview did not contain all the required fields (such as options, to, bcc, etc.), which caused it to fail in `verify_options()` and `verify_recipient()`.

Also, unless I’m missing something, this is how the preview is currently handled in newreply and newthread:
```php
	// Preview a post that was written.
	$preview = '';
	if(!empty($mybb->input['previewpost']))
	{
	....
	....
		$posthandler->set_data($post);

		// Now let the post handler do all the hard work.
		$valid_post = $posthandler->verify_message();
		$valid_subject = $posthandler->verify_subject();

		...
		...

		$post_errors = array();
		// Fetch friendly error messages if this is an invalid post
		if(!$valid_post || !$valid_subject || !$valid_username)
		{	
```

That said, I have now updated the PM preview validation to include the missing data, so we can call `validate_pm()` as suggested.

private.php
```diff
@@ -783,6 +786,20 @@
...
+
+		$send_errors = '';
+		if(!$pmhandler->validate_pm())
+		{
+			$send_errors = $pmhandler->get_friendly_errors();
+			if(!empty($send_errors))
+			{
+				$send_errors = inline_error($send_errors);
+			}
+		}
```
Sama34, line 804:
```suggestion

			$sendpm['preview'] = false;
		}
```

private.php
```diff
@@ -783,6 +786,20 @@
...
+
 		$postbit = build_postbit($post, 2);
```
Sama34:
```suggestion
		if($sendpm['preview'])
		{
				$postbit = build_postbit($post, 2);
		}
```

Sama34 (changes requested):
Te changes are necessary to not render the post bit :

Image

With suggestions :

Image

Sama34:
The suggested changes also make it consistent with how post previews are (not) rendered when there are errors :
image

> Also, unless I’m missing something, this is how the preview is currently handled in newreply and newthread:

You are correct, I mixed something somewhere. (But it seems nicer to do it better when possible :D)

PR #5180, opened 2025-12-04
Allow overriding task titles and descriptions via language keys

arcalys:
### Added

Added the ability to override titles and descriptions for planned tasks using language keys/files. This would eliminate the need to maintain `tasks.xml` in language packs. There may be a few additional places where this should be applied, but I wanted to check first if this is something people would be interested in.

### DIFF
--- a/inc/functions_task.php
+++ b/inc/functions_task.php
@@ -18,6 +18,8 @@ function run_task($tid=0)
 {
 	global $db, $mybb, $cache, $plugins, $task, $lang;
 
+	$lang->load('tools_tasks');
+
 	// Run a specific task
 	if($tid > 0)
 	{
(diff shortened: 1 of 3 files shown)


inc/functions_task.php
Sama34, line 19:
```suggestion
	global $db, $mybb, $cache, $plugins, $task, $lang;
	
	$lang->load('tools_tasks');
```

Sama34 (changes requested):

inc/functions_task.php
```diff
@@ -68,6 +70,15 @@ function run_task($tid=0)
 		);
 		$db->update_query("tasks", $updated_task, "tid='{$task['tid']}'");
 
+		if(!empty($task['file']))
```
Sama34, line 73: Foremost, this should be outside the if statement.

Regardless, seems like this file needs no edition, as the task title for logging seems to be managed in the following block:

https://github.com/mybb/mybb/blob/7e0247a0874cc28c5b5c44c98d18139738f2140f/admin/modules/tools/tasks.php#L677-L694

As you can see, the code fetches the task name directly from the database (it even ignores the possibility of the task no longer existing...).

Regards.

Sama34 (changes requested):

inc/functions_task.php
arcalys: I will add the translation to this for the tasks logs which I forgot, but the change in `functions_task` was actually to translate the name of the task being sent by email when a task fail.

Sama34 (approved):

Issue #4863, opened 2024-11-30
Add global variables from plugin hooks to Twig context


dvz:
### MyBB ≤ 1.8
In MyBB ≤ 1.8, templates rendered using `eval()` have access to all variables available in the executing scope.

Custom plugin variables for use in templates are, therefore, usually set in:
- scope relevant to target templates, by modifying existing core variables (through hook function arguments or return values),
- the global scope (where some core templates are rendered, e.g. top-level `*.php` files).

### MyBB 1.9
With the introduction of the Twig template engine, only explicitly provided context variables are accessible in templates.

Core variables are provided locally ([example](https://github.com/mybb/mybb/blob/976321e8f9474540925d0d6d09e0cfdc0d79848d/showthread.php#L1422-L1439)), and the Theme System provides a [helper function](https://github.com/mybb/mybb/blob/976321e8f9474540925d0d6d09e0cfdc0d79848d/inc/src/View/functions.php#L137-L147) to make variables available in all rendered templates.

## Current State
Depending on local core code changes between 1.8 and 1.9, modifying core variables (e.g. adding another key to the [`$post` variable](https://github.com/mybb/mybb/blob/976321e8f9474540925d0d6d09e0cfdc0d79848d/inc/functions_post.php#L913)) is expected to work without changes.

For new global variables, updating a plugin from MyBB 1.8 to MyBB 1.9 would involve explicitly sharing the variable using the helper function, instead of simply declaring it as global:
```diff
function my_hook()
{
-   global $my_widget;

    $my_widget = '...';

+   \MyBB\View\set([
+       'my_widget' => $my_widget,
+   ]);
}
```

## Improvements
Global variables added during hook execution in [`pluginSystem::run_hooks()`](https://github.com/mybb/mybb/blob/976321e8f9474540925d0d6d09e0cfdc0d79848d/inc/class_plugins.php#L158-L177) can be collected automatically in a manner similar to:
```php
$existingGlobals = array_keys($GLOBALS);

// run hooks...

$newGlobals = array_diff_key(
    $GLOBALS,
    array_flip($existingGlobals),
);

\MyBB\View\set($newGlobals);
```

The same technique may be used to collect variables set by plugins outside of hook functions in [`pluginSystem::load()`](https://github.com/mybb/mybb/blob/976321e8f9474540925d0d6d09e0cfdc0d79848d/inc/class_plugins.php#L39).

The performance impact may be proportional to the number of effectively used hooks provided by the core.

[...]

PR #5186, opened 2025-12-06
Add config flag check for collecting plugins' global variables

dvz:
Adds a config flag to compatibility feature added in #4863, similarly to #5165

Enabled by default; can be disabled with a [Configuration FIle](https://docs.mybb.com/1.8/administration/configuration-file/) flag:
```php
$config['compat_plugin_globals'] = false;
```

### DIFF
--- a/inc/class_plugins.php
+++ b/inc/class_plugins.php
@@ -148,6 +148,8 @@ class pluginSystem
 	 */
 	function run_hooks(string $hook, &$arguments = "")
 	{
+		global $mybb;
+
 		if(!isset($this->hooks[$hook]) || !is_array($this->hooks[$hook]))
 		{
 			return $arguments;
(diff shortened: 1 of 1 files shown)


arcalys (approved):

inc/class_plugins.php
```diff
@@ -157,7 +159,10 @@ function run_hooks(string $hook, &$arguments = "")
 
 		$this->current_hook_stack[] = $this->current_hook;
 
-		$existing_globals = array_keys($GLOBALS);
+		if($mybb->config['compat_plugin_globals'] ?? true)
```
Sama34, line 162: Doesn't this mean it will be enabled regardless of the set value in config ? I noticed the same in the other flag.

dvz: `true` would be used if the value is not defined or is `null` (similar to `isset()` behavior; different from `empty()` behavior)

https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce

PR #5198, opened 2025-12-08
Add the ability to suspend a user's avatar

### LINKED ISSUE #5184: Add the ability to suspend a user's avatar
Add the ability to suspend a user's avatar through both the moderator CP and admin CP, similar to the existing option for suspending signatures.

Original discussion: https://community.mybb.com/thread-194657.html

arcalys:
### Added

Added the ability to suspend a user's avatar privileges. When suspended, the user's current avatar is removed so that it is automatically replaced with the default avatar.

image
image
image

Resolves #5184


Sama34:
It looks weird by default:

image

After just ticking the checkbox, it does not save (no error shown), yet the following message is shown:

image

The item menu should disappear if the user has no access to the page :
image

If I set `1` and `Permanent` the following is shown :
image

arcalys:
If you tick the box and do not change the other values, it saves with "permanent". 
Not saying this is not weird, but this is 100% the same behaviour and display as the other items:
image
image
image


Maybe we need to refactor that page for all moderation items...

Sama34:
I see, but it is indeed weird that it says "suspended until 1 minute ago".

Fix what you can fix in the scope of this PR and let me know to test again.

I think the menu hiding is a valid suggestion ?

We can open a new issue to fix whatever is causing this broken logic.

arcalys:
> I see, but it is indeed weird that it says "suspended until 1 minute ago".
> 
> Fix what you can fix in the scope of this PR and let me know to test again.
> 
> I think the menu hiding is a valid suggestion ?
> 
> We can open a new issue to fix whatever is causing this broken logic.

When you say the menu item should disappear, are you referring to the “avatar” tab? Because if that’s the case, the “signature” tab doesn’t disappear when a user’s signature is suspended. Why should this one behave differently?

Sama34:
I think both should disappear, so my opinion is that you implement the proper behaviour here for the avatar permission.

Edit: Maybe @justinsoltesz has some input regarding UX

arcalys:
> I think both should disappear, so my opinion is that you implement the proper behaviour here for the avatar permission.
> 
> Edit: Maybe @justinsoltesz has some input regarding UX

Updated to hide the profile menu item if suspended.
There is a `TODO` comment for the signature saying `{# TO DO: need to check if user can change signature #}`, so I guess you are right in saying it should be hidden.

PR #5210, opened 2025-12-15
Update moderation suspension options

### LINKED ISSUE #4422: Set default value for moderation/suspension duration options in Mod CP's Edit Profile
Set the default duration to _Permanent_ for the following options in the Mod CP's profile editor to improve usability:
- _Moderate this user's posts_
- _Suspend this user's posting privileges_
- _Suspend this user's signature_

arcalys:
### Added

- Added `No change` option (set to `""`) to avoid overriding existing suspensions when existing.

### Changed

- If no existing suspension, `Permanent` is the default option in the list.

### Fixed

- Incorrect conditional for signature suspension message: `suspendsigtime` should be `suspendsignature` because time is set to 0 for permanent suspension.
- Missing `|raw` for permanent suspension messages that contain HTML tags.

Resolves #4422

### DIFF
--- a/inc/themes/core.base/frontend/templates/modcp/editprofile.twig
+++ b/inc/themes/core.base/frontend/templates/modcp/editprofile.twig
@@ -196,7 +196,7 @@
                             {% if user.suspendsigtime %}
                                 <p class="field__value">
                                     {% if user.suspendsigtime == 0 %}
-                                        {{ lang.suspendsignature_perm }}
+                                        {{ lang.suspendsignature_perm|raw }}
                                     {% else %}
                                         {{ trans('suspendsignature_for', user.suspendsigtime|my_date) }}
                                     {% endif %}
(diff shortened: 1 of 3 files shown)


inc/themes/core.base/frontend/templates/modcp/editprofile.twig
```diff
@@ -193,10 +193,10 @@
                             <label><input type="checkbox" name="suspendsignature" value="1"{% if user.suspendsignature or (mybb.input.suspendsignature and errors) %} checked{% endif %} class="checkbox" onclick="toggleAction();" /> {{ lang.suspend_signature }}</label>
                         </p>
                         <div class="field__block" id="suspend_action">
-                            {% if user.suspendsigtime %}
+                            {% if user.suspendsignature %}
                                 <p class="field__value">
                                     {% if user.suspendsigtime == 0 %}
-                                        {{ lang.suspendsignature_perm }}
+                                        {{ lang.suspendsignature_perm|raw }}
```
Sama34, line 199: I think we should stop using any HTML in templates, even breaks.

No language string should have to use the `|raw` option.

Sama34 (changes requested):

inc/themes/core.base/frontend/templates/modcp/editprofile.twig
arcalys: I'd suggest keeping it here for the sake of this PR. I will refactor the language keys and `raw` separately ;-)

Issue #5171, opened 2025-11-28
Minimize reliance on raw filter in Twig templates


arcalys:
Most templates currently rely on the `|raw` filter to display URLs and content. The idea would be to:
- Update the `get_*_link` functions to use `urlencode()` instead of `htmlspecialchars_uni()`.
- Call the `get_*_link` functions directly from the templates (e.g., replace `{{ post.profilelink|raw }}` with `{{ get_profile_link(post.uid) }}`).

PR #5211, opened 2025-12-15
Remove raw usage for links and dates in index template

arcalys:
### Removed

- Computed links and values for profile links and dates.

### Changed

- Replaced `|raw` filter usage with appropriate Twig filters.

Part of #5171


inc/themes/core.base/frontend/templates/index/index.twig
```diff
@@ -66,8 +68,14 @@
                 <p class="board-statistics__body">
                     {{ lang.stats_posts_threads }}<br />
                     {{ lang.stats_numusers }}<br />
-                    {{ lang.stats_newestuser|raw }}<br />
-                    {{ lang.stats_mostonline|raw }}
+                    {{ lang.stats_newestuser }},
+                    {% if newestmember == 0 %}
+                        {{ lang.nobody }}
+                    {% else %}
+                        <a href="{{ get_profile_link(newestmember) }}">{{ newestmember|format_name }}</a>
```
Sama34: Why not add a helper for `build_profile_link()`? I'm concerned about the inconsistencies on formatted built links.

index.php
```diff
@@ -314,14 +302,7 @@
...
 	// Check who's the newest member.
-	if(!$stats['lastusername'])
-	{
-		$newestmember = $lang->nobody;;
-	}
-	else
-	{
-		$newestmember = build_profile_link($stats['lastusername'], $stats['lastuid']);
-	}
+	$newestmember = !$stats['lastusername'] ? 0 : $stats['lastuid'];
```
Sama34: This variable seems redundant, maybe inline `$stats['lastuid'] ?? 0` ?

inc/themes/core.base/frontend/templates/index/index.twig
arcalys: I'm not sure if this is the right approach, but even if we added build_profile_link, it would still be implemented separately, wouldn’t it?

Sama34: Yes, it would be implemented separately.

Alternatively, maybe the decrease of usage of the `build_profile_link` may be the desired approach instead of the contrary.

Since we are trying to move some basic logic to templates, the way you are doing it seems reasonable.

inc/themes/core.base/frontend/templates/index/index.twig
```diff
@@ -51,7 +53,7 @@
                 <p class="board-statistics__body">
                     {% apply spaceless %}
                         {% for user in birthdays %}
-                            {{ user.profilelink|raw }}{% if user.age %} ({{ user.age }}) {% endif %}{% if loop.last != true %}{{ lang.comma }} {% endif %}
+                            <a href="{{ get_profile_link(user.uid) }}">{{ user.uid|format_name }}<a/>{% if user.age %} ({{ user.age }}) {% endif %}{% if loop.last != true %}{{ lang.comma }} {% endif %}
```
Sama34, line 56: Just noticed this tag is broken `<a/>`.

arcalys: Pushed a fix, thanks!

PR #5216, opened 2025-12-18
Add extra hooks to warnings datahandler

### LINKED ISSUE #5202: Add hooks to warning data handler
Before:
https://github.com/mybb/mybb/blob/25edd3ca6c2abbe8398804fad4e4c6114b405943/inc/datahandlers/warnings.php#L342

https://github.com/mybb/mybb/blob/25edd3ca6c2abbe8398804fad4e4c6114b405943/inc/datahandlers/warnings.php#L666

(Though these 2 methods should probably use the user data handler...)

https://github.com/mybb/mybb/blob/25edd3ca6c2abbe8398804fad4e4c6114b405943/inc/datahandlers/warnings.php#L698

arcalys:
### Added 

- Added extra hooks to `update_user()`, `expire_warnings()` and `insert_warning()` in the warnings data handler.

Resolves #5202

### DIFF
--- a/inc/datahandlers/warnings.php
+++ b/inc/datahandlers/warnings.php
@@ -56,6 +56,20 @@ class WarningsHandler extends DataHandler
 	public $friendly_action = '';
 
 	/**
+	 * The list of warnings that are due to expire.
+	 *
+	 * @var array
+	 */
+	public array $expiring_warnings = array();
+
+	/**
+	 * The list of users affected by expired warnings and their updated warning points.
+	 *
+	 * @var array
+	 */
+	public array $affected_users = array();
+
+	/**
 	* Validate a warning user assets.
 	*
 	* @return boolean True when valid, false when invalid.
@@ -314,8 +328,6 @@ class WarningsHandler extends DataHandler
 	{
 		global $db, $plugins;
 
-		$users = array();
-
 		$query = $db->query("
 			SELECT w.wid, w.uid, w.points, u.warningpoints
 			FROM ".TABLE_PREFIX."warnings w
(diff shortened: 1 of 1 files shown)


inc/datahandlers/warnings.php
```diff
@@ -339,6 +339,7 @@ function expire_warnings()
 			}
 		}
 
+		$plugins->run_hooks("datahandler_warnings_expire_warnings", $this);
```
Sama34: I correct myself; this should go before the query above because the while loop already modified the rows at this point.

Here, it is impossible to query for the warning identifiers to be affected.

Sama34: Maybe another solution would be to catch and pass on the modified warning identifiers, but I think just moving the hook is enough.

Sama34 (changes requested):

inc/datahandlers/warnings.php
arcalys: I refactored `expire_warnings()` slightly to add a **before** and an **after** hook, so that plugins can access warnings **before they expire** and **after they have expired**. If this isn’t what you think is best, I can revert to simply moving the original hook before the query.

inc/datahandlers/warnings.php
```diff
@@ -312,9 +326,10 @@ function get($wid)
 	*/
 	function expire_warnings()
 	{
-		global $db;
+		global $db, $plugins;
 
-		$users = array();
+		$this->affected_users = array();
```
Sama34:
```suggestion
```

inc/datahandlers/warnings.php
```diff
@@ -312,9 +326,10 @@ function get($wid)
...
+		$this->expiring_warnings = array();
```
Sama34:
```suggestion
```

PR #5217, opened 2025-12-19
Fix showtimespentonline type error in ACP

arcalys:
### Fixed

Fixes an issue raised by @Sama34  when saving new `showtimespentonline` option in the ACP, throwing a SQL error for non PDO connections.

### DIFF
--- a/inc/datahandlers/user.php
+++ b/inc/datahandlers/user.php
@@ -662,6 +662,7 @@ class UserDataHandler extends DataHandler
 		$this->verify_yesno_option($options, 'pmnotice', 1);
 		$this->verify_yesno_option($options, 'pmnotify', 1);
 		$this->verify_yesno_option($options, 'invisible', 0);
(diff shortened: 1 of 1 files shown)


dvz:
Looking at other options, https://github.com/mybb/mybb/blob/ed8294756bdc364f0646b0084ab4501dd7fca75d/inc/datahandlers/user.php#L657 could be better.

We may also want it in https://github.com/mybb/mybb/blob/ed8294756bdc364f0646b0084ab4501dd7fca75d/inc/datahandlers/user.php#L1133

arcalys:
> Looking at other options,
> 
> https://github.com/mybb/mybb/blob/ed8294756bdc364f0646b0084ab4501dd7fca75d/inc/datahandlers/user.php#L657
> could be better.
> 
> We may also want it in
> 
> https://github.com/mybb/mybb/blob/ed8294756bdc364f0646b0084ab4501dd7fca75d/inc/datahandlers/user.php#L1133

Updated to use the data handler.

PR #5226, opened 2025-12-23
Remove RE: prefix from posts/threads subjects and align length limit

### LINKED ISSUE #4416: Enhancement: $subject_length > 85

arcalys:
### Added

- Added `max_subject_length` class attribute to post data handler.

### Changed

- Updated length limit to 120 characters to match database column.

### Removed

- Removed `RE:` prefixes from threads/posts subjects (and post subjects are deprecated anyway).

Resolves #4416


inc/languages/english/datahandler_post.lang.php
```diff
@@ -11,7 +11,7 @@
 $l['postdata_missing_message'] = 'The message is missing. Please enter a message.';
 $l['postdata_message_too_long'] = 'The message is too long. Please enter a message shorter than {1} characters (currently {2}).';
 $l['postdata_message_too_short'] = 'The message is too short. Please enter a message longer than {1} characters.';
-$l['postdata_subject_too_long'] = 'The subject is too long. Please enter a subject shorter than 85 characters (currently {1}).';
+$l['postdata_subject_too_long'] = 'The subject is too long. Please enter a subject shorter than 120 characters (currently {1}).';
```
Sama34: Since this is used directly in the data handler, maybe pass on the limit to the language string.

arcalys: I was supposed to, forgot about it. 
Updated thank you :)

inc/datahandlers/post.php
```diff
@@ -212,21 +219,10 @@ function verify_subject()
...
-				$subject_length = $subject_length - 4;
-			}
-		}
-
-		if($subject_length > 85)
+		if(my_strlen($subject) > $this->max_subject_length)
 		{
 			// Subject is too long
-			$this->set_error('subject_too_long', my_strlen($subject));
+			$this->set_error('subject_too_long', [$this->max_subject_length, my_strlen($subject)]);
```
Sama34: I think it should be `my_number_format(this->max_subject_length)`

Sama34 (changes requested):

inc/datahandlers/post.php
arcalys: There will never be a decimal point or a thousands separator though, is it really necessary?

PR #5228, opened 2025-12-24
Deprecate format_avatar()

### LINKED ISSUE #5194: Function `format_avatar()` should return integer width and height values
For compatibility reasons, `width_height` should be kept, but the function should return integer values instead of hardocded `width="n" height="n"`

All twig templates should be updated then.
`/inc/themes/core.base/frontend/templates/usercp/home.twig`
`/inc/themes/core.base/frontend/templates/postbit/postbit.twig`
`/inc/themes/core.base/frontend/templates/member/profile.twig`

Related: #3645

arcalys:
### Changed

- Deprecated `format_avatar()`.

Resolves #5194

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -1048,8 +1048,7 @@ EOF;
 	{
 		if(!my_validate_url($user['avatar']))
 		{
-			$avatar = format_avatar($user['avatar'], $user['avatardimensions']);
-			$user['avatar'] = $avatar['image'];
+			$user['avatar'] = htmlspecialchars_uni($mybb->get_asset_url($user['avatar']));
 		}
 	}
 	else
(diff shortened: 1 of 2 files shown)


Sama34:
`format_avatar()` is used in `/admin/modules/user/users.php`, so it should be removed from there too before fully deprecating.

dvz:
[`Deprecated`](https://www.php.net/manual/en/class.deprecated.php) is available from PHP 8.4 (lints don't catch it, tests run on 8.4 - it looks like it's already included in 2 places on `dev-1.9`)

arcalys:
> [`Deprecated`](https://www.php.net/manual/en/class.deprecated.php) is available from PHP 8.4 (lints don't catch it, tests run on 8.4 - it looks like it's already included in 2 places on `dev-1.9`)

Yeah we did not think of that when I added those two in earlier PRs. Do you want them removed as well?

Sama34:
I see no reason not to use the Deprecated attribute since it works as a comment in php < 8.4

dvz:
Yes, we can just use `@deprecated` for the time being, and correct the other ones here or separately.

arcalys:
> `format_avatar()` is used in `/admin/modules/user/users.php`, so it should be removed from there too before fully deprecating.

What would you recommend doing for the ACP part?

arcalys:
Removed the ACP calls to format_avatar() @Sama34

Sama34:
Seems like the AdminCP change removed fallback to the default avatar.

admin/modules/user/users.php
```diff
@@ -3938,9 +3937,23 @@ function build_users_view($view)
 				$max_dimensions = '34x34';
 			}
 
-			$avatar = format_avatar($user['avatar'], $user['avatardimensions'], $max_dimensions);
+			$avatar_url = htmlspecialchars_uni($mybb->get_asset_url($user['avatar']));
```
Sama34, line 3947:
```suggestion
			if(empty($user['avatar']))
			{
				$user['avatar'] = $mybb->settings['useravatar'];

				$user['avatardimensions'] = $mybb->settings['useravatardims'];
			}

			$avatar_url = htmlspecialchars_uni($mybb->get_asset_url($user['avatar']));
```

admin/modules/user/users.php
```diff
@@ -3938,9 +3937,23 @@ function build_users_view($view)
...
+				require_once MYBB_ROOT."inc/functions_image.php";
+				$scaled = scale_image($width, $height, $max_width, $max_height);
+				$dimensions = " width=\"{$scaled['width']}\" height=\"{$scaled['height']}\"";
+			}
+			elseif(!empty($user['avatardimensions']))
+			{
+				list($width, $height) = preg_split('/[|x]/', $user['avatardimensions']);
+				$dimensions = " width=\"{$width}\" height=\"{$height}\"";
+			}
+			$user['view']['avatar'] = "<img src=\"{$avatar_url}\" alt=\"\"{$dimensions} />";
```
Sama34:
```suggestion
			if(!empty($user['avatardimensions']))
			{
				list($width, $height) = preg_split('/[|x]/', $user['avatardimensions']);
				list($max_width, $max_height) = preg_split('/[|x]/', $max_dimensions);
				require_once MYBB_ROOT."inc/functions_image.php";
				$scaled = scale_image($width, $height, $max_width, $max_height);
				$dimensions = " width=\"{$scaled['width']}\" height=\"{$scaled['height']}\"";
			}

			$user['view']['avatar'] = "<img src=\"{$avatar_url}\" alt=\"\"{$dimensions} />";
```

Sama34 (changes requested):

PR #5256, opened 2026-01-09
Format last visit date in portal.twig

arcalys:
### Fixed

- Missing closing `</li>` tag.
- Properly formatted last visit date on portal.

### DIFF
--- a/private.php
+++ b/private.php
@@ -761,8 +761,8 @@ if($mybb->input['action'] == "send")
 		$post['message'] = $mybb->get_input('message');
 		$post['subject'] = htmlspecialchars_uni($mybb->get_input('subject'));
 		$post['icon'] = $mybb->get_input('icon', MyBB::INPUT_INT);
-		$post['to'] = $to;
-		$post['bcc'] = $bcc;
+		$post['to'] = $sendpm['to'];
+		$post['bcc'] = $sendpm['bcc'];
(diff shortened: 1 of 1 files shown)


inc/themes/core.base/frontend/templates/portal/portal.twig
```diff
@@ -22,7 +22,7 @@
                 <section class="block block--sidebar">
                     <h2 class="title title--sidebar">{{ lang.welcome }}</h2>
                     <ul class="list list--divided list--stats">
-                        <li class="list__item">{{ lang.member_welcome_lastvisit }} {{ portal.lastvisit|raw }}
+                        <li class="list__item">{{ lang.member_welcome_lastvisit }} {{ portal.lastvisit|my_date }}
```
Sama34: It seems like this may also be unset ? I can't see the definition in the `portal.php` file.

arcalys: It's defined here: https://github.com/mybb/mybb/blob/dev-1.9/portal.php#L149

arcalys: Checked again, its indeed defined for logged in user, and block is showed only to logged in users

arcalys:
Fixed in the themes PR.

PR #5224, opened 2025-12-22
Update 1.9 theme style and associated fixes

justinsoltesz:
Includes a number of changes to the theme to update, standardize, and otherwise clean up the design. Also includes a number of fixes associated with these changes.

I am still working through some additional clean up and testing, so making this a draft pull request for now.

Resolves #5201

PR #5259, opened 2026-01-10
Render forum rules as raw HTML in forumdisplay.twig

arcalys:
### Fixed

- HTML is displayed instead of rendered for forum rules in forumdisplay (other locations, such as newthread, already use `|raw`).


justinsoltesz:
I can just include this in my theme update PR, since I already am making several changes to `forumdisplay.twig`

arcalys:
To be included in https://github.com/mybb/mybb/pull/5224

PR #5298, opened 2026-02-15
Fix/absolute attachment urls

Sama34:
This is a fix for when rending/linking attachments in the portal; because the portal can be placed outside the scope of the forum, it is necessary to make attachment links absolute.

I just made all links absolute for consistency across the theme.

### DIFF
--- a/inc/themes/core.base/frontend/templates/misc/attachments_attachment.twig
+++ b/inc/themes/core.base/frontend/templates/misc/attachments_attachment.twig
@@ -5,7 +5,7 @@
             {% if attachment.visible %}
                 {{ attachment.filename }}
             {% else %}
-                <a href="attachment.php?aid={{ attachment.aid }}" target="_blank">{{ attachment.filename }}</a>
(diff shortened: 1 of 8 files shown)


inc/themes/core.base/frontend/templates/misc/attachments_attachment.twig
```diff
@@ -5,7 +5,7 @@
             {% if attachment.visible %}
                 {{ attachment.filename }}
             {% else %}
-                <a href="attachment.php?aid={{ attachment.aid }}" target="_blank">{{ attachment.filename }}</a>
+                <a href="{{ mybb.settings.bburl }}/attachment.php?aid={{ attachment.aid }}" target="_blank">{{ attachment.filename }}</a>
```
arcalys: Could be replaced with a call to `url()`.

arcalys: Same comment for all other occurences.

Sama34: You are correct, since I'm updating all of them I can as well do so, thanks

arcalys (approved):

PR #5306, opened 2026-03-01
Implement magic login feature using email links

Sama34:
Implements #5311 

The feature adds a new setting under the "Login and Registration Options" settings group.
- Enable Magic Login
 
Enabling the feature doesn't affect other login methods.

When enabled, a new login option should appear in both the quick login modal and the login page. A simple validation for the email is done, but the user receives the same successful message regardless of the email address being assigned to a user or not.

The feature will fetch the first found user with an email, which may cause confusion for emails shared across different forum accounts (i.e: the "Allow emails to be registered multiple times?" setting is enabled).

Requesting magic links and logging in using magic links should be allowed actions when the board is closed.

image
image
image

<details>
<summary>Generative AI Summary (Le Chat)</summary>
This feature introduces passwordless authentication via "magic links" sent to users' email addresses, addressing challenges like password fatigue and security risks. Controlled by an admin setting, it includes a database table for storing links, rate limiting, and security notifications. Users request a link, click it to log in, and are notified of the login attempt. The system ensures compatibility with existing password logins and includes measures to mitigate risks such as email interception and brute force attacks. Testing covers functionality, rate limiting, security notifications, and database integrity. Documentation updates are required for user, admin, and developer guides. The implementation involves modifications to several files including database schemas, templates, and upgrade routines.
</details>

### DIFF
--- a/inc/datahandlers/login.php
+++ b/inc/datahandlers/login.php
@@ -301,7 +301,7 @@ class LoginDataHandler extends DataHandler
 	/**
 	 * @return bool true
 	 */
-	function complete_login()
+	function complete_login(bool $update_hash = true)
 	{
 		global $plugins, $db, $mybb, $session;
 
@@ -333,7 +333,7 @@ class LoginDataHandler extends DataHandler
 			$this->captcha->invalidate_captcha();
 		}
 
-		if (user_password_needs_rehash($user)) {
+		if ($update_hash && user_password_needs_rehash($user)) {
 			$data = create_password($this->data['password']);
 			$db->update_query('users', $data, 'uid=' . (int)$user['uid']);
 		}
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "symfony/console": "^7.0",
         "symfony/filesystem": "^7.1",
         "symfony/polyfill-mbstring": "^1.32",
+        "symfony/uid": "^7.0",
         "twig/twig": "^3.10"
     },
     "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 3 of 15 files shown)


dvz:
Why UUID?

Sama34:
> Why UUID?

Because of its uniqueness. For instance, the `mybb_awaitingactivation` table still requires a primary code making the assumption of the `code` uniqueness questionable.

I designed the table as a general purpose table and one that will probably not be pruned for logging purposes.

If you are worried about the relying on an additional dependency, I could implement a global function instead.

If you are worried on the length of the UUID well, I'm assuming the user won't be typing the code themselves. Why would they? That would seem contrary to the "usefulness and easy of use" reasoning that motivates this feature. User should be able to just open the link with a click and at worst they can copy/paste it.

dvz:
We'd use `random_str()` in most places - looks like it would suffice here.

Sama34:
I prefer UUIDs, but I'm willing to keep the standardized use of `random_str`.

inc/datahandlers/login.php
```diff
@@ -301,7 +301,7 @@ function validate_login()
 	/**
 	 * @return bool true
 	 */
-	function complete_login()
+	function complete_login(bool $update_hash = true)
```
dvz, line 304: Can the class check provided login data instead?

dvz:
We can store only a HMAC of this value instead (using `encryption_key`). It can additionally depend on user ID and expiration time.

There may be cases where the link is accessed automatically before the user does - a POST form may help (and address phishing issues around unknowingly logging into an account).

Dropping the `url` logic may make sense (and reduce duplication + reduce phishing opportunities).

There may be opportunities to move some of the logic into `LoginDataHandler`, making the code more cohesive (https://github.com/mybb/mybb/pull/5306/changes#r2880807697).

Sama34:
I agree the `url` parameter should be removed, I just kept it for consistency.

I will take your points in consideration, I already have some changes in local that I need to test.

There are also some issues regarding rate limiting I need to consider too.

PR #5323, opened 2026-03-13
Switch ACP 2FA to chillerlan libs and expand authsecret

### LINKED ISSUE #5260: 2FA QR code should be generated locally
This shouldn't be sent to a third party for security reasons.

arcalys:
Removes ACP 2FA’s dependency on external QR providers and switches 2FA internals to chillerlan libraries.

### Added

Added `chillerlan/php-authenticator` and `chillerlan/php-qrcode` dependencies.

### Changed

- Replaced ACP 2FA secret/verification flow to use `chillerlan/php-authenticator` through a thin wrapper.
- Replaced QR code rendering with local rendering via `chillerlan/php-qrcode`.
- Increased adminoptions.authsecret column width from varchar(16) to varchar(64).

Resolves #5260


dvz:
Any issues with pre-existing codes?

Looks like we can remove `GoogleAuthenticator.php` now.

Some components may be locked to higher versions than necessary:
```
symfony/string     v8.0.6 requires php (>=8.4)
symfony/var-dumper v8.0.6 requires php (>=8.4)
```
(related: #5327)

arcalys:
> Any issues with pre-existing codes?
> 
> Looks like we can remove `GoogleAuthenticator.php` now.
> 
> Some components may be locked to higher versions than necessary:
> 
> ```
> symfony/string     v8.0.6 requires php (>=8.4)
> symfony/var-dumper v8.0.6 requires php (>=8.4)
> ```
> 
> (related: #5327)

I don't remember any issues when I tested, even with pre-existing codes, but I'll do another round of tests.
I'll downgrade the components to avoid the PHP 8.4+ dependency.

arcalys:
@dvz 

Downgraded the dependencies and re-tested using a code generated before the library change. Everything looks good on my end. If you or @Sama34 could test it as well, that would be ideal.

PR #5343, opened 2026-05-08
Fix logic flaw in usercp.php regarding 'invisible' status permission validation

### LINKED ISSUE #5339: Invisible mode permission bypass in usercp.php
**Description:**
There is a logic flaw in `usercp.php `where the `invisible` status of a user is updated without checking if their usergroup has the `canbeinvisible` permission.

**Steps to Reproduce:**

1. Login with a user whose usergroup has `canbeinvisible` set to No.
2. Manually send a POST request to `usercp.php?action=do_options` with `invisible=1` or manipulate the HTML to add the checkbox.
3. Observe that the user's `invisible` status is updated in the database regardless of group permissions.

**Expected Behavior:**
The system should verify the `canbeinvisible` permission before processing the update. If the user doesn't have permission, the value should be forced to `0`. Additionally, the option should not be visible in the UserCP UI.

Proposed Solution:

Add a backend check in `usercp.php` for `$mybb->usergroup['canbeinvisible'].`

Update the relevant template to hide the checkbox for unauthorized users.

emregeldegul:
Set default value for 'invisible' option if usergroup allows it.

Resolves #5339 for 1.8

Currently, usercp.php updates a user's invisible status without verifying if their usergroup has the canbeinvisible permission. While the default MyBB 1.8 templates use conditional variables to hide the option in the UI, there is no backend validation. This allows users to bypass the restriction by sending direct POST requests or using custom themes that might have the option hardcoded.

This PR adds a backend check to ensure that the invisible status is only updated if the user has the required permission. If not, the value is forced to 0. This ensures consistent permission enforcement regardless of the theme or method used to submit the form.

### DIFF
--- a/usercp.php
+++ b/usercp.php
@@ -785,7 +785,7 @@ if($mybb->input['action'] == "do_options" && $mybb->request_method == "post")
 		"allownotices" => $mybb->get_input('allownotices', MyBB::INPUT_INT),
 		"hideemail" => $mybb->get_input('hideemail', MyBB::INPUT_INT),
 		"subscriptionmethod" => $mybb->get_input('subscriptionmethod', MyBB::INPUT_INT),
-		"invisible" => $mybb->get_input('invisible', MyBB::INPUT_INT),
+		"invisible" => 0,
 		"dstcorrection" => $mybb->get_input('dstcorrection', MyBB::INPUT_INT),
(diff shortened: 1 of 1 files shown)


arcalys:
Thanks @emregeldegul

PR #5344, opened 2026-05-08
Fix #5292: Add error message when uploading avatar while avatar is suspended

### LINKED ISSUE #5292: Uploading an avatar from the AdminCP fails when user has avatar suspended
This is more of a UIX issue. When users have their avatar privilege suspended, but admins try to update their avatar (i.e: upload), the admin doesn't receive any clarification about the avatar not being uploaded and instead is misleading into thinking there was no issue whatsoever.

It is of course confusing to not get any error message about this and still the avatar is not uploaded.

I think the avatar tab should be dynamically hidden using CSS based on the value of the `Suspend avatar use` option.

emregeldegul:
### Resolves
Closes #5292 

### Description
This PR addresses the silent failure issue that occurs when an admin attempts to upload a new avatar for a user while the "Suspend Avatar" option is checked in the AdminCP. 

Previously, the system would successfully upload the new avatar, but the subsequent moderation block would immediately delete it without throwing any errors, misleading the admin into thinking the process was entirely successful.

While the original issue suggested dynamically hiding the avatar tab using CSS, I believe a backend validation approach is much more robust and provides better UX by explicitly informing the admin of the conflict.

### Changes Made
* Added a new language string `suspendavatar_upload_error` to `admin/user_users.lang.php`.
* Added validation in `admin/modules/user/users.php` to check if an avatar file/URL is being submitted concurrently with the `suspendavatar` action. 
* If a conflict is detected, it now halts the process and throws an explicit inline error, prompting the admin to untick the suspend option before proceeding with the upload.

### DIFF
--- a/admin/modules/user/users.php
+++ b/admin/modules/user/users.php
@@ -631,123 +631,132 @@ if($mybb->input['action'] == "edit")
 			// Are we uploading a new avatar?
 			if($_FILES['avatar_upload']['name'])
 			{
-				if($user['suspendavatar'])
+				if($mybb->get_input('suspendavatar'))
 				{
-					flash_message($lang->suspendavatar_upload_error, 'error');
-					admin_redirect("index.php?module=user-users&action=edit&uid={$user['uid']}");
-				}
-				$avatar = upload_avatar($_FILES['avatar_upload'], $user['uid']);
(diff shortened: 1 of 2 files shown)


emregeldegul:
fixed @arcalys

PR #5349, opened 2026-05-16
Fix default avatar display in user CP home

arcalys:
### Fixed

- Default avatar initial not displaying in user CP home (incorrect `alt` reference)


arcalys:
Will be handled in #5353

PR #5350, opened 2026-05-16
Allow using font icons for post icons

### LINKED ISSUE #5188: Update post icons to use font icons
https://github.com/mybb/mybb/issues/3500#issuecomment-439635990

The post icons feature should support font icons.

I'm adding some issues to 1.9 by default, but feel free to suggest a target version.

arcalys:
Introduces CSS/font-based post icons while preserving legacy image-path icons for backward compatibility.

### Added
- Added `icon_class` column to the post icons schema.
- Added upgrade logic to add icon_class for existing installs.
- Added class-based default post icons for new installs only.

### Changed
- Extended ACP add, edit, and manage flows to support both icon modes.
- Validation now requires at least one of `icon_class` or `path`.
- Updated frontend rendering to use icon data:
  - render `icon_class` first when available
  - otherwise render `path`

I intentionally did not add CSS/font icon rows in the upgrade to avoid duplicating post icons when upgrading from 1.8; class-based defaults are only seeded on new installs.

Resolves #5188


emregeldegul:
Thanks for the PR, this is a great feature! I’d also suggest adding a 'default post icon.' This way, users can set a default icon to be used as a fallback. If no specific icon is selected (i.e., it's left empty), the default one will be displayed. Just to clarify, I’m not talking about assigning a default value in the database, but rather handling it on the frontend/display side.

PR #5360, opened 2026-05-20
Fix: Prevent thread subscriptions from being deleted during quick/inline edits

emregeldegul:
### Description
This PR fixes a bug where a user's thread subscription is unexpectedly removed when a moderator (or the user themselves) performs a Quick Edit (`xmlhttp.php`) on a post.

### Cause
In `PostDataHandler::update_post()`, the subscription logic falls back to a destructive `else` block that deletes the user's subscription if `$post['options']['subscriptionmethod']` is empty. Since inline/quick edits do not submit subscription preferences, this condition is always triggered, wiping out the author's subscription if their `$uid` is present.

### Solution
Modified the fallback `else` block to an explicit `else if` statement. The subscription will now only be deleted from the database if `subscriptionmethod` is explicitly set to `"none"`. Missing subscription arrays (like those in AJAX quick edits) will safely bypass the deletion logic, preserving existing subscriptions.

### DIFF
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -2000,7 +2000,6 @@ class PostDataHandler extends DataHandler
 		$db->update_query("posts", $this->post_update_data, "pid='".(int)$post['pid']."'");
 
 		// Automatic subscription to the thread
-		// Automatic subscription to the thread
 		if($post && !empty($post['options']['subscriptionmethod']) && $uid > 0)
 		{
 			switch($post['options']['subscriptionmethod'])
--- a/inc/datahandlers/post.php
+++ b/inc/datahandlers/post.php
@@ -2000,6 +2000,7 @@ class PostDataHandler extends DataHandler
 		$db->update_query("posts", $this->post_update_data, "pid='".(int)$post['pid']."'");
 
 		// Automatic subscription to the thread
+		// Automatic subscription to the thread
 		if($post && !empty($post['options']['subscriptionmethod']) && $uid > 0)
 		{
 			switch($post['options']['subscriptionmethod'])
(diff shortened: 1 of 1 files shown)


BettyIsBoop:
My opinion: $uid is bad here.
All permission test are done using $uid, even if moderator update ...

See https://github.com/emregeldegul/mybb/blob/712d6bef3d1935b63467befa304731fcbcd3b16d/inc/datahandlers/post.php#L1883 for example.

uid must be the current user !

My fix is 

```
		// Automatic subscription to the thread
		if($post && !empty($post['options']['subscriptionmethod']) && $mybb->user['uid'] > 0)
		{
			switch($post['options']['subscriptionmethod'])
			{
				case "pm":
					$notification = 2;
					break;
				case "email":
					$notification = 1;
					break;
				default:
					$notification = 0;
			}
			require_once MYBB_ROOT."inc/functions_user.php";
			add_subscribed_thread($post['tid'], $notification, $mybb->user['uid']);
		}
		elseif ($mybb->user['uid'] > 0)
		{
			$db->delete_query("threadsubscriptions", "uid='".(int)$mybb->user['uid']."' AND tid='".(int)$post['tid']."'");
		}
```

BettyIsBoop:
There are 2 issues her to fix

1. Edition by moderator update user subscription (my fix)
2. Quick edition remove subscription

Your fix are not OK too : ==="none" mean "Subscription without email"

We need it to be set, but it is empty, i report boith issue and the fix 

```
		// Automatic subscription to the thread
		if($post && !empty($post['options']['subscriptionmethod']) && $mybb->user['uid'] > 0)
		{
			switch($post['options']['subscriptionmethod'])
			{
				case "pm":
					$notification = 2;
					break;
				case "email":
					$notification = 1;
					break;
				default:
					$notification = 0;
			}
			require_once MYBB_ROOT."inc/functions_user.php";
			add_subscribed_thread($post['tid'], $notification, $mybb->user['uid']);
		}
		elseif (isset($post['options']['subscriptionmethod']) && $mybb->user['uid'] > 0)
		{
			$db->delete_query("threadsubscriptions", "uid='".(int)$mybb->user['uid']."' AND tid='".(int)$post['tid']."'");
		}
```

PR #5378, opened 2026-06-05
#5365: Add random password generator to registration form

### LINKED ISSUE #5365: Generate random password at registration
Add a button to generate a random password at registration.

Original thread: https://community.mybb.com/thread-221632.html

peripherium:
This PR adds a random password generator to the registration form.

A new control allows users to generate a secure random password without leaving the registration page. The generated password is automatically applied to both password fields and can be regenerated as needed.

(1) Uses the browser's cryptographically secure random number generator (crypto.getRandomValues)

(2) The JavaScript random_str() implementation follows the same approach as the existing PHP random_str() in inc/functions.php

(3) Ensures at least one digit, one uppercase letter, and one lowercase letter are present in the generated password
Users can toggle the visibility of the generated password to make it easier to save or record

(4) Additional passwords can be generated with a single click if the user prefers a different one

(5) Does not change the existing registration workflow

Resolves #5365

### DIFF
--- a/inc/languages/english/global.lang.php
+++ b/inc/languages/english/global.lang.php
@@ -78,6 +78,7 @@ $l['username'] = "Username";
 $l['username1'] = "Email";
 $l['username2'] = "Username/Email";
 $l['password'] = "Password";
+$l['random_password'] = "Or generate random password";
 $l['login_username'] = "Username";
 $l['login_username1'] = "Email";
 $l['login_username2'] = "Username/Email";
--- a/inc/themes/core.base/frontend/templates/member/register.twig
+++ b/inc/themes/core.base/frontend/templates/member/register.twig
@@ -2,6 +2,32 @@
 
 {% block head %}
     <title>{{ lang.registration }} - {{ mybb.settings.bbname }}</title>
+    <style>
+.field-container {
+    position: relative;
+}
+
+[data-password-peek] {
+    position: absolute;
+    top: 0;
+    right: 0;
+    margin-top: 1px;
+    padding: 8px 9px;
+    cursor: pointer;
+}
+
+[data-password-peek] .icon {
+    color: #aaa;
+}
+
+[data-password-peek][aria-pressed="true"] .icon {
+    color: #222;
+}
+
+input[data-password-peekable] {
+    padding-right: 35px;
+}
+    </style>
 {% endblock head %}
 
 {% block body %}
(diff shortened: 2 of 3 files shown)


arcalys:
I haven't checked the JavaScript part, but instead of inlining the script, you can include it via the `assets.json` manifest. For example:
```
		"./jscripts/random_password.js": {
			"attached_to": [
				{
					"script": "member.php"
				}
			]
		}
```
Also, should we generate this on the backend side instead (e.g. using `Random\Randomizer::getBytesFromString()`)?

peripherium:
Agreed. This version is much clearer and easier to follow. Thanks for the suggestion.

> Also, should we generate this on the backend side instead (e.g. using Random\Randomizer::getBytesFromString())?

I chose a client-side implementation intentionally, since users will likely regenerate passwords multiple times until they get one they personally consider memorable, convenient, or otherwise suitable.

Generating passwords directly in the browser allows this without page reloads or additional server requests.

The implementation uses the Web Crypto API (`crypto.getRandomValues())`, which provides cryptographically secure randomness client-side.

I may be overlooking something, so please let me know if this approach could cause issues elsewhere or if there are implementation details I should take into account.

dvz:
https://github.com/mybb/mybb/issues/5365#issuecomment-4663248408

With both features present, we can keep the generation mechanism separate from revealing.

The generation can use the same approach to events as in `form.js` to skip embedded scripts altogether (possibly in the same file, eventually).

A _Generate_ label alone may be better when placed next to password fields.

Symbols like `randompasswd` should be expanded for readability. The JS methods can be wrapped in an object/class similarly to existing scripts in `jscripts/`.

Client-side generation is good.

`randomInt()` is currently affected by [modulo bias](https://codeql.github.com/codeql-query-help/javascript/js-biased-cryptographic-random/) that should be addressed for a generic helper that may be reused later.

peripherium:
> The generation can use the same approach to events as in `form.js` to skip embedded scripts altogether (possibly in the same file, eventually).

Okay, I've inserted that into `form.js` now. If we’re using this "peek" mechanism, that's probably a good place for it.

> A _Generate_ label alone may be better when placed next to password fields.

I've incorporated a space-saving solution that matches the rest of the design. Maybe that's exactly what you meant regarding the `<label>`?

> `randomInt()` is currently affected by [modulo bias](https://codeql.github.com/codeql-query-help/javascript/js-biased-cryptographic-random/) that should be addressed for a generic helper that may be reused later.

I've fixed that now. With uniformly distributed random numbers, the "largest possible multiple of the range" approach shouldn't make any noticeable difference to the user.

However, a small problem has arisen with the new changes, and I haven't been able to come up with a solution yet: the "eye" icon for the password fields is appearing outside the input fields (see screenshot). Can anyone spot what I'm doing differently compared to the setup script? I would be very grateful for any suggestions.
eye-icon

peripherium:
Please disregard my previous comment regarding the design question. With my latest push, the changes should now align with the expected forum style based on the earlier feedback.

@dvz & @arcalys  I'd appreciate it if you could take another look and let me know whether anything else should be adjusted.
rand-pw

Issue #3662, opened 2019-05-22
Extended session tracking


dvz:
Most important changes:
- track user sessions per-device in a database table (currently a `sessions` table is used for visit statistics; this will allow to list and terminate sessions on user and global level in case of a breach),
- split session keys (with one part designated for querying rows, to prevent database timing attacks on the 2nd part),
- store hashes/HMACs of session keys only (the 2nd part when split; this will prevent extracting and using keys from database leaks, and HMACs can limit exploitation of maliciously inserted values),
- use IP and useragent string comparison to limit hijacking attempts,
- use session-specific anti-CSRF tokens.

Possibly built on top of [`\Illuminate\Session`](https://github.com/laravel/framework/tree/5.8/src/Illuminate/Session).

Target branch: 1.9 or 1.10.

effone:
Shall it provide encryption as well using OpenSSL with `AES-256-CBC`?

dvz:
It might at some point; HMAC and/or encryption keys for sessions as well as [passwords](https://github.com/mybb/mybb/issues/3530) would require a storage system, preferably a common one with support for multiple keys, first.

PR #5379, opened 2026-06-05
Expire awaitingactivation tokens after 24 hours

unknownhad:
# Expire awaitingactivation tokens after 24 hours

## Problem

Password reset (`type='p'`), email-change confirmation (`type='e'`) and ACP
unlock (`type='l'`) codes are stored in `awaitingactivation` together with a
`dateline`, but the `dateline` was never checked when the code was consumed.
Every such code therefore stayed valid **forever** (single-use, but with an
unlimited window) and there was no task cleaning the rows up.

The dead `$now = TIME_NOW;` in the reset-password handler (member.php) shows an
age check was once intended but never finished.

Impact:
- A password-reset link works a year later — a stale/leaked link is a permanent
  account-takeover primitive.
- An email-change confirmation link works indefinitely. If the pending address
  is later abandoned and re-registered by someone else, they can complete the
  change and take over the account.
- A stale ACP unlock link clears an admin lockout at any time, and the unlock
  endpoint runs before the login rate-limiter, so the 8-char token could be
  ground on.

For reference, WordPress expires reset links after 24h, phpBB after 1h.

## Changes

- **member.php (`resetpassword`)** — fetch `dateline` alongside the code,
  reject + delete codes older than 24h, and remove the dead `$now = TIME_NOW;`.
- **member.php (`activate`, type `e`)** — reject email-change confirmations
  older than 24h before applying the new address. Scoped to `type='e'` only, so
  account-activation (`r`) and board-validation (`b`) links — which can
  legitimately be days old — are unaffected.
- **admin/index.php (`unlock`)** — add a 24h `dateline` bound to the unlock
  token lookup, and raise the lockout token from `random_str()` (8 chars) to
  `random_str(30)` to match the reset-code length.
- **inc/tasks/hourlycleanup.php** — purge stale `p`/`e`/`l` rows hourly; `r`/`b`
  are intentionally left untouched.
- **inc/languages/english/member.lang.php** — two new error strings.

The 24h lifetime is hardcoded (matches phpBB/WP/Discourse). Happy to convert it
to a board setting if preferred.

## Not included / follow-ups

- **Rate limiting on the ACP unlock endpoint.** The `action=unlock` block runs
  before the `do=login` limiter. With the token now 30 chars brute force is
  infeasible, but a dedicated limiter on the unlock action is still worth
  considering and is an architectural decision better left to maintainers.
- **Login-key rotation on logout** (separate report item) is intentionally NOT
  touched here, as it's tracked in #3662 and slated for a session-key redesign.

## Testing

Installed MyBB 1.8 (this branch) in Docker (PHP 8.1 + MariaDB) and verified, via
direct HTTP requests, that for each of the three token types a 1-year-old code
is now rejected while a fresh code still works, that legitimate account
activation (`r`) is unaffected, and that hourly cleanup removes stale `p`/`e`/`l`
rows while keeping `r`/`b`. Reverse-applying the patch reproduces the original
indefinite-validity behaviour (password reset, email hijack, lockout clear all
succeed with a year-old token).

## Disclosure

This change was prepared with AI assistance and has been reviewed and tested by
the @unknownhad , per the MyBB contribution policy on generated content.

### DIFF
--- a/inc/tasks/hourlycleanup.php
+++ b/inc/tasks/hourlycleanup.php
@@ -49,8 +49,7 @@ function task_hourlycleanup($task)
 	$cut = TIME_NOW-(60*60*24*7);
 	$db->delete_query("questionsessions", "dateline < '".(int)$time['question']."'");
 
-	// Delete expired password reset ('p'), email change ('e') and admin lockout ('l') tokens (older than 24 hours).
-	// Account activation ('r') and board validation ('b') tokens are intentionally left untouched.
+	// Delete expired password reset ('p'), email change ('e') and admin lockout ('l') tokens (older than 24 hours)
 	$db->delete_query("awaitingactivation", "type IN ('p','e','l') AND dateline < '".(int)(TIME_NOW-(60*60*24))."'");
 
 	add_task_log($task, $lang->task_hourlycleanup_ran);
--- a/inc/tasks/hourlycleanup.php
+++ b/inc/tasks/hourlycleanup.php
@@ -49,5 +49,9 @@ function task_hourlycleanup($task)
 	$cut = TIME_NOW-(60*60*24*7);
 	$db->delete_query("questionsessions", "dateline < '".(int)$time['question']."'");
 
+	// Delete expired password reset ('p'), email change ('e') and admin lockout ('l') tokens (older than 24 hours).
+	// Account activation ('r') and board validation ('b') tokens are intentionally left untouched.
+	$db->delete_query("awaitingactivation", "type IN ('p','e','l') AND dateline < '".(int)(TIME_NOW-(60*60*24))."'");
+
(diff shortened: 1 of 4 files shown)


inc/tasks/hourlycleanup.php
```diff
@@ -49,5 +49,9 @@ function task_hourlycleanup($task)
 	$cut = TIME_NOW-(60*60*24*7);
 	$db->delete_query("questionsessions", "dateline < '".(int)$time['question']."'");
 
+	// Delete expired password reset ('p'), email change ('e') and admin lockout ('l') tokens (older than 24 hours).
+	// Account activation ('r') and board validation ('b') tokens are intentionally left untouched.
```
dvz:
```suggestion
	// Delete expired password reset ('p'), email change ('e') and admin lockout ('l') tokens (older than 24 hours)
```

PR #5392, opened 2026-06-17
Fix undefined array key in get_post_attachments after portal.php

### LINKED ISSUE #5386: portal.php - Warning [2] Undefined array key "candlattachments"
Another regression from #5124.

`Warning [2] Undefined array key "candlattachments" - Line: 1079`

In portal.php, permissions are stored per forum ID:
```php
$forumpermissions[$thread['fid']] = forum_permissions($thread['fid']);
```
So the actual permission value would be located at:
```php
$forumpermissions[$fid]['candlattachments']
```
But in inc/functions_post.php, `$forumpermissions` is treated as a flat permission array:
```php
if(!isset($forumpermissions))
{
    $forumpermissions = forum_permissions($post['fid']);
}
```
and later accessed as:
```php
$forumpermissions['candlattachments']
```
If `$forumpermissions` was already initialized by portal.php, this key does not exist at the top level, because the array is nested by forum ID. This results in the undefined array key warning.

SAY-5:
Resolves #5386

When `portal.php` runs first it populates the global `$forumpermissions` as an array keyed by forum ID, but `get_post_attachments()` treats it as a flat permission row. Its `isset($forumpermissions)` guard then sees the nested array and skips the lookup, so reading `$forumpermissions['candlattachments']` warns. This resolves the permissions for the post's own forum from whichever shape the global currently holds.

### DIFF
--- a/inc/functions_post.php
+++ b/inc/functions_post.php
@@ -1026,9 +1026,20 @@ function get_post_attachments($id, &$post)
 	];
 	$post['hasattachments'] = false;
 
-	if(!isset($forumpermissions))
+	// $forumpermissions may be unset, a flat permission array, or (as built by
+	// portal.php) an array keyed by forum ID. Resolve this post's permissions
+	// from whichever shape is present so a missing key does not warn.
+	if(isset($forumpermissions[$post['fid']]) && is_array($forumpermissions[$post['fid']]))
 	{
-		$forumpermissions = forum_permissions($post['fid']);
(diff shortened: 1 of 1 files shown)


SAY-5:
Opened #5406 with the same fix against feature. Happy for this one to be closed, or kept if you want the 1.9 line covered too.

arcalys:
We will port the commits to 1.9 when the next 1.8 patch is released :)
Closing this one.

PR #5393, opened 2026-06-19
Validate email change activation against requested user

arcalys:
### Changed

Ensures email change activation is validated against the requested user.

Based on a report from Aikido Security (aikido.dev).


dvz:
Regression from 8e73fe1f4526086320b2d035c9c3eef6a0b05210

PR #5410, opened 2026-09-03
Update Twitter profile field to X

emregeldegul:
## Summary

Update the default social profile field to use the current X branding instead of Twitter.

## Changes

- Rename the default profile field from `Twitter` to `X`.
- Update its description to refer to an X username.

## Validation

- `php -l inc/seeds/db_inserts.php`
- `git diff --check`

## Note
Spider seed display name is now `X`, while the `useragent` remains `Twitterbot` intentionally because the agent identifier has not changed.


arcalys:
We should probably also add a migration to update it on existing installations.

emregeldegul:
> We should probably also add a migration to update it on existing installations.

Could you provide some guidance on how migrations are typically handled here? Are they kept in a separate repository? Also, it would be really helpful to know the approach taken when the contact fields were moved to custom profile fields—could you point me in the right direction?

arcalys:
> > We should probably also add a migration to update it on existing installations.
> 
> Could you provide some guidance on how migrations are typically handled here? Are they kept in a separate repository? Also, it would be really helpful to know the approach taken when the contact fields were moved to custom profile fields—could you point me in the right direction?

The upgrade script for 1.9 is located here: [upgrade script](https://github.com/mybb/mybb/blob/dev-1.9/inc/upgrades/upgrade100.php). There are already existing field updates (including the profile fields you mention), but let me know if this is not clear enough :)

emregeldegul:
> > > We should probably also add a migration to update it on existing installations.
> > 
> > 
> > Could you provide some guidance on how migrations are typically handled here? Are they kept in a separate repository? Also, it would be really helpful to know the approach taken when the contact fields were moved to custom profile fields—could you point me in the right direction?
> 
> The upgrade script for 1.9 is located here: [upgrade script](https://github.com/mybb/mybb/blob/dev-1.9/inc/upgrades/upgrade100.php). There are already existing field updates (including the profile fields you mention), but let me know if this is not clear enough :)

Hi, thank you. I looked into this a bit more. As far as I can tell, the Twitter profile field was introduced during the ongoing 1.9 development, and it wasn’t part of the feature branch. Since 1.9 hasn’t been released yet, I don’t think there are existing 1.9 installations that would need this migration. Because of that, I removed the migration and kept the change in the seed data only. Does this approach sound right, or am I missing an upgrade path?

dvz:
We wanted to reduce out-of-the-box profile fields to avoid these issues - not sure if this one was discussed, but IMO removing it altogether would make sense.

emregeldegul:
I think it could be nice to keep a few popular platforms by default as examples. However, as you mentioned, keeping these kinds of platform-specific fields up to date can become a headache from time to time. Ultimately, it's your call, and I'm happy to proceed in whichever direction you think is best.

Stefan-MyBB:
MyBB used to have ICQ, AIM, Yahoo and MSN fields that were the most popular platforms at that time. Given the long release cycles of MyBB I'm also for keeping it as minimal as possible. Forum owners can always add the platforms that are relevant for them.

